PR #3, opened 2025-06-07
[Draft]Add grouped query attention fwd and bwd kernels

xuchangtolearn:
Add grouped query attention fwd and bwd kernels


chengyupku:
LGTM, but I think we don't need profiling the performance in the check function :-)

xuchangtolearn:
> LGTM, but I think we don't need profiling the performance in the check function :-)

Thanks for your comment. I have removed profiling the performance from the check function.

PR #9, opened 2025-06-11
[Dev] Add MHA fwd/bwd kernels & refine format style

Rachmanino:
1. Add `type: ignore` to linear attention kernels to improve code style.
2. Add MHA fwd/bwd kernels for BSHD layout w/o wgmma


tla/__init__.py
chengyupku, line 1: Please stick to either camelCase or snake_case for naming, but avoid mixing both styles in the same codebase.

PR #33, opened 2025-09-28
[Fix]: Disable tma lower in mha decode

RMLYC:
1. fix dir path bug in ci-test.sh
2. disable tma lower in mha decode


RMLYC:
Fix in #34

PR #42, opened 2025-10-11
[RFC][Refactor] Refactor TileOps for better hierarchical abstraction and designs

Rachmanino:
This pull request introduces a new modular kernel and function structure for multi-head attention (MHA), refactors imports to support the new organization, and adds comprehensive testing and profiling capabilities. The main focus is on making the MHA kernel implementation more extensible and easier to autotune, while providing a standardized interface for kernel and function classes.

### MHA kernel and function modularization

* Added `Kernel` base class (`top/kernels/kernel.py`) and refactored the MHA forward kernel into a new class `mha_fwd_kernel_sm80` with default configs and autotuning support (`top/kernels/mha.py`). [[1]](diffhunk://#diff-aaffaac3959b29514c78c786315e40c2e06e9f1a61352e689b2c3d1fe430d1d8R1-R25) [[2]](diffhunk://#diff-36df7a8303a04b3ab43d8fadce67e5954229f733069f43921cdfb437b2703206R1-R157)
* Introduced `Function` base class and implemented the `mha_fwd` function class, which wraps the kernel, provides reference implementation, input generation, correctness checks, and profiling (`top/functions/function.py`, `top/functions/mha.py`). [[1]](diffhunk://#diff-0dd5bec4811cd58c851c4dba1e881a9d24e2853cc06ab975b8a97a7edb7a1f8bR1-R25) [[2]](diffhunk://#diff-eda6e0fd3387d295cb4c5272e8f67bb741c517b806c4b8be32ffb32f7d9a7c93R1-R60)

### Refactoring and import changes

* Updated package-level `__init__.py` files to use the new `kernels` and `functions` modules, simplifying imports and exposing the new API (`top/__init__.py`, `top/functions/__init__.py`, `top/kernels/__init__.py`). [[1]](diffhunk://#diff-52b943d35a75e1c19f000d326368a98ec6d7ed7e490e545bc00447ba3b069b0aL1-R19) [[2]](diffhunk://#diff-1bdca0670164c5ffc18d8c6375567892e2ec3d551c375d37fd4f998b55566308R1-R4) [[3]](diffhunk://#diff-3da25043a7d121238f449e39fe7dcabd2c10f3d245e6ffc222b91407ce6fdb32R1-R4)

### Testing and profiling

* Added a new test script `tests/v2/test_mha.py` that allows command-line configuration of MHA parameters and runs correctness and performance checks for the new kernel and function classes.

These changes lay the groundwork for future kernel and function extensions, enable easier autotuning, and provide a more robust testing and profiling workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,5 +115,14 @@ mla = MLAKernel(batch, heads, kv_heads, kv_ctx, dim, pe_dim, block_N, block_H, n
 
 out = mla(q, q_pe, kv, k_pe)
 ```
+## Design
+
+### Hierarchical Concepts
+There are 4 hierarchical key concepts in TileOPs, each layer stands for a different level of abstraction and is built from or delegates to the next lower level.
+
+- **Layer**: **Op**s that are equipped with runtime elements (e.g. Weights, Buffers, etc.) 
+- **Op**: defines computation and is implemented **Function**s for forward and backward.
+- **Function**: determines the implementation for a given shape and hardware, dispatching to the correct **Kernel** and providing unit test and benchmark.
+- **Kernel**: Tilelang-based kernels for hardware-specific logic.
 
 ## Acknowledgments


top/functions/mha.py
```diff
@@ -0,0 +1,60 @@
...
+        self.dim = dim
+        self.is_causal = is_causal
+
+        flops_per_matmul = 2.0 * batch * heads * seq_len * seq_len * dim
+        self.total_flops = 2 * flops_per_matmul
+        if is_causal:
+            self.total_flops *= 0.5
+
+        # TODO: dispatch to different kernels based on archs and inputs
+        self.kernel = mha_fwd_kernel_sm80(batch, heads, seq_len, dim, is_causal)
```
xiayuqing0622, line 24: For future kernel dispatching, it might be useful to keep a whitelist of supported architectures per function. Since TileLang kernels can typically run across multiple architectures, we could register the supported architectures for each kernel.

top/functions/mha.py
```diff
@@ -0,0 +1,60 @@
...
+        print("All checks passed.✅")
+
+    def profile(self, warmup=100, rep=100):
+        # TODO: support cupti backend for better accuracy
+        Q, K, V = self.gen_inputs()
+        with torch.no_grad():
+            tl_latency = do_bench(lambda: self.forward(Q, K, V), warmup=warmup, rep=rep)
+        
+        print(f"Tilelang latency: {tl_latency:.2f} ms")
+        print(f"Tilelang TFlops: {self.total_flops / tl_latency * 1e-9:.2f} TFlops")
```
xiayuqing0622, line 60: These three functions are primarily used for benchmarking (both correctness and performance). They should be an independent object rather than being coupled with `Function`.

top/functions/mha.py
```diff
@@ -0,0 +1,60 @@
...
+
+        flops_per_matmul = 2.0 * batch * heads * seq_len * seq_len * dim
+        self.total_flops = 2 * flops_per_matmul
+        if is_causal:
+            self.total_flops *= 0.5
+
+        # TODO: dispatch to different kernels based on archs and inputs
+        self.kernel = mha_fwd_kernel_sm80(batch, heads, seq_len, dim, is_causal)
+
+    def forward(self, Q: torch.Tensor, K: torch.Tensor, V: torch.Tensor) -> torch.Tensor:
```
xiayuqing0622, line 26: We need to add asserts (either here or in the kernel file) to check that the inputs — shape, data type, etc. — are supported. This will prevent undefined behavior and make it clear to users why a failure occurs.

top/functions/function.py
```diff
@@ -0,0 +1,25 @@
...
+class Function:
+
+    def autotune(self):
+        for attr in dir(self):
+            if isinstance(attr, Kernel) and hasattr(attr, 'autotune'):
+                attr.autotune()
+
+    def forward(self, *args, **kwargs):
+        raise NotImplementedError("forward method is not implemented")
+
```
xiayuqing0622, line 12: what about backward?

tests/v2/test_mha.py
```diff
@@ -0,0 +1,21 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--batch', type=int, default=8, help='batch size')
+    parser.add_argument('--seq_len', type=int, default=1024, help='sequence length')
+    parser.add_argument('--heads', type=int, default=32, help='num heads')
+    parser.add_argument('--dim', type=int, default=64, help='head dim')
+    parser.add_argument('--causal', action='store_true', default=False, help='causal attention')
+    args = parser.parse_args()
+    B, S, H, D, causal = args.batch, args.seq_len, args.heads, args.dim, args.causal
+
+    test_mha_kernel(B, S, H, D, causal)
```
xiayuqing0622, line 21: We could provide some common input configurations along with a script to generate the benchmark results.

top/functions/function.py
Rachmanino: As the document explained, `function` here does not have backward impl, and its `forward()` method is merely an alias of `__call__()`. We use a pair of fwd&bwd functions to compose an op(erator)

top/functions/mha.py
Rachmanino: Great. But sometimes kernels for different archs have distinct computation schedules, and we may have to dispatch in `function` if one kernel is not sufficiently compatible with all archs?

top/functions/mha.py
Rachmanino: I agree, but will it bring unnecessary runtime overhead? Or we may use an argument to control whether to perform such assertations and checks?

Issue #69, opened 2025-12-18
[BUG] Autotune failed in tilelang==0.1.7


RMLYC:
The test_autotune test fails due to tilelang version upgrade.
```bash
Traceback (most recent call last):
  File "/home/lyc/Project/TileOPs/tests/test_autotune.py", line 25, in <module>
    test_mha_kernel_autotune(args.batch, args.seq_len, args.heads, args.dim, args.causal,
  File "/home/lyc/Project/TileOPs/tests/test_autotune.py", line 8, in test_mha_kernel_autotune
    op = mha_fwd(B, H, S, D, causal, dtype, tune=True)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lyc/Project/TileOPs/top/ops/mha.py", line 33, in __init__
    self.kernel = self.kernel_map["mha_fwd_kernel"](
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lyc/Project/TileOPs/top/kernels/flash_attn/fwd.py", line 365, in __init__
    self.init_config(config, tune)
  File "/home/lyc/Project/TileOPs/top/kernels/kernel.py", line 24, in init_config
    self.autotune()
  File "/home/lyc/Project/TileOPs/top/kernels/kernel.py", line 63, in autotune
    tuned_kernel = autotuned_kernel_fn()
                   ^^^^^^^^^^^^^^^^^^^^^
  File "/home/lyc/miniconda3/envs/tileops-release/lib/python3.11/site-packages/tilelang/autotuner/tuner.py", line 680, in __call__
    artifact = autotuner.run()
               ^^^^^^^^^^^^^^^
  File "/home/lyc/miniconda3/envs/tileops-release/lib/python3.11/site-packages/tilelang/autotuner/tuner.py", line 318, in run
    assert isinstance(cell.cell_contents, (int, float, str, bool, type(None))), \
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Cell contents [8, 1024, 32, 128] is not serializable: <class 'list'>
```

lcy-seso:
Closing this issue due to prolonged inactivity. If you believe this is still relevant, please feel free to reopen it.

PR #71, opened 2025-12-19
[BugFix]: avoid closure-captured mutable shapes in flash_attn and sparse_mla kernel

RMLYC:
1. Prevent AutoTuner crash caused by mutable/closure-captured shape objects by moving shape definitions into the [@tilelang.jit] scope
2. AutoTune.run() asserts closure cell contents are serializable; mutable lists captured by the kernel factory caused assertion failures and broke autotuning/caching. Moving shapes into the JIT scope and using tuples avoids closure capture and keeps cache keys stable.
3. related #69

### DIFF
--- a/top/kernels/deepseek_mla/deepseek_dsa_decode.py
+++ b/top/kernels/deepseek_mla/deepseek_dsa_decode.py
@@ -46,7 +46,7 @@ def _sparse_mla_kernel(batch,
         sm_scale = sm_scale * 1.44269504  # log2(e)
 
     head_kv = heads // kv_group
-    ori_heads = heads
+    head_q = heads
     indices_dtype = "int32"
     accum_dtype = "float"
 
@@ -62,9 +62,9 @@ def _sparse_mla_kernel(batch,
     )
     def _sparse_mla_fwd_func(block_I, threads):
 
-        q_shape = (batch, seq_len, ori_heads, dim + tail_dim)
+        q_shape = (batch, seq_len, head_q, dim + tail_dim)
         kv_shape = (batch, seq_len_kv, kv_group, dim + tail_dim)
-        o_shape = (batch, seq_len, ori_heads, dim)
+        o_shape = (batch, seq_len, head_q, dim)
         indices_shape = (batch, seq_len, kv_group, topk)
 
         heads = head_kv


top/kernels/deepseek_mla/sparse_mla.py
```diff
@@ -59,12 +56,15 @@ def _sparse_mla_kernel(batch,
     )
     def _sparse_mla_fwd_func(block_I, threads):
 
+        q_shape = [batch, seq_len, ori_heads, dim + tail_dim]
+        kv_shape = [batch, seq_len_kv, kv_group, dim + tail_dim]
+        o_shape = [batch, seq_len, ori_heads, dim]
+        indices_shape = [batch, seq_len, kv_group, topk]
```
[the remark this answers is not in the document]
RMLYC: done

xiayuqing0622:
Lines 26–31: The comment is too long and should be reformatted into multiple lines. We should also add lint / pre-commit checks to enforce line-length limits.

Regarding the content, please split this into a concise high-level comment followed by short, focused notes or TODOs. For example:
TODO:
- ...
- ...
- ...

top/kernels/deepseek_mla/sparse_mla.py
```diff
@@ -40,10 +40,7 @@ def _sparse_mla_kernel(batch,
         sm_scale = sm_scale * 1.44269504  # log2(e)
 
     head_kv = heads // kv_group
-    q_shape = [batch, seq_len, heads, dim + tail_dim]
-    kv_shape = [batch, seq_len_kv, kv_group, dim + tail_dim]
-    o_shape = [batch, seq_len, heads, dim]
-    indices_shape = [batch, seq_len, kv_group, topk]
+    ori_heads = heads
```
xiayuqing0622: rename `ori_heads` → `head_q` for clarity.

RMLYC:
> Lines 26–31: The comment is too long and should be reformatted into multiple lines. We should also add lint / pre-commit checks to enforce line-length limits.
> 
> Regarding the content, please split this into a concise high-level comment followed by short, focused notes or TODOs. For example: TODO:
> 
> * ...
> * ...
> * ...

Which file? Currently, the code uses pre-commit for formatting checks.

top/ops/sparse_mla.py
```diff
@@ -39,9 +39,12 @@ def __init__(self,
...
         if q_start_index_s != 0:
-            assert q_start_index_s > kv_stride, "If it is because each cp has too short length, "
-            "you should fix the logic involving CP0 (cp_rank == 0), to make sure q with pos < KV_Stride - 1 is masked "
-            "(or you may just ignore how this is handled if nan in these q's Out would not effect others, which is reported to be likely to happen by wangding)"
+            assert q_start_index_s > kv_stride, (
+                "If it is because each cp has too short length, "
+                "you should fix the logic involving CP0 (cp_rank == 0) "
+                "to make sure q with pos < KV_Stride - 1 is masked. "
+                "(Or you may ignore this handling if NaN in these q's outputs "
+                "does not affect other results; this was reported by wangding.)")
```
xiayuqing0622, line 47: The assertion message is overly verbose, mixes multiple concerns, and includes informal wording and personal attribution, which makes it hard to read and not suitable for production code.

Assertions should clearly state the invariant being checked, not embed long explanations, speculative handling, or historical context. This logic also encodes a non-obvious contract between` q_start_index_s`, `kv_stride`, and CP0 masking that should be documented explicitly rather than hidden inside an assert string.
I suggest:
1. Make the assertion concise and factual. Keep the assert focused on what must be true, not why;
2. Move the explanation into a comment (or docstring). Explain why this invariant exists using structured comments, not an assert message.
```
# When q_start_index_s != 0, CP0 is expected to mask queries with
# position < kv_stride - 1. Otherwise, short CP lengths may produce
# invalid (NaN) outputs for early queries.
if q_start_index_s != 0:
    assert q_start_index_s > kv_stride, (
        "q_start_index_s must be greater than kv_stride when non-zero"
    )
```

top/kernels/deepseek_mla/deepseek_dsa_decode.py
```diff
@@ -59,12 +60,16 @@ def _sparse_mla_kernel(batch,
...
+        indices_shape = (batch, seq_len, kv_group, topk)
+
         heads = head_kv
-        # print(f'heads = {heads}')
         padded_H = max(tilelang.math.next_power_of_2(head_kv), 16)
         if padded_H != heads:
-            assert kv_group == 1, 'here we solve the heads padding automatically, other wise you should handle Q copy and Output copy with your mask (when kv_group == 1, use g_i * padded_H:(g_i+1) * padded_H would be handled automatically)'
-        # print(f'padded_H = {padded_H}, heads = {heads}')
+            assert kv_group == 1, ("Automatic heads padding is only supported when kv_group == 1. "
+                                   "Otherwise handle Q and Output copies with an appropriate mask.")
```
xiayuqing0622, line 74: The assertion message mixes a policy decision (“only supported”) with implementation guidance (“handle Q and Output copies with a mask”). This makes the intent unclear and puts design instructions inside an assert. The invariant should be stated succinctly, while the rationale and future work should be documented separately.

suggested rewrite:
```
# Automatic head padding is only valid when kv_group == 1.
# For kv_group > 1, Q and output tensors must be masked explicitly
# to handle padded heads correctly.
# TODO: Support automatic head padding for kv_group > 1
# by applying appropriate masks during Q and output copies.
if padded_H != heads:
    assert kv_group == 1, (
        "Automatic head padding requires kv_group == 1"
    )
```

top/kernels/deepseek_mla/deepseek_dsa_decode.py
```diff
@@ -23,12 +23,16 @@ def _sparse_mla_kernel(batch,
...
+
+    Note that the first kv_stride - 1 token outputs may be NaN. Since these
+    outputs are typically unused we assume this is acceptable. Still, in the
+    backward pass you may need to handle these carefully to avoid propagation
+    of NaNs (e.g. 'dout * nan'). It may be acceptable to set such NaNs to
+    zero, but we leave a reminder to handle 'delta = out * dout' appropriately.
+    The behavior may change if CP0 logic is fixed (there is a compiler-related
+    bug affecting that logic).
+    """
     assert dim == tilelang.math.next_power_of_2(
```
xiayuqing0622, line 38: The original docstring mixes background, assumptions, speculative fixes, and historical context in a single paragraph, which makes it hard to scan.
suggested rewrite:
```
"""
Implements sparse attention.

Note:
- The first (kv_stride - 1) token outputs may be NaN due to current CP0 handling.
- These outputs are typically unused and assumed to be safe to ignore.
- During the backward pass, care must be taken to prevent NaN propagation
  (e.g., in expressions like `delta = out * dout`).
- A common mitigation is to zero out invalid outputs before backward
  computation.

This behavior is a known limitation and may change once the CP0 logic
is corrected (currently affected by a compiler-related issue).
"""
```

PR #79, opened 2026-01-04
[Feat] Add native sparse attention op

### LINKED ISSUE #70: [Feature Request] DeepSeek Native Sparse Attention
https://arxiv.org/abs/2502.11089

P1

michaelwithu:
1. Implemented the nsa_fwd kernel and wrapped it according to the TileOps interface hierarchy.
2. Developed a mean pooling kernel using TileLang, and verified its correctness by comparing outputs against the Triton implementation.

resolve https://github.com/tile-ai/TileOPs/issues/70

todo：

- [ ] Remove temporary/test-only code that will be invalid in the final implementation  
- [ ] Integrate mean pooling kernel into the TileOp interface layer  
- [ ] Reuse FlashAttention’s infrastructure to support compression and sliding window attention functionality  
- [ ] Implement the complete NSA pipeline


top/kernels/deepseek_nsa/nsa_torch.py
```diff
@@ -0,0 +1,380 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2023-2025, Songlin Yang, Yu Zhang
```
lcy-seso: I’m curious why these implementations live in the tileops repo.
What are they used for in this context?
Is there a reason we don’t depend on the original project directly?

PR #123, opened 2026-01-12
[Bug] Fix the bug where the nvcc version is not equal to 12.9 during CI runtime.

RMLYC:
## Description
During CI runtime, the nvcc version is required to be 12.9, but it is actually running version 12.4.

## Type of Change

- [x] Bug fix
- [ ] New operator implementation
- [ ] Performance improvement
- [ ] Documentation update
- [ ] Infrastructure/CI

## Checklist

- [ ] I have run `pre-commit run --all-files` and fixed all linting issues.
- [ ] I have verified that my changes pass local unit tests.
- [ ] **(For new ops)** I have added the corresponding `Benchmark` class in `benchmarks/`.
- [ ] **(For new ops)** I have reported benchmark results in the tracking issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,7 @@ jobs:
           conda activate tileops-release
           export PYTHONPATH="$(pwd):$PYTHONPATH"
           echo "PYTHONPATH=$PYTHONPATH"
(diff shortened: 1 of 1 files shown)


RMLYC:
This bug will be fixed in other pr

PR #225, opened 2026-02-27
[CI] Validate GPU frequency before GPU jobs

### LINKED ISSUE #216: [CI] CI should verify GPU frequency to ensure correct execution
## Summary

During CI runs, we should check the GPU frequency before executing GPU-dependent tests/workloads.

## Motivation

GPU frequency can affect runtime behavior and result correctness/stability in some workloads. If the frequency is not at the expected level (for example, due to power management, throttling, or environment misconfiguration), CI may produce unreliable results or false failures/passes.

## Proposed change

Add a CI pre-check step that:

- reads the current GPU frequency (or relevant clock information),
- validates it against the expected range/configuration,
- fails early (with clear logs) when the GPU frequency is not suitable for the test run.
- GPU frequency should be lock to 1830MHz

## Expected benefit

This will improve CI reliability and help ensure that performance-sensitive or hardware-dependent tests run under valid GPU conditions.

RMLYC:
Closes #216

## Summary
- add a GPU pre-check step in CI GPU jobs to read current graphics clock
- validate all detected GPUs are running at 1830 MHz with retries and clear per-GPU logs
- fail early when GPU frequency is not 1830 MHz to avoid unreliable test runs

## Test plan
- [x] PRE_COMMIT_HOME=/tmp/pre-commit-cache pre-commit run --all-files
- [x] GPU frequency check execution on self-hosted runner (to be validated in CI)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,50 @@ jobs:
         with:
           python-version: "3.11"
 
+      - name: Validate GPU frequency
+        run: |
+          set -euo pipefail
+          TARGET_CLOCK_MHZ=1830
+          RETRIES=5
+          SLEEP_SECONDS=2
+
+          if ! command -v nvidia-smi >/dev/null 2>&1; then
+            echo "::error::nvidia-smi is not available on this runner."
+            exit 1
+          fi
+
+          echo "Detected GPUs:"
+          nvidia-smi -L
+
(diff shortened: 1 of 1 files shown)


xZacky:
Checking GPU frequency before testing is not feasible, because after frequency locking, if there is no workload on the GPU, the frequency will still remain very low.

PR #480, opened 2026-03-13
[CI] Configure persistent compile caches in self-hosted CI

### LINKED ISSUE #465: [PERF][CI] explicitly configure compile cache directories for self-hosted ci
### Symptom / Motivation
TileOPs GPU CI jobs still pay repeated TileLang and Triton compilation cost even when the self-hosted runner already has persistent cache directories under `/home/ci-runner/.tilelang/cache` and `/home/ci-runner/.triton/cache`. The current workflows do not make cache usage explicit or validate that the test process resolves the expected cache paths at runtime.

### Root Cause Analysis
The CI workflows rely on implicit default cache resolution instead of exporting the cache-related environment variables in the job environment. That makes cache behavior harder to verify from logs and easier to regress when `HOME`, process environment, or workflow structure changes. In addition, TileLang documents `TILELANG_CACHE_DIR` as the compile cache root and `TILELANG_TMP_DIR` as a temp directory under that root, but the workflows do not currently pin those variables explicitly.

### Related Files
- `.github/workflows/ci.yml`
- TileLang compile cache environment variables: `TILELANG_CACHE_DIR`, `TILELANG_TMP_DIR`
- Triton compile cache environment variable: `TRITON_CACHE_DIR`

## Goal
Make self-hosted CI jobs use explicit and verifiable compile cache directories so repeated PR and nightly runs can reuse existing TileLang and Triton compilation artifacts.

## Plan

1. Add a workflow step in GPU CI jobs to export explicit cache environment variables before tests run.
2. Pin TileLang cache paths to `/home/ci-runner/.tilelang/cache` and `/home/ci-runner/.tilelang/cache/tmp`.
3. Pin Triton cache path to `/home/ci-runner/.triton/cache`.
4. Optionally pin related PyTorch JIT cache directories to stable runner-local paths.
5. Print resolved cache directories in CI logs so cache behavior is observable and regressions are easier to diagnose.
6. Update both `.github/workflows/ci.yml` and `.github/workflows/nightly.yml` consistently.

## Constraints
- Do not add a separate cache warmup workflow.
- Keep the solution compatible with the existing self-hosted runner layout under `/home/ci-runner`.
- Prefer explicit environment variables over wrapper scripts or implicit `HOME`-based behavior.

## Acceptance Criteria
- [ ] `.github/workflows/ci.yml` exports explicit TileLang and Triton cache directories before GPU tests run.
- [ ] `.github/workflows/nightly.yml` exports the same cache directories before benchmark and packaging test jobs run.
- [ ] CI logs print the resolved cache directories seen by the Python process.

RMLYC:
Closes #465

## Summary

- explicitly pin TileLang and Triton compile cache directories for self-hosted CI jobs
- print the runtime cache environment in CI, benchmark, and packaging test paths

## Test plan

- [x] pre-commit run --files .github/workflows/ci.yml .github/workflows/nightly.yml
- [x] git diff --cached --check
- [x] GitHub Actions not run locally

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,7 +265,7 @@ jobs:
         run: |
           set -euo pipefail
 
-          VENV_PREFIX="tileops_release_venv"
+          VENV_PREFIX="tileops_ci_venv"
           HASH_INPUT="$(
             {
               echo "python=3.11"
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -99,7 +99,7 @@ jobs:
         run: |
           set -euo pipefail
 
-          VENV_PREFIX="tileops_benchmark_venv"
+          VENV_PREFIX="tileops_ci_venv"
           HASH_INPUT="$(
             {
               echo "python=3.11"
@@ -120,6 +120,17 @@ jobs:
           else
             echo "No matching venv found. Will create: ${VENV_PATH}"
             echo "VENV_REUSED=false" >> "$GITHUB_ENV"
+
+            for old_venv in "${{ runner.tool_cache }}/${VENV_PREFIX}"_*; do
+              if [ ! -d "${old_venv}" ]; then
+                continue
+              fi
+              if [ "${old_venv}" = "${VENV_PATH}" ]; then
+                continue
+              fi
+              echo "Removing stale venv: ${old_venv}"
+              rm -rf "${old_venv}"
+            done
           fi
         shell: bash
 
@@ -270,7 +281,7 @@ jobs:
         run: |
           set -euo pipefail
 
-          VENV_PREFIX="tileops_wheel_ci_venv"
+          VENV_PREFIX="tileops_packaging_venv"
           HASH_INPUT="$(
             {
               echo "python=3.11"
@@ -291,6 +302,17 @@ jobs:
           else
             echo "No matching venv found. Will create: ${VENV_PATH}"
             echo "VENV_REUSED=false" >> "$GITHUB_ENV"
+
+            for old_venv in "${{ runner.tool_cache }}/${VENV_PREFIX}"_*; do
+              if [ ! -d "${old_venv}" ]; then
+                continue
+              fi
+              if [ "${old_venv}" = "${VENV_PATH}" ]; then
+                continue
+              fi
+              echo "Removing stale venv: ${old_venv}"
+              rm -rf "${old_venv}"
+            done
           fi
         shell: bash
 
@@ -306,7 +328,8 @@ jobs:
           python -m venv "${VENV_PATH}"
           # shellcheck source=/dev/null
           source "${VENV_PATH}/bin/activate"
-          python -m pip install --upgrade pip setuptools wheel
+          python -m pip install --upgrade pip setuptools wheel --no-user
+          PIP_NO_BUILD_ISOLATION=1 pip install -e '.[dev]'
         shell: bash
 
       - name: Install required dependencies
@@ -317,7 +340,7 @@ jobs:
             echo "Cached packaging venv already prepared at ${VENV_PATH}"
             exit 0
           fi
-          pip install build pytest
+          echo "Packaging dependencies already installed via .[dev]"
         shell: bash
 
       - name: Build wheel package


RMLYC:
Verification update based on CI results for this PR:

The venv-reuse experiment is effective.

Comparison on this PR:
- Before (`run 23040114422`, job `66916439005`, runner `tile-ops-mig-1`)
  - `Ensure persistent venv and install dependencies`: about 1m33s
  - `Run tests`: about 17m44s
- After (`run 23043318153`, job `66928492385`, runner `tile-ops-mig-0`)
  - `Resolve persistent venv`: about 0s
  - `Ensure persistent venv and install dependencies`: about 0s
  - `Run tests`: about 53s

This shows the dependency-hash venv reuse is working and the test runtime improvement is significant once the runner hits an existing matching environment.

Current observation:
- The main remaining bottleneck in the successful rerun is now `Checkout code` (~11m13s), not test execution.
- `tile-ops-mig-0` and `tile-ops-mig-1` are runner-local and independent, so first-hit behavior can still differ by runner.

.github/workflows/ci.yml
```diff
@@ -239,26 +261,63 @@ jobs:
...
+          echo "VENV_DIR=${VENV_DIR}" >> "$GITHUB_ENV"
+          echo "VENV_PATH=${VENV_PATH}" >> "$GITHUB_ENV"
+
+          if [ -x "${VENV_PATH}/bin/python" ]; then
+            echo "Reusing existing venv: ${VENV_PATH}"
+            echo "VENV_REUSED=true" >> "$GITHUB_ENV"
+          else
+            echo "No matching venv found. Will create: ${VENV_PATH}"
+            echo "VENV_REUSED=false" >> "$GITHUB_ENV"
+          fi
```
lcy-seso, line 289: Stale venvs with the same prefix but a different hash are never cleaned up. Each `pyproject.toml` change leaves an orphaned 1-2 GB directory. Consider adding GC after resolution:

```bash
# Remove stale venvs with same prefix but different hash
for old_venv in "${{ runner.tool_cache }}"/${VENV_PREFIX}_*; do
  if [ -d "$old_venv" ] && [ "$old_venv" != "${VENV_PATH}" ]; then
    echo "Removing stale venv: $old_venv"
    rm -rf "$old_venv"
  fi
done
```

Same applies to the two venv resolve steps in `nightly.yml`.

.github/workflows/ci.yml
```diff
@@ -195,6 +195,28 @@ jobs:
         with:
           python-version: "3.11"
 
+      - name: Configure persistent compile caches
```
lcy-seso, line 198: This 22-line block is duplicated verbatim in all 3 jobs (here + benchmark + packaging in `nightly.yml`). The "Resolve persistent venv" step (~28 lines) is similarly duplicated 3x with only the prefix and install command differing.

Consider extracting both into composite actions:
- `.github/actions/configure-caches/action.yml`
- `.github/actions/resolve-venv/action.yml` (parameterized by `prefix` and `install-command`)

This would single-source the hash logic and make the GC fix above a one-place change.

.github/workflows/nightly.yml
```diff
@@ -29,6 +29,28 @@ jobs:
         with:
           python-version: "3.11"
 
+      - name: Configure persistent compile caches
```
lcy-seso, line 32: `TILELANG_CACHE_DIR` and `TRITON_CACHE_DIR` grow indefinitely as kernels evolve. Since nightly already runs daily, consider adding a periodic eviction step:

```bash
# Evict compile cache entries not accessed in 30 days
find "${TILELANG_CACHE_DIR}" -atime +30 -type f -delete 2>/dev/null || true
find "${TRITON_CACHE_DIR}" -atime +30 -type f -delete 2>/dev/null || true
```

RMLYC: Agreed that the minimal implementation can be as small as the two `find ... -delete` lines you suggested. I am intentionally not folding it into this PR because the open question is policy rather than code size: whether we should use `atime` or another signal, what retention window is safe, and where nightly eviction should live without regressing the warm-cache speedup that this PR restores.

I split that follow-up into a dedicated issue so the eviction policy can be decided and validated independently of the already-verified venv reuse fix: #486.

PR #656, opened 2026-03-24
[Perf][Conv2d] Generalize Conv2dKernel dispatch and add 5x5 coverage

### LINKED ISSUE #434: [New Op] Conv2d
## Operator Description

Implement the forward `Conv2d` operator in TileOPs.

For input tensor `x \in R^{N \times C_in \times H \times W}`, weight tensor `w \in R^{C_out \times C_in \times K_h \times K_w}`, and optional bias `b \in R^{C_out}`, the output tensor `y \in R^{N \times C_out \times H_out \times W_out}` is defined as:

`y[n, c_out, h, w] = b[c_out] + sum_{c_in, kh, kw} x[n, c_in, h * stride_h + kh - pad_h, w * stride_w + kw - pad_w] * weight[c_out, c_in, kh, kw]`

This issue tracks standard dense `Conv2d` forward delivery as the first convolution operator in the broader conv and pooling family tracked by #402. The implementation should cover the common PyTorch `nn.Conv2d` forward semantics for stride, padding, and optional bias, but does not need to support dilation in this issue. Related tracking context: #402. Related implementation-pattern reference: #407. PyTorch operator semantics reference: https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html.

## Related Files

- `tileops/kernels/conv2d/` — add: TileLang kernel implementation for dense forward Conv2d
- `tileops/ops/conv2d.py` — add: `Op` wrapper and public operator interface for Conv2d
- `tileops/ops/__init__.py` — modify: export the new Conv2d op class through the package surface
- `tests/ops/test_conv2d.py` — add: correctness coverage against the PyTorch Conv2d reference across representative parameter combinations
- `benchmarks/ops/bench_conv2d.py` — add: latency / throughput / bandwidth benchmarking on representative workloads

### 1. Kernel Implementation (L1)

Implement the TileLang forward kernel under `tileops/kernels/conv2d/` for dense `groups=1` Conv2d.

- [ ] **Kernel**: Implement the forward Conv2d kernel in `tileops/kernels/conv2d/`
  - [ ] Support standard forward `Conv2d` semantics with configurable stride, padding, and optional bias
  - [ ] Explicitly exclude dilation support from this issue scope
  - [ ] Restrict scope to dense `groups=1` forward Conv2d; exclude transposed, grouped, and depthwise variants from this issue
- [ ] **Verification**: Pass functional correctness checks against the agreed reference implementation
  - [ ] Validate representative `NCHW` workloads such as ResNet-style `3x3` and `1x1` convolutions

### 2. Op Definition (L2)

Add the operator wrapper, test coverage, and benchmark entry needed to expose and validate Conv2d end to end.

- [ ] **Interface**: Define the `Op` wrapper and public interface in `tileops/ops/conv2d.py`
  - [ ] Update package exports in `tileops/ops/__init__.py`
- [ ] **Unit Tests**: Implement `tests/ops/test_conv2d.py` and compare against the PyTorch reference
  - [ ] Cover FP16 and BF16
  - [ ] Cover combinations of kernel size, stride, padding, and bias/no-bias
  - [ ] Exclude dilation test cases from this issue scope
- [ ] **Benchmarks**: Implement `benchmarks/ops/bench_conv2d.py`
  - [ ] Report latency, TFLOPS, and DRAM bandwidth on representative workloads
  - [ ] Compare performance against the PyTorch `nn.Conv2d` baseline

## Goal

- Data types: FP16, BF16
- Maximum relative error: FP16 `1e-3`, BF16 `1.6e-2`
- Maximum absolute error: FP16 `1e-3`, BF16 `1.6e-2`
- Accuracy reference implementation: PyTorch `torch.nn.Conv2d` with dilation disabled
- Speed reference implementation: PyTorch `torch.nn.Conv2d` with dilation disabled
- Speed reference repository: https://github.com/pytorch/pytorch
- Other tracked metrics: TFLOPS, DRAM bandwidth

## Acceptance Criteria

- [ ] The dense forward Conv2d kernel is added under `tileops/kernels/conv2d/` and matches PyTorch `torch.nn.Conv2d` semantics for stride, padding, and optional bias
- [ ] Dilation is explicitly out of scope for this issue and is not required by the kernel, interface, tests, or benchmarks
- [ ] Kernel correctness is verified on representative `NCHW` workloads for both FP16 and BF16 within the confirmed relative/absolute error tolerances
- [ ] The Conv2d op wrapper is added in `tileops/ops/conv2d.py` and exported through `tileops/ops/__init__.py` without introducing a new `Op` pattern
- [ ] `tests/ops/test_conv2d.py` covers the agreed parameter combinations and passes against the PyTorch reference implementation
- [ ] `benchmarks/ops/bench_conv2d.py` reports latency, TFLOPS, and DRAM bandwidth, and includes comparison against the PyTorch baseline
- [ ] The issue references and related-file mapping are sufficient for a follow-up implementation pass without additional clarification

## Constraints

- Must follow the `Op` base-class pattern — no new patterns
- Must not change any existing public API or break existing tests
- Kernel must be written in Tile-lang (consistent with all other TileOPs kernels)
- CUDA only; CPU support is out of scope

## Reference

- PyTorch `Conv2d` API: https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html
- PyTorch repository: https://github.com/pytorch/pytorch
- TileOPs family tracking for convolution and pooling operators: #402
- Related TileOPs implementation-planning example: #407

RMLYC:
Close #434 

## Summary

- unify conv2d dispatch so `1x1 + stride=1 + pad=0` uses `Conv2d1x1Kernel`, and all other cases fall back to `Conv2dKernel`
- rename the generic NxN path to `_conv2d_kernel` / `Conv2dKernel` and update upstream/downstream interfaces accordingly
- keep the public Conv2d input layout as NHWC and align kernel wrappers with that contract
- add representative 5x5 coverage to unit tests and benchmarks
- remove `groups` and `dilation` from `Conv2dOp`; they are not planned for support in this op

## Test plan

- [x] `python -m pytest tests/ops/test_conv2d.py -vvs`
- [x] `python -m pytest benchmarks/ops/bench_conv2d.py -vvs`
- [x] Verified benchmark report generated on NVIDIA H200 with Torch 2.9.0+cu128
- [ ] `scripts/validate.sh --pre`
  `scripts/validate.sh` is not present in this repository

## Benchmark

Latest benchmark report: `profile_run.log` on NVIDIA H200

| Case | TileOPs TFLOPS | Torch TFLOPS | Relative to Torch |
| --- | ---: | ---: | ---: |
| resnet-3x3-fp16 | 51.79 | 13.81 | 3.75x |
| stem-3x3-s2-fp16 | 1.37 | 0.55 | 2.49x |
| stage-transition-3x3-s2-fp16 | 37.53 | 12.72 | 2.95x |
| highres-3x3-s1-fp16 | 458.53 | 91.34 | 5.02x |
| midres-5x5-s1-fp16 | 96.25 | 34.49 | 2.79x |
| stage-transition-5x5-s2-fp16 | 54.91 | 29.19 | 1.88x |
| stride2-bf16 | 5.28 | 2.09 | 2.53x |
| resnet-1x1-fp16 | 44.60 | 5.19 | 8.59x |
| bottleneck-expand-1x1-fp16 | 53.52 | 8.56 | 6.25x |
| bottleneck-reduce-1x1-fp16 | 39.65 | 8.37 | 4.74x |
| late-stage-1x1-fp16 | 23.96 | 4.94 | 4.85x |
| classifier-1x1-fp16 | 20.99 | 4.52 | 4.64x |
| resnet-1x1-bf16 | 46.88 | 5.21 | 9.00x |

## Regression

- existing conv2d unit coverage passes after dispatch changes
- `test_conv2d_does_not_dispatch_1x1_kernel_with_padding` now verifies fallback to `Conv2dKernel` instead of raising
- zero-bias and 3x3 / 5x5 dispatch behavior are covered in `tests/ops/test_conv2d.py`

## Additional context

- `Conv2dKernel` is the generic path for non-fast-path cases; `Conv2d1x1Kernel` remains the specialized fast path
- this PR keeps the operator focused on the supported contract and removes unused `groups` / `dilation` API surface

### DIFF
--- a/tileops/kernels/conv2d/conv2d.py
+++ b/tileops/kernels/conv2d/conv2d.py
@@ -85,21 +85,18 @@ def _conv2d_1x1_kernel(
                 for i, j in T.Parallel(block_m, block_n):
                     m_idx = by * block_m + i
                     oc = bx * block_n + j
-                    out_shared[i, j] = T.if_then_else(
-                        (m_idx < hw) & (oc < c_out),
-                        T.cast(
-                            out_local[i, j] + T.if_then_else(
-                                has_bias,
-                                T.cast(
-                                    T.if_then_else(oc < c_out, bias[oc], T.cast(0.0, dtype)),
-                                    accum_dtype,
-                                ),
-                                T.cast(0.0, accum_dtype),
-                            ),
-                            dtype,
-                        ),
-                        T.cast(0.0, dtype),
-                    )
+                    if has_bias:
+                        out_shared[i, j] = T.if_then_else(
+                            (m_idx < hw) & (oc < c_out),
+                            T.cast(out_local[i, j] + T.cast(bias[oc], accum_dtype), dtype),
+                            T.cast(0.0, dtype),
+                        )
+                    else:
+                        out_shared[i, j] = T.if_then_else(
+                            (m_idx < hw) & (oc < c_out),
+                            T.cast(out_local[i, j], dtype),
+                            T.cast(0.0, dtype),
+                        )
 
                 T.copy(out_shared, out_flat[bz, by * block_m, bx * block_n])
 
@@ -205,21 +202,18 @@ def _conv2d_kernel(
                 for i, j in T.Parallel(block_m, block_n):
                     m_idx = by * block_m + i
                     oc = bx * block_n + j
-                    out_shared[i, j] = T.if_then_else(
-                        (m_idx < n * out_h * out_w) & (oc < c_out),
-                        T.cast(
-                            out_local[i, j] + T.if_then_else(
-                                has_bias,
-                                T.cast(
-                                    T.if_then_else(oc < c_out, bias[oc], T.cast(0.0, dtype)),
-                                    accum_dtype,
-                                ),
-                                T.cast(0.0, accum_dtype),
-                            ),
-                            dtype,
-                        ),
-                        T.cast(0.0, dtype),
-                            )
+                    if has_bias:
+                        out_shared[i, j] = T.if_then_else(
+                            (m_idx < n * out_h * out_w) & (oc < c_out),
+                            T.cast(out_local[i, j] + T.cast(bias[oc], accum_dtype), dtype),
+                            T.cast(0.0, dtype),
+                        )
+                    else:
+                        out_shared[i, j] = T.if_then_else(
+                            (m_idx < n * out_h * out_w) & (oc < c_out),
+                            T.cast(out_local[i, j], dtype),
+                            T.cast(0.0, dtype),
+                        )
 
                 if use_hopper_im2col:
                     T.copy(out_shared, out_flat[by * block_m, bx * block_n])
@@ -334,17 +328,6 @@ def _(
     out_w = (w + 2 * pad_w - kernel_w) // stride_w + 1
     return torch.empty((n, out_h, out_w, c_out), dtype=inputs[0].dtype, device=inputs[0].device)
 
-
-def _conv2d_shared_memory_for_nxn(
-    block_m: int,
-    block_n: int,
-    block_k: int,
-    num_stages: int,
-    dtype: torch.dtype,
-) -> int:
-    return _conv2d_shared_memory_bytes(block_m, block_n, block_k, num_stages, dtype)
-
-
 class Conv2dKernel(Kernel):
     supported_archs: list[int] = [80, 86, 89, 90]
 
@@ -444,7 +427,7 @@ class Conv2dKernel(Kernel):
         )
         valid_configs = []
         for block_m, block_n, block_k, num_stages, threads, enable_rasteration in configs:
-            shared_memory_bytes = _conv2d_shared_memory_for_nxn(
+            shared_memory_bytes = _conv2d_shared_memory_bytes(
                 block_m, block_n, block_k, num_stages, self.dtype)
             if shared_memory_bytes > _HOPPER_SHARED_MEMORY_LIMIT_BYTES:
                 continue
@@ -466,6 +449,7 @@ class Conv2dKernel(Kernel):
     ) -> torch.Tensor:
         if bias is None:
             bias = torch.zeros(self.c_out, device=x.device, dtype=x.dtype)
+        # OIHW -> HWIO to match the kernel layout (kernel_h, kernel_w, c_in, c_out).
         weight_hwcf = weight.permute(2, 3, 1, 0).contiguous()
         return _conv2d_wrapped_kernel(
             self.n,


tileops/ops/conv2d.py
```diff
@@ -0,0 +1,90 @@
...
+            and self.padding == (0, 0)
+            and "conv2d_1x1_kernel" in self.kernel_map
+        ):
+            kernel_name = "conv2d_1x1_kernel"
+        elif "conv2d_kernel" in self.kernel_map:
+            kernel_name = "conv2d_kernel"
+        if kernel_name == "conv2d_kernel":
+            kernel_kwargs["kernel_h"] = self.kernel_size[0]
+            kernel_kwargs["kernel_w"] = self.kernel_size[1]
+        self.kernel = self.kernel_map[kernel_name](**kernel_kwargs)
```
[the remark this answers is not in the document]
RMLYC: Fixed in commit d29736d. I removed the temporary `kernel_name` flow and now assign `self.kernel` directly inside the dispatch branches. If neither `conv2d_1x1_kernel` nor `conv2d_kernel` is present in a custom `kernel_map`, `Conv2dOp` now raises an explicit `NotImplementedError` instead of falling through to a `KeyError`.

benchmarks/ops/bench_conv2d.py
```diff
@@ -0,0 +1,89 @@
+from typing import Optional, Tuple
+
+import pytest
+import torch
+
+from benchmarks.benchmark import BenchmarkBase, BenchmarkReport
+from tests.ops.test_conv2d import Conv2dTest
```
lcy-seso: Benchmark imports `Conv2dTest` from `tests.ops.test_conv2d`. This couples benchmarks to the test module — if test fixtures are refactored, benchmarks silently break. Consider extracting the shared shape struct to a common location or duplicating the minimal definition here. Low priority.

tileops/kernels/conv2d/conv2d.py
```diff
@@ -0,0 +1,621 @@
...
+    block_n: int,
+    block_k: int,
+    num_stages: int,
+    threads: int,
+    enable_rasteration: bool,
+    x: torch.Tensor,
+    weight: torch.Tensor,
+    bias: torch.Tensor,
+) -> torch.Tensor:
+    return _conv2d_kernel(
```
superAngGao, line 299: Wanted to flag a potential performance issue here.

Each `forward()` call goes through this custom_op, which calls `_conv2d_kernel(params)(config)(x, w, b)` — that re-creates the `@tilelang.jit` closure, re-hashes, and hits the cache lookup every single invocation. I did a quick local test on H200 and the overhead is roughly **~60ms per call**:

- Direct call to pre-compiled `JITKernel` → **0.01ms** (45.9 TFLOPS)
- Via `_conv2d_wrapped_kernel` → **63ms** (0.01 TFLOPS)

Same issue applies to `_conv2d_1x1_wrapped_kernel` (line 235).

I noticed there's a `fix/jitimpl-lru-cache` branch in the repo — is that already addressing this? Adding `@functools.lru_cache` on `_conv2d_kernel` and `_conv2d_1x1_kernel` should fix it. What do you think?

```python
@functools.lru_cache(maxsize=32)
def _conv2d_kernel(n, c_in, h, w, c_out, kernel_h, kernel_w, ...):
    ...
```

benchmarks/ops/bench_conv2d.py
```diff
@@ -0,0 +1,89 @@
...
+        stride=stride,
+        padding=padding,
+        bias=True,
+        dtype=dtype,
+        tune=tune,
+    )
+    result = bm.profile(op, *inputs, warmup=5, rep=10)
+    BenchmarkReport.record("conv2d", locals(), result, tag="tileops")
+
+    result_bl = bm.profile(test.ref_program, *inputs, warmup=5, rep=10)
```
superAngGao: The current baseline uses `test.ref_program`, which includes two `.permute().contiguous()` calls (NHWC→NCHW before conv, NCHW→NHWC after). Those layout copies get counted toward the baseline latency, but TileOPs is native NHWC with zero conversion, so the comparison might be a bit skewed.

I ran a few extra baselines on the same H200 (`cudnn.benchmark=True`, pre-compiled TileOPs kernels):

| Case | TileOPs | torch-nchw | torch-nhwc | vs nchw | vs nhwc | vs perm |
|---|---:|---:|---:|---:|---:|---:|
| resnet-3x3-fp16 | 32.26 | 15.95 | 28.50 | 2.02x | 1.13x | 2.64x |
| highres-3x3-s1-fp16 | 212.61 | 304.02 | 438.11 | 0.70x | 0.49x | 1.14x |
| bottleneck-expand-1x1 | 24.80 | 14.90 | 13.61 | 1.66x | 1.82x | 3.29x |
| resnet-1x1-bf16 | 18.78 | 12.69 | 13.38 | 1.48x | 1.40x | 3.30x |

A few observations:
- **1x1 kernels genuinely win** — 1.2x–1.8x over cuDNN NHWC, solid improvement
- **NxN small-spatial** — ahead vs NCHW, roughly on par with cuDNN NHWC
- **highres-3x3 (large spatial + large channel)** — cuDNN NHWC is actually ~2x faster, likely Winograd advantage

Would it make sense to add `torch-nchw` and `torch-nhwc` (channels_last) baselines alongside the current one? That way the speedup numbers are more transparent and can also help guide where to focus optimization next.

tileops/kernels/conv2d/conv2d.py
```diff
@@ -0,0 +1,621 @@
...
+
+                T.use_swizzle(10, enable=enable_rasteration)
+                T.clear(out_local)
+
+                for k_iter in T.Pipelined(T.ceildiv(c_in, block_k), num_stages=num_stages):
+                    T.copy(x_flat[bz, by * block_m, k_iter * block_k], data_shared)
+                    T.copy(weight[bx * block_n, k_iter * block_k], weight_shared)
+                    T.gemm(data_shared, weight_shared, out_local, transpose_B=True)
+
+                for i, j in T.Parallel(block_m, block_n):
```
superAngGao, line 85: Quick question: `has_bias` is a Python-level closure constant, but the kernel evaluates it at runtime via `T.if_then_else(has_bias, ...)`. Does tilelang constant-fold this during codegen? If so, no worries. If not, it might be worth splitting into with/without-bias kernel variants at the Python level to avoid the per-element branch.

tileops/kernels/conv2d/conv2d.py
```diff
@@ -0,0 +1,621 @@
...
+                            )
+                            data_shared[i, j] = T.if_then_else(
+                                in_bound,
+                                x[batch, ih, iw, ci],
+                                T.cast(0.0, dtype),
+                            )
+
+                    T.copy(weight_flat[k_iter * block_k, bx * block_n], weight_shared)
+
+                    T.gemm(data_shared, weight_shared, out_local)
```
superAngGao, line 200: It looks like the bounds check happens twice here — once when writing to `out_shared` (line 200–216), and again when copying from `out_shared` to `out_flat` (line 222–226). The second check alone should be sufficient since stray zero values in shared memory are harmless.

Not sure if tilelang optimizes this away automatically — curious to hear your take.

tileops/kernels/conv2d/conv2d.py
```diff
@@ -0,0 +1,621 @@
...
+    out_h = (h + 2 * pad_h - kernel_h) // stride_h + 1
+    out_w = (w + 2 * pad_w - kernel_w) // stride_w + 1
+    return torch.empty((n, out_h, out_w, c_out), dtype=inputs[0].dtype, device=inputs[0].device)
+
+
+def _conv2d_shared_memory_for_nxn(
+    block_m: int,
+    block_n: int,
+    block_k: int,
+    num_stages: int,
```
superAngGao: nit: This function just forwards to `_conv2d_shared_memory_bytes` with the same signature and no extra logic. Is this reserved for future extension, or could we use `_conv2d_shared_memory_bytes` directly?

tileops/kernels/conv2d/conv2d.py
```diff
@@ -0,0 +1,621 @@
...
+            })
+        return valid_configs
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        weight: torch.Tensor,
+        bias: Optional[torch.Tensor] = None,
+    ) -> torch.Tensor:
+        if bias is None:
```
superAngGao, line 450: nit: The permute here is OIHW→HWIO, which is correct, but it's easy to misread as HWOI at first glance. A short comment might help readability:

```python
# OIHW -> HWIO to match kernel layout (kernel_h, kernel_w, c_in, c_out)
weight_hwio = weight.permute(2, 3, 1, 0).contiguous()
```

benchmarks/ops/bench_conv2d.py
RMLYC: Addressed in 76c6ebb. I removed the benchmark dependency on `tests.ops.test_conv2d` by adding a local `Conv2dBenchCase` in `benchmarks/ops/bench_conv2d.py` and moving the minimal `gen_inputs()` / `ref_program()` logic there. This keeps the benchmark self-contained while preserving the same input generation and baseline path.

benchmarks/ops/bench_conv2d.py
RMLYC: Addressed in 1c61e18. I removed the old permute-based baseline and split the benchmark into two explicit references: `torch-nchw` and `torch-nhwc` (`channels_last`), with layout preparation moved outside the timed region. I also ran the full `bench_conv2d` suite on H200 with the new setup; TileOPs is ahead on all 13 current cases, with average TFLOPS ratios of about 2.43x vs `torch-nchw` and 2.10x vs `torch-nhwc`.

tileops/kernels/conv2d/conv2d.py
RMLYC: Addressed in 1c61e18. I added `@functools.lru_cache(maxsize=32)` to both `_conv2d_kernel` and `_conv2d_1x1_kernel`, so repeated calls now reuse the tilelang builder instead of reconstructing the JIT closure on every wrapper invocation.

benchmarks/ops/bench_conv2d.py
RMLYC: Adding the per-case ratio table from the full `bench_conv2d` run on H200 for reference (TFLOPS-based):

| Case | tileops / torch-nchw | tileops / torch-nhwc |
|---|---:|---:|
| resnet-3x3-fp16 | 2.89x | 2.15x |
| stem-3x3-s2-fp16 | 2.01x | 2.74x |
| stage-transition-3x3-s2-fp16 | 2.16x | 1.16x |
| highres-3x3-s1-fp16 | 1.86x | 1.64x |
| midres-5x5-s1-fp16 | 2.17x | 1.66x |
| stage-transition-5x5-s2-fp16 | 1.44x | 1.12x |
| stride2-bf16 | 2.04x | 1.60x |
| resnet-1x1-fp16 | 3.07x | 3.05x |
| bottleneck-expand-1x1-fp16 | 2.84x | 2.80x |
| bottleneck-reduce-1x1-fp16 | 1.71x | 1.94x |
| late-stage-1x1-fp16 | 3.15x | 2.34x |
| classifier-1x1-fp16 | 3.21x | 1.91x |
| resnet-1x1-bf16 | 3.00x | 3.23x |

Average across all 13 cases: 2.43x vs `torch-nchw`, 2.10x vs `torch-nhwc`.

RMLYC: Adding the full H200 `bench_conv2d` summary with both absolute TFLOPS and relative ratios:

| Case | TileOPs TFLOPS | torch-nchw TFLOPS | torch-nhwc TFLOPS | tileops / nchw | tileops / nhwc |
|---|---:|---:|---:|---:|---:|
| resnet-3x3-fp16 | 51.31 | 17.77 | 23.91 | 2.89x | 2.15x |
| stem-3x3-s2-fp16 | 1.45 | 0.72 | 0.53 | 2.01x | 2.74x |
| stage-transition-3x3-s2-fp16 | 36.72 | 17.02 | 31.69 | 2.16x | 1.16x |
| highres-3x3-s1-fp16 | 330.21 | 177.13 | 201.40 | 1.86x | 1.64x |
| midres-5x5-s1-fp16 | 89.86 | 41.42 | 54.01 | 2.17x | 1.66x |
| stage-transition-5x5-s2-fp16 | 47.27 | 32.92 | 42.29 | 1.44x | 1.12x |
| stride2-bf16 | 4.95 | 2.43 | 3.10 | 2.04x | 1.60x |
| resnet-1x1-fp16 | 37.83 | 12.32 | 12.39 | 3.07x | 3.05x |
| bottleneck-expand-1x1-fp16 | 41.42 | 14.58 | 14.80 | 2.84x | 2.80x |
| bottleneck-reduce-1x1-fp16 | 31.00 | 18.11 | 15.94 | 1.71x | 1.94x |
| late-stage-1x1-fp16 | 20.70 | 6.58 | 8.84 | 3.15x | 2.34x |
| classifier-1x1-fp16 | 15.71 | 4.90 | 8.22 | 3.21x | 1.91x |
| resnet-1x1-bf16 | 37.38 | 12.48 | 11.57 | 3.00x | 3.23x |

Average ratio across all 13 cases: 2.43x vs `torch-nchw`, 2.10x vs `torch-nhwc`.

tileops/kernels/conv2d/conv2d.py
RMLYC: Addressed in 4bb629b. I added a short `OIHW -> HWIO` comment next to the `permute(2, 3, 1, 0)` so the intended kernel layout is explicit.

tileops/kernels/conv2d/conv2d.py
RMLYC: Addressed in 4bb629b. I moved the bias handling to a Python-level branch so the kernel now emits separate biased / bias-free paths instead of carrying `has_bias` through an element-wise `T.if_then_else`.

tileops/kernels/conv2d/conv2d.py
RMLYC: Addressed in 4bb629b. `_conv2d_shared_memory_for_nxn` was just a pass-through helper, so I removed it and now call `_conv2d_shared_memory_bytes(...)` directly from the autotune config filter.

tileops/kernels/conv2d/conv2d.py
RMLYC: I looked at this one while updating the kernel paths. I left the double check in place for now: although the second writeback guard looks sufficient at first glance, I did not want to remove the first guard without validating the generated tilelang code path and edge-tile behavior more carefully. The other cleanups in 4bb629b are in; I can revisit this one separately with a focused micro-benchmark/codegen check if needed.

tileops/kernels/conv2d/conv2d.py
RMLYC: done

PR #665, opened 2026-03-25
[Feat] Add ssd_decode and da_cumsum_fwd kernels for Mamba-2

stelladuyx:
## Description

- **`ssd_decode`**: TileLang kernel + op for the Mamba-2 single-step recurrent SSM decode. Updates the SSM state in-place (`state[b,h,p,n] ← dA·state + dt·B·x`) and accumulates the output (`y[b,h,p] = Σ_n state·C`). Supports arbitrary `ngroups`. The D skip connection, output gate (SiLU/RMSNormGated), causal-conv1d step, and `out_proj` are intentionally left to the caller (documented in-file with a full diff vs. the official `Mamba2.step()`).
- **`da_cumsum_fwd`**: TileLang kernel + op for the chunk-local inclusive prefix sum `dA_cumsum[b,h,c,l] = Σ_{i≤l} dt[b,c·Q+i,h] · A[h]`, matching `A_cumsum` from the Mamba-2 `ssd_minimal_discrete` reference. Consumed by `ssd_chunk_scan_fwd`, `ssd_chunk_state_fwd`, and `ssd_state_passing_fwd`.

Both kernels include op wrappers, tests (smoke + full param sweeps), and benchmarks.

## Test plan

- [x] `pytest tests/ops/test_ssd_decode.py -m smoke` — passes
- [x] `pytest tests/ops/test_da_cumsum_fwd.py -m smoke` — passes
- [x] `pytest tests/ops/test_ssd_decode.py tests/ops/test_da_cumsum_fwd.py -m full` — full param sweep
- [x] `python benchmarks/ops/bench_ssd_decode.py` — benchmark vs PyTorch reference
- [x] `python benchmarks/ops/bench_da_cumsum_fwd.py` — benchmark vs PyTorch reference


tileops/ops/ssd_decode.py
```diff
@@ -0,0 +1,106 @@
...
+            raise ValueError("x must be a CUDA tensor")
+        if x.dtype != self.dtype:
+            raise ValueError(f"Expected x dtype {self.dtype}, got {x.dtype}")
+        if dt.dtype != torch.float32:
+            raise ValueError(f"Expected float32 dt, got {dt.dtype}")
+        if state.dtype != torch.float32:
+            raise ValueError(f"Expected float32 state, got {state.dtype}")
+
+        A = A.contiguous()
+        dt = dt.contiguous()
```
superAngGao, line 102: `state = state.contiguous()` will silently create a **copy** when the input tensor is non-contiguous (e.g. a slice or view). The kernel then mutates the copy in-place, and the caller's original `state` tensor is **never updated** — a silent correctness bug.

Tests don't catch this because `torch.randn()` produces contiguous tensors, but in a real model the state may come from a pre-allocated buffer or a view.

Suggested fix — either assert contiguity:
```python
if not state.is_contiguous():
    raise ValueError("state must be contiguous for in-place update")
```
or copy back after the kernel call (less ideal, as it hides a performance pitfall).

tileops/kernels/mamba/da_cumsum_fwd.py
```diff
@@ -0,0 +1,178 @@
...
+                    running[0] = running[0] + dt_val * dA_head
+                    dA_cumsum[bb, bh, bc, l] = running[0]
+
+        return main
+
+    return kernel_func
+
+
+@torch.library.custom_op("top::da_cumsum_fwd", mutates_args=())
+def _da_cumsum_fwd_wrapped(
```
superAngGao, line 97: The default config uses `threads=1`, running a serial scan over Q positions. This is correct and simple, but:

1. **No `autotune_configs`** — unlike `SsdDecodeKernel`, calling this kernel with `tune=True` has no search space to explore.
2. **Potential perf uplift with warp-level scan** — for Q=64, two warps with `__shfl_up`-style scan can finish in ~6 steps vs 64 serial iterations, with minimal shared-memory overhead. Since `A[h]` is constant across the chunk, you can factor it out: do a prefix sum on `dt` alone, then multiply by `A[h]` once.

For large grid sizes (B×H×C >> SM count) the serial approach is fine — but for small-batch / few-head configs, intra-block parallelism would help occupancy.

Worth adding at least a few `autotune_configs` with different thread counts so that `tune=True` can explore this.

superAngGao (approved):
LGTM.

Issue #671, opened 2026-03-25
[Feat][MoE] Add KimiMoEOp (Kimi K2 MoE variant)


michaelwithu:
## Background

Kimi K2 uses a MoE FFN architecture that differs from the existing `Qwen3MoENopadOp` in its routing mechanism:

- **Routing**: sigmoid (not softmax) + per-expert `correction_bias` — bias is added to sigmoid scores for top-k selection only; output weights are gathered from the original (unbiased) sigmoid scores
- **`routed_scaling_factor`**: scalar multiplier applied to the routed output after unpermute (Kimi K2: 2.872)

### `FusedTopKKernel` / `FusedTopKOp`
- Add `with_correction_bias` compile-time variant to `FusedTopKKernel`
- Warp all-reduce carries original sigmoid score alongside biased score; `topk_weights` output uses unbiased scores
- Backward compatible: default `with_correction_bias=False` path unchanged, Qwen3MoE unaffected

### `KimiMoENopadOp` (new)
- Tight layout (T×K rows), GPU tile scheduler
- Pipeline: `FusedTopKOp` → `MoePermuteNopadOp` → gate+up GEMM → SwiGLU → down GEMM → `MoeUnpermuteOp`

### `KimiMoEPaddedOp` (new)
- Same semantics, block\_m-aligned padded layout with `GroupedGemmOp` as comparison baseline

### Tests & Benchmark
- Correctness tests against PyTorch reference, covering correction\_bias precision, routed\_scaling\_factor, bf16/fp16
- Benchmark vs vLLM on Kimi K2 production config (E=384, K=6, H=7168, F=2048)

## Parent Issue

Part of #404.

PR #672, opened 2026-03-25
[Feat][MoE] Add KimiMoENopadOp / KimiMoEPaddedOp with correction_bias routing (Kimi K2)

michaelwithu:
## Summary

Implement MoE FFN operator for Kimi K2, closing #671.

Compared to `Qwen3MoENopadOp`:
- **Routing**: sigmoid + per-expert `correction_bias` — bias is used for top-k selection only; output weights are the original (unbiased) sigmoid scores
- **`routed_scaling_factor`**: scalar multiplier applied to the routed output after unpermute (Kimi K2: 2.872)
- **`shared_experts_fn`**: optional callable for shared expert output (DeepSeekV3 `n_shared_experts=1`; Kimi K2 passes `None`)

## Changes

- **`tileops/kernels/moe/fused_topk.py`**: add `with_correction_bias` compile-time variant; warp all-reduce tracks `(l_best_val, l_best_orig, l_best_idx)` so `topk_weights` reflects original sigmoid, not biased scores. Default path unchanged.
- **`tileops/ops/moe/fused_topk.py`**: expose `with_correction_bias` and `correction_bias` in `forward()`.
- **`tileops/ops/moe/kimi_moe.py`**: `KimiMoENopadOp` (tight T×K layout, GPU tile scheduler) + `KimiMoEPaddedOp` (padded layout, comparison baseline).
- **`tests/ops/test_moe_kimi_moe.py`**: correctness vs PyTorch reference.
- **`benchmarks/ops/bench_moe_kimi_moe.py`**: benchmark vs vLLM on Kimi K2 production config (E=384, K=6, H=7168, F=2048).

## Benchmark (H200, E=384, K=6, bf16)

> vLLM baseline uses pre-computed routing, benchmarking GEMM+SwiGLU only. TileOPs includes the full routing step.

### H=7168, F=2048 (Kimi K2 actual scale)

| T | TileOPs nopad (ms) | vs padded | vs vLLM |
|---|---|---|---|
| 512  | 7.74 / 34.97 TFLOPs | **1.30×** faster | -2% |
| 2048 | 12.32 / 87.86 TFLOPs | **1.23×** faster | +2% |
| 4096 | 17.17 / 126.05 TFLOPs | **1.11×** faster | -4% |

### H=2048, F=1024 (ablation)

| T | TileOPs nopad (ms) | vs padded | vs vLLM |
|---|---|---|---|
| 512  | 0.90 / 42.73 TFLOPs | **1.55×** faster | **+13%** |
| 2048 | 1.20 / 129.31 TFLOPs | **1.57×** faster | **+28%** |
| 4096 | 1.89 / 164.04 TFLOPs | **1.52×** faster | **+38%** |

At full Kimi K2 scale (H=7168), TileOPs nopad matches vLLM end-to-end despite including routing overhead. At smaller hidden sizes (H=2048), nopad outperforms vLLM by 13–38%.

## Test plan

```bash
# Backward compat
python -m pytest tests/ops/test_moe_fused_topk.py tests/ops/test_moe_qwen3_moe.py -vvs
# New tests
python -m pytest tests/ops/test_moe_kimi_moe.py -vvs
# Benchmark
python -m pytest benchmarks/ops/bench_moe_kimi_moe.py -vvs
```


tileops/kernels/moe/fused_topk.py
```diff
@@ -80,138 +93,221 @@ def _func(TOKENS_PER_BLOCK):
...
+                gating_output: T.Tensor([num_tokens, num_experts], "float32"),  # noqa: F821
+                topk_weights: T.Tensor([num_tokens, top_k], "float32"),         # noqa: F821
+                topk_ids: T.Tensor([num_tokens, top_k], "int32"),               # noqa: F821
+            ):
+                with T.Kernel(num_blocks, threads=TOKENS_PER_BLOCK * WARP_SIZE) as (block_id,):
+                    tx = T.get_thread_binding()
+                    warp_id = tx // WARP_SIZE
+                    lane_id = tx % WARP_SIZE
+                    # Each warp handles one token.
+                    token_id = block_id * TOKENS_PER_BLOCK + warp_id
```
superAngGao, line 221: **[Major]** The `with_correction_bias` branch duplicates the entire standard kernel (~120 lines). Steps 1 (load), step 2 (scoring — the softmax path is dropped entirely here), and step 3 (K-pass argmax + warp all-reduce + masking) are largely copy-pasted.

I understand TileLang requires separate `@T.prim_func` definitions for different signatures, but the duplication makes future bug fixes error-prone (e.g. a fix to the tie-breaking logic must be manually mirrored).

Consider:
- Extracting the standard (non-bias) kernel as-is and having the bias variant call into shared helper macros / inline functions for the common parts (load + masking loops)
- Or at minimum, adding a comment at the top of the `else` branch noting it is an exact copy of the original and any changes must be synchronized

tileops/kernels/moe/fused_topk.py
```diff
@@ -80,138 +93,221 @@ def _func(TOKENS_PER_BLOCK):
...
-                    inv_row_sum = T.alloc_var(T.float32)
-                    inv_row_sum = T.float32(1)  # default (sigmoid / no renorm)
-
-                    if scoring_func == "softmax":
-                        l_max = T.alloc_var(T.float32)
-                        l_sum = T.alloc_var(T.float32)
-
-                        # Warp max all-reduce (shfl_xor, no barrier)
-                        l_max = -T.infinity("float32")
+        if with_correction_bias:
```
superAngGao, line 96: **[Nit]** The correction_bias variant hardcodes sigmoid scoring — it silently ignores `scoring_func`. The `__init__` validates `scoring_func=="sigmoid"` already, but it would be clearer to add an assertion or comment here inside `_func` that this branch always uses sigmoid, so a reader does not wonder where the softmax path went.

benchmarks/ops/bench_moe_kimi_moe.py
```diff
@@ -0,0 +1,262 @@
...
+        return hidden, gating, correction_bias, w_gate_up, w_down
+
+    def ref_program(self, *args):
+        return None
+
+
+class KimiMoEBenchFixture(FixtureBase):
+    """Kimi K2 production-scale configs: E=384, K=6, H=7168, F=2048."""
+    PARAMS = [
+        (
```
superAngGao, line 113: **[Minor]** Per `.claude/rules/benchmark.md`: "If an external baseline is conditional on library availability, add `test.ref_program` as a torch fallback in the `else` branch."

`ref_program` returns `None` here, but the torch-ref fallback is implemented inline in the test function (lines 243-260). Consider moving the per-expert GEMM loop into `ref_program()` so the pattern matches the convention and is reusable.

benchmarks/ops/bench_moe_kimi_moe.py
```diff
@@ -0,0 +1,262 @@
...
+    # Uses pre-computed Kimi K2 routing (FusedTopKOp) + vLLM fused_experts for
+    # a fair GEMM/SwiGLU/unpermute comparison.
+    if _VLLM_AVAILABLE:
+        def _vllm_fn(hidden, gating, correction_bias, w_gate_up, w_down):
+            return _vllm_fused_experts(
+                hidden, w_gate_up, w_down,
+                topk_weights, topk_ids,
+            )
+
+        _vllm_fn(hidden, gating, correction_bias, w_gate_up, w_down)  # warmup
```
superAngGao, line 228: **[Nit]** `_vllm_fn` and `_ref_fn` accept `(hidden, gating, correction_bias, w_gate_up, w_down)` but only use `hidden` and the weights — `gating` and `correction_bias` are unused (routing is pre-computed via `topk_weights`/`topk_ids` from closure). This works for `bm.profile()` signature matching, but a short comment explaining the unused args would help readability.

tileops/ops/moe/kimi_moe.py
```diff
@@ -0,0 +1,360 @@
+"""Kimi K2 / DeepSeekV3-variant MoE FFN operator.
```
superAngGao, line 1: **[Minor]** Should a follow-up issue be opened for the `ops_manifest.yaml` entry? The existing convention (e.g. `c192503` for retnet) adds manifest entries alongside or shortly after the op PR.

PR #677, opened 2026-03-26
[CI] Reuse trusted hash venv for fork gpu smoke

### LINKED ISSUE #676: [PERF][CI] reduce fork pr gpu-smoke cold-start overhead
### Symptom / Motivation
`gpu-smoke` is consistently much slower on fork pull requests than on trusted `push` runs, even when the PR path only runs the `smoke` marker set. Current evidence points to environment cold-start overhead dominating the runtime gap.

### Root Cause Analysis
In `.github/workflows/gpu-smoke.yml`, fork PR runs execute in an isolated per-run runtime root under `${RUNNER_TEMP}`. The workflow already copies trusted cache directories into that temporary root, but the fork path still falls back to creating a fresh virtual environment and reinstalling dependencies whenever no reusable environment is available in the temporary location.

Trusted `push` runs can reuse a hash-keyed persistent venv under `${{ runner.tool_cache }}`, so their environment setup cost is much lower. Fork PR runs therefore pay an avoidable cold-start penalty even when the actual pytest scope is smaller.

A workflow change has now been prepared to improve this path: for fork PRs, compute the same dependency hash used by trusted runs, check whether the matching trusted venv exists, and if it does, copy that venv into the fork PR runtime root instead of reinstalling dependencies. If the trusted hash venv does not exist, the workflow falls back to the current fresh-create-and-install behavior.

### Related Files

## Goal
Reduce fork PR `gpu-smoke` runtime by avoiding repeated dependency installation when a matching trusted hash venv already exists, while preserving fork PR isolation.

## Plan

1. Land the fork-PR trusted-venv-copy workflow change.
2. Measure post-change fork PR runs and compare setup time against trusted `push` runs.
3. Add clearer timing and reuse diagnostics to CI logs if the remaining bottleneck is still unclear.
4. Revisit `security-policy` fallback scope only after environment setup cost is under control.

## Constraints
- Preserve isolated fork PR runtime state and cleanup semantics.
- Do not allow fork PR jobs to mutate the trusted shared venv.
- Keep a safe fallback path that recreates and reinstalls dependencies when the trusted hash venv is unavailable.

## Acceptance Criteria
- [ ] Fork PR `gpu-smoke` does not reinstall dependencies when a matching trusted hash venv exists.
- [ ] Fork PR `gpu-smoke` still succeeds by creating a fresh venv when no matching trusted hash venv exists.
- [ ] CI logs clearly show whether the fork path copied a trusted venv or fell back to fresh install.
- [ ] Median fork PR `gpu-smoke` runtime improves relative to the current baseline.

RMLYC:
Closes #676

## Summary

- reuse the trusted hash-keyed venv for fork PR gpu-smoke runs by copying it into the per-run runtime root when available
- fall back to the existing fresh venv creation and dependency install path when no matching trusted venv exists
- add a lightweight copied-venv validation step before continuing to tests

## Test plan

- [x] pre-commit passed
- [ ] pytest passed

## Benchmark

- not run locally; this change targets CI environment setup latency rather than kernel runtime

## Regression

- fork PR jobs still keep isolated runtime state under RUNNER_TEMP
- fork PR jobs still fall back to fresh environment creation when the trusted hash venv is unavailable

## Additional context

- issue: #676

### DIFF
--- a/.github/workflows/gpu-smoke.yml
+++ b/.github/workflows/gpu-smoke.yml
@@ -515,12 +515,7 @@ jobs:
             source "${VENV_PATH}/bin/activate"
             python --version
             python -m pip --version
-            python - <<'PY'
-            import tileops
-            import torch
-
-            print("Copied fork venv import check ok")
-            PY
+            python -c "import tileops, torch; print('Copied fork venv import check ok')"
             exit 0
           fi
 


.github/workflows/gpu-smoke.yml
```diff
@@ -490,13 +500,28 @@ jobs:
...
+            rm -rf "${VENV_PATH}"
+            if command -v rsync >/dev/null 2>&1; then
+              rsync -a --delete "${TRUSTED_VENV_PATH}/" "${VENV_PATH}/"
+            else
+              mkdir -p "${VENV_PATH}"
+              cp -a "${TRUSTED_VENV_PATH}/." "${VENV_PATH}/"
+            fi
+            # Validate the copied venv before continuing. A byte-for-byte copy can
+            # still be sensitive to absolute paths embedded by venv tooling.
+            # shellcheck source=/dev/null
```
lcy-seso, line 514: **[INFO]** The validation checks `import tileops` but doesn't verify the editable install resolves to the current workspace checkout. If the workspace path ever diverges between trusted and fork runners, the fork could silently test stale code.

Consider adding a path assertion:
```bash
python -c "import tileops, pathlib; assert str(pathlib.Path(tileops.__file__).resolve()).startswith('$GITHUB_WORKSPACE'), f'tileops resolved outside workspace: {tileops.__file__}'"
```

.github/workflows/gpu-smoke.yml
```diff
@@ -379,7 +391,12 @@ jobs:
             PIP_CACHE_DIR="${RUNTIME_ROOT}/pip/cache"
             WHEEL_DIR="${RUNTIME_ROOT}/wheels"
             VENV_PATH="${RUNTIME_ROOT}/venv"
-            VENV_REUSED="false"
+            if [[ -x "${TRUSTED_VENV_PATH}/bin/python" ]]; then
```
lcy-seso, line 394: **[INFO]** `VENV_REUSED` is set to `true` here when the trusted venv exists, but the actual copy happens in a later step. If the copy fails, the exported env var is semantically wrong (harmless due to `set -euo pipefail`, but misleading). Consider setting it after the copy succeeds, or using a name like `VENV_COPY_PENDING` here.

Issue #696, opened 2026-03-28
[Bug][Bench] NSA topk benchmark failures (3 tests)


superAngGao:
Parent: #689

## Problem

All 8 tests across `bench_deepseek_nsa_topk`, `bench_deepseek_nsa_cmp_fwd`, and `bench_deepseek_nsa_fwd` fail during the benchmark run. The failures also occurred with the old `do_bench` approach (not caused by `bench_kernel` migration).

## Failing tests

- `test_nsa_topk_bench[mainstream-fp16]`
- `test_nsa_topk_bench[shorter-seq]`
- `test_nsa_topk_bench[long-sequence]`
- `test_nsa_cmp_fwd_bench[mainstream-fp16]`
- `test_nsa_cmp_fwd_bench[long-sequence-fp16]`
- `test_nsa_fwd_bench[single-block]`
- `test_nsa_fwd_bench[long-context]`
- `test_nsa_fwd_bench[multi-selected-blocks]`

## Current workaround

NSA benchmark files are skipped via `collect_ignore_glob` in `benchmarks/conftest.py` to unblock the nightly run. This was introduced in the benchmark overhaul PR (ref #689, #692).

```python
collect_ignore_glob = [
    "ops/bench_deepseek_nsa_*.py",
]
```

## Notes

- The `NsaTopkTest` import chain pulls in `tileops.ops.NSATopkVarlenOp`, triggering kernel compilation at import time
- Needs investigation whether the failure is in the op itself, the test fixture, or the benchmark harness

## Reproduction

```bash
python -m pytest benchmarks/ops/bench_deepseek_nsa_topk.py -v
```

Issue #691, opened 2026-03-28
[Doc][Bench] CUPTI per-kernel duration drifts with n_repeat


superAngGao:
## Problem

When benchmarking with CUPTI (`torch.profiler` + `ProfilerActivity.CUDA`), the **per-iteration kernel duration changes depending on `n_repeat`** — the total number of iterations profiled in a single active step. This is a profiler-level artifact, not a real kernel behavior change.

## Reproduction

All tests use `torch.profiler.schedule(wait=0, warmup=1, active=1, repeat=1)` with a warmup step (discarded) followed by an active step (measured). L2 is flushed (60MB zero) before each iteration. GPU clocks locked at 1830 MHz on H200.

Tested functions are **stock PyTorch ops** (no TileLang/Triton):
- `torch.mm` (cuBLAS GEMM)
- `torch.softmax`
- `torch.relu`

#### `torch.mm` — 1024×1024×1024 FP16

| n_repeat | per_iter (us) | CUDA events | events/iter | notes |
|---|---|---|---|---|
| 5 | **6.41** | 10 | 2.0 | 1 GEMM kernel per iter |
| 50 | **6.59** | 100 | 2.0 | stable |
| 500 | **7.04** | 1058 | 2.1 | extra events appear |
| 2000 | **7.30** | 4383 | 2.2 | +14% drift from n=5 |

At large n_repeat, cuBLAS occasionally inserts auxiliary CUDA kernels (workspace ops, etc.), inflating the per-iter sum. The extra ~500 events at n=2000 vs n=500 are not FillFunctor (those are excluded).

#### `torch.mm` — 4096×4096×4096 FP16

| n_repeat | per_iter (us) | CUDA events | events/iter | dominant kernel |
|---|---|---|---|---|
| 5 | **282.5** | 24 | 4.8 | 8×175.7us |
| 50 | **329.0** | 279 | 5.6 | 93×176.0us |
| 500 | **329.1** | 2520 | 5.0 | 840×195.0us |
| 2000 | **249.0** | 7020 | 3.5 | 2340×211.8us |

**Non-monotonic**: n=50 is 17% higher than n=5, while n=2000 is 12% lower. cuBLAS appears to switch tiling strategies at high iteration counts — each sub-kernel gets slower (176→212us) but fewer sub-kernels are launched per GEMM (4.8→3.5 per iter).

#### `torch.softmax` — 256×4096 FP16

| n_repeat | per_iter (us) | CUDA events | events/iter |
[...]

Issue #689, opened 2026-03-28
[Perf][Bench] Benchmark infrastructure overhaul — align with NVIDIA SOL-ExecBench


superAngGao:
## Goal

Overhaul the TileOPs nightly benchmark infrastructure to reduce wall time from ~54 min to ~16 min while improving measurement accuracy.

## Background: NVIDIA SOL-ExecBench protocol

NVIDIA's [SOL-ExecBench](https://arxiv.org/abs/2603.19173) establishes a rigorous GPU kernel benchmarking protocol:

- **Fixed iteration count**: 10 warmup + 50 timed iterations × 3 trials
- **L2 cache flush**: 256 MB buffer zeroed before every timed iteration
- **Input diversity**: tensor arguments cloned each iteration with 256-byte address shift
- **Clock locking**: GPU frequencies locked via nvidia-smi
- **Timing**: CUDA events
- **Reporting**: mean of 3 trials

This fixed-count approach avoids the pitfalls of time-based iteration scaling, where tiny kernels accumulate 2000+ iterations and profiler overhead dominates.

### Part 1: Aligning with SOL-ExecBench

We adopt the core protocol with our `bench_kernel()` function:

| Parameter | Old (`do_bench`) | New (`bench_kernel`) | SOL-ExecBench |
|---|---|---|---|
| Warmup | `100ms / estimate` → 1000+ | **10 iterations** | 10 |
| Timed repeats | `200ms / estimate` → 2000+ | **50 iterations** | 50 |
| Trials | 1 | **3** | 3 |
| L2 flush | 256 MB hardcoded | **`L2_cache_size` dynamic** | 256 MB |
| Input handling | Reuse same tensor | **3-copy clone pool** | Clone + 256B shift |
| Clock locking | nvidia-smi | nvidia-smi | nvidia-smi |
| Reporting | Single mean | **Median of 3 trial means** | Mean of 3 trials |

We use median instead of mean for trial aggregation — with only 3 samples, median is strictly more robust to outliers.

### Part 2: Improvements over SOL-ExecBench

**1. CUPTI instead of CUDA events for pure kernel timing.**
TileLang kernels have ~45 us launch overhead due to the TVM runtime dispatch layer, compared to ~12 us for native PyTorch ops. CUDA events measure launch + execution, which would show TileLang as *slower* than torch — even though the GPU kernel itself is faster. CUPTI isolates pure GPU execution time, enabling accurate SOL/roofline analysis.

[...]

Issue #694, opened 2026-03-28
[Refactor][Bench] Migrate 6 legacy do_bench benchmarks to bench_kernel


superAngGao:
Parent: #689

## Problem

6 benchmark files define their own `_profile_manual()` that calls `do_bench()` directly with time-based iteration (`warmup=100ms, rep=100ms`), bypassing the new `bench_kernel()` protocol. This means they still run 2000+ iterations per measurement and incur the old profiler overhead.

## Affected files

- `benchmarks/ops/bench_deltanet_chunkwise.py`
- `benchmarks/ops/bench_gated_deltanet_chunkwise.py`
- `benchmarks/ops/bench_gla_chunkwise.py`
- `benchmarks/ops/bench_gla_recurrence.py`
- `benchmarks/ops/bench_moe_permute_align.py`
- `benchmarks/ops/bench_gqa_sliding_window_varlen_fwd.py`

## Fix

Replace `_profile_manual()` / direct `_do_bench()` calls with `BenchmarkBase.profile()` which uses `bench_kernel()`. This requires:
1. Converting the benchmark class to inherit `BenchmarkBase`
2. Passing `fn` and `args` separately (for input cloning support)
3. Removing the inline `_do_bench` import and fallback logic

PR #697, opened 2026-03-28
[Perf][Bench] Overhaul benchmark infrastructure with SOL-ExecBench protocol

### LINKED ISSUE #690: [Perf][Bench] Bypass key_averages() parsing with raw Kineto C++ events
## Problem

When `do_bench(backend='cupti')` is called for small/fast kernels, nearly **half the wall time** is spent parsing profiler events — not running the actual kernel.

For example, benchmarking a tiny ReLU (1K elements, `rep=100`, `n_repeat=1280`):

| Phase | Time | % | Notes |
|---|---|---|---|
| Kernel loop (n_repeat iterations) | 273ms | 52% | Actual GPU work |
| **`key_averages()` parsing** | **328ms** | **48%** | **Kineto event aggregation — the bottleneck** |
| Warmup, estimation, CUPTI init/teardown | ~10ms | <2% | Negligible |

### Root cause

Calling `profiler.key_averages()` (or `profiler.events()`) triggers a **lazy parsing chain** inside PyTorch's profiler:

1. **`_parse_kineto_results()`** (~150ms) — iterates all C++ `_KinetoEvent` objects and constructs a Python `FunctionEvent` (24 fields each) for every single event, including memory record association and correlation ID mapping.
2. **`_build_tree()`** (~30ms) — sorts all events and builds parent-child relationship trees.
3. **`key_averages()` aggregation** (~50ms) — groups events by key and accumulates stats via `FunctionEventAvg`.

For `n_repeat=1280`, the profiler processes ~7400 total events (3500+ CUDA kernel events). All three stages scale with event count, but `do_bench` only needs **total CUDA kernel duration** — none of the per-event Python objects, tree structure, or grouped aggregation are used.

### Note on `events()` vs `key_averages()`

Both trigger the same `_parse_kineto_results()` + `_build_tree()` pipeline. Using `events()` instead of `key_averages()` saves only the final aggregation step (~50ms) but not the dominant parsing cost (~180ms).

## Solution

Access the raw C++ Kineto events directly via `profiler.profiler.kineto_results.events()`. Each `_KinetoEvent` C++ object exposes `device_type()`, `name()`, and `duration_ns()` — exactly what `do_bench` needs. This completely bypasses Python `FunctionEvent` construction, tree building, and aggregation.

```python
from torch.autograd.profiler import DeviceType

def _sum_kernel_time_us(kineto_results):
    total_us = 0.0
    for evt in kineto_results.events():
        if evt.device_type() == DeviceType.CUDA:
            name = evt.name()
            if "vectorized_elementwise" in name and "FillFunctor" in name:
                continue  # exclude cache.zero_() kernels
            total_us += evt.duration_ns() / 1000.0
    return total_us
```

### Benchmark (n_repeat=1280, 7414 events)

| Approach | Parsing time | Kernel result |
|---|---|---|
| `key_averages()` (current) | 328ms | 0.0028ms |
| Raw Kineto C++ iteration | **15ms** | 0.0028ms |
| **Speedup** | **21x** | Identical |

### Impact

- **Per `do_bench` call**: saves ~300ms on small kernels (where `n_repeat` is large)
- **Nightly benchmark suite** (~1500 calls): estimated **~7 minutes** saved per run
- **`CuptiSession` shared path**: `_on_trace_ready` callback also benefits from the same optimization

### Affected code

- `benchmarks/benchmark.py`: `CuptiSession._on_trace_ready()` and `_bench_with_cupti_fixed()` both call `profiler.key_averages()`
- Upstream `tilelang/profiler/bench.py`: `_bench_with_cupti()` has the same pattern (already monkey-patched by TileOPs)

## Validated

- All 9 `bench_dropout` parametrized tests pass with the optimization applied
- Kernel time values are bit-identical between old and new approaches

### LINKED ISSUE #692: [Perf][Bench] Adopt NVIDIA SOL-ExecBench benchmark protocol (bench_kernel)
Parent: #689

## Summary

Replace the time-based `do_bench(warmup=100ms, rep=200ms, backend='cupti')` with a new `bench_kernel()` function following the NVIDIA SOL-ExecBench protocol (arxiv.org/abs/2603.19173).

## Protocol

| Parameter | Old (`do_bench`) | New (`bench_kernel`) | NVIDIA SOL-ExecBench |
|---|---|---|---|
| Warmup | `100ms / estimate` → 1000+ iters | **10 iterations** | 10 |
| Timed repeats | `200ms / estimate` → 2000+ iters | **50 iterations** | 50 |
| Trials | 1 | **3** | 3 |
| Timing backend | CUPTI (shared session) | **CUPTI (per-trial profiler)** | CUDA events |
| L2 flush | 256MB hardcoded | **`L2_cache_size` dynamic** (60MB on H200) | 256MB |
| Input handling | Reuse same tensor | **3-copy clone pool** (>1GB skip) | Clone + 256B address shift |
| Reporting | Single mean | **Median of 3 trial means** | Mean of 3 trials |
| Memory cleanup | None | **`torch.cuda.empty_cache()`** per call | N/A |

## Why CUPTI over CUDA events

TileLang kernels have ~45us launch overhead (vs torch ~12us) due to the TVM runtime dispatch layer. CUDA events would mask TileLang's kernel-level advantage. CUPTI measures pure GPU execution time, enabling accurate SOL/roofline analysis.

## Result

918 tests in **16 min** (was 54 min) — **3.4x speedup**, 0 failures.

## References

- NVIDIA SOL-ExecBench paper: https://arxiv.org/abs/2603.19173
- NVIDIA SOL-ExecBench repo: https://github.com/NVIDIA/SOL-ExecBench

### LINKED ISSUE #693: [Fix][Bench] Remove warmup_cupti fixtures that conflict with bench_kernel
Parent: #689

## Problem

7 benchmark files define a `warmup_cupti` session-scoped fixture that creates its own `torch.profiler.profile` to pre-initialize CUPTI. This conflicts with `bench_kernel()` which creates its own profiler per trial — Kineto is a global singleton and throws `RuntimeError: Can't disable Kineto profiler when it's not running`.

## Affected files

- `benchmarks/ops/bench_moe_fused_topk.py`
- `benchmarks/ops/bench_moe_kimi_moe.py`
- `benchmarks/ops/bench_moe_permute.py`
- `benchmarks/ops/bench_moe_permute_align.py`
- `benchmarks/ops/bench_moe_qwen3_moe.py`
- `benchmarks/ops/bench_moe_unpermute.py`
- `benchmarks/ops/bench_gqa_sliding_window_varlen_fwd.py`

## Fix

`bench_kernel` includes a profiler warmup step in its schedule (`warmup=1, active=1`), so external CUPTI pre-initialization is unnecessary. The fixtures should be removed or made no-op.

superAngGao:
## Summary

- Replace time-based `do_bench(warmup=100ms, rep=200ms)` with `bench_kernel()` following the [NVIDIA SOL-ExecBench protocol](https://arxiv.org/abs/2603.19173): fixed 10 warmup + 50 repeat × 3 trials, CUPTI pure kernel timing, L2 flush sized to actual cache, input tensor cloning for address diversity
- Bypass `profiler.key_averages()` Python parsing with direct Kineto C++ event iteration (`_sum_kernel_time_us`): 330ms → 15ms per call (21x)
- Disable `warmup_cupti` fixtures in 7 MOE/GQA benchmark files that conflict with `bench_kernel`'s per-trial profiler
- Skip NSA benchmarks (`collect_ignore_glob` in conftest) until op failures are resolved (#696)
- Add `torch.cuda.empty_cache()` after each `bench_kernel` call to prevent memory accumulation

## Result

918 tests, **0 failures**, **16 min** (was 54 min) — **3.4x speedup** on H200 at 1830 MHz.

## Test plan

- [x] Full benchmark suite (excl. NSA) passes: 918 passed, 0 errors
- [x] Measurement accuracy validated: CUPTI values match across flush sizes (60MB vs 256MB) and iteration counts (50 vs 2000) within 3%
- [x] MOE benchmarks pass with disabled `warmup_cupti` fixture
- [x] Memory stays stable across 918 tests (empty_cache prevents accumulation)
- [ ] Nightly CI run on main after merge

Closes #690, closes #692, closes #693.
Refs #689, #691, #696.

### DIFF
--- a/benchmarks/ops/bench_gqa_sliding_window_varlen_fwd.py
+++ b/benchmarks/ops/bench_gqa_sliding_window_varlen_fwd.py
@@ -9,7 +9,6 @@ from benchmarks.benchmark import BenchmarkBase, BenchmarkReport
 from tests.ops.test_gqa_sliding_window_varlen_fwd import GqaSlidingWindowVarlenFwdTest
 from tileops.ops import GqaSlidingWindowVarlenFwdOp
 
-
 _GQA_SLIDING_WINDOW_VARLEN_FWD_BENCH_PARAMS = [
     pytest.param(1, [3000], [3000], 32, 8, 128, True, -1, -1, torch.float16, False, id="single-seq-causal"),
     pytest.param(2, [1537, 3073], [1537, 3073], 32, 8, 128, True, 2000, -1, torch.float16, False, id="mixed-length-left-window"),


superAngGao:
Thanks for the review! Addressing each point:

**`CuptiSession` schedule `repeat=0` (critical):** This is actually correct — `repeat=0` means *infinite* repetition in `torch.profiler.schedule` ([PyTorch docs](https://pytorch.org/docs/stable/profiler.html#torch.profiler.schedule): "0 means that the cycles will continue until the profiling is finished"). The `CuptiSession` is a legacy component retained for the 6 benchmark files that still call `do_bench` directly (#694). `bench_kernel` does not use it.

**Warmup step using `_run(0)`:** Good catch — fixed to use `_run(i)` for consistency.

**Dead code in `warmup_cupti` fixtures:** Agreed, `if True: return` is ugly. Removed the entire fixture body from all 7 files.

lcy-seso (approved):

PR #717, opened 2026-03-31
[Bench][Op] Add FLA baseline for GatedDeltaNetDecode

### LINKED ISSUE #716: [Bench][Op] GatedDeltaNetDecode benchmark missing FLA baseline
## Summary

`bench_gated_deltanet_recurrence.py` unconditionally uses `torch-ref` as the only baseline for `GatedDeltaNetDecodeOp`. FLA provides `fused_recurrent_gated_delta_rule` which is the proper external baseline for comparison.

Other similar benchmarks (GLA decode, DeltaNet chunkwise, GatedDeltaNet chunkwise) all use FLA as the primary baseline with `torch-ref` as fallback when FLA is not installed.

## Expected

- Use `fla.ops.gated_delta_rule.fused_recurrent_gated_delta_rule` (T=1) as primary baseline, tagged `"fla"`
- Fall back to `torch-ref` when FLA is not installed
- Follow the same pattern as `bench_gla_recurrence.py`

superAngGao:
## Summary

- Add `fused_recurrent_gated_delta_rule` (T=1) from FLA as primary baseline for `GatedDeltaNetDecodeOp` benchmark
- Fall back to `torch-ref` when FLA is not installed
- Follows the same pattern as `bench_gla_recurrence.py`

Previously the benchmark unconditionally used `torch-ref`, which is a pure-PyTorch reference — not a meaningful performance comparison.

Closes #716

## Test plan

- [x] `pytest benchmarks/ops/bench_gated_deltanet_recurrence.py` — passes with FLA installed (uses `"fla"` tag)
- [ ] Verify fallback to `"torch-ref"` when FLA is not installed

### DIFF
--- a/benchmarks/ops/bench_gated_deltanet_recurrence.py
+++ b/benchmarks/ops/bench_gated_deltanet_recurrence.py
@@ -78,10 +78,12 @@ def test_gated_deltanet_decode_bench(
         g_fla = g.unsqueeze(1)       # [B, H] -> [B, 1, H]
         beta_fla = beta.unsqueeze(1)
 
+        state_fla = state.contiguous()
+
         def fla_decode():
             return fused_recurrent_gated_delta_rule(
                 q_fla, k_fla, v_fla, g=g_fla, beta=beta_fla,
-                initial_state=state.contiguous(),
+                initial_state=state_fla,
                 output_final_state=True,
             )
 


superAngGao:
Good catch — moved `state.contiguous()` outside the closure in e91091e so it's not repeated inside the profiled loop.

lcy-seso (approved):

Issue #720, opened 2026-03-31
[Bench] Add FlashInfer as attention benchmark baseline


superAngGao:
## Summary

Add FlashInfer as a benchmark baseline for GQA/MHA attention kernels, complementing the existing Flash-Attention (FA3) baseline.

Parent issue: #719

## Baseline Assignment Strategy

FA3 remains the baseline for standard prefill fwd/bwd (its strongest area). FlashInfer covers decode variants and feature combinations that FA3 does not support.

| Benchmark | Current baseline | New baseline | Fallback chain |
|-----------|-----------------|-------------|----------------|
| `bench_gqa.py` fwd/bwd | FA3 | **FA3** (keep) | FA3 → torch-sdpa |
| `bench_mha.py` fwd/bwd | FA3 | **FA3** (keep) | FA3 → torch-sdpa |
| `bench_gqa_decode.py` | torch-sdpa | **FlashInfer** (upgrade) | FlashInfer → torch-sdpa |
| `bench_gqa_decode_paged.py` | torch-ref | **FlashInfer** (upgrade) | FlashInfer → torch-ref |
| `bench_mha_decode.py` | FA3 | **FA3** (keep) | FA3 → torch-sdpa |
| `bench_mha_decode_paged.py` | torch-ref | **FlashInfer** (upgrade) | FlashInfer → torch-ref |
| `bench_gqa_sliding_window_fwd.py` | FA3 / torch-sdpa | **FlashInfer** (upgrade) | FlashInfer → FA3 → torch-sdpa |
| `bench_gqa_sliding_window_varlen_fwd.py` | FA3 / torch-sdpa | **FlashInfer** (upgrade) | FlashInfer → FA3 → torch-sdpa |
| Future RoPE/FP8/LogitsClip benchmarks | N/A | **FlashInfer** | FlashInfer → torch-ref |

## Fallback Pattern

Files where FlashInfer is primary:

```python
if _FLASHINFER_AVAILABLE:
    # FlashInfer baseline (preferred)
    ...
    BenchmarkReport.record(op, locals(), result, tag="flashinfer")
elif _FA3_AVAILABLE:
    # FA3 baseline (fallback, where applicable)
    ...
    BenchmarkReport.record(op, locals(), result, tag="fa3")
else:
    # torch baseline (last resort)
    ...
    BenchmarkReport.record(op, locals(), result, tag="torch-sdpa")
```

[...]

PR #721, opened 2026-03-31
[CI][Bench] Add flashinfer-python to benchmark dependencies

superAngGao:
## Summary

- Add `flashinfer-python>=0.6.6` to `pyproject.toml` `[bench]` optional dependencies
- Add to both Dockerfiles (`Dockerfile`, `.github/runner/Dockerfile`)

## Verified

- FlashInfer 0.6.6 installed in `tileops-runner:latest` via `docker commit` (incremental layer)
- Smoke tested in Docker: decode, prefill, prefill+RoPE all pass on H200
- PyTorch 2.10.0+cu128, CUDA 12.8, JIT compilation mode

Partial fix for #720

## Test plan

- [x] `docker run tileops-runner:latest python -c "import flashinfer"` succeeds
- [x] FlashInfer decode/prefill/RoPE smoke tests pass in container

### DIFF
--- a/.github/runner/Dockerfile
+++ b/.github/runner/Dockerfile
@@ -54,11 +54,13 @@ RUN pip install --no-cache-dir \
         "pytest-xdist>=3.0" \
         "pyyaml>=6.0"
 
-RUN MAX_JOBS=64 NVCC_THREADS=64 pip install --no-cache-dir --no-build-isolation "flash-attn==2.8.3" \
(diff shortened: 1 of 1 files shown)


zhen8838:
If the repo-root `Dockerfile` is also meant to be a runner image, we should keep a single runner Dockerfile instead of maintaining two near-duplicate ones. If it is meant to be a standalone TileOPs image, it should be made cleanly standalone instead of keeping runner-specific setup (`actions-runner`, runner entrypoint, runner workdir). Right now the split is ambiguous.

Issue #719, opened 2026-03-31
[Feature][Op] GQA kernel variants: RoPE, ALiBi, FP8, LogitsClip, SplitKV


superAngGao:
## Sub-issues

| # | Issue | Scope | Depends on | Priority |
|---|-------|-------|-----------|----------|
| 1 | #722 | SplitKV refactor (decode kernels) | — | P0 (prerequisite) |
| 2 | #723 | RoPE kernels + Op dispatch | #722 | P0 |
| 3 | #724 | LogitsClip kernels | #723 | P2 |
| 4 | #725 | Sliding window file reorganization | — | Independent |
| 5 | #726 | FP8 + RoPE kernels | #723 | P1 |
| 6 | #727 | ALiBi kernels | — | P4 |
| 7 | #720 | FlashInfer benchmark baselines | #723+ | Parallel |

### Dependency graph

```
#725 (SW reorg)           independent
#722 (SplitKV) → #723 (RoPE) → #724 (LogitsClip)
                    │
                    ├──→ #726 (FP8+RoPE)
                    └──→ #720 (FlashInfer baselines)
#727 (ALiBi)              independent
```

---

## Summary

Add comprehensive GQA kernel variants to support RoPE, ALiBi, FP8 quantization, logits soft capping, and explicit split-KV, covering the full combination matrix needed by production LLMs.

## Motivation

Current GQA kernels only support basic attention (no positional encoding, no quantization, no logits clipping). All major LLMs require at least one of these features:

| Model | RoPE | ALiBi | FP8 | LogitsClip | Sliding Window |
|-------|:----:|:-----:|:---:|:----------:|:--------------:|
| LLaMA 3 | ✓ | | ✓ (inference) | | |
| Mistral/Mixtral | ✓ | | ✓ | | ✓ |
| Qwen2.5 | ✓ | | ✓ | | ✓ |
| Gemma 2 | ✓ | | | ✓ | |
| DeepSeek-V3 | ✓ | | ✓ | | ✓ |
| BLOOM | | ✓ | | | |
| MPT | | ✓ | | | |

## Interface Design (FlashInfer-style)

All GQA Ops share a unified parameter interface. RoPE frequencies are computed inside the kernel from `rope_theta` + `rope_scale` (no external cos/sin tables needed). ALiBi slopes are derived from `num_heads` internally.

```python
class GQAFwdOp(Op):
    def __init__(
        self,
        batch, heads, heads_kv, seq_len, dim,
[...]

PR #728, opened 2026-03-31
[Refactor][Op] Split decode kernels into no-split, splitKV, and combine

### LINKED ISSUE #722: [Refactor][Op] Split decode kernels into no-split, splitKV, and combine
## Summary

Refactor existing GQA/MHA decode kernels to separate the no-split and split-KV code paths into independent kernel classes, plus a shared combine kernel.

Parent issue: #719

## Motivation

Current decode kernels auto-dispatch split vs no-split internally within a single class. This causes:
- Mixed grid dimensions and output layouts in one kernel
- Different autotune search spaces conflated
- Code duplication when adding feature combinations (each feature must implement both paths)
- Potential bugs during kernel registration

## Current state

| Kernel class | Internal paths |
|-------------|---------------|
| `gqa_decode_kernel` | no-split + split+combine |
| `gqa_decode_paged_kernel` | no-split + split+combine |
| `mha_decode_kernel` | no-split + split+combine |
| `mha_decode_paged_kernel` | no-split + split+combine |

The split kernel and combine are already separate macros (`_gqa_decode_split` + `combine`) inside the same `prim_func`. This refactor promotes them to independent kernel classes.

## Target state

| New kernel class | Responsibility |
|-----------------|---------------|
| `GQADecodeKernel` | No-split, short sequences |
| `GQADecodeSplitKVKernel` | Split, outputs partial_o + partial_lse |
| `GQADecodeCombineKernel` | Merges partial results via online softmax (shared) |
| `GQADecodePagedKernel` | No-split paged |
| `GQADecodePagedSplitKVKernel` | Split paged |
| `GQADecodePagedCombineKernel` | Combine paged (shared) |
| `MHADecodeKernel` | No-split |
| `MHADecodeSplitKVKernel` | Split |
| `MHADecodeCombineKernel` | Combine (shared) |
| `MHADecodePagedKernel` | No-split paged |
| `MHADecodePagedSplitKVKernel` | Split paged |
| `MHADecodePagedCombineKernel` | Combine paged (shared) |

Also renames snake_case kernel classes to PascalCase per naming convention.

### `num_split` alignment between split and combine

`num_split` is determined by the split kernel's autotune config. The Op reads `num_split` from the split kernel's config after autotune and uses it to construct the combine kernel:

```python
class GQADecodeOp(Op):
    def __init__(self, batch, heads, heads_kv, seqlen_kv, dim, dtype, tune=False):
        self.no_split_kernel = GQADecodeKernel(batch, heads, heads_kv, seqlen_kv, dim, dtype)
        self.split_kernel = GQADecodeSplitKVKernel(batch, heads, heads_kv, seqlen_kv, dim, dtype, tune=tune)
        
        # After autotune, read num_split from split kernel's config
        num_split = self.split_kernel.config["num_split"]
        self.combine_kernel = GQADecodeCombineKernel(batch, heads, dim, num_split, dtype)
        
        # Pre-allocate intermediate tensors
        self.glse = torch.empty(batch, heads, num_split, dtype=dtype, device="cuda")
        self.partial_o = torch.empty(batch, heads, num_split, dim, dtype=dtype, device="cuda")
```

### Autotune strategy

- **Split kernel**: autotunes over `{block_H, block_N, num_split, num_stages, threads}` (same as current)
- **Combine kernel**: no autotune needed — it is a simple reduction over `num_split` partial results, performance is not sensitive to tuning
- **No-split kernel**: autotunes over `{block_H, block_N, num_stages, threads}` independently

### Intermediate tensor pre-allocation

`glse` and `Output_partial` shapes depend on `num_split`. Currently they are allocated in `forward()` every call. After refactor, the Op pre-allocates them in `__init__()` based on the autotuned `num_split`:

```python
# Pre-allocated in __init__:
self.glse = torch.empty(batch, heads, num_split, dtype=dtype, device="cuda")
self.partial_o = torch.empty(batch, heads, num_split, dim, dtype=dtype, device="cuda")
```

### Combine kernel is universal

The combine kernel only depends on `{batch, heads, num_split, dim}` and performs online softmax merging. It does **not** depend on any attention feature (RoPE, FP8, LogitsClip, etc.), so a single combine kernel class is shared across all feature combinations.

### Op dispatch (runtime)

```python
class GQADecodeOp(Op):
    def forward(self, Q, K, V, real_seqlen_kv):
        num_split = self.split_kernel.config["num_split"]
        block_N = self.split_kernel.config["block_N"]

        if real_seqlen_kv < num_split * block_N:
            return self.no_split_kernel(Q, K, V, real_seqlen_kv)

        self.split_kernel(Q, K, V, real_seqlen_kv, self.glse, self.partial_o)
        return self.combine_kernel(self.partial_o, self.glse)
```

### Split kernel output contract

| Output | Shape | Description |
|--------|-------|-------------|
| `partial_o` | `[batch, heads, num_split, dim]` | Per-split weighted output (already divided by local logsum) |
| `glse` | `[batch, heads, num_split]` | Per-split log-sum-exp in log2 space: `log2(logsum) + scores_max * scale` |

The combine kernel uses these to reconstruct the final output via:
```
lse_max = max(glse[..., :])
w_k = exp2(glse[..., k] - lse_logsum)
output = sum(w_k * partial_o[..., k, :])
```

## Scope

- 4 existing kernel classes → 12 kernel classes
- No new features, pure refactor
- Tests and benchmarks must continue to pass
- This is a prerequisite for #719 feature kernels (RoPE, FP8, etc.)

superAngGao:
## Summary

Refactor all 4 decode kernel classes from monolithic split/no-split dispatch into independent kernel classes. Each old class becomes three:

| Old | New (no-split) | New (split) | New (combine) |
|-----|---------------|-------------|---------------|
| `gqa_decode_kernel` | `GQADecodeKernel` | `GQADecodeSplitKVKernel` | `GQADecodeCombineKernel` |
| `gqa_decode_paged_kernel` | `GQADecodePagedKernel` | `GQADecodePagedSplitKVKernel` | `GQADecodePagedCombineKernel` |
| `mha_decode_kernel` | `MHADecodeKernel` | `MHADecodeSplitKVKernel` | `MHADecodeCombineKernel` |
| `mha_decode_paged_kernel` | `MHADecodePagedKernel` | `MHADecodePagedSplitKVKernel` | `MHADecodePagedCombineKernel` |

Also renames snake_case kernel classes to PascalCase per naming convention. Old names kept as backward-compat aliases.

Closes #722

## Motivation

- Independent kernels avoid mixed grid dimensions and output layouts in one class
- Separate autotune search spaces (no-split tunes `{block_H, block_N}`, split tunes `{block_H, block_N, num_split}`)
- Combine kernel is feature-agnostic — shared across future RoPE/FP8/ALiBi variants (#719)
- Eliminates potential bugs during kernel registration

## Design

- `num_split` alignment: Op reads from split kernel's autotuned config, passes to combine kernel constructor
- Autotune: only on split kernel; combine is a simple reduction (fixed threads=128)
- Intermediate tensors (`glse`, `Output_partial`): allocated by tilelang via `out_idx`
- Op dispatch: `real_seqlen_kv < num_split * block_N` → no-split, else → split + combine

## Test plan

- [x] `test_gqa_decode.py` — 3/3 passed
- [x] `test_gqa_decode_paged.py` — 7/7 passed
- [x] `test_mha_decode.py` — 3/3 passed
- [x] `test_mha_decode_paged.py` — 4/4 passed
- [x] All 17 decode tests pass together (no cross-contamination)


Gabbering (commented):
## goose goose review — `b8f5fc61`

*honk.* The master has split monolithic kernels into three pieces each. Admirable ambition. The split logic, however, has been copy-pasted into the Op layer with a fresh bug baked in, and the backward-compat alias points somewhere that will confuse anyone who actually relied on the old API.

### goose Bugs

- `tileops/ops/gqa_decode_paged.py:73-74` / `tileops/ops/mha_decode_paged.py:73-74` — The split kernel's `forward()` signature now expects a raw `split_length` tensor, but the Op computes `acc_split_length = torch.cumsum(split_length, dim=1)` and passes **that** as the `split_length` argument. The kernel internally reads `split_length_shared[sid]` to compute `loop_range = T.ceildiv(split_length_shared[sid], block_N)` — it expects *per-split lengths*, not *cumulative* lengths. Passing cumulative sums means the first split processes the correct amount, but every subsequent split processes way too many iterations (the cumulative total up to that point). This will produce wrong attention outputs for every batch element that takes the split path. The old monolithic kernel in `gqa_decode_paged` *did* pass `acc_split_length` because the old macro consumed it as cumulative — but now the paged split kernel (`gqa_decode_paged.py:144` and `mha_decode_paged.py:208`) uses `split_length_shared[sid]` directly as a length (with `offset = split_length_shared[sid - 1] // block_N`). So the paged kernels actually *do* expect cumulative prefix sums. That means the variable is confusingly named `split_length` at the Op call site but the kernel consumes it as cumulative offsets. Either rename or — more critically — verify which semantic the kernel actually expects, because the non-paged GQA/MHA ops pass *non-cumulative* per-split lengths for those same-named parameters. If the paged kernel expects cumulative and the non-paged kernel expects raw lengths, you have an inconsistent abstraction that will blow up the moment anyone swaps one for the other.

- `tileops/ops/gqa_decode_paged.py:67` / `tileops/ops/mha_decode_paged.py:67` — `chunk_size = real_max // (num_split * block_N) * block_N`. When `real_max` is exactly `threshold` (i.e. `num_split * block_N`), `chunk_size = block_N`, and `split_length[:, -1] = real_max - (num_split - 1) * block_N = block_N`. Fine. But when `real_max` is, say, `threshold + 1`, `chunk_size` is still `block_N` (integer division floors), and `split_length[:, -1] = threshold + 1 - (num_split - 1) * block_N = block_N + 1`. That last split now has one extra element — which is correct — but the identical logic in the non-paged ops (`gqa_decode.py:78`, `mha_decode.py:78`) uses `real_seqlen_kv` (a Python int) in the same formula. The real concern: when `real_max < num_split * block_N` but `real_max >= threshold`, you get `chunk_size = 0` because `real_max // (num_split * block_N) = 0`, meaning all `num_split - 1` splits get length 0, and the last split gets the entire `real_max`. You just launched `num_split` GPU thread blocks where `num_split - 1` of them do nothing. Not wrong, but wasteful — and the threshold check that's supposed to prevent this (`real_max < threshold` where `threshold = num_split * block_N`) is identical to the condition that produces `chunk_size = 0`, so this case literally cannot occur. That's fine. But master, please realize the threshold and the chunk formula are tautologically coupled — they can't diverge, or you get silent waste.

- `tileops/kernels/flash_decode/gqa_decode.py:525` / `tileops/kernels/flash_decode/mha_decode.py:601` — The backward-compat alias `gqa_decode_kernel = GQADecodeSplitKVKernel` / `mha_decode_kernel = MHADecodeSplitKVKernel`. The old `gqa_decode_kernel.forward()` accepted `(Q, K, V, real_seqlen_kv)` — four args — and internally handled the split-vs-nosplit dispatch plus `split_length` construction. The new `GQADecodeSplitKVKernel.forward()` requires `(Q, K, V, real_seqlen_kv, split_length)` — five args. Any existing caller using the old name `gqa_decode_kernel()(q, k, v, seqlen)` will now get a `TypeError: forward() missing 1 required positional argument: 'split_length'`. That's not backward compatibility, that's a booby trap. Same applies to all four aliases.

- `tileops/kernels/flash_decode/__init__.py:3-4` — `gqa_decode_kernel` is no longer exported from `gqa_decode.py` (the `__all__` was changed), yet `__init__.py` still tries to import `gqa_decode_paged_kernel` from `gqa_decode_paged` (which *does* export it as an alias) and `mha_decode_kernel` / `mha_decode_paged_kernel` from their respective modules. But `gqa_decode_kernel` is **not** imported anywhere in the flash_decode `__init__.py`, and `tileops/kernels/__init__.py:39` still lists it in `__all__` but never imports it. So `from tileops.kernels import gqa_decode_kernel` will raise `ImportError`. The old name survived for paged/mha variants but was dropped for plain GQA.

### feather Performance

- `tileops/ops/gqa_decode.py:71` / `tileops/ops/mha_decode.py:71` / `tileops/ops/gqa_decode_paged.py:60` / `tileops/ops/mha_decode_paged.py:60` — `split_length` tensor allocation happens on every `forward()` call that takes the split path. This is a small `torch.full` + indexing on GPU, but it's a host-side kernel launch + sync per decode step in the hot serving path. The split lengths are deterministic given `(real_seqlen_kv, num_split, block_N)` — cache them or at minimum pre-allocate the buffer once in `__init__` and fill in-place.

- `tileops/ops/gqa_decode_paged.py:55` / `tileops/ops/mha_decode_paged.py:55` — `real_seqlen_kv.max().item()` is a device-to-host sync on every forward call. The old code had this same problem, but now that you're refactoring anyway, this is the goose's moment to point out: you're stalling the entire GPU pipeline to read one integer back to the CPU just to decide a dispatch branch. Consider passing `max_seqlen_kv` as a pre-computed host-side value.

### egg Test gaps

- No tests exercise the **Op-level** dispatch logic (`gqa_decode.py`, `mha_decode.py`, `gqa_decode_paged.py`, `mha_decode_paged.py`). The PR says tests pass for the kernel-level test files, but the split-length computation, the threshold dispatch, and the split→combine pipeline are all new code in the Op layer. A `real_seqlen_kv` that is exactly at the threshold boundary, or one element above/below, is never tested at the Op level — and that's where the cumulative-vs-raw `split_length` mismatch lives.

- The backward-compat aliases (`gqa_decode_kernel`, `mha_decode_kernel`, etc.) have no import or smoke test. Given the alias now points to a different signature, a single `from tileops.kernels import gqa_decode_kernel` in a test would have caught the breakage immediately.

superAngGao:
Closed since the repo has too many commits about refactors.

PR #731, opened 2026-03-31
[Feat][POOL] Implement averagepool1d, averagepool2d, averagepool3d

### LINKED ISSUE #678: [FEAT][POOL] implement averagepool1d, averagepool2d, averagepool3d
### Symptom / Motivation
Issue #402 tracks the implementation of the conv & pooling operator family. Standard average pooling is still missing from TileOPs because there are no generic `averagepool1d`, `averagepool2d`, or `averagepool3d` operators yet.

These operators are still important coverage gaps:
- `averagepool1d` is required for temporal, audio, and sequence downsampling workloads
- `averagepool2d` is required for standard vision and CNN workloads
- `averagepool3d` is required for video, volumetric, and spatiotemporal workloads

Completing them brings the standard average-pooling family in TileOPs closer to parity with the three core PyTorch average pooling APIs.

### Root Cause Analysis
N/A - Feature implementation (new operators).

The repository already contains convolution kernels for 1D/2D/3D and a specialized mean-pooling kernel for DeepSeek NSA, but there is no generic average-pooling kernel/op/test/benchmark path for the standard PyTorch-style pooling APIs.

### Related Files
- `tileops/kernels/conv/conv1d.py` — existing sliding-window reference for 1D spatial indexing
- `tileops/kernels/conv/conv2d.py` — existing sliding-window reference for 2D spatial indexing
- `tileops/kernels/conv/conv3d.py` — existing sliding-window reference for 3D spatial indexing
- `tileops/kernels/deepseek_nsa/mean_pooling_fwd.py` — existing reduction-style reference for mean pooling
- `tileops/kernels/pool/` — add: kernel package for generic average pooling
- `tileops/ops/avg_pool1d.py` — add: Op wrapper for AveragePool1d
- `tileops/ops/avg_pool2d.py` — add: Op wrapper for AveragePool2d
- `tileops/ops/avg_pool3d.py` — add: Op wrapper for AveragePool3d
- `tileops/kernels/__init__.py` — modify: export new kernel classes or helpers if needed
- `tileops/ops/__init__.py` — modify: export new ops
- `tests/ops/test_avg_pool1d.py` — add: correctness tests
- `tests/ops/test_avg_pool2d.py` — add: correctness tests
- `tests/ops/test_avg_pool3d.py` — add: correctness tests
- `benchmarks/ops/bench_avg_pool1d.py` — add: performance benchmarks
- `benchmarks/ops/bench_avg_pool2d.py` — add: performance benchmarks
- `benchmarks/ops/bench_avg_pool3d.py` — add: performance benchmarks

## Goal
Implement `averagepool1d`, `averagepool2d`, and `averagepool3d` following the existing TileOPs 2-layer architecture (Kernel L1 -> Op L2), reusing the current convolution sliding-window structure where appropriate and validating correctness against PyTorch.

## Plan

1. Study the existing `conv1d`, `conv2d`, and `conv3d` kernels to extract reusable sliding-window indexing, launch, and API patterns.
2. Study the existing specialized mean-pooling kernel to extract reusable reduction and averaging patterns.
3. Implement TileLang kernels for `averagepool1d`, `averagepool2d`, and `averagepool3d` with semantics aligned with `torch.nn.functional.avg_pool1d`, `torch.nn.functional.avg_pool2d`, and `torch.nn.functional.avg_pool3d`.
4. Add L2 op wrappers for all three operators with the same integration style as existing TileOPs ops.
5. Update exports in `tileops/kernels/__init__.py` and `tileops/ops/__init__.py` as needed.
6. Add correctness tests covering representative shapes, stride/padding settings, and comparison against PyTorch reference.
7. Add benchmarks reporting latency and DRAM bandwidth for representative 1D, 2D, and 3D average-pooling workloads.

## Constraints
- Must follow TileOPs 2-layer architecture (Kernel L1 -> Op L2)
- Must follow existing convolution and reduction structure and naming conventions where applicable
- CUDA only; CPU support is out of scope
- Cannot change any existing public APIs or break existing tests
- Parent issue: #402

## Acceptance Criteria
- [ ] AC-1: AveragePool1d forward kernel and Op are implemented and pass PyTorch reference tests
- [ ] AC-2: AveragePool2d forward kernel and Op are implemented and pass PyTorch reference tests
- [ ] AC-3: AveragePool3d forward kernel and Op are implemented and pass PyTorch reference tests
- [ ] AC-4: All three ops are exported through the standard TileOPs op surface
- [ ] AC-5: All three ops include benchmarks reporting latency and DRAM bandwidth
- [ ] AC-6: The standard average-pooling family tracked by #402 is complete for `averagepool1d`, `averagepool2d`, and `averagepool3d`

RMLYC:
## Summary
- add generic TileLang average pooling kernels for 1D, 2D, and 3D workloads under the TileOPs L1/L2 architecture
- export the new avg pool ops and add correctness coverage against PyTorch
- add avg pool benchmarks and move PyTorch layout permutes outside `ref_program` so baseline timing excludes data-layout conversion

## Validation
- `python -m pytest tests/ops/test_avg_pool1d.py tests/ops/test_avg_pool2d.py tests/ops/test_avg_pool3d.py -v`
- `python -m pytest benchmarks/ops/bench_avg_pool1d.py -v`
- `python -m pytest benchmarks/ops/bench_avg_pool2d.py -v`
- `python -m pytest benchmarks/ops/bench_avg_pool3d.py -v`
- reran benchmark comparisons after removing baseline permute cost

## Benchmark Snapshot

Source: `profile_run.log` generated on 2026-04-01 17:13:52. Values below follow the rounded display precision in that report.

### avg_pool1d

| Case | TileOps latency (ms) | Torch latency (ms) | TileOps TFLOPS | Torch TFLOPS | TileOps BW (TB/s) | Torch BW (TB/s) |
| audio-downsample-fp16 | 0.01 | 0.02 | 0.42 | 0.14 | 0.84 | 0.28 |
| long-temporal-fp16 | 0.03 | 0.10 | 0.75 | 0.20 | 1.50 | 0.39 |
| ceil-bf16 | 0.00 | 0.01 | 0.24 | 0.15 | 0.36 | 0.23 |

### avg_pool2d

| Case | TileOps latency (ms) | Torch latency (ms) | TileOps TFLOPS | Torch TFLOPS | TileOps BW (TB/s) | Torch BW (TB/s) |
| vision-3x3-s2 | 0.01 | 0.01 | 0.31 | 0.27 | 0.34 | 0.30 |
| vision-5x5-s2 | 0.01 | 0.01 | 0.56 | 0.36 | 0.22 | 0.14 |
| ceil-divisor-bf16 | 0.01 | 0.01 | 0.34 | 0.32 | 0.22 | 0.21 |

### avg_pool3d

| Case | TileOps latency (ms) | Torch latency (ms) | TileOps TFLOPS | Torch TFLOPS | TileOps BW (TB/s) | Torch BW (TB/s) |
| video-2x2x2 | 0.00 | 0.01 | 0.40 | 0.12 | 0.90 | 0.26 |
| ceil-video | 0.01 | 0.02 | 0.28 | 0.16 | 0.21 | 0.12 |
| divisor-bf16 | 0.00 | 0.01 | 0.08 | 0.05 | 0.10 | 0.07 |

Closes #678.

### DIFF
--- a/tileops/kernels/pool/common.py
+++ b/tileops/kernels/pool/common.py
@@ -1,6 +1,6 @@
+import warnings
 from collections.abc import Sequence
 from typing import Tuple
-import warnings
 
 
 def _normalize_pool_dims(name: str, value: int | Sequence[int], ndim: int) -> tuple[int, ...]:


tileops/kernels/pool/avg_pool2d.py
```diff
@@ -0,0 +1,267 @@
...
+                        start_w = ow * stride_w - pad_w
+                        end_h = start_h + kernel_h
+                        end_w = start_w + kernel_w
+                        valid_h = T.max(T.min(end_h, h_in) - T.max(start_h, 0), 0)
+                        valid_w = T.max(T.min(end_w, w_in) - T.max(start_w, 0), 0)
+                        valid_count = valid_h * valid_w
+                        padded_h = T.max(T.min(end_h, h_in + pad_h) - T.max(start_h, -pad_h), 0)
+                        padded_w = T.max(T.min(end_w, w_in + pad_w) - T.max(start_w, -pad_w), 0)
+                        padded_count = padded_h * padded_w
+                        divisor = T.max(
```
lcy-seso: This changes the public avg_pool2d interface semantics relative to PyTorch. `divisor_override` is only invalid when it is `0`; negative values are accepted and participate in the division. Here the final divisor is clamped with `max(..., 1)`, so any negative override is silently rewritten to `1` and produces different outputs from `F.avg_pool2d`. Please split the two paths: keep the safety clamp only for the auto-computed valid/padded counts, and validate `divisor_override != 0` separately instead of clamping it.

tileops/kernels/pool/common.py
```diff
@@ -0,0 +1,37 @@
+from typing import Tuple
+
+
+def normalize_1d(value: int | Tuple[int]) -> int:
```
lcy-seso, line 25: The API boundary is still missing parameter validation, and this helper makes that visible. For 1D specifically, `normalize_1d((3, 4))` will silently become `3`, whereas `torch.avg_pool1d` rejects any tuple whose arity is not exactly 1. More broadly, the new avg_pool ops also do not validate `stride > 0`, illegal padding, or `divisor_override == 0`, so a number of invalid user inputs will either be accepted silently or fail with implementation-specific errors instead of stable API errors. Please add shared normalization + validation helpers here and use them from all three avg_pool ops.

tileops/kernels/pool/avg_pool3d.py
```diff
@@ -0,0 +1,316 @@
...
+                        end_w = start_w + kernel_w
+                        valid_d = T.max(T.min(end_d, d_in) - T.max(start_d, 0), 0)
+                        valid_h = T.max(T.min(end_h, h_in) - T.max(start_h, 0), 0)
+                        valid_w = T.max(T.min(end_w, w_in) - T.max(start_w, 0), 0)
+                        valid_count = valid_d * valid_h * valid_w
+                        padded_d = T.max(T.min(end_d, d_in + pad_d) - T.max(start_d, -pad_d), 0)
+                        padded_h = T.max(T.min(end_h, h_in + pad_h) - T.max(start_h, -pad_h), 0)
+                        padded_w = T.max(T.min(end_w, w_in + pad_w) - T.max(start_w, -pad_w), 0)
+                        padded_count = padded_d * padded_h * padded_w
+                        divisor = T.max(
```
lcy-seso: Same issue here as in 2D: clamping the final divisor with `max(..., 1)` changes `divisor_override` semantics. In PyTorch, `divisor_override=-1` is legal; only `0` is rejected. This implementation silently converts negative overrides into `1`, so the op is no longer API-compatible. Please handle `divisor_override` as an explicit validated override rather than feeding it through the generic nonzero-count clamp.

tileops/ops/avg_pool2d.py
```diff
@@ -0,0 +1,70 @@
...
+from tileops.kernels.kernel import Kernel
+from tileops.kernels.pool import AvgPool2dKernel
+from tileops.kernels.pool.common import normalize_2d
+
+from .op import Op
+
+__all__ = ["AvgPool2dOp"]
+
+
+class AvgPool2dOp(Op):
```
lcy-seso, line 18: The layout contract is implicit right now. This op takes `(N, H, W, C)` tensors, but that is only discoverable by reading the kernel shape and the tests that permute to/from PyTorch `NCHW`. There is no `layout` parameter, no docstring, and no input-shape validation explaining the expected format. For a public operator API this is very easy to misuse, because the name and argument list look like `torch.avg_pool2d` while the tensor layout contract is different. Please make the layout explicit at the API boundary, either by documenting/validating channels-last inputs or by adding an explicit `layout` contract.

tileops/kernels/pool/avg_pool2d.py
RMLYC: Addressed in `03321ea`.

Updated the 2D divisor path so only the auto-computed valid/padded count is clamped to be nonzero. Explicit `divisor_override` now bypasses that clamp, and `divisor_override == 0` is rejected separately at the API boundary. This preserves PyTorch-compatible behavior for negative overrides.

tileops/ops/avg_pool2d.py
RMLYC: Addressed in `03321ea`.

I made the channels-last contract explicit at the op boundary. The avg_pool ops now document the expected layouts (`NLC`, `NHWC`, `NDHWC`) in their docstrings, and `forward()` validates the input shape so layout misuse fails with a clear error instead of remaining implicit. I kept the API surface unchanged rather than adding a new `layout` parameter in this pass.

tileops/kernels/pool/avg_pool3d.py
RMLYC: Addressed in `03321ea`.

Applied the same fix to 3D: explicit `divisor_override` no longer flows through the generic `max(..., 1)` clamp, and only `divisor_override == 0` is rejected at validation time. Negative overrides are preserved so the 3D public semantics stay aligned with PyTorch.

tileops/kernels/pool/common.py
RMLYC: Addressed in `03321ea`.

I factored shared pool parameter handling into `tileops/kernels/pool/common.py` and now route 1D/2D/3D avg_pool ops through shared normalization and validation helpers. The shared checks now cover tuple arity, `stride > 0`, bounded non-negative padding, and `divisor_override != 0`, with regression tests added for the invalid-input cases.

tileops/kernels/pool/common.py
```diff
@@ -0,0 +1,98 @@
+from collections.abc import Sequence
+from typing import Tuple
+
+
+def _normalize_pool_dims(name: str, value: int | Sequence[int], ndim: int) -> tuple[int, ...]:
+    if isinstance(value, int):
```
lcy-seso, line 10: This still treats `bool` as a valid pooling dimension because `bool` is a subclass of `int` in Python. From a public operator API perspective that is too permissive: `kernel_size=True`, `stride=True`, and `padding=True` are currently accepted here, while PyTorch rejects them with a type error. The same issue also exists for `divisor_override`, which is only checked for `== 0` and therefore still accepts values like `1.5` or `True`. I would tighten the shared validation so these parameter types are rejected deterministically at the API boundary.

tileops/kernels/pool/common.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def validate_channels_last_input(
+    *,
+    op_name: str,
+    x_shape: tuple[int, ...],
+    expected_shape: tuple[int, ...],
+    layout: str,
+) -> None:
+    if x_shape != expected_shape:
```
lcy-seso, line 84: This only validates shape equality, not layout. That means the new “channels-last contract” still does not fail reliably for user mistakes: an incorrect `NCHW` tensor will silently pass whenever the dimensions are ambiguous, e.g. `(N, C, H, W) == (N, H, W, C)` for shapes like `(1, 8, 8, 8)`. From a library-user perspective that is still an unsafe API boundary, because the misuse produces wrong semantics instead of a clear error. At minimum this limitation should be acknowledged and covered by regression tests; ideally the contract should be made explicit in a way that does not rely on shape coincidence.

lcy-seso (changes requested):
This PR is clearly better than the initial version, and the follow-up commits did address several important semantic issues. In particular, the 2D/3D `divisor_override` behavior is now much closer to PyTorch, the shared normalization/validation helpers are a good direction, and the channels-last contract is at least documented explicitly now.

The main remaining concerns are:

- The layout contract is documented, but not actually enforced reliably. `forward()` only compares shapes, so an incorrect `NCHW` input can still silently pass as `NHWC` whenever the dimensions are ambiguous (for example `(1, 8, 8, 8)`). From a user perspective, silent semantic misuse is much worse than an eager failure.
- The shared parameter validation is still not stable / PyTorch-compatible at the type boundary. Today the ops still accept inputs like `kernel_size=True`, `stride=True`, `padding=True`, and `divisor_override=1.5/True`, while PyTorch rejects them. For a public operator API, these cases should be rejected deterministically at construction time.
- The regression tests cover several obvious invalid-input cases, but they still miss the highest-risk user-facing paths: ambiguous-layout inputs and invalid-but-accepted parameter types.

I think this is close, but I still recommend tightening the API boundary before merge.

tileops/kernels/pool/common.py
RMLYC: done

tileops/kernels/pool/common.py
RMLYC: report warnings when (N, C, H, W) == (N, H, W, C)

PR #732, opened 2026-03-31
[CI] align gpu-smoke warmup with smoke targets

### LINKED ISSUE #729: [PERF][CI] align gpu-smoke warmup with selected smoke targets
### Symptom / Motivation
`gpu-smoke` currently pays kernel compilation / autotune setup cost inside the main correctness job. Unlike `nightly`, it has no warmup pre-stage in the same workflow, so targeted smoke runs still spend time compiling kernels before executing assertions. In addition, the trusted venv cache key is derived from the full `pyproject.toml`, which causes unnecessary venv invalidation when non-dependency fields change.

### Root Cause Analysis
- `.github/workflows/gpu-smoke.yml` does not have a pre-job that warms the same pytest subset selected by `security-policy`.
- The runtime state hash uses the whole `pyproject.toml` instead of dependency-only inputs, so harmless metadata/config edits can force a fresh venv.
- The existing warmup helper (`scripts/warmup_kernel_cache.py`) is benchmark-oriented and cannot directly consume the exact `pytest_targets` subset that `gpu-smoke` will run.
- Without aligned warmup scope, preheating unrelated tests wastes time, while skipping warmup leaves `gpu-smoke` to absorb first-run JIT cost.

### Related Files
- `scripts/warmup_kernel_cache.py`
- `scripts/conftest_warmup.py`
- `scripts/dep_hash.py`

## Goal
Reduce `gpu-smoke` wall time without introducing Docker by adding a same-workflow `cache-warmup` pre-job that warms exactly the smoke-marked pytest subset selected for `gpu-smoke`, while improving trusted venv reuse with a dependency-only cache key.

## Plan

1. Add a `cache-warmup` job to `.github/workflows/gpu-smoke.yml` and make `gpu-smoke` depend on it.
2. Reuse `security-policy.outputs.pytest_targets` so `cache-warmup` warms the exact subset that `gpu-smoke` will execute.
3. Restrict both warmup and the main GPU smoke run to `-m smoke`.
4. Switch the trusted venv cache key from full-`pyproject.toml` hashing to `scripts/dep_hash.py`.
5. Extend `scripts/warmup_kernel_cache.py` to accept explicit pytest targets and a marker expression.
6. Keep the existing trusted/fork cache isolation model unchanged.

## Constraints
- Do not introduce Docker for `gpu-smoke`.
- Keep warmup in the same `gpu-smoke.yml` workflow as a prerequisite job.
- Warm only the selected smoke subset; do not expand to `full` coverage in this change.

## Acceptance Criteria
- [ ] `gpu-smoke` contains a `cache-warmup` prerequisite job in the same workflow.
- [ ] `cache-warmup` warms the same `pytest_targets` subset that `gpu-smoke` runs.
- [ ] Both warmup and `gpu-smoke` execute only `-m smoke`.
- [ ] Trusted venv reuse is keyed by dependency-only inputs via `scripts/dep_hash.py`.
- [ ] Modified files pass unit tests or workflow validation relevant to the change.

RMLYC:
Closes #729

## Summary

- add a `cache-warmup` prerequisite job to `gpu-smoke.yml` and make `gpu-smoke` depend on it
- align warmup scope with `security-policy.outputs.pytest_targets` so warmup and smoke run the same `-m smoke` subset
- switch the trusted gpu-smoke venv cache key to `scripts/dep_hash.py` and reuse local wheels when available
- extend `scripts/warmup_kernel_cache.py` and `scripts/conftest_warmup.py` so correctness smoke subsets can be warmed without benchmark-only assumptions

## Test plan

- [x] pre-commit passed during `git commit`
- [x] `python scripts/warmup_kernel_cache.py --help`
- [ ] GitHub Actions workflow execution not run locally

## Additional context

- `docs/CONTRIBUTING.md` and `scripts/validate.sh` referenced by the local PR skill are not present in the current checkout, so this PR was created using the repo's existing title conventions and PR template structure.

### DIFF
--- a/.github/workflows/gpu-smoke.yml
+++ b/.github/workflows/gpu-smoke.yml
@@ -335,7 +335,7 @@ jobs:
             echo "reason=$reason"
           } >> "$GITHUB_OUTPUT"
 
-  gpu-smoke:
+  cache-warmup:
     needs: [ci-prereq, security-policy]
     if: ${{ github.repository == 'tile-ai/TileOPs' && needs.ci-prereq.outputs.skip != 'true' && needs.security-policy.outputs.skip_gpu_smoke != 'true' }}
     runs-on: [self-hosted, tile-ops, venv]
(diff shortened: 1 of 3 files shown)


.github/workflows/gpu-smoke.yml
```diff
@@ -580,10 +832,7 @@ jobs:
               print(f"{key}={os.environ.get(key)}")
           PY
 
-          TEST_MARK_EXPR="smoke or full"
-          if [[ "$EVENT_NAME" == "pull_request" ]]; then
-            TEST_MARK_EXPR="smoke"
-          fi
+          TEST_MARK_EXPR="smoke"
```
lcy-seso, line 835: **[Needs data]** This changes push-to-main from `-m "smoke or full"` (~1871 tests, ~11 min) to `-m "smoke"` only.

`full`-tier tests cover broader dtypes, larger shapes, and `tune=True` paths. With this change, those are deferred to nightly (~24h gap). Could you share how long a smoke-only run takes? If the saving is marginal (e.g. 11 min → 8 min), the coverage tradeoff is hard to justify. If it's significant (e.g. 11 min → 3 min), the gap is more defensible.

scripts/warmup_kernel_cache.py
```diff
@@ -59,19 +53,24 @@ def main():
...
           f"x {args.max_workers} compile threads each")
 
-    # Collect benchmark files and shard.
-    bench_dir = os.path.join(os.path.dirname(__file__), "..", "benchmarks", "ops")
-    all_files = sorted(glob.glob(os.path.join(bench_dir, "bench_*.py")))
-    shard_files = all_files[args.shard::args.total_shards]
+    if args.pytest_targets:
+        selected_targets = args.pytest_targets
+        print(f"Using explicit pytest targets ({len(selected_targets)}):")
+        for target in selected_targets:
```
lcy-seso, line 59: **[Code]** Agree with the earlier review: when `--pytest-targets` is provided, `--shard`/`--total-shards` is silently ignored. If a future CI matrix passes explicit targets into a sharded job, every shard runs the full list redundantly.

Suggestion: unify into one list, then apply sharding unconditionally:
```python
if args.pytest_targets:
    all_targets = args.pytest_targets
else:
    all_targets = sorted(glob.glob(...))
selected_targets = all_targets[args.shard::args.total_shards]
```

.github/workflows/gpu-smoke.yml
```diff
@@ -335,7 +335,7 @@ jobs:
             echo "reason=$reason"
           } >> "$GITHUB_OUTPUT"
 
-  gpu-smoke:
+  cache-warmup:
     needs: [ci-prereq, security-policy]
     if: ${{ github.repository == 'tile-ai/TileOPs' && needs.ci-prereq.outputs.skip != 'true' && needs.security-policy.outputs.skip_gpu_smoke != 'true' }}
```
lcy-seso, line 340: **[Code]** The `cache-warmup` job has no `timeout-minutes`. The `gpu-smoke` job (and nightly's `cache-warmup`) both set explicit timeouts. A hung warmup (GPU deadlock, infinite compilation loop) will run until the runner's global limit.

Suggestion: add `timeout-minutes: 30` (or match the existing gpu-smoke timeout).

.github/workflows/gpu-smoke.yml
```diff
@@ -369,11 +369,258 @@ jobs:
...
+          echo "Resolved warmup targets: ${PYTEST_TARGETS}"
+          read -r -a TARGETS <<< "${PYTEST_TARGETS}"
+          echo "Warming pytest -m \"smoke\" on ${#TARGETS[@]} target(s)"
+          python scripts/warmup_kernel_cache.py \
+            --pytest-targets "${TARGETS[@]}" \
+            --marker smoke \
+            --max-workers 64 \
+            -n 8
+
+      - name: Cleanup isolated fork state
```
lcy-seso, line 580: **[Code]** The ~100-line "Resolve runtime state" block (lines 372-466) is duplicated nearly verbatim between `cache-warmup` and `gpu-smoke`. Updating one without the other will cause subtle cache-key or path divergence.

Suggestion: extract into a composite action (e.g. `.github/actions/resolve-runtime/action.yml`) or a shared shell script.

lcy-seso (commented):
## Summary

The PR correctly addresses the stated problem: warmup scope was misaligned with test scope, causing wasted compilation time. The warmup script generalization (`--pytest-targets`, `-m marker`) and conftest `ImportError` handling are clean. The `dep_hash.py` cache key switch is a clear improvement.

However, **the `TEST_MARK_EXPR` change needs explicit justification**. Dropping `full` tests from push-to-main runs (~1871 smoke+full tests in ~11 min) creates a coverage gap between merge and nightly. The issue spec prescribes this but doesn't quantify the time saved or argue that the 24-hour delay is acceptable. Before merging, the author should provide smoke-only run time so reviewers can evaluate the tradeoff.

Inline comments below flag code-level issues.

PR #778, opened 2026-04-03
[Feat][MoE] SharedFusedMoE with native shared expert kernel and TP support

### LINKED ISSUE #777: [Feat][MoE] SharedFusedMoE with native shared expert kernel and TP support
### Motivation

`SharedFusedMoE` previously accepted a user-provided `shared_experts_fn: Callable` (an external `nn.Module`) to compute the shared expert output. This design has two problems:

1. **No kernel ownership**: the shared expert computation bypasses TileOPs' kernel layer entirely, losing performance visibility and benchmark coverage.
2. **No TP support**: there is no mechanism to shard shared expert weights across TP ranks, which is required for multi-GPU inference (e.g. Kimi K2 / DeepSeek-V3 with tp_size=8).

### Solution

Replace the callback-based design with a kernel-native path:

- Add `SharedExpertMLPKernel` (TileLang): gate+up GEMM → SiLU-mul → down GEMM, operating on a weight shard of size `[2*F_s/tp, H]` / `[H, F_s/tp]`.
- Extend `SharedFusedMoE` with `shared_ffn_size`, `tp_size`, `tp_rank` parameters. The op accepts **complete** weights and shards them internally (ColumnParallel on `gate_up` dim=0, RowParallel on `down` dim=1).
- Return a **partial** `shared_out` when `tp_size > 1`; the caller is responsible for `dist.all_reduce`.

### Related Files

- `tileops/kernels/moe/shared_expert_mlp.py` — new `SharedExpertMLPKernel`
- `tileops/kernels/moe/__init__.py` — export `SharedExpertMLPKernel`
- `tileops/ops/moe/shared_fused_moe.py` — refactor interface, add TP sharding
- `tests/ops/test_moe_shared_fused_moe.py` — single-GPU correctness tests
- `tests/ops/test_moe_shared_fused_moe_distributed.py` — multi-GPU TP tests (torchrun)
- `benchmarks/ops/bench_moe_shared_fused_moe.py` — op-level benchmark
- `benchmarks/kernels/bench_moe_shared_expert_mlp.py` — kernel-level benchmark

## Plan

1. **`SharedExpertMLPKernel`**: TileLang kernel implementing `hidden [T,H] × w_gate_up [2F,H]^T → SiLU-mul → × w_down [H,F]^T → [T,H]`. Operates on the local shard size `F_s // tp_size`.

2. **`SharedFusedMoE` interface change**:
   - Remove `shared_experts_fn: Callable`
   - Add `shared_ffn_size: Optional[int]`, `tp_size: int = 1`, `tp_rank: int = 0`
   - Validate `shared_ffn_size % tp_size == 0`
   - Internally shard weights via `narrow().contiguous()` in `forward()`
   - Return `(shared_out_partial, routed_out)`; partial when `tp_size > 1`

3. **Single-GPU tests** (`test_moe_shared_fused_moe.py`):
   - `test_shared_fused_moe_basic`: shared_out vs float32 math reference, routed_out vs `FusedMoe`
   - `test_shared_fused_moe_none`: `shared_ffn_size=None` → `shared_out is None`
   - `test_shared_fused_moe_tp`: simulate TP on single GPU, accumulate partials, compare to per-shard float32 reference

4. **Multi-GPU tests** (`test_moe_shared_fused_moe_distributed.py`):
   - `smoke-tp2` (T=64, H=128, F_s=64): quick 2-GPU sanity check
   - `kimi-k2-tp2` / `kimi-k2-tp8` (T=512, H=7168, F_s=18432): real Kimi K2 scale
   - Compare TileOPs partial + manual `dist.all_reduce` vs vLLM `DeepseekV2MLP` (internal all-reduce)
   - Skip automatically when GPU count < tp_size or vLLM unavailable

5. **Benchmarks**: op-level and kernel-level benchmarks with vLLM baseline.

## Constraints

- TP support covers **shared expert only**; routed expert TP (EP via `expert_map`) is unchanged.
- Caller is responsible for `dist.all_reduce` on the partial shared output — the op does not communicate.
- `layout="padded"` is not supported for the new path (planned for removal).
- Requires `shared_ffn_size % tp_size == 0`.

## Acceptance Criteria

- [ ] `SharedExpertMLPKernel` implemented and exported from `tileops/kernels/moe`.
- [ ] `SharedFusedMoE` accepts `shared_ffn_size`, `tp_size`, `tp_rank`; old `shared_experts_fn` removed.
- [ ] Single-GPU smoke tests pass: `pytest tests/ops/test_moe_shared_fused_moe.py -m smoke`.
- [ ] Multi-GPU smoke test passes: `torchrun --nproc_per_node=2 -m pytest tests/ops/test_moe_shared_fused_moe_distributed.py -m smoke`.
- [ ] Full Kimi K2 scale TP8 test passes: `torchrun --nproc_per_node=8 -m pytest tests/ops/test_moe_shared_fused_moe_distributed.py -m full`.
- [ ] Benchmarks record TileOPs and at least one non-TileOPs baseline with non-None FLOPs/memory.

michaelwithu:
## Summary

This PR implements `SharedExpertMLPKernel` (TileLang) and refactors `SharedFusedMoE` to support Tensor Parallelism for the shared expert path. Closes #777.

---

## Motivation

The previous `SharedFusedMoE` accepted a user-provided `shared_experts_fn: Callable` (an external `nn.Module`) to compute the shared expert output. This design had two problems:

1. **No kernel ownership**: shared expert computation bypassed TileOPs' kernel layer entirely, losing performance visibility and benchmark coverage.
2. **No TP support**: no mechanism to shard shared expert weights across TP ranks, which is required for multi-GPU inference (e.g. Kimi K2 / DeepSeek-V3 with `tp_size=8`).

---

### New files

| File | Description |
|------|-------------|
| `tileops/kernels/moe/shared_expert_mlp.py` | `SharedExpertMLPKernel`: TileLang kernel implementing gate+up GEMM → SiLU-mul → down GEMM |
| `tests/ops/test_moe_shared_fused_moe_distributed.py` | Multi-GPU TP tests via `torchrun`, comparing TileOPs vs vLLM `DeepseekV2MLP` |
| `benchmarks/ops/bench_moe_shared_fused_moe.py` | Op-level benchmark with vLLM baseline |
| `benchmarks/kernels/bench_moe_shared_expert_mlp.py` | Kernel-level benchmark |

### Modified files

**`tileops/ops/moe/shared_fused_moe.py`** — interface refactor + TP support

- Remove `shared_experts_fn: Callable` parameter
- Add `shared_ffn_size: Optional[int]`, `tp_size: int = 1`, `tp_rank: int = 0`
- Op accepts **complete** weights and shards them internally:
  - `shared_w_gate_up [2*F_s, H]` → ColumnParallel, split along `dim=0`
  - `shared_w_down [H, F_s]` → RowParallel, split along `dim=1`
- Returns `(shared_out_partial, routed_out)`; when `tp_size > 1`, `shared_out_partial` is a partial sum — caller does `dist.all_reduce`

**`tileops/kernels/moe/__init__.py`** — export `SharedExpertMLPKernel`

**`tests/ops/test_moe_shared_fused_moe.py`** — rewritten with 3 tests (see Testing section)

**`tests/ops/test_moe_fused_moe_distributed.py`** — fix stale parameter name `shared_experts_fn` → `shared_experts`

---

### TP split convention (matches vLLM `DeepseekV2MLP`)

TileOPs mirrors vLLM's weight layout exactly. The caller passes complete weights; the op slices them with `narrow().contiguous()` per `tp_rank`.

```
gate_up_proj: MergedColumnParallelLinear  →  each rank holds [2*F_s/tp, H]
down_proj:    RowParallelLinear           →  each rank holds [H, F_s/tp]
```

### Partial output contract

```python
# tp_size > 1 usage
shared_partial, routed_out = op(hidden, gating, w_gate_up, w_down,
                                 shared_w_gate_up=w_gu,  # complete [2*F_s, H]
                                 shared_w_down=w_d)       # complete [H, F_s]
dist.all_reduce(shared_partial)  # caller's responsibility
final = shared_partial + routed_out
```

The routed expert path is **not affected** by TP sharding.

---

### Single-GPU (`pytest -m smoke`)

| Test | What it checks |
|------|----------------|
| `test_shared_fused_moe_basic` | `shared_out` vs float32 math reference (`atol=1e-2`); `routed_out` vs `FusedMoe` (`atol=1e-5`) |
| `test_shared_fused_moe_none` | `shared_ffn_size=None` → `shared_out is None` |
| `test_shared_fused_moe_tp` | Simulate `tp_size=2` on single GPU: accumulate partials, compare to per-shard float32 reference |

All 3 passed ✅

### Multi-GPU (`torchrun`)

| Test | Config | Command |
|------|--------|---------|
| `smoke-tp2` | T=64, H=128, F_s=64, tp_size=2 | `torchrun --nproc_per_node=2 -m pytest ... -m smoke` |
| `kimi-k2-tp2` | T=512, H=7168, F_s=18432, tp_size=2 | `torchrun --nproc_per_node=2 -m pytest ... -m full` |
| `kimi-k2-tp8` | T=512, H=7168, F_s=18432, tp_size=8 | `torchrun --nproc_per_node=8 -m pytest ... -m full` |

Compares TileOPs `SharedFusedMoE` partial + manual `dist.all_reduce` against vLLM `DeepseekV2MLP` (internal all-reduce). All passed ✅

Tests skip cleanly when not launched via `torchrun` (CI single-GPU environments).

---

## Benchmark

**Hardware**: NVIDIA H200 | **Config**: Kimi K2 scale — E=384, K=8, H=7168, F=2048, F_s=18432, bf16, sigmoid routing

| num_tokens | TileOPs (ms) | TileOPs (TFLOPS) | vLLM (ms) | vLLM (TFLOPS) | Ratio |
|-----------|-------------|-----------------|----------|--------------|-------|
| 1 | 0.95 | 1.58 | 0.50 | 3.01 | 0.53x |
| 32 | 5.46 | 8.77 | 7.50 | 6.39 | **1.37x** |
| 512 | 10.44 | 73.45 | 17.01 | 45.08 | **1.63x** |
| 2048 | 19.99 | 153.39 | 24.06 | 127.48 | **1.20x** |
| 4096 | 32.63 | 187.96 | 30.33 | 202.19 | 0.93x |

TileOPs leads at medium batch sizes (T=32~2048). At T=1 vLLM is faster (memory-bound decode, lower cuBLAS overhead); at T=4096 vLLM's Triton kernel edges ahead in the compute-bound regime.

---

## Notes

- `layout="padded"` is not supported in the new path (planned for removal).
- `shared_ffn_size % tp_size == 0` is validated at construction time.
- Routed expert TP (EP via `expert_map`) is a separate concern and unchanged.

## Follow-up

- #828 — Add `pre_sharded` fast path to eliminate per-forward ~95MB weight copies in TP mode
- #829 — Fix `enable_rasteration` → `enable_rasterization` typo across 40+ kernel configs

Suggestions: `TPFixture` should inherit `FixtureBase` from `workloads.base`; consider separate GEMM tuning configs for gate_up vs down in `SharedExpertMLPKernel`.

### DIFF
--- a/benchmarks/kernels/bench_moe_shared_expert_mlp.py
+++ b/benchmarks/kernels/bench_moe_shared_expert_mlp.py
@@ -4,11 +4,11 @@ import pytest
 import torch
 
 from benchmarks.benchmark import BenchmarkBase, BenchmarkReport
-from tests.test_base import FixtureBase, TestBase
 from tileops.kernels.moe import SharedExpertMLPKernel
+from workloads.base import FixtureBase, WorkloadBase
 
 
-class SharedMLPBenchTest(TestBase):
+class SharedMLPBenchTest(WorkloadBase):
     def __init__(self, num_tokens, hidden_size, ffn_size, dtype):
         self.num_tokens = num_tokens
         self.hidden_size = hidden_size
@@ -21,12 +21,6 @@ class SharedMLPBenchTest(TestBase):
         w_gate_up = torch.randn(self.ffn_size * 2, self.hidden_size, dtype=self.dtype, device=device)
         w_down = torch.randn(self.hidden_size, self.ffn_size, dtype=self.dtype, device=device)
         return hidden, w_gate_up, w_down
-
-    def ref_program(self, hidden, w_gate_up, w_down):
-        gate = torch.nn.functional.linear(hidden, w_gate_up[:self.ffn_size])
-        up = torch.nn.functional.linear(hidden, w_gate_up[self.ffn_size:])
-        gate_up = torch.nn.functional.silu(gate) * up
-        return torch.nn.functional.linear(gate_up, w_down)
 
 
 class SharedMLPBenchFixture(FixtureBase):
@@ -44,11 +38,11 @@ class SharedMLPBenchFixture(FixtureBase):
 
 class SharedMLPBenchmark(BenchmarkBase):
     def calculate_flops(self):
-        t = self.test
+        t = self.workload
         return 2 * t.num_tokens * t.hidden_size * t.ffn_size * 3  # gate + up + down
 
     def calculate_memory(self):
-        t = self.test
+        t = self.workload
         elem = 2  # bf16
         return elem * (t.num_tokens * t.hidden_size + 3 * t.ffn_size * t.hidden_size + 3 * t.num_tokens * t.ffn_size)
 
--- a/benchmarks/ops/bench_moe_shared_fused_moe.py
+++ b/benchmarks/ops/bench_moe_shared_fused_moe.py
@@ -36,15 +36,15 @@ except ImportError:
     _VLLM_AVAILABLE = False
 
 from benchmarks.benchmark import BenchmarkBase, BenchmarkReport
-from tests.test_base import FixtureBase, TestBase
 from tileops.ops.moe import FusedTopKOp, SharedFusedMoE
+from workloads.base import FixtureBase, WorkloadBase
 
 # ---------------------------------------------------------------------------
 # Test / fixture types
 # ---------------------------------------------------------------------------
 
 
-class SharedFusedMoEBenchTest(TestBase):
+class SharedFusedMoEBenchTest(WorkloadBase):
     def __init__(
         self,
         num_tokens,
@@ -101,9 +101,6 @@ class SharedFusedMoEBenchTest(TestBase):
         ) * 0.02
         return hidden, gating, correction_bias, w_gate_up, w_down, shared_w_gate_up, shared_w_down
 
-    def ref_program(self, *args):
-        return None
-
 
 class SharedFusedMoEBenchFixture(FixtureBase):
     PARAMS = [
@@ -146,7 +143,7 @@ class SharedFusedMoEBenchFixture(FixtureBase):
 class SharedFusedMoEBenchmark(BenchmarkBase):
 
     def calculate_flops(self) -> Optional[float]:
-        t = self.test
+        t = self.workload
         routed = t.num_tokens * t.top_k * (
             2 * t.ffn_size * t.hidden_size * 2   # gate+up
             + t.hidden_size * t.ffn_size * 2      # down
@@ -158,7 +155,7 @@ class SharedFusedMoEBenchmark(BenchmarkBase):
         return routed + shared
 
     def calculate_memory(self) -> Optional[float]:
-        t = self.test
+        t = self.workload
         elem = 2  # bf16 = 2 bytes
         routed_w = (
             t.num_experts * 2 * t.ffn_size * t.hidden_size


tileops/ops/moe/shared_fused_moe.py
```diff
@@ -96,24 +137,50 @@ def forward(
...
+                raise ValueError(
+                    "shared_w_gate_up and shared_w_down must be provided "
+                    "when shared_ffn_size is set"
+                )
+            # TP sharding: ColumnParallel on gate_up (dim=0), RowParallel on down (dim=1)
+            if self.tp_size > 1:
+                F_s = self.shared_ffn_size
+                shard_size = F_s // self.tp_size
+                gate_up_shard = shared_w_gate_up.narrow(0, self.tp_rank * shard_size * 2, shard_size * 2).contiguous()
+                down_shard = shared_w_down.narrow(1, self.tp_rank * shard_size, shard_size).contiguous()
```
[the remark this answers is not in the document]
michaelwithu: Good catch on the principle. A clarification on the actual cost here:

- `narrow()` is O(1) — it returns a view with no data copy.
- `gate_up_shard` (row slice, dim=0) is already contiguous, so `.contiguous()` is a no-op.
- `down_shard` (column slice, dim=1) is non-contiguous, so `.contiguous()` does copy ~`H × F_s/tp × 2` bytes once per forward.

The reason weights are passed per-forward rather than cached at init is intentional: `SharedFusedMoE` follows the same stateless-weight convention as `FusedMoe` and `FusedMoeExperts` in this codebase — the op holds no weight tensors, making it easier to use with weight-sharing, quantization, and EP frameworks that manage weight lifecycle externally.

That said, the `down_shard` copy is a real cost. The right fix is to let the caller pre-shard and pass already-sharded weights (i.e. `shared_w_gate_up [2*F_s/tp, H]` and `shared_w_down [H, F_s/tp]` directly), which eliminates the copy entirely. We can add a `pre_sharded=True` fast path in a follow-up. Tracking in #777.

tileops/kernels/moe/shared_expert_mlp.py
```diff
@@ -0,0 +1,123 @@
...
+                w_down: torch.Tensor) -> torch.Tensor:
+        T_dim = self.num_tokens
+        F = self.ffn_size
+
+        # [T, H] @ [2F, H]^T -> [T, 2F]
+        gate_up_out = self._gemm_gate_up(hidden, w_gate_up)
+
+        # Split: .contiguous() is a no-op when T=1 (stride[0] stays 2F), use reshape instead
+        gate = gate_up_out[:, :F].reshape(T_dim, F).contiguous()
+        up = gate_up_out[:, F:].reshape(T_dim, F).contiguous()
```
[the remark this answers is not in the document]
michaelwithu: Adopted. Refactored `_silu_mul_split_kernel` into `_silu_mul_fused_kernel` which accepts the combined `[T, 2F]` tensor directly and reads gate/up via index offset (`gate_up[:, j]` and `gate_up[:, j+N]`) inside the kernel. The two intermediate `[T, F]` buffer allocations are now eliminated entirely. Fixed in d109a58.

tests/ops/test_moe_fused_moe_distributed.py
```diff
@@ -212,7 +212,7 @@ def test_shared_fused_moe_ep_distributed(T, E_global, K, H, F, shared_F, world_s
         routed_scaling_factor=2.827,
         layout="nopad", dtype=dtype,
         expert_map=expert_map,
-        shared_experts_fn=shared_mlp,
+        shared_experts=shared_mlp,
```
superAngGao: **Bug (blocking):** `shared_experts` is not a valid parameter for the refactored `SharedFusedMoE`. The old `shared_experts_fn: Callable` was removed and replaced by `shared_ffn_size: int` + weight tensors in `forward()`. This rename (from one invalid name to another) will `TypeError` at construction time.

Two options:
1. Convert this EP test to the new weight-based API (pass `shared_ffn_size`, `shared_w_gate_up`, `shared_w_down`).
2. Keep a backward-compatible `shared_experts: Callable` kwarg in `SharedFusedMoE` for EP scenarios where the shared expert runs as a whole module.

Same issue on line 230.

tileops/kernels/moe/shared_expert_mlp.py
```diff
@@ -0,0 +1,119 @@
...
+import tilelang
+import tilelang.language as T
+import torch
+
+from tileops.kernels.gemm.gemm import GemmKernel
+from tileops.kernels.kernel import Kernel
+
+__all__ = ["SharedExpertMLPKernel"]
+
+_DEFAULT_CONFIG = {"block_m": 128, "block_n": 256, "block_k": 64, "num_stages": 3, "threads": 256, "enable_rasteration": True}
```
superAngGao, line 15: **Nit:** This config is shared by both `_gemm_gate_up` (`[T, H] @ [2F, H]^T`) and `_gemm_down` (`[T, F] @ [H, F]^T`). For Kimi K2 (F_s=18432, H=7168), the N dimensions differ by ~2.5x. `block_m=128` is also wasteful for decode (T=1). Per-GEMM configs or expanding `autotune_configs` would help — fine as a follow-up.

tileops/ops/moe/shared_fused_moe.py
```diff
@@ -83,7 +100,31 @@ def __init__(
...
             expert_map=expert_map,
         )
-        self.shared_experts_fn = shared_experts_fn
+
+        if tp_size < 1:
+            raise ValueError(f"tp_size must be >= 1, got {tp_size}")
+        if not (0 <= tp_rank < tp_size):
+            raise ValueError(f"tp_rank must be in [0, tp_size), got tp_rank={tp_rank}, tp_size={tp_size}")
+        if shared_ffn_size is not None and shared_ffn_size % tp_size != 0:
+            raise ValueError(
```
superAngGao, line 110: **Nit:** PR description says `layout="padded"` is not supported with the shared expert path. Consider adding a guard here:
```python
if shared_ffn_size is not None and layout == "padded":
    raise ValueError("layout='padded' is not supported with shared experts")
```
Prevents silent misbehavior.

tileops/ops/moe/shared_fused_moe.py
superAngGao: The `down_shard` column slice is the only real copy here (`gate_up_shard` row slice is already contiguous). For Kimi K2 with tp_size=8 this copies ~32 MB per forward. The planned `pre_sharded=True` fast path will fix this — just flagging so it stays on the radar.

tests/ops/test_moe_fused_moe_distributed.py
michaelwithu: Fixed in 7ef3526. Migrated the EP test to the new weight-based API: replaced `nn.Sequential` + `shared_experts=` with `shared_ffn_size=shared_F` at construction and `shared_w_gate_up` / `shared_w_down` passed in `forward()`. In EP mode `tp_size=1` (default), so each rank computes the full shared expert independently — same semantics as before.

tileops/ops/moe/shared_fused_moe.py
michaelwithu: Added in 7ef3526: `if shared_ffn_size is not None and layout == 'padded': raise ValueError(...)`.

tileops/ops/moe/shared_fused_moe.py
michaelwithu: Confirmed — only `down_shard` (column slice) triggers a real copy, ~32 MB for Kimi K2 tp8. The `pre_sharded` fast path is tracked in #777.

tests/ops/test_moe_fused_moe_distributed.py
michaelwithu: Thanks for catching this! Fixed in 7ef3526 — migrated the EP test to the new weight-based API: replaced `nn.Sequential` + `shared_experts=` with `shared_ffn_size=shared_F` at construction and `shared_w_gate_up` / `shared_w_down` passed in `forward()`. In EP mode `tp_size=1` (default), so each rank computes the full shared expert independently — same semantics as before.

tileops/kernels/moe/shared_expert_mlp.py
michaelwithu: Good point. Worth noting that `_DEFAULT_CONFIG` is only the fallback for `tune=False` — when autotune is enabled (`tune=True`), each `GemmKernel` searches independently and will find separate optimal configs for gate_up and down. That said, improving the default config for the non-tuned path is a valid follow-up. Tracking in #777.

tileops/ops/moe/shared_fused_moe.py
michaelwithu: Good suggestion, added in 7ef3526. Now raises `ValueError` at construction time when `shared_ffn_size is not None and layout == 'padded'`.

tileops/ops/moe/shared_fused_moe.py
michaelwithu: Confirmed — `gate_up_shard` (row slice) is already contiguous so its `.contiguous()` is a no-op, only `down_shard` (column slice) triggers a real copy, ~32 MB for Kimi K2 tp8. The `pre_sharded` fast path to eliminate this is tracked in #777.

tileops/kernels/moe/shared_expert_mlp.py
michaelwithu: Good point. Worth noting that `_DEFAULT_CONFIG` is only the fallback for `tune=False` — when autotune is enabled (`tune=True`), each `GemmKernel` searches independently and will find separate optimal configs for gate_up and down.

We also ran autotune locally for Kimi K2 scale (T=512, H=7168, F_s=18432) and both GEMMs converged to the same config (`block_n=256`), so the current default is already optimal for this workload. Per-GEMM defaults for other scales (e.g. decode T=1) are tracked in #777.

tests/ops/test_moe_shared_fused_moe.py
```diff
@@ -84,12 +84,80 @@ def test_shared_fused_moe_none():
...
+    w_down = torch.randn(E, H, F, dtype=dtype, device=dev) * 0.02
+    shared_w_gate_up = torch.randn(F_s * 2, H, dtype=dtype, device=dev) * 0.02
+    shared_w_down = torch.randn(H, F_s, dtype=dtype, device=dev) * 0.02
+
+    # Float32 math reference: compute each TP shard's contribution manually
+    # This matches exactly what each rank's kernel computes (column-parallel on gate_up, row-parallel on down)
+    shard_size = F_s // tp_size
+    partial_sum_ref = torch.zeros(T, H, dtype=torch.float32, device=dev)
+    for tp_rank in range(tp_size):
+        gate_up_shard = shared_w_gate_up[tp_rank * shard_size * 2: (tp_rank + 1) * shard_size * 2]  # [2*shard, H]
```
lcy-seso: The reference path is reproducing the same sharding bug as the implementation.

Since `shared_w_gate_up` is laid out as `[gate_all ; up_all]`, the per-rank reference shard should be built by slicing the gate half and up half separately, then concatenating them. Using a single contiguous slice here makes the test agree with the current implementation even if the implementation is wrong.

I would strongly recommend rewriting the reference shard construction so this test can actually catch TP layout mistakes.

tests/ops/test_moe_shared_fused_moe_distributed.py
```diff
@@ -0,0 +1,223 @@
...
+            vllm_mlp = DeepseekV2MLP(
+                hidden_size=H,
+                intermediate_size=F_s,
+                hidden_act="silu",
+                reduce_results=True,
+                prefix="shared_experts",
+            ).to(dtype=dtype, device=dev)
+
+            # gate_up: ColumnParallel → each rank holds [2*shard, H]
+            gate_up_shard = shared_w_gate_up[tp_rank * shard_size * 2 : (tp_rank + 1) * shard_size * 2]
```
lcy-seso: Same issue in the distributed reference setup: this copies the same contiguous `gate_up` slice into vLLM, so the cross-framework test is validating a shared bug rather than the intended TP contract.

If the declared full-weight layout is `[gate_all ; up_all]`, each TP rank should receive:

- its shard from the gate half
- its shard from the up half

and then concatenate those two local shards before loading them into `gate_up_proj.weight`.

Otherwise this test will pass even when the TileOPs TP slicing logic is incorrect.

tileops/ops/moe/shared_fused_moe.py
```diff
@@ -96,24 +139,50 @@ def forward(
...
+            if shared_w_gate_up is None or shared_w_down is None:
+                raise ValueError(
+                    "shared_w_gate_up and shared_w_down must be provided "
+                    "when shared_ffn_size is set"
+                )
+            # TP sharding: ColumnParallel on gate_up (dim=0), RowParallel on down (dim=1)
+            if self.tp_size > 1:
+                F_s = self.shared_ffn_size
+                shard_size = F_s // self.tp_size
+                gate_up_shard = shared_w_gate_up.narrow(0, self.tp_rank * shard_size * 2, shard_size * 2).contiguous()
```
lcy-seso: The TP sharding for `shared_w_gate_up` looks incorrect here.

`shared_w_gate_up` is documented as `[2 * F_s, H]` with the first `F_s` rows = gate and the second `F_s` rows = up. With that layout, TP rank `r` should receive:

- `gate[r * shard : (r + 1) * shard]`
- `up[r * shard : (r + 1) * shard]`

concatenated back into `[2 * shard, H]`.

The current contiguous slice

`shared_w_gate_up.narrow(0, tp_rank * shard_size * 2, shard_size * 2)`

instead takes one block out of `[gate_all ; up_all]`, which mixes the logical partitioning. For example, with `F_s=8, tp_size=2`, rank 0 currently gets `gate[0:8]`, but it should get `gate[0:4] + up[0:4]`.

That means the TP path will compute the wrong shared-expert output when the caller passes full weights in the declared layout.

tests/ops/test_moe_shared_fused_moe_distributed.py
michaelwithu: Fixed. Same correction applied to the distributed reference: gate/up halves sliced separately before loading into vLLM. The distributed test now uses vLLM as a genuine independent ground truth. (commit 7ceaa76)

tileops/ops/moe/shared_fused_moe.py
michaelwithu: Confirmed. The contiguous slice was wrong — with `F_s=8, tp_size=2` rank 0 was getting the entire gate half instead of `gate[0:4] + up[0:4]`.

Fixed by slicing gate and up halves separately:

```python
gate_shard = shared_w_gate_up[r * s : (r + 1) * s]
up_shard   = shared_w_gate_up[F_s + r * s : F_s + (r + 1) * s]
gate_up_shard = torch.cat([gate_shard, up_shard], dim=0).contiguous()
```

tests/ops/test_moe_shared_fused_moe.py
michaelwithu: Confirmed. The reference was using the same contiguous slice as the impl, so the test was only validating internal consistency rather than correctness — it would pass even when the TP layout was wrong.

Rewrote the reference to slice gate and up halves separately:

```python
gate_up_shard = torch.cat([
    shared_w_gate_up[tp_rank * shard_size : (tp_rank + 1) * shard_size],
    shared_w_gate_up[F_s + tp_rank * shard_size : F_s + (tp_rank + 1) * shard_size],
], dim=0)
```

Now the reference is an independent ground truth and will correctly catch TP layout mistakes.

tests/ops/test_moe_shared_fused_moe_distributed.py
michaelwithu: Confirmed. The distributed reference was loading the same contiguous slice into vLLM, so the cross-framework test was validating a shared bug rather than the actual TP contract — vLLM's `MergedColumnParallelLinear` would compute the wrong result too since it received the wrong weights.

Fixed by constructing the vLLM shard correctly:

```python
gate_up_shard = torch.cat([
    shared_w_gate_up[tp_rank * shard_size : (tp_rank + 1) * shard_size],
    shared_w_gate_up[F_s + tp_rank * shard_size : F_s + (tp_rank + 1) * shard_size],
], dim=0)
```

vLLM now receives the correct per-rank shard, making it a genuine independent ground truth.

tileops/ops/moe/shared_fused_moe.py
```diff
@@ -1,56 +1,71 @@
...
+        tp_size=tp_size, tp_rank=tp_rank,
+    )
+    # Pass complete weights; op shards them internally per tp_rank.
+    # shared_out is a partial result — caller must all-reduce across TP ranks.
+    shared_out_partial, routed_out = op(
+        hidden, gating, w_gate_up, w_down,
+        shared_w_gate_up=shared_w_gate_up,  # [2*F_s, H]  complete
+        shared_w_down=shared_w_down,         # [H, F_s]   complete
+    )
+    # dist.all_reduce(shared_out_partial)  ← caller's responsibility
```
Ibuki-wind: Please make the example explicit that this reduction must happen on the TP group, not via a bare `dist.all_reduce(...)` on the default group. Once this op is used together with EP/DP, the current wording is easy to misread and encourages an incorrect collective scope.

tileops/ops/moe/shared_fused_moe.py
```diff
@@ -96,24 +139,56 @@ def forward(
...
+                caller must all-reduce across TP ranks.
         """
-        # Compute shared expert output
-        shared_out = (
-            self.shared_experts_fn(hidden_states)
-            if self.shared_experts_fn is not None
-            else None
-        )
+        if self._shared_mlp_kernel is not None:
+            if shared_w_gate_up is None or shared_w_down is None:
```
Ibuki-wind, line 168: This new interface really needs explicit shape validation in addition to the `None` check. In TP mode the contract is that callers still pass full `shared_w_gate_up [2 * F_s, H]` and full `shared_w_down [H, F_s]`, not TP-local shards. Without a direct check here, a caller can easily pass local shards and then fail later inside slicing/GEMM with a much less actionable error. I would strongly prefer to reject that misuse here with a precise message such as `expected full shared-expert weights, not TP-local shards`.

tests/ops/test_moe_shared_fused_moe.py
```diff
@@ -84,12 +84,83 @@ def test_shared_fused_moe_none():
...
 
     shared_out, routed_out = op(hidden, gating, w_gate_up, w_down)
 
     assert shared_out is None
     assert routed_out.shape == (T, H)
-    print("PASS SharedFusedMoE with shared_experts_fn=None")
+    print("PASS SharedFusedMoE with shared_ffn_size=None")
+
+
+@pytest.mark.smoke
```
Ibuki-wind, line 97: Please add a negative test for the TP contract above: construct `SharedFusedMoE(tp_size > 1, ...)`, pass TP-local shared weights (or any wrong shape), and assert that we raise the intended `ValueError`. Right now the smoke suite only covers the happy path, so the most important API-clarity guarantee of this refactor is still untested.

tileops/ops/moe/shared_fused_moe.py
michaelwithu: Fixed. The docstring example now explicitly notes the TP group requirement:
```python
# dist.all_reduce(shared_out_partial, group=tp_group)  ← caller's responsibility
# Must use the TP process group, not the default group (important in EP/DP setups).
```
(commit c8c020f)

tileops/ops/moe/shared_fused_moe.py
michaelwithu: Fixed. Added explicit shape validation before the TP slicing block. The check validates both `shared_w_gate_up` and `shared_w_down` against their expected full-weight shapes and raises a `ValueError` with a clear message pointing callers to the contract:

```python
if shared_w_gate_up.shape != (2 * F_s, H):
    raise ValueError(
        f"shared_w_gate_up must be full weights with shape ({2 * F_s}, {H}), "
        f"got {tuple(shared_w_gate_up.shape)}. "
        "Pass complete weights; the op shards them internally per tp_rank."
    )
```

(commit c8c020f)

tests/ops/test_moe_shared_fused_moe.py
michaelwithu: Added `test_shared_fused_moe_tp_rejects_local_shards`. It constructs a `SharedFusedMoE(tp_size=2)` then passes TP-local shards for both `shared_w_gate_up` and `shared_w_down` separately, asserting `ValueError` with message matching `'full weights'` in each case. (commit c8c020f)

benchmarks/kernels/bench_moe_shared_expert_mlp.py
```diff
@@ -0,0 +1,86 @@
+"""Benchmark for SharedExpertMLPKernel vs PyTorch MLP."""
+
+import pytest
+import torch
+
+from benchmarks.benchmark import BenchmarkBase, BenchmarkReport
+from tests.test_base import FixtureBase, TestBase
```
Ibuki-wind: **Blocking — #787 migration required.**

#787 (merged 2026-04-05) moved `FixtureBase` to `workloads.base` and eliminated all `tests.*` imports from `benchmarks/`. This line re-introduces the coupling that #787 removed.

Also: `SharedMLPBenchTest` inherits `TestBase` on line 11, which forces a `ref_program` ABC implementation. After switching to `WorkloadBase`, the `ref_program` becomes unnecessary — move the PyTorch baseline into the benchmark function as a local closure (matching the pattern in other post-#787 benchmarks).

```suggestion
from workloads.base import WorkloadBase, FixtureBase
```

benchmarks/kernels/bench_moe_shared_expert_mlp.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class SharedMLPBenchmark(BenchmarkBase):
+    def calculate_flops(self):
+        t = self.test
+        return 2 * t.num_tokens * t.hidden_size * t.ffn_size * 3  # gate + up + down
+
+    def calculate_memory(self):
+        t = self.test
+        elem = 2  # bf16
```
Ibuki-wind, line 46: **Blocking — `AttributeError` on current `main`.**

`BenchmarkBase` was renamed from `self.test` to `self.workload` in #787. This crashes at runtime. Same issue on line 56.

```suggestion
        t = self.workload
```

benchmarks/ops/bench_moe_shared_fused_moe.py
```diff
@@ -0,0 +1,305 @@
...
+    from vllm.model_executor.layers.fused_moe.router.fused_topk_router import (
+        fused_topk as _vllm_fused_topk,
+    )
+    _VLLM_AVAILABLE = True
+except ImportError:
+    _VLLM_AVAILABLE = False
+
+from benchmarks.benchmark import BenchmarkBase, BenchmarkReport
+from tests.test_base import FixtureBase, TestBase
+from tileops.ops.moe import FusedTopKOp, SharedFusedMoE
```
Ibuki-wind, line 39: **Blocking — same #787 migration as the kernel benchmark.**

`tests.test_base` imports must become `workloads.base` imports. `SharedFusedMoEBenchTest` should inherit `WorkloadBase` instead of `TestBase`.

```suggestion
from workloads.base import WorkloadBase, FixtureBase
```

benchmarks/ops/bench_moe_shared_fused_moe.py
```diff
@@ -0,0 +1,305 @@
...
+        shared_w_gate_up = torch.randn(
+            self.shared_ffn_size * 2, self.hidden_size, dtype=self.dtype, device=dev
+        ) * 0.02
+        shared_w_down = torch.randn(
+            self.hidden_size, self.shared_ffn_size, dtype=self.dtype, device=dev
+        ) * 0.02
+        return hidden, gating, correction_bias, w_gate_up, w_down, shared_w_gate_up, shared_w_down
+
+    def ref_program(self, *args):
+        return None
```
Ibuki-wind: **Nit** — This dummy exists solely to satisfy `TestBase`'s ABC. Once migrated to `WorkloadBase` (which has no `ref_program` abstract method), delete this entirely.

benchmarks/ops/bench_moe_shared_fused_moe.py
```diff
@@ -0,0 +1,305 @@
...
+
+# ---------------------------------------------------------------------------
+# Benchmark class
+# ---------------------------------------------------------------------------
+
+
+class SharedFusedMoEBenchmark(BenchmarkBase):
+
+    def calculate_flops(self) -> Optional[float]:
+        t = self.test
```
Ibuki-wind: **Blocking — `AttributeError` on current `main`.**

`BenchmarkBase.test` was renamed to `BenchmarkBase.workload` in #787. All `self.test` references in this class will crash.

```suggestion
        t = self.workload
```

tileops/kernels/moe/shared_expert_mlp.py
```diff
@@ -0,0 +1,119 @@
...
+
```
Ibuki-wind, line 16: **Nit** — `enable_rasteration` — is this the correct TileLang config key, or a typo for `enable_rasterization`? If typo, the flag is silently ignored and the intended rasterization optimization is not applied.

tests/ops/test_moe_shared_fused_moe_distributed.py
```diff
@@ -0,0 +1,226 @@
...
+        dist.destroy_process_group()
+
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+class TPFixture:
+    """Fixture for shared expert TP tests."""
+    PARAMS = [
```
Ibuki-wind, line 110: **Suggestion** — `TPFixture` hand-rolls what `FixtureBase` (via `FixtureMeta`) already provides. Inheriting `FixtureBase` from `workloads.base` keeps the fixture contract consistent across the repo and avoids maintaining a parallel parametrize path.

```python
from workloads.base import FixtureBase

class TPFixture(FixtureBase):
    PARAMS = [
        (
            "T, H, F_s, tp_size",
            [
                pytest.param(64, 128, 64, 2, marks=pytest.mark.smoke, id="smoke-tp2"),
                pytest.param(512, 7168, 18432, 2, marks=pytest.mark.full, id="kimi-k2-tp2"),
                pytest.param(512, 7168, 18432, 8, marks=pytest.mark.full, id="kimi-k2-tp8"),
            ],
        )
    ]
```

benchmarks/kernels/bench_moe_shared_expert_mlp.py
michaelwithu: Fixed. Migrated to `workloads.base`:
- `from tests.test_base import FixtureBase, TestBase` → `from workloads.base import FixtureBase, WorkloadBase`
- `SharedMLPBenchTest(TestBase)` → `SharedMLPBenchTest(WorkloadBase)`
- Dropped `ref_program`; PyTorch baseline already lives as a local closure in the benchmark function.

Also rebased onto current `main` to pick up the `workloads/` layer. (commit d742852)

benchmarks/kernels/bench_moe_shared_expert_mlp.py
michaelwithu: Fixed. `self.test` → `self.workload` in both `calculate_flops` and `calculate_memory`. Same fix applied to `bench_moe_shared_fused_moe.py`. (commit d742852)

benchmarks/kernels/bench_moe_shared_expert_mlp.py
michaelwithu: Fixed in d742852. Migrated `bench_moe_shared_expert_mlp.py` to the workloads layer: `tests.test_base` → `workloads.base`, `TestBase` → `WorkloadBase`, `ref_program` removed, PyTorch baseline inlined as a local closure inside the benchmark function.

benchmarks/kernels/bench_moe_shared_expert_mlp.py
michaelwithu: Fixed in d742852. Both `self.test` references (lines 46 and 56) renamed to `self.workload`.

Issue #783, opened 2026-04-04
[Refactor] Fully detach benchmarks/ from tests/ via workloads/ shared layer


lcy-seso:
## Description

`benchmarks/` still depends on `tests/` across the repo. This is not a single-file problem; it is a structural dependency bug.

Current coupling forms:
- `benchmarks/benchmark.py` imports `tests.test_base.TestBase`
- benchmark files import workload classes from `tests/ops/test_*.py`
- some benchmark files import private helpers from test modules
- benchmark code still reads workload fields through `self.test.*`

Current repo state confirms this is systemic:
- 81 `tests.*` imports under `benchmarks/`
- 67 benchmark files affected
- 42 `self.test.*` accesses remain

This coupling means benchmark code breaks when test-only code changes. That is the wrong dependency direction. Benchmarks and tests may share workload definitions, but benchmarks must not depend on test-layer correctness code or test module structure.

## Decision

Fix this as a single repo-wide refactor.

Introduce one neutral shared layer: `workloads/`.

`workloads/` owns only:
- workload parameters
- `gen_inputs()`
- shared fixture / parametrization metadata currently needed by both tests and benchmarks
- pure helper functions that benchmarks currently import from `tests/`

`workloads/` must not own:
- `ref_program()`
- `check()`
- tolerance / assertion logic
- other correctness-only semantics

Required dependency direction after this issue:
- `tests/` -> `workloads/`: allowed
- `benchmarks/` -> `workloads/`: allowed
- `benchmarks/` -> `tests/`: forbidden

Required contract change:
- `BenchmarkBase` must depend on `WorkloadBase`, not `TestBase`
- benchmark code must use `self.workload`, not `self.test`

This issue is not a pilot and not a staged cleanup. It is done only when the benchmark tree is fully detached from the test tree.

## Scope

In scope:
- add `workloads/base.py` with `WorkloadBase`, `FixtureMeta`, and `FixtureBase`
- move shared workload definitions out of `tests/ops/test_*.py` into `workloads/ops/*.py`
- move benchmark-used pure helpers out of `tests/` into `workloads/`
- update `benchmarks/benchmark.py` to use `WorkloadBase`
- rename benchmark-side `self.test` usage to `self.workload`
- update all benchmark imports so they no longer import any `tests.*`
- update tests to import workload definitions from `workloads/` while keeping correctness logic in `tests/`
- update `tests/test_base.py` so it no longer defines `FixtureMeta` / `FixtureBase`

[...]

PR #784, opened 2026-04-04
[Chore][Manifest] Add cleanup gate and orchestrator discipline to spec skills

### LINKED ISSUE #786: [META][MANIFEST] Harden spec-pipeline: cleanup gate, orchestrator discipline, migration hygiene
### Symptom / Motivation

PR #782 (softmax family spec migration, batch 1) was the first real execution of the spec-pipeline. All 3 ops migrated successfully, but the run exposed three pipeline design gaps:

1. **No cleanup mechanism for dual-path debt**: when spec-implement rewrites a base class, it creates a dual-path `__init__` (legacy + spec execution paths) to keep unmigrated siblings' tests passing. After migration completes, no mechanism triggers cleanup — temporary code becomes permanent code.

2. **Sub-agent commit loss**: logsumexp's spec-test and spec-implement agents reported successful commits, but changes remained in the working tree. The orchestrator proceeded without verification, so the next agent ran in a dirty environment.

3. **Test-bench framework coupling** (separate issue #783): `BenchmarkBase` directly depends on `TestBase`. When spec-test modifies a TestBase constructor, benchmarks break. Root cause is codebase architecture, not skill design.

### Root Cause Analysis

**S1 (test-bench coupling)**: `benchmarks/benchmark.py` imports `tests.test_base.TestBase`; benchmark helpers import from test files. The only shared concern is `gen_inputs()`, which should be extracted into an independent data layer. → Split to issue #783.

**S2 (dual-path)**: the pipeline loops per-op (TEST→IMPLEMENT), so when the first op rewrites the base class, sibling tests still use the old interface. Two approaches compared:

- **REWRITE_INTERFACE** (code-first): new skill that atomically rewrites the base class before per-op loop. Eliminates dual-path but **breaks TDD** — structural changes happen without test coverage. The logsumexp reshape bug was caught by tests; without test coverage such bugs are uncontrollable.
- **Cleanup Gate** (TDD + deferred cleanup): preserve per-op TDD, accept temporary dual-path, explicitly trigger cleanup after all siblings complete. Main risk (blocked ops prevent cleanup) is mitigable via timeout policy.

**Decision: adopt Cleanup Gate.** Rationale: the TDD safety net is more valuable than eliminating temporary complexity. Risk is more controllable, implementation cost is lower.

**S3 (commit loss)**: sub-agent git commits may silently fail due to pre-commit hooks. The orchestrator trusted agent return reports without verifying worktree state.

### Related Files

- `.claude/skills/spec-pipeline/SKILL.md`
- `.claude/skills/spec-implement/SKILL.md`
- `.gitignore`

## Goal

Codify lessons from PR #782 into skill definitions: add cleanup gate, orchestrator discipline, and dual-path policy so the next pipeline execution (batch 2: reduction ops) does not repeat these issues.

## Plan

1. Add GROUP_BY_BASE phase and CLEANUP_GATE/CLEANUP steps to spec-pipeline skill
2. Add orchestrator discipline section to spec-pipeline (clean worktree verification + dual-path definition)
3. Add dual-path policy section to spec-implement skill
4. Add `.foundry/migrations/` to `.gitignore` (local working state, not tracked)
5. Sync TileOpsGov design doc (companion PR: AIGCIC/TileOpsGov#17)

## Constraints

- No changes to op implementation code or tests
- No changes to manifest
- Skill definition changes are additive (new steps, no removal of existing steps)

## Acceptance Criteria

- [ ] AC-1: spec-pipeline SKILL.md contains GROUP_BY_BASE, CLEANUP_GATE, CLEANUP steps
- [ ] AC-2: spec-pipeline SKILL.md contains orchestrator discipline section (clean worktree + dual-path definition)
- [ ] AC-3: spec-implement SKILL.md contains dual-path policy section
- [ ] AC-4: `.foundry/migrations/` is in `.gitignore`
- [ ] AC-5: TileOpsGov design doc is consistent with skill definitions (AIGCIC/TileOpsGov#17)

lcy-seso:
Closes #786

## Summary

- Add cleanup gate (CLEANUP_GATE → CLEANUP) to spec-pipeline: removes dual-path legacy code after all siblings in a base-class group are promoted or blocked
- Add orchestrator worktree discipline: verify clean worktree between sub-agent dispatches, commit on behalf if needed
- Add GROUP_BY_BASE phase to track per-base-class migration progress
- Add dual-path policy to spec-implement: acceptable during migration, cleanup is orchestrator's responsibility
- Add `.foundry/migrations/` to `.gitignore`

## Context

Design decisions from PR #782 (softmax family migration) retrospective:

| Issue | Decision | Rationale |
|---|---|---|
| S1: test-bench coupling | Codebase fix (issue #783), not skill fix | BenchmarkBase importing TestBase is the root cause |
| S2: dual-path __init__ | Cleanup gate, not REWRITE_INTERFACE | Preserves TDD invariant; REWRITE_INTERFACE breaks it |
| S3: sub-agent commit loss | Orchestrator worktree verification | Sub-agents may fail silently on pre-commit hooks |

## Test plan

- [x] Skill definitions parse correctly (no syntax errors in SKILL.md)
- [ ] Next pipeline execution (batch 2: reduction ops) validates the new flow

### DIFF
--- a/.claude/skills/spec-audit/SKILL.md
+++ b/.claude/skills/spec-audit/SKILL.md
@@ -82,7 +82,7 @@ Per-op entry. **Example — field set is not exhaustive. Add fields that aid the
 }
 ```
 
-Required fields: `classification`, `source_op`, `manifest_signature`. Gap report is a starting point — agent reads live code and manifest during downstream skills.
+Required fields: `classification`, `source_op`, `base_class`, `manifest_signature`. `base_class` is the immediate parent class name (e.g., `_SoftmaxBaseOp`); if the op inherits `Op` directly, use `"Op"`. The orchestrator uses `base_class` for GROUP_BY_BASE grouping. Gap report is a starting point — agent reads live code and manifest during downstream skills.
 
 `pytorch_equivalent`: corresponding PyTorch function, or `null`. Not every op has one.
 


.claude/skills/spec-pipeline/SKILL.md
```diff
@@ -51,6 +73,12 @@ stateDiagram-v2
 
 Gap report written to `.foundry/migrations/<family>.json`.
 
+### 1b. GROUP_BY_BASE
+
+Group ops by `base_class` from the gap report. Each group is a set of sibling ops sharing a base class. Process groups in order; within each group, process ops in order (first op likely fixes the base class, subsequent ops validate).
```
lcy-seso, line 78: `GROUP_BY_BASE` now makes `base_class` a required upstream dependency, but the current pipeline contract does not guarantee that field will exist in every gap-report entry. As written, this step is relying on `spec-audit` output that `spec-audit` still documents as optional/non-required. I think this section needs to either define a fallback grouping rule or explicitly tighten the upstream gap-report contract so `base_class` is guaranteed.

.claude/skills/spec-pipeline/SKILL.md
```diff
@@ -108,6 +136,30 @@ Orchestrator (not a sub-skill) changes manifest:
...
+### 7c. CLEANUP
+
+Remove dual-path legacy code from the base class. This step fires once per base-class group, after all siblings are done.
+
+Actions:
+
+1. Remove legacy `__init__` branch (`if M is not None and N is not None` path)
+1. Remove `_legacy` flag and `_forward_legacy` method
+1. Remove `M`, `N` keyword-only parameters from `__init__`
+1. Run full test suite for all ops in the group: `python -m pytest <all source_test files> -v`
```
lcy-seso: This cleanup contract is still inconsistent for groups containing blocked ops. The gate allows cleanup once siblings are `promoted` or `blocked`, but this step then requires the full test suite for *all* ops in the group. For blocked ops, those tests may still fail by definition. I think this needs a stricter rule about what is validated here (for example, only promoted ops plus cleanup-specific regression checks) instead of `all source_test files`.

.claude/skills/spec-pipeline/SKILL.md
```diff
@@ -108,6 +136,30 @@ Orchestrator (not a sub-skill) changes manifest:
 - Commit the manifest change
 - Update gap report: add `promoted_at` timestamp (keep `classification` unchanged)
 
+### 7b. CLEANUP_GATE
+
+After each FLIP_STATUS, check group completion:
+
+- All siblings in the current base-class group are `promoted` or `blocked`? → trigger CLEANUP
```
lcy-seso, line 145: The state machine still only reaches `CLEANUP_GATE` from `FLIP_STATUS`. That means if the last sibling in a group finishes via `REPORT_BLOCKED`, cleanup is never re-evaluated even though this section defines group completion as `promoted or blocked`. I think cleanup evaluation needs to happen after any terminal per-op outcome, not only after promotion.

.claude/skills/spec-pipeline/SKILL.md
```diff
@@ -35,12 +36,33 @@ stateDiagram-v2
...
 ```
 
+## Orchestrator Discipline
+
+### Clean worktree between sub-agents
+
+After each sub-agent returns and before dispatching the next, verify:
+
+```bash
+git diff --quiet && git diff --cached --quiet
```
lcy-seso: This worktree check still misses untracked files. `git diff --quiet && git diff --cached --quiet` only covers tracked changes, so a sub-agent can still leave newly created files uncommitted and the orchestrator will treat the tree as clean. I think this discipline needs an explicit untracked-file check as well.

.claude/skills/spec-pipeline/SKILL.md
lcy-seso: Accept. Fixed: `base_class` is now documented as a required field in the gap report. For ops inheriting `Op` directly, `base_class` is `"Op"` — cleanup gate is a no-op for this group.

.claude/skills/spec-pipeline/SKILL.md
lcy-seso: Accept. Fixed: CLEANUP now runs tests and --check-op for promoted ops only. Blocked ops are excluded from cleanup validation.

.claude/skills/spec-pipeline/SKILL.md
lcy-seso: Accept. Fixed: state machine now has REPORT_BLOCKED --> CLEANUP_GATE transition. Cleanup evaluation happens after any terminal per-op outcome.

.claude/skills/spec-pipeline/SKILL.md
lcy-seso: Accept. Fixed: replaced git diff --quiet with git status --porcelain to catch untracked files too.

.claude/skills/spec-pipeline/SKILL.md
```diff
@@ -51,6 +73,14 @@ stateDiagram-v2
 
 Gap report written to `.foundry/migrations/<family>.json`.
 
+### 1b. GROUP_BY_BASE
+
+Group ops by `base_class` from the gap report. Each group is a set of sibling ops sharing a base class. Process groups in order; within each group, process ops in order (first op likely fixes the base class, subsequent ops validate).
+
+`base_class` is a required field in the gap report. spec-audit must populate it for every op entry. If an op inherits `Op` directly (no intermediate base class), its `base_class` is `"Op"` — these ops form a single group but are independent (no shared base class to rewrite, so cleanup gate is a no-op for this group).
```
lcy-seso, line 80: Re-reviewing the latest head: this section is much better now, but it still leaves one remaining blocker. `spec-pipeline` now explicitly treats `base_class` as a required grouping key, including the `"Op"` fallback for direct subclasses. That means the upstream contract must guarantee `base_class` on every gap-report entry. Right now `spec-audit` still only shows `base_class` in the example payload and does not list it as a required field, so the two skill contracts are still not aligned. I think this PR should not merge until that upstream requirement is documented as well.

lcy-seso: Already fixed in commit `5a81d56` (pushed before this comment). spec-audit SKILL.md line 85 now reads: **Required fields: `classification`, `source_op`, `base_class`, `manifest_signature`.** The two contracts are aligned.

PR #794, opened 2026-04-05
[Feat][Reduction] Migrate softmax family to spec-conformant interface

lcy-seso:
## Summary

- Migrate 3 softmax-family ops (`softmax_fwd`, `log_softmax_fwd`, `logsumexp_fwd`) from legacy `__init__(M, N, dtype)` to manifest-specified `__init__(dtype=..., dim=..., keepdim=...)`.
- M/N are now derived from input tensor shape + `dim` at forward time; kernels cached by `(M, N)` to avoid TIR rebuilds.
- Status changes: `softmax_fwd` and `log_softmax_fwd` promoted to `implemented`. `logsumexp_fwd` remains `spec-only` (manifest requires `dim: int | list[int]`, impl supports single int only).

## Changes by file

| File | Change |
|------|--------|
| `_softmax_base.py` | Rewritten: `__init__(dtype, dim)`, M/N computed at forward time, kernel caching by (M,N), `movedim` for non-last dim, dim range validation matching PyTorch `IndexError` |
| `softmax.py` | Docstring cleanup (inherits everything from base) |
| `log_softmax.py` | Docstring cleanup (inherits everything from base) |
| `logsumexp.py` | Overrides `__init__` to add `keepdim` param (only op in family that exposes it); removed legacy `_reshape_output` override |
| `workloads/ops/softmax.py` | All 3 workloads: `(m, n, dtype)` → `(shape, dtype)` |
| `tests/ops/test_softmax.py` | All tests rewritten for spec interface; added `dim=0`/`dim=1` coverage; added `keepdim=True` tests for logsumexp |
| `benchmarks/ops/bench_softmax.py` | Manifest-driven: uses `load_workloads()` + `eval_roofline()` from `tileops.manifest`; skips unsupported kernel shapes with narrowed `ValueError` check |
| `ops_manifest.yaml` | `softmax_fwd`, `log_softmax_fwd`: `spec-only` → `implemented`; `logsumexp_fwd`: remains `spec-only` |

## Design decisions

1. **All-keyword `__init__`** (`*` separator) — all existing callers already passed params as kwargs, so backward-compatible during dual-path phase. Legacy path removed in cleanup commit.

2. **Kernel caching by `(M, N)` tuple** — TIR kernels bake in M and N at compile time. `dispatch_kernel` called once in `__init__`; cache lookup on each forward.

3. **`movedim` transpose for non-last dim** — when `dim != -1`, we `movedim(dim, -1)` before reshape to (M, N), run kernel, then `movedim(-1, dim)` back.

4. **Dim range validation** — raises `IndexError` matching PyTorch's message format for out-of-range dims, instead of silently coercing via modulo.

5. **`keepdim` scoped to `LogSumExpOp` only** — not in the shared base class. `SoftmaxOp` and `LogSoftmaxOp` do not expose `keepdim` since their manifest entries don't include it.

6. **`logsumexp_fwd` remains `spec-only`** — manifest specifies `dim: "int | list[int]"` but the current kernel only supports single-int dim. Per trust model: "Implementation does not conform to spec → set `status: spec-only`."

7. **Manifest-driven benchmarks** — workload shapes from `load_workloads()`, FLOPs/memory from `eval_roofline()`. Validator passes with 0 warnings. Unsupported shapes (N=32768, N=102400) are skipped with a narrowed `ValueError` check matching only `"No configurations to tune"`.

## Known gaps surfaced by manifest-driven benchmarks

9 out of 15 manifest workloads are skipped because the kernel does not support their reduction-dim size:
- `attn-weights-32k` (N=32768): autotune fails with "No configurations to tune"
- `lm-head-logits` (N=102400): same kernel limitation

These are pre-existing kernel coverage gaps, not regressions from this PR.

## Validation

- 105/105 tests pass (smoke + full, including keepdim tests)
- 15/15 benchmarks: 6 passed, 9 skipped (kernel limitation)
- Manifest validator: 0 errors, 0 warnings for all 3 ops
- Pre-commit hooks clean

## Context

This is **batch 1** of the reduction family spec migration (19 ops total, 6 groups). Softmax group (3 ops) was chosen as the validation batch. Remaining 5 groups tracked in `.foundry/migrations/reduction.json`.

## Test plan

- [x] `python -m pytest tests/ops/test_softmax.py -v` — 105 passed
- [x] `python -m pytest benchmarks/ops/bench_softmax.py -v` — 6 passed, 9 skipped
- [x] `python scripts/validate_manifest.py --check-op softmax_fwd --check-op log_softmax_fwd --check-op logsumexp_fwd` — 0 errors, 0 warnings
- [ ] CI preflight

### DIFF
--- a/benchmarks/ops/bench_softmax.py
+++ b/benchmarks/ops/bench_softmax.py
@@ -122,8 +122,10 @@ def test_softmax_bench(shape: tuple, dtype: torch.dtype) -> None:
     op = SoftmaxOp(dtype=dtype, dim=-1, tune=True)
     try:
         result = bm.profile(op, *inputs)
-    except (ValueError, Exception) as exc:
-        pytest.skip(f"Kernel does not support this shape: {exc}")
+    except ValueError as exc:
+        if "No configurations to tune" in str(exc):
+            pytest.skip(f"Kernel does not support this shape: {exc}")
+        raise
     BenchmarkReport.record(op, locals(), result, tag="tileops")
 
     def baseline_fn(x):
@@ -147,8 +149,10 @@ def test_log_softmax_bench(shape: tuple, dtype: torch.dtype) -> None:
     op = LogSoftmaxOp(dtype=dtype, dim=-1, tune=True)
     try:
         result = bm.profile(op, *inputs)
-    except (ValueError, Exception) as exc:
-        pytest.skip(f"Kernel does not support this shape: {exc}")
+    except ValueError as exc:
+        if "No configurations to tune" in str(exc):
+            pytest.skip(f"Kernel does not support this shape: {exc}")
+        raise
     BenchmarkReport.record(op, locals(), result, tag="tileops")
 
     def baseline_fn(x):
@@ -172,8 +176,10 @@ def test_logsumexp_bench(shape: tuple, dtype: torch.dtype) -> None:
     op = LogSumExpOp(dtype=dtype, dim=-1, tune=True)
     try:
         result = bm.profile(op, *inputs)
-    except (ValueError, Exception) as exc:
-        pytest.skip(f"Kernel does not support this shape: {exc}")
+    except ValueError as exc:
+        if "No configurations to tune" in str(exc):
+            pytest.skip(f"Kernel does not support this shape: {exc}")
+        raise
     BenchmarkReport.record(op, locals(), result, tag="tileops")
 
     def baseline_fn(x):


tests/ops/test_softmax.py
```diff
@@ -6,6 +6,10 @@
 
 Smoke tests (1 per function, first param) use small data for quick CI.
 Full tests use small data for config breadth + large data for stress.
+
+Softmax tests target the spec-conformant interface:
+  SoftmaxOp(dtype=dtype, dim=dim)  -- no M/N params
+Log_softmax and logsumexp tests still use the legacy interface.
```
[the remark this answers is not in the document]
lcy-seso: Fixed in 007fdaa. Updated the module docstring to reflect that all three ops now use the spec-conformant interface.

tests/ops/test_softmax.py
```diff
@@ -35,206 +39,287 @@ def _get_tolerances(dtype: torch.dtype) -> tuple[float, float]:
...
 
 @LogSumExpFixture
-def test_logsumexp_op(m: int, n: int, dtype: torch.dtype, tune: bool) -> None:
-    test = LogSumExpTest(m, n, dtype)
-    op = LogSumExpOp(M=m, N=n, dtype=dtype, tune=tune)
+def test_logsumexp_op(shape: tuple, dim: int, dtype: torch.dtype, tune: bool) -> None:
+    test = LogSumExpTest(shape, dtype, dim=dim)
+    op = LogSumExpOp(dtype=dtype, dim=dim, tune=tune)
     atol, rtol = _get_tolerances(dtype)
     test.check(op, *test.gen_inputs(), atol=atol, rtol=rtol)
```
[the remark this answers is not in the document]
lcy-seso: Valid point — added 6 keepdim=True test cases in 007fdaa covering dim=-1 (2D, 3D), dim=0, and dim=1, validating both shape and values against `torch.logsumexp(..., keepdim=True)`.

tileops/ops/reduction/_softmax_base.py
```diff
@@ -2,8 +2,13 @@
 
 Provides the shared validate -> reshape -> pad -> kernel -> trim -> reshape
 pattern for softmax, log_softmax, and logsumexp ops.
+
+Construction: ``op(dtype, dim=-1, keepdim=False)``.  M and N are derived
```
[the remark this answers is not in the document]
lcy-seso: Good catch — fixed in 007fdaa. Docstring now shows `op(dtype=..., dim=-1, keepdim=False)` to match the keyword-only signature.

tileops/ops/reduction/_softmax_base.py
```diff
@@ -38,73 +43,124 @@ class _SoftmaxBaseOp(Op):
...
-            N,
-            self._op_kind,
-            dtype,
-            tune=tune,
-        )
+        self.dim = dim
+        self.keepdim = keepdim
+        self._kernel_map_override = kernel_map
+        self._tune = tune
+        self._kernel_cache: Dict[tuple, object] = {}
```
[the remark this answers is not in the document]
lcy-seso: Agreed on both points — fixed in 007fdaa. `dispatch_kernel` is now called once in `__init__`, and removed from the cache-miss path in `_get_or_create_kernel`.

Regarding unbounded cache growth: acknowledged in the PR description as an observation. For current LLM inference workloads (few distinct shapes), this is fine. If variable-length training becomes a use case, we'll add LRU eviction.

tileops/ops/reduction/_softmax_base.py
```diff
@@ -38,73 +43,124 @@ class _SoftmaxBaseOp(Op):
...
+            y = y[:, :N] if y.ndim == 2 else y
+
+        return self._reshape_output(y, orig_shape, dim, needs_transpose)
+
+    def _get_or_create_kernel(self, M: int, N: int) -> object:
+        """Return a cached kernel for (M, N), creating one if needed."""
+        key = (M, N)
+        if key not in self._kernel_cache:
+            self.dispatch_kernel(self._kernel_map_override)
+            kernel_cls = self.kernel_map[self._kernel_key]
```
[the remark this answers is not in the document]
lcy-seso: Fixed in the same commit (007fdaa) — `dispatch_kernel` moved to `__init__`, removed from `_get_or_create_kernel`.

tests/ops/test_softmax.py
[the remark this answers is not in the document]
lcy-seso: Fixed in 007fdaa — docstring now reads "All operators use the spec-conformant interface."

tileops/ops/reduction/_softmax_base.py
```diff
@@ -38,73 +43,123 @@ class _SoftmaxBaseOp(Op):
...
         self._validate(x)
         orig_shape = x.shape
 
-        # Flatten leading dims: (..., N) -> (M, N)
-        x = x.contiguous().reshape(-1, self.N)
-        M_actual = x.shape[0]
-        if M_actual != self.M:
-            raise ValueError(f"Expected M={self.M} (product of leading dims), got {M_actual}")
+        # Normalize dim.
+        dim = self.dim % x.ndim
```
lcy-seso, line 93: This modulo normalization changes the public API semantics. For a 3D input, `dim=-4` becomes `2` and `dim=3` becomes `0`, so TileOPs silently runs where PyTorch raises `IndexError`. I verified both cases locally on GPU. Since the manifest requires PyTorch interface alignment, this should validate the dim range instead of coercing out-of-range values back into range.

tileops/ops/reduction/_softmax_base.py
```diff
@@ -38,73 +43,123 @@ class _SoftmaxBaseOp(Op):
 
     def __init__(
         self,
-        M: int,
-        N: int,
+        *,
         dtype: torch.dtype,
+        dim: int = -1,
+        keepdim: bool = False,
```
lcy-seso: Pushing `keepdim` into the shared base `__init__` makes it part of the public constructor for `SoftmaxOp` and `LogSoftmaxOp` too, even though their manifest entries only expose `dim`. That weakens the spec boundary for the two same-shape ops. I would either split the reduction-only parameter out of the common constructor, or override the concrete op signatures so `keepdim` is only public on `LogSumExpOp`.

lcy-seso:
GPU revalidation on local H200 passed (`tests/ops/test_softmax.py`: 105/105, `benchmarks/ops/bench_softmax.py`: 12/12), but I still cannot treat this as ready because the implementation and the spec are not actually aligned yet.

The main issue is that `logsumexp_fwd` was promoted to `implemented` while its manifest still advertises `dim: int | list[int]`, but the op only supports a single integer axis. I verified this locally: `LogSumExpOp(dtype=torch.float16, dim=[1, 2])(x)` raises `TypeError`, while `torch.logsumexp(x, dim=[1, 2])` works and returns shape `[2]`.

The second functional mismatch is invalid-dimension handling. `_SoftmaxBaseOp` normalizes `dim` with `% x.ndim`, so values like `dim=-4` or `dim=3` on a 3D input silently execute instead of raising. I verified this locally too: TileOPs succeeds for both cases, while PyTorch raises `IndexError`. That breaks the manifest rule that Op semantics must align with the PyTorch interface.

There is also still a process-level gap in the spec-driven claim: the benchmark file is not manifest-driven yet. It hardcodes workloads and roofline math instead of using `load_workloads` / `eval_roofline`, and the targeted manifest validator still warns about exactly that.

I left inline comments on the concrete fix points. Once the manifest, runtime semantics, and benchmark pattern are brought back into sync, this will be in much better shape as the first migration PR.

tileops/ops_manifest.yaml
```diff
@@ -1398,7 +1398,7 @@ ops:
 
   logsumexp_fwd:
     family: reduction
-    status: spec-only
+    status: implemented
```
lcy-seso: The status flip here is premature relative to the current manifest contract. `logsumexp_fwd` is now marked `implemented`, but the manifest still exposes `dim: "int | list[int]"` while the migrated op only supports a single integer axis. I verified this locally on GPU: `LogSumExpOp(dtype=torch.float16, dim=[1, 2])(x)` fails with `TypeError`, while `torch.logsumexp(x, dim=[1, 2])` succeeds. Either implement multi-dim reduction semantics before flipping status, or narrow the spec in a separate explicit manifest change.

benchmarks/ops/bench_softmax.py
```diff
@@ -77,11 +79,11 @@ def calculate_memory(self) -> Optional[float]:
 
 @pytest.mark.parametrize("m, n, dtype, tune", _SOFTMAX_BENCH_PARAMS)
 def test_softmax_bench(m: int, n: int, dtype: torch.dtype, tune: bool) -> None:
-    test = SoftmaxTest(m, n, dtype)
+    test = SoftmaxTest((m, n), dtype)
```
lcy-seso: This benchmark is still hardcoded, so the PR is not actually completing the manifest-driven migration loop yet. `docs/manifest.md` requires benchmark files to use `load_workloads()` and `eval_roofline()`, and `python scripts/validate_manifest.py --check-op softmax_fwd` still emits warnings for this file. For a spec-driven pilot PR, I think this should be brought into the manifest-driven pattern rather than left as a known gap.

tileops/ops/reduction/_softmax_base.py
lcy-seso: Valid — fixed in f4f791e. Now raises `IndexError` with the same message format as PyTorch for out-of-range dims, instead of silently coercing via modulo.

```python
# PyTorch-matching behavior:
if self.dim < -x.ndim or self.dim >= x.ndim:
    raise IndexError(
        f"Dimension out of range (expected to be in range of "
        f"[{-x.ndim}, {x.ndim - 1}], but got {self.dim})"
    )
```

tileops/ops/reduction/_softmax_base.py
lcy-seso: Valid — fixed in f4f791e. `keepdim` is removed from `_SoftmaxBaseOp.__init__` and now only exposed through `LogSumExpOp.__init__` which overrides the base constructor to add it. `SoftmaxOp` and `LogSoftmaxOp` no longer accept `keepdim` in their public API.

tileops/ops_manifest.yaml
lcy-seso: Valid — reverted `logsumexp_fwd` to `spec-only` in f4f791e. The manifest specifies `dim: "int | list[int]"` but the current implementation only supports single-int dim. Per the trust model: "Implementation does not conform to spec → set status: spec-only, fix code in follow-up PR."

Multi-dim reduction support for logsumexp is a kernel-level change and will be addressed in a separate PR.

benchmarks/ops/bench_softmax.py
lcy-seso: Valid — fully addressed in f4f791e. The benchmark is now manifest-driven:

- **`load_workloads()`**: shapes come from `ops_manifest.yaml`, not hardcoded params
- **`eval_roofline()`**: FLOPs/memory derived from manifest roofline expressions, not hand-coded formulas
- **Validator passes clean**: 0 warnings for all 3 ops

Additional finding: manifest workloads include shapes the kernel can't support yet (`attn-weights-32k` N=32768, `lm-head-logits` N=102400). These are gracefully skipped with `pytest.skip()` rather than failing. This surfaces the kernel coverage gap — 9 out of 15 manifest workloads are skipped due to kernel limitations.

benchmarks/ops/bench_softmax.py
```diff
@@ -19,70 +21,109 @@
...
+    test = SoftmaxTest(shape, dtype)
     bm = SoftmaxBenchmark(test)
     inputs = test.gen_inputs()
 
-    op = SoftmaxOp(M=m, N=n, dtype=dtype, tune=tune)
-    result = bm.profile(op, *inputs)
+    op = SoftmaxOp(dtype=dtype, dim=-1, tune=True)
+    try:
+        result = bm.profile(op, *inputs)
+    except (ValueError, Exception) as exc:
```
lcy-seso: These three handlers are still too broad: `except (ValueError, Exception)` is effectively `except Exception`, so any benchmark regression here will be silently reclassified as a skip. In my local rerun, the current skips were all the expected `No configurations to tune` cases, but this pattern will also hide unrelated failures. Please narrow this to the specific unsupported-shape/config exception you expect, or preflight the workload before profiling instead of swallowing arbitrary runtime errors.

lcy-seso: Valid — narrowed in 8e3f9a5. Now only catches `ValueError` with message containing `"No configurations to tune"` (the specific autotune failure for unsupported shapes). All other exceptions propagate as real failures.

```python
except ValueError as exc:
    if "No configurations to tune" in str(exc):
        pytest.skip(f"Kernel does not support this shape: {exc}")
    raise
```

PR #797, opened 2026-04-06
[CI] Remove commit-level validation; enforce format via PR title + squash merge

lcy-seso:
## Summary

- Remove `validate-commits` CI job — redundant when squash merge is enforced and PR title is the single source of commit subject on main
- Remove `tests/ci/test_merge_commit_skip.sh` — tested the removed commit validation logic
- Add PR title pre-flight validation gate to `.foundry/mold/pre-create-checks.md` — sources `types.sh` to validate locally before `gh pr create`, guaranteeing CI passes
- Promote three validated feedback items into domain-rules (`benchmark.md`, `manifest-spec.md`, `testing-budget.md`)
- Add two collaboration rules to `CLAUDE.md` (no issue annotations in source files, sub-agent worktree verification)

## Test plan

- [x] pre-commit passed
- [x] `validate-pr-title` CI check remains — covers all PRs (agent and human)
- [x] Repo settings confirmed: squash-only, `PR_TITLE` as commit subject, `validate-pr-title` is required check in branch ruleset

### DIFF
--- a/.claude/domain-rules/manifest-spec.md
+++ b/.claude/domain-rules/manifest-spec.md
@@ -24,4 +24,4 @@ ______________________________________________________________________
   - `variant_of` is one level only. Variant → primary. Primary must not have `variant_of`.
   - Variants share `source.kernel` and `source.op`. Each has its own `signature`, `workloads`, `roofline`.
 - Tensor layout defaults to contiguous row-major. When an op requires non-default layout (e.g., `channels_last`), add `layout` field to the tensor declaration. `shape` dimension names reflect actual memory order.
-- Never modify manifest to match non-conforming code. If code doesn't match spec: set `status: spec-only`, fix implementation in a follow-up PR. Never remove params, vars, or shape_rules to silence validator errors.
+- Never modify manifest to match non-conforming code. If code doesn't match spec: set `status: spec-only` and add a comment explaining the discrepancy, then fix implementation in a follow-up PR. Never remove params, vars, or shape_rules to silence validator errors.
--- a/.github/workflows/preflight.yml
+++ b/.github/workflows/preflight.yml
@@ -153,38 +153,17 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
-          fetch-depth: 0
-
-      - name: Check if manifest changed
-        id: changed
-        env:
-          EVENT_NAME: ${{ github.event_name }}
-          BASE_SHA: ${{ github.event.pull_request.base.sha }}
-          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
-        run: |
-          if [[ "$EVENT_NAME" == "pull_request" ]]; then
-            if git diff --name-only "$BASE_SHA" "$HEAD_SHA" | grep -q '^tileops/ops_manifest\.yaml$'; then
-              echo "skip=false" >> "$GITHUB_OUTPUT"
-            else
-              echo "Manifest not changed; skipping validation."
-              echo "skip=true" >> "$GITHUB_OUTPUT"
-            fi
-          else
-            echo "skip=false" >> "$GITHUB_OUTPUT"
-          fi
+          fetch-depth: 1
 
       - name: Set up Python
-        if: steps.changed.outputs.skip != 'true'
         uses: actions/setup-python@v5
         with:
           python-version: "3.10"
 
       - name: Install TileOPs and validator dependencies
-        if: steps.changed.outputs.skip != 'true'
         run: pip install -e .
 
       - name: Validate ops_manifest.yaml
-        if: steps.changed.outputs.skip != 'true'
         run: python scripts/validate_manifest.py --levels schema,shape,dtype,bench
 
   actionlint:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -36,7 +36,7 @@ TileOPs is a high-performance LLM operator library built on TileLang. The goal i
 - Add necessary docs and tests when introducing files/interfaces.
 - Response should include: change summary, affected paths, validation steps, and next suggestions.
 - No GitHub issue numbers or ad-hoc issue annotations in source/test files. Issue refs belong in commit messages and PR descriptions only.
-- When orchestrating sub-agents: verify clean worktree (`git diff --quiet && git diff --cached --quiet`) after each sub-agent returns. Commit on behalf if dirty.
+- When orchestrating sub-agents: verify clean worktree (`git diff --quiet && git diff --cached --quiet`) after each sub-agent returns. If dirty, commit on behalf using a descriptive message (e.g., `Sub-agent [name]: [task summary]`).
 
 ## Domain Rules (load on demand)
 
--- a/.github/workflows/gpu-smoke.yml
+++ b/.github/workflows/gpu-smoke.yml
@@ -46,7 +46,7 @@ jobs:
             exit 0
           fi
 
-          required_checks=(validate-commits pre-commit gitleaks actionlint)
+          required_checks=(pre-commit gitleaks actionlint)
           deadline=$((SECONDS + 900))
 
           while true; do


.github/workflows/preflight.yml
```diff
@@ -18,53 +18,6 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
```
haruhi55, line 20: `validate-commits` cannot be removed as-is. This is not just a policy relaxation: `.github/workflows/gpu-smoke.yml` still waits for a check named `validate-commits`, so this change can leave downstream gating in a permanently pending/timeout path and silently skip GPU smoke for PRs. It also removes server-side enforcement of commit format entirely. Please either keep the job, or update every consumer and branch-protection dependency in the same PR with equivalent coverage.

.github/workflows/preflight.yml
```diff
@@ -193,28 +146,49 @@ jobs:
...
+
+      - name: Check if manifest changed
+        id: changed
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          if [[ "$EVENT_NAME" == "pull_request" ]]; then
+            if git diff --name-only "$BASE_SHA" "$HEAD_SHA" | grep -q '^tileops/ops_manifest\.yaml$'; then
```
haruhi55: The new skip condition is too narrow for what `validate_manifest.py` actually verifies. That validator checks repository-wide invariants, including code/bench integration, not only diffs to `tileops/ops_manifest.yaml`. With this filter, a PR can break manifest validation by changing validator logic, benchmark files, or manifest-related code paths while this job is skipped. Please gate on the full dependency set, or keep this job always-on.

tests/ci/test_merge_commit_skip.sh
```diff
@@ -1,83 +0,0 @@
-#!/bin/bash
```
haruhi55, line 1: Removing the only regression test around CI skip/validation behavior is not acceptable in the same PR that rewrites preflight semantics. If the old test is obsolete, replace it with coverage for the new behavior before merging. CI control-flow changes need explicit regression tests.

haruhi55 (changes requested):
This PR changes shared CI behavior for all PRs, so the safety bar is high. In its current form it introduces a concrete cross-workflow dependency break (`gpu-smoke` still requires `validate-commits`), weakens server-side enforcement, and adds a manifest-skip heuristic that can miss real failures. Please fix the downstream dependency, restore equivalent enforcement, and add regression coverage before merge.

.github/workflows/preflight.yml
```diff
@@ -145,7 +98,7 @@ jobs:
           fi
 
   pre-commit:
-    needs: [validate-commits, validate-pr-title, ci-gate]
+    needs: [validate-pr-title, ci-gate]
```
haruhi55, line 101: `validate-commits` cannot be removed as-is. `.github/workflows/gpu-smoke.yml` still waits for a check named `validate-commits`, so this change can leave downstream gating in a pending/timeout path and silently skip GPU smoke. It also removes server-side enforcement of commit format entirely. Please either keep the job, or update every consumer and branch-protection dependency in the same PR with equivalent coverage.

.github/workflows/preflight.yml
haruhi55: The new skip condition is too narrow for what `validate_manifest.py` actually verifies. That validator checks repository-wide invariants, including code/bench integration, not only diffs to `tileops/ops_manifest.yaml`. With this filter, a PR can break manifest validation by changing validator logic, benchmark files, or manifest-related code paths while this job is skipped. Please gate on the full dependency set, or keep this job always-on.

tests/ci/test_merge_commit_skip.sh
haruhi55, line 1: Removing the only regression test around CI skip/validation behavior is not acceptable in the same PR that rewrites preflight semantics. If the old test is obsolete, replace it with coverage for the new behavior before merging. CI control-flow changes need explicit regression tests.

.claude/domain-rules/manifest-spec.md
```diff
@@ -24,3 +24,4 @@ ______________________________________________________________________
   - `variant_of` is one level only. Variant → primary. Primary must not have `variant_of`.
   - Variants share `source.kernel` and `source.op`. Each has its own `signature`, `workloads`, `roofline`.
 - Tensor layout defaults to contiguous row-major. When an op requires non-default layout (e.g., `channels_last`), add `layout` field to the tensor declaration. `shape` dimension names reflect actual memory order.
+- Never modify manifest to match non-conforming code. If code doesn't match spec: set `status: spec-only`, fix implementation in a follow-up PR. Never remove params, vars, or shape_rules to silence validator errors.
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Added the discrepancy comment requirement to the rule. See c6b3f36.

CLAUDE.md
```diff
@@ -35,6 +35,8 @@ TileOPs is a high-performance LLM operator library built on TileLang. The goal i
 - If unrelated failures appear, report them but do not fix them in the same task.
 - Add necessary docs and tests when introducing files/interfaces.
 - Response should include: change summary, affected paths, validation steps, and next suggestions.
+- No GitHub issue numbers or ad-hoc issue annotations in source/test files. Issue refs belong in commit messages and PR descriptions only.
+- When orchestrating sub-agents: verify clean worktree (`git diff --quiet && git diff --cached --quiet`) after each sub-agent returns. Commit on behalf if dirty.
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Added descriptive commit message format for sub-agent commits. See c6b3f36.

.github/workflows/preflight.yml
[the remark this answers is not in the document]
lcy-seso: Good catch. Fixed by removing `validate-commits` from gpu-smoke.yml's `required_checks` array instead of adding a no-op shim. See c6b3f36.

.github/workflows/preflight.yml
```diff
@@ -145,7 +98,7 @@ jobs:
...
     if: ${{ github.repository == 'tile-ai/TileOPs' && needs.ci-gate.outputs.skip != 'true' }}
     runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
lcy-seso: Already addressed — PR title was updated to `[CI]` and description now covers the full scope of changes including CI and test removals.

.github/workflows/preflight.yml
lcy-seso: Valid — gpu-smoke.yml dependency was missed. Fixed in c6b3f36: removed `validate-commits` from the `required_checks` array in gpu-smoke.yml.

Regarding server-side enforcement: commit-level format is intentionally dropped. Repo enforces squash-merge-only with PR title as commit subject (`squash_merge_commit_title=PR_TITLE`), and `validate-pr-title` is a required check in the branch ruleset. Every commit landing on main is format-validated through this chain.

.github/workflows/preflight.yml
lcy-seso: Agreed — the validator checks cross-file invariants (code/bench integration, shape rules), not just the YAML file itself. Reverted validate-manifest to always-on in c6b3f36.

tests/ci/test_merge_commit_skip.sh
lcy-seso: Disagree. `test_merge_commit_skip.sh` tested the merge-commit skip logic inside `validate-commits` — a job that no longer exists. There is no new CI control-flow behavior introduced that needs regression coverage: the only change is *removal* of a job and updating downstream `needs` arrays. The remaining `validate-pr-title` job is unchanged and already covered by its own CI run. Writing a test for "a job was removed" would be testing CI configuration, not behavior.

.github/workflows/preflight.yml
lcy-seso: Same as above — gpu-smoke.yml dependency fixed in c6b3f36. Server-side enforcement preserved via PR title validation + squash-merge-only repo setting.

.github/workflows/preflight.yml
lcy-seso: Same as above — reverted to always-on in c6b3f36.

tests/ci/test_merge_commit_skip.sh
lcy-seso: Same as above — the test covered removed functionality. No new behavior requires regression coverage.

haruhi55: The functional regressions are fixed, but I still do not consider this ready under a strict CI bar. This PR removes repo-local enforcement/test coverage and makes correctness depend on external repository settings (`squash`-only + `PR_TITLE` + required `validate-pr-title`). That contract is real today, but it is no longer guarded from within the repo. Please add repo-local regression protection for the new contract, or keep an in-repo compatibility check, before merge.

lcy-seso: The repo-local contract is already guarded:

1. `validate-pr-title` job in `preflight.yml` — sources `types.sh` and validates PR title format. This is in-repo code, runs on every PR.
2. `.foundry/mold/pre-create-checks.md` — agent pre-flight gate that validates title against the same `types.sh` before `gh pr create`.
3. Branch ruleset makes `validate-pr-title` a required check — PRs cannot merge without it passing.

The only thing not guarded in-repo is the squash-merge-only setting, which is a GitHub repo admin config. But that's true of *all* branch protection — it has never been testable from within the repo. Writing a test that asserts "the repo is configured for squash merge" would be testing GitHub settings, not code behavior, and would require admin API access that CI doesn't have.

The enforcement chain is: in-repo CI (`validate-pr-title`) guarantees format → repo setting (squash-only) guarantees the validated title becomes the commit subject. The first half is fully in-repo; the second half is infrastructure config, same as branch protection itself.

haruhi55: I verified the updated workflow and the repository enforcement chain. With squash-only merges, `PR_TITLE` as the squash commit title, and `validate-pr-title` required in the ruleset, I’m satisfied that the final commit-subject guarantee remains enforced. The earlier workflow regressions were fixed, so I’m clearing this blocker.

PR #834, opened 2026-04-07
[Feat][Reduction] Support dim=None in all reduction ops for full reduction

### LINKED ISSUE #814: [Feat][Reduction] Support dim=None for reduce-over-all-dimensions
### Symptom / Motivation
Discovered during PR #812 ([Feat][Reduction] Add list[int] dim support for multi-dim reduction ops).

PyTorch reduction ops (`torch.sum`, `torch.mean`, `torch.amax`, etc.) accept `dim=None` to reduce over all dimensions, returning a scalar. The current `normalize_dim()` helper in `_multidim.py` raises a `TypeError` on `None` input, and the ops default to `dim=-1` rather than supporting the full-reduction semantic.

A reviewer (gemini-code-assist) flagged this gap during PR #812 review. It was resolved as out-of-scope since the PR focused on `list[int]` support.

### Root Cause Analysis
Adding `dim=None` support would require changes to both the `_multidim.py` helper and each op's `forward()` method to handle the scalar-output case. This scope expansion didn't fit within the `list[int]` PR.

### Related Files
- `tileops/ops/reduction/_multidim.py` — `normalize_dim()` needs `None` handling
- `tileops/ops/reduction/reduce.py` — all `_SimpleReduceOp` and `_WelfordReduceOp` subclasses
- `tileops/ops/reduction/all_op.py`, `any_op.py`, `count_nonzero.py` — logical reduce ops
- `tileops/ops/reduction/l1_norm.py`, `l2_norm.py`, `inf_norm.py` — norm ops

## Goal
Allow `dim=None` in reduction ops to reduce over all dimensions, matching PyTorch's public API.

## Plan

1. Update `normalize_dim()` to accept `None` and return `list(range(ndim))`.
2. Update `ops_manifest.yaml` dim type to include `None` (separate manifest PR per trust model).
3. Add test cases for `dim=None` across representative ops.

## Constraints
- Must not regress changes from PR #812
- Must match PyTorch semantics: `dim=None` reduces all dims, returns scalar (or size-1 tensor with `keepdim=True`)
- Manifest changes require separate PR per trust model

## Acceptance Criteria
- [ ] AC-1: `SumOp(dim=None)(x)` produces same result as `torch.sum(x)`
- [ ] AC-2: All reduction ops accept `dim=None` without error
- [ ] AC-3: Tests cover `dim=None` with and without `keepdim`

lcy-seso:
## Summary

Support `dim=None` for full reduction in all applicable reduction ops, matching PyTorch's public API. When `dim=None`, `normalize_dim()` returns `list(range(ndim))`, routing through the existing multi-dim flatten path.

**Affected ops** (13 total): `sum`, `mean`, `amax`, `amin`, `prod`, `var`, `std`, `var_mean`, `all`, `any`, `count_nonzero`, `l1_norm`, `l2_norm`, `inf_norm`.

**Excluded** (by design):
- `logsumexp` — `torch.logsumexp` does not support `dim=None`; manifest stays `int | list[int]`
- `softmax`, `log_softmax` — PyTorch requires an explicit int `dim`; existing forward-time `ValueError` rejection path preserved
- `argmax`, `argmin`, `cumsum`, `cumprod` — different return semantics (index tensors / scan ops)

Closes #814

## Changes

- `_multidim.py`: `normalize_dim(dim=None, ndim)` → returns `list(range(ndim))`
- `_softmax_base.py`: Narrowed `dim` type to `Union[int, List[int]]` (no `None`)
- `reduce.py`: `_ReduceOpBase.__init__` accepts `dim: Union[int, List[int], None]`
- `all_op.py`, `any_op.py`, `count_nonzero.py`, `l1_norm.py`, `l2_norm.py`, `inf_norm.py`: Accept `dim=None` in `__init__`, route through multi-dim path
- `test_reduce_dim_none.py`: 16 new test functions covering all 13 ops with `dim=None`, parametrized across shapes, keepdim, and dtypes (fp16/bf16/fp32, bool for logical ops)

## Test plan

- [x] AC-1: `SumOp(dim=None)(x)` produces same result as `torch.sum(x)` for all parametrized cases
- [x] AC-2: All 13 applicable reduction ops accept `dim=None` without error
- [x] AC-3: Tests cover `dim=None` with both `keepdim=True` and `keepdim=False`
- [x] AC-4: Float32 coverage for numeric/norm ops, bool for `all`/`any`, fp16/bf16 for `count_nonzero`

## Follow-up

No follow-up issues or suggestions.

### DIFF
--- a/tests/ops/test_reduce_dim_none.py
+++ b/tests/ops/test_reduce_dim_none.py
@@ -231,25 +231,6 @@ def test_var_mean_dim_none(
     assert torch.allclose(mean_out, ref_mean, **tol), f"mean err: {(mean_out - ref_mean).abs().max()}"
 
 
-# ---------------------------------------------------------------------------
-# LogSumExp
-# ---------------------------------------------------------------------------
-
-
-@DimNoneFixture
-def test_logsumexp_dim_none(
-    shape: tuple, keepdim: bool, dtype: torch.dtype,
-) -> None:
-    from tileops.ops.reduction.logsumexp import LogSumExpOp
-
-    x = torch.randn(*shape, dtype=dtype, device="cuda")
-    op = LogSumExpOp(dtype=dtype, dim=None, keepdim=keepdim)
-    dims = _all_dims(shape)
-    ref = torch.logsumexp(x.float(), dim=dims, keepdim=keepdim).to(dtype)
-    y = op(x)
-    tol = _tol(dtype)
-    assert y.shape == ref.shape, f"shape mismatch: {y.shape} vs {ref.shape}"
-    assert torch.allclose(y, ref, **tol), f"max err: {(y - ref).abs().max()}"
 
 
 # ---------------------------------------------------------------------------
--- a/tileops/ops/reduction/_softmax_base.py
+++ b/tileops/ops/reduction/_softmax_base.py
@@ -46,15 +46,10 @@ class _SoftmaxBaseOp(Op):
         self,
         *,
         dtype: torch.dtype,
-        dim: Union[int, List[int], None] = -1,
+        dim: Union[int, List[int]] = -1,
         kernel_map: Optional[Dict[str, Kernel]] = None,
         tune: bool = False,
     ):
-        if not self._supports_multidim and not isinstance(dim, int):
-            raise TypeError(
-                f"{self.__class__.__name__} expects 'dim' to be an int; "
-                f"got {type(dim).__name__}."
-            )
         self.dtype = dtype
         self.dim = dim
         self.keepdim = False
--- a/tests/ops/test_softmax.py
+++ b/tests/ops/test_softmax.py
@@ -491,16 +491,23 @@ def test_logsumexp_1d(n: int, dtype: torch.dtype) -> None:
 
 @pytest.mark.smoke
 def test_softmax_rejects_multidim_before_kernel() -> None:
-    """SoftmaxOp must raise TypeError for non-int dim at construction."""
-    with pytest.raises(TypeError, match="expects 'dim' to be an int"):
-        SoftmaxOp(dtype=torch.float32, dim=[-1, 0])
+    """SoftmaxOp must raise ValueError for list dim before touching the kernel."""
+    x = torch.randn(4, 8, device="cuda", dtype=torch.float32)
+    op = SoftmaxOp(dtype=torch.float32, dim=[-1, 0])
+    with pytest.raises(ValueError, match="does not support multi-dim"):
+        op(x)
+    # Verify no kernel was built (cache must remain empty).
+    assert len(op._kernel_cache) == 0
 
 
 @pytest.mark.smoke
 def test_log_softmax_rejects_multidim_before_kernel() -> None:
-    """LogSoftmaxOp must raise TypeError for non-int dim at construction."""
-    with pytest.raises(TypeError, match="expects 'dim' to be an int"):
-        LogSoftmaxOp(dtype=torch.float32, dim=[-1, 0])
+    """LogSoftmaxOp must raise ValueError for list dim before touching the kernel."""
+    x = torch.randn(4, 8, device="cuda", dtype=torch.float32)
+    op = LogSoftmaxOp(dtype=torch.float32, dim=[-1, 0])
+    with pytest.raises(ValueError, match="does not support multi-dim"):
+        op(x)
+    assert len(op._kernel_cache) == 0
 
 
 @pytest.mark.smoke


benchmarks/ops/bench_moe_fused_moe.py
```diff
@@ -223,6 +223,7 @@ def _vllm_fn(hidden, gating, correction_bias, w_gate_up, w_down):
                 topk=top_k,
                 renormalize=renormalize,
                 scoring_func=scoring_func,
+                correction_bias=correction_bias,
             )
```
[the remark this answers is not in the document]
lcy-seso: This file (`benchmarks/ops/bench_moe_fused_moe.py`) is not modified by this PR's commit (993ec54). The change appears in the branch diff because of prior commits on `main`. This comment is not actionable for this PR.

Current HEAD: 1aa3662

benchmarks/ops/bench_moe_shared_fused_moe.py
```diff
@@ -234,6 +234,7 @@ def _vllm_fn(hidden, gating, correction_bias, w_gate_up, w_down,
                 topk=top_k,
                 renormalize=renormalize,
                 scoring_func=scoring_func,
+                correction_bias=correction_bias,
```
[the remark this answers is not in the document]
lcy-seso: This file (`benchmarks/ops/bench_moe_shared_fused_moe.py`) is not modified by this PR's commit (993ec54). The change appears in the branch diff because of prior commits on `main`. This comment is not actionable for this PR.

Current HEAD: 1aa3662

tileops/ops/reduction/_softmax_base.py
```diff
@@ -87,8 +87,8 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         self._validate(x)
         orig_shape = x.shape
 
-        # --- multi-dim path ---
-        if isinstance(self.dim, (list, tuple)):
+        # --- multi-dim path (includes dim=None for full reduction) ---
+        if isinstance(self.dim, (list, tuple)) or self.dim is None:
             if not self._supports_multidim:
```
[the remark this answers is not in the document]
lcy-seso: Reviewer determined this is a false positive:

The copilot reviewer claims PyTorch raises TypeError for dim=None on softmax, but PyTorch actually raises RuntimeError ('Please look up dimensions by name, got: name = None.'). The current code correctly rejects dim=None for SoftmaxOp/LogSoftmaxOp (where _supports_multidim=False) with a ValueError. The error message 'does not support multi-dim reduction. Use a scalar dim.' is technically correct -- it tells the user to pass a scalar dim. While the message could be slightly more specific for the dim=None case, this is cosmetic and the suggested TypeError is not aligned with PyTorch's own behavior. The copilot's suggested code restructuring would also duplicate multi-dim path logic unnecessarily.

Current HEAD: 1aa3662

tileops/ops/reduction/reduce.py
```diff
@@ -58,7 +58,7 @@ def __init__(
         self,
         *,
         dtype: torch.dtype,
-        dim: Union[int, List[int]] = -1,
+        dim: Union[int, List[int], None] = -1,
         keepdim: bool = False,
         kernel_map: Optional[Dict[str, Kernel]] = None,
         tune: bool = False,
```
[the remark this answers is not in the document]
lcy-seso: Acknowledged — docstring updates are a valid follow-up but out of scope for this PR which focuses on the functional `dim=None` support. Will address in a separate doc PR.

tileops/ops/reduction/reduce.py
```diff
@@ -58,7 +58,7 @@ def __init__(
         self,
         *,
         dtype: torch.dtype,
-        dim: Union[int, List[int]] = -1,
+        dim: Union[int, List[int], None] = -1,
         keepdim: bool = False,
```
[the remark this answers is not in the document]
lcy-seso: ArgmaxOp/ArgminOp are intentionally excluded — they return index tensors, not reduced values. `dim=None` for argmax/argmin has different semantics (flattened index) and is tracked separately. The AC specifies "all reduction ops" meaning value-reducing ops.

tileops/ops/reduction/_multidim.py
```diff
@@ -38,6 +38,9 @@ def normalize_dim(
         IndexError: If any dim is out of range.
         ValueError: If duplicate dims are given or dim is an empty list.
     """
+    if dim is None:
+        return list(range(ndim))
+
     dims = [dim] if isinstance(dim, int) else list(dim)
```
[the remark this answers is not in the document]
lcy-seso: The `list(dim)` call handles both `list` and `tuple` inputs correctly at runtime. Adding `tuple[int, ...]` to the type hint is a stylistic preference — the function already works with tuples. Low priority, can be addressed in a typing cleanup PR.

tileops/ops/reduction/all_op.py
```diff
@@ -108,8 +108,8 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
 
         orig_shape = x.shape
 
-        # --- multi-dim path ---
-        if isinstance(self.dim, (list, tuple)):
+        # --- multi-dim path (includes dim=None for full reduction) ---
+        if isinstance(self.dim, (list, tuple)) or self.dim is None:
```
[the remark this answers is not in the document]
lcy-seso: The isinstance check is a 1-line guard in each op's forward() — centralizing it would require either a mixin or decorator that adds complexity without clear benefit. Each op may diverge in the future. Keeping it inline is intentional.

benchmarks/ops/bench_moe_fused_moe.py
```diff
@@ -223,6 +223,7 @@ def _vllm_fn(hidden, gating, correction_bias, w_gate_up, w_down):
                 topk=top_k,
                 renormalize=renormalize,
                 scoring_func=scoring_func,
+                correction_bias=correction_bias,
```
Ibuki-wind: This `correction_bias` baseline repair is unrelated to `[Feat][Reduction]` and mixes a Benchmark-stage fix into a reduction API PR, which breaks the trust-model scope boundary and makes the regression surface harder to attribute. -> Remove this benchmark change from PR #834 and send it as a dedicated benchmark-scoped PR.

benchmarks/ops/bench_moe_shared_fused_moe.py
Ibuki-wind: This is the same scope leak as the non-shared benchmark: an unrelated Benchmark-stage baseline fix is being bundled into a reduction feature PR. -> Split the shared-MoE `correction_bias` benchmark change into its own benchmark-scoped PR instead of landing it here.

tileops/ops/reduction/reduce.py
```diff
@@ -58,7 +58,7 @@ def __init__(
         self,
         *,
         dtype: torch.dtype,
-        dim: Union[int, List[int]] = -1,
+        dim: Union[int, List[int], None] = -1,
```
Ibuki-wind, line 61: `_ReduceOpBase.__init__` now accepts `dim=None`, which widens the public contract for `sum/mean/amin/amax/prod/var/std/var_mean`, but `tileops/ops_manifest.yaml` still declares `dim` without `None` for these ops and keeps `prod_fwd.dim` as `int` only. The manifest is the source of truth for op interfaces. -> Land the matching manifest/interface update first (or keep the implementation on the manifest's current signature until that PR merges).

tests/ops/test_reduce_dim_none.py
```diff
@@ -0,0 +1,378 @@
...
+import torch
+
+from tests.test_base import FixtureBase
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+
+class DimNoneFixture(FixtureBase):
```
Ibuki-wind, line 21: `DimNoneFixture` only covers fp16/bf16, so the new `dim=None` path is still untested for supported `float32` numeric reductions and vector norms. That leaves part of the dtype contract uncovered even though this PR makes `dim=None` a public API. -> Add at least one float32 case for each affected numeric/norm op family, or split the fixture per family so coverage matches the manifest-declared dtypes.

tests/ops/test_reduce_dim_none.py
```diff
@@ -0,0 +1,378 @@
...
+    assert y.shape == ref.shape, f"shape mismatch: {y.shape} vs {ref.shape}"
+    assert torch.allclose(y, ref, **tol), f"max err: {(y - ref).abs().max()}"
+
+
+# ---------------------------------------------------------------------------
+# Logical reduce ops: all, any, count_nonzero
+# ---------------------------------------------------------------------------
+
+
+class DimNoneLogicalFixture(FixtureBase):
```
Ibuki-wind, line 241: `DimNoneLogicalFixture` only exercises float32, so `all/any` never hit their supported `bool` path, and `count_nonzero` is only checked for float32 even though its manifest contract includes fp16/bf16. -> Expand the logical/count `dim=None` tests to cover each supported dtype (`bool` for `all`/`any`, fp16 and bf16 for `count_nonzero`).

benchmarks/ops/bench_moe_fused_moe.py
lcy-seso: Adopted. Reverted the benchmark fix from this branch (1d2613e). Will land as a dedicated benchmark-scoped PR. See c6fc350.

benchmarks/ops/bench_moe_shared_fused_moe.py
lcy-seso: Adopted. Same revert as above — benchmark correction_bias fix removed from this branch. See c6fc350.

tileops/ops/reduction/reduce.py
lcy-seso: Valid point. Out of scope for this PR — the manifest-trust-model rule requires manifest changes in a separate PR with human review. The issue constraints explicitly state this: "Manifest changes require separate PR per trust model." Will follow up.

tests/ops/test_reduce_dim_none.py
lcy-seso: Adopted. Added float32 to DimNoneFixture (numeric/norm ops). See c6fc350.

tests/ops/test_reduce_dim_none.py
lcy-seso: Adopted. Added bool to DimNoneLogicalFixture for all/any, plus fp16/bf16 parametrized test for count_nonzero. 74 tests now pass (up from 54). See c6fc350.

tileops/ops/reduction/reduce.py
Ibuki-wind, line 61: `_ReduceOpBase.__init__` still accepts `dim=None`, which widens the public contract for `sum/mean/amin/amax/prod/var/std/var_mean`, but `tileops/ops_manifest.yaml` still declares `dim` as `int | list[int]` for most of these ops and `int` only for `prod_fwd`. The manifest is the source of truth for op interfaces and is meant to land as a separate PR. -> Merge the matching manifest/interface change first, or keep this implementation on the manifest's current signature until that manifest PR lands.

Ibuki-wind (changes requested):
1. tileops/ops/reduction/reduce.py:61 - `dim=None` still widens the public reduction signature while `tileops/ops_manifest.yaml` continues to declare `dim` without `None` for these ops, and `prod_fwd` remains `int`-only -> merge the matching manifest/interface change first, or keep this PR aligned with the current manifest contract.

Ibuki-wind (commented):
PR #834 is no longer blocked on benchmark scope or dtype coverage; those points were addressed. The remaining blocker is narrower and more fundamental: this PR introduces a new user-visible interface contract (`dim=None`) before that contract exists in `tileops/ops_manifest.yaml`. Under the trust model, this is not just an implementation detail or a test expansion; it is a manifest-level API widening.

As long as the manifest is the source of truth for op interfaces, implementation cannot be the place where a wider public contract becomes true first and the manifest catches up later. → Merge the corresponding manifest PR first, or explicitly gate this PR on an already approved manifest change that adds `dim=None` to the reduction interface contract.

tileops/ops/reduction/reduce.py
lcy-seso: Valid point. Out of scope for this PR — the issue plan (step 2) and constraints explicitly require manifest changes in a separate PR per trust-model rules. The implementation intentionally lands first; the manifest update to add `None` to dim type declarations is tracked as a follow-up. Per `manifest-trust-model.md`: implementation ahead of spec is the expected sequence when the manifest PR needs independent human review.

Ibuki-wind, line 61: The latest reply makes the trust-boundary problem clearer, not smaller: this PR is intentionally landing a wider public contract (`dim=None`) in implementation first and treating the manifest update as follow-up work. Under the current model, `ops_manifest.yaml` is supposed to approve interface widening before implementation makes it true. -> Do not merge this implementation-first API change until the corresponding manifest change is already merged or explicitly approved as the upstream contract.

lcy-seso: Resolved. The manifest update landed in PR #843 (merged), which declares `dim: int | list[int] | None` for all 13 applicable reduction ops with `status: spec-only`. This PR now aligns implementation with the approved manifest contract.

tileops/ops/reduction/_softmax_base.py
```diff
@@ -46,7 +46,7 @@ def __init__(
         self,
         *,
         dtype: torch.dtype,
-        dim: Union[int, List[int]] = -1,
+        dim: Union[int, List[int], None] = -1,
         kernel_map: Optional[Dict[str, Kernel]] = None,
         tune: bool = False,
     ):
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Added early TypeError in `_SoftmaxBaseOp.__init__` when `_supports_multidim=False` and `dim` is not int. See 91cb021.

tileops/ops/reduction/logsumexp.py
```diff
@@ -44,7 +44,7 @@ def __init__(
         self,
         *,
         dtype: torch.dtype,
-        dim: Union[int, List[int]] = -1,
+        dim: Union[int, List[int], None] = -1,
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Narrowed `LogSumExpOp.dim` to `Union[int, List[int]]` — `torch.logsumexp` does not support `dim=None`, and PR #843 confirmed the manifest stays at `int | list[int]`. See 91cb021.

tileops/ops/reduction/_softmax_base.py
```diff
@@ -46,10 +46,15 @@ def __init__(
...
+        if not self._supports_multidim and not isinstance(dim, int):
```
Ibuki-wind: This eager `TypeError` changes the existing softmax/log_softmax contract and regresses current smoke coverage: on this branch, `pytest tests/ops/test_softmax.py -m smoke -q` now fails because `SoftmaxOp(dim=[...])` and `LogSoftmaxOp(dim=[...])` raise in `__init__()` instead of the prior forward-time `ValueError`. -> Revert the constructor guard / `dim=None` widening in `_SoftmaxBaseOp` for this PR so unsupported softmax-family ops keep their existing rejection path.

tests/ops/test_reduce_dim_none.py
```diff
@@ -0,0 +1,423 @@
...
+    assert torch.allclose(mean_out, ref_mean, **tol), f"mean err: {(mean_out - ref_mean).abs().max()}"
+
+
+# ---------------------------------------------------------------------------
+# LogSumExp
+# ---------------------------------------------------------------------------
+
+
+@DimNoneFixture
+def test_logsumexp_dim_none(
```
Ibuki-wind: This `logsumexp dim=None` coverage is now out of date with `upstream/main`: the merged manifest still declares `logsumexp_fwd.dim` as `int | list[int]`, not `None`, so this PR is asserting and implementing an interface the spec does not currently allow. -> Remove the `LogSumExpOp(dim=None)` test (and the corresponding `_softmax_base.py` support) from this PR; keep the rest of the dim-None implementation scoped to the ops whose manifest already includes `None`.

tileops/ops/reduction/_softmax_base.py
lcy-seso: Adopted. Reverted constructor guard and `dim=None` type widening in `_SoftmaxBaseOp`. Restored original forward-time `ValueError` rejection path and tests. See 6cd5c57.

tests/ops/test_reduce_dim_none.py
lcy-seso: Adopted. Removed `LogSumExpOp(dim=None)` test — manifest confirms `logsumexp_fwd.dim` stays `int | list[int]`. See 6cd5c57.

tests/ops/test_reduce_dim_none.py
```diff
@@ -0,0 +1,404 @@
...
+
+
+class DimNoneLogicalFixture(FixtureBase):
+    PARAMS = [
+        (
+            "shape, keepdim, dtype",
+            [
+                pytest.param(
+                    (4, 8, 256), False, torch.float32,
+                    marks=pytest.mark.smoke,
```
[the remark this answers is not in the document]
lcy-seso: Already removed in 6cd5c57. The `logsumexp dim=None` test was deleted because the manifest declares `logsumexp_fwd.dim` as `int | list[int]` only (no `None`).

tileops/ops/reduction/_softmax_base.py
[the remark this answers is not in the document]
lcy-seso: Reverted in 6cd5c57 per human reviewer feedback. The existing forward-time `ValueError` rejection path is preserved instead of an init-time guard, keeping the current softmax/log_softmax contract intact.

tileops/ops/reduction/logsumexp.py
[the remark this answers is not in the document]
lcy-seso: `LogSumExpOp.dim` narrowed to `Union[int, List[int]]` (retained from 91cb021). Additionally, `_SoftmaxBaseOp.dim` also narrowed and logsumexp `dim=None` test removed in 6cd5c57.

PR #838, opened 2026-04-08
[Benchmark][CI] Resolve nightly perf history artifact lookup

### LINKED ISSUE #837: [BUG][CI] fix nightly perf history artifact lookup across workflow runs
### Symptom / Motivation
The nightly workflow fails in the `op_test` job with `Error: Unable to download artifact(s): Artifact not found for name: tileops_perf_history`.[Bug](https://github.com/tile-ai/TileOPs/actions/runs/24099357438/job/70318963750)

This breaks the perf-history bootstrap path for `scripts/nightly_report.py` and adds avoidable CI noise even when the rest of the nightly run is healthy.

### Root Cause Analysis
The `Download previous perf history` step in `.github/workflows/nightly.yml` uses `actions/download-artifact@v4` without a `run-id`, so GitHub Actions looks for `tileops_perf_history` in the current workflow run.

That artifact is uploaded later in the same job, which means the download step cannot succeed. The intended source is the previous successful nightly run, not the current run.

### Related Files
- `scripts/nightly_report.py`
- Issue #618

## Goal
Update the nightly workflow so perf history is resolved from a previous successful nightly run, or skipped cleanly when no history artifact exists yet.

## Plan

1. Add a lookup step before artifact download to resolve a previous successful `nightly.yml` run that contains `tileops_perf_history`.
2. Gate `actions/download-artifact@v4` on the resolved `run-id` and download from that run instead of the current run.
3. Keep `nightly_report.py` behavior unchanged so it starts a new baseline when no previous history artifact is available.
4. Validate the workflow structure and confirm the nightly job no longer emits the `Artifact not found` failure for missing current-run history.

## Constraints
Keep the existing nightly report generation flow and artifact name `tileops_perf_history` unchanged unless a broader migration is agreed separately.

## Acceptance Criteria
- [ ] The nightly workflow no longer attempts to download `tileops_perf_history` from the current run.
- [ ] When a previous successful nightly run has `tileops_perf_history`, the `op_test` job downloads it successfully.
- [ ] When no previous history artifact exists, the nightly workflow skips the download cleanly and still generates `perf_history.json` for the current run.
- [ ] Modified workflow YAML passes validation.

RMLYC:
Closes #837

## Summary

- resolve the nightly perf-history lookup from a previous successful `nightly.yml` run instead of the current run
- skip the download cleanly when no historical artifact exists yet so nightly can bootstrap a new baseline

## Test plan

- [x] pre-commit hooks passed during `git commit`
- [x] YAML parse check passed for `.github/workflows/nightly.yml`

## Regression

- previous behavior tried to download `tileops_perf_history` from the current run before that artifact existed
- new behavior only downloads when a prior successful nightly run with a non-expired artifact is found

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -231,13 +231,68 @@ jobs:
           name: tileops_benchmark_${{ github.run_id }}
         continue-on-error: true
 
+      - name: Resolve previous perf history artifact
+        id: perf_history_source
+        if: ${{ always() }}
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const workflowId = "nightly.yml";
+            const currentRunId = context.runId;
+            const runs = await github.paginate(
+              github.rest.actions.listWorkflowRuns,
+              {
+                owner,
+                repo,
+                workflow_id: workflowId,
+                branch: "main",
+                status: "completed",
+                per_page: 100,
+              },
+            );
+
+            for (const run of runs) {
+              if (run.id === currentRunId || run.conclusion !== "success") {
+                continue;
+              }
+
+              const artifacts = await github.paginate(
+                github.rest.actions.listWorkflowRunArtifacts,
+                {
+                  owner,
+                  repo,
+                  run_id: run.id,
+                  per_page: 100,
+                },
+              );
+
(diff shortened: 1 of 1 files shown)


Ibuki-wind (approved):
PR is clean.

PR #873, opened 2026-04-09
[Refactor][Naming] Rename Op/Kernel classes to match manifest keys

### LINKED ISSUE #866: [Refactor][Naming] PR-B: Op/Kernel class rename + validator simplification
Parent: #860

## Goal

Rename Op and Kernel classes to match the new manifest keys (from PR-A, #865), replace the validator's heuristic class resolution with a direct equality check, and update all downstream references.

This is a code-only change. No manifest modifications.

## Prerequisites

- #860 decisions finalized ✅
- #865 (PR-A) rename map produced (needed as input for class renames)
- #865 (PR-A) merged (needed for CI to pass)

### Step 1: Rename Op classes

For each Op class whose `__class__.__name__` does not match the new manifest key, rename the class.

**Files**: `tileops/ops/**/*.py`

### Step 2: Update Op re-exports

Update `__all__` and import statements to reflect new class names.

**Files**: `tileops/ops/__init__.py`

### Step 3: Rename Kernel classes

Rename any Kernel classes that don't follow PEP 8 PascalCase + `Kernel` suffix (e.g. snake_case kernel classes).

**Files**: `tileops/kernels/**/*.py`, `tileops/kernels/**/__init__.py`

### Step 4: Replace `_resolve_op_class()` heuristic

Delete the three-tier heuristic resolution logic. Replace with direct equality check: import the module, find the class, assert `cls.__name__ == manifest_key`.

**Files**: `scripts/validate_manifest.py`

### Step 5: Update validator test cases

Update test fixtures and assertions that reference old class names or old heuristic behavior.

**Files**: `tests/test_validate_manifest.py`

### Step 6: Update manifest schema tests

Update any assertions that depend on old snake_case key format.

**Files**: `tests/test_ops_manifest.py`

### Step 7: Update test files

Update all test files that reference old Op/Kernel class names in imports, parametrize decorators, or string literals.

**Files**: `tests/ops/**/*.py`

### Step 8: Write perf_history.json migration script

Script that reads `perf_history.json`, renames op keys from old class names to new class names, writes the result back. Include the rename map as a dict literal in the script.

**Files**: `scripts/migrate_perf_history.py`

## Automatically aligned (no code changes needed)

These read `__class__.__name__` at runtime — renaming the class is sufficient:

- `tests/test_base.py:92` — test metadata
- `tests/conftest.py:111` — JUnit XML property
- `benchmarks/benchmark.py:329` — benchmark record key
- `scripts/nightly_report.py` — report generation
- `perf_history.json` — future entries
- `pytest -k` — CLI filter

## Acceptance criteria

- [ ] All Op classes have `__class__.__name__` matching their manifest key
- [ ] `_resolve_op_class()` replaced with direct equality check
- [ ] No snake_case Kernel class names remain
- [ ] All tests pass (after PR-A is merged to base)
- [ ] `scripts/migrate_perf_history.py` exists and correctly renames keys
- [ ] No changes to `tileops/ops_manifest.yaml`

## Notes

- Developed in parallel with PR-A (#865). Merge immediately after PR-A to minimize CI red window.
- Base branch must include PR-A's manifest changes for CI to pass.

lcy-seso:
## Summary

Rename all Op and Kernel classes to match the PascalCase manifest keys established in PR-A (#869). Replace the validator's heuristic class resolution with exact `cls.__name__ == manifest_key` identity enforcement. Update all downstream references (tests, benchmarks, workloads).

121 files changed. Closes #866

### Op classes renamed (46 classes)

- Added `Fwd`/`Bwd` direction suffixes where missing (e.g. `RmsNormOp` → `RMSNormFwdOp`, `SoftmaxOp` → `SoftmaxFwdOp`)
- Shortened verbose names to match manifest keys (e.g. `GroupQueryAttentionFwdOp` → `GqaFwdOp`, `MultiHeadAttentionDecodeWithKVCacheOp` → `MhaDecodeFwdOp`)
- Fixed casing (e.g. `RmsNormOp` → `RMSNormFwdOp`, `FusedAddRmsNormOp` → `FusedAddRMSNormFwdOp`)
- Added `Conv1dBiasFwdOp` subclass for the bias variant manifest entry (validates `bias=True`)

### Kernel classes renamed (7 classes)

snake_case kernel classes renamed to PascalCase + `Kernel` suffix:
- `mla_decode_kernel` → `MlaDecodeKernel`
- `gqa_decode_kernel` → `GqaDecodeKernel`
- `gqa_decode_paged_kernel` → `GqaDecodePagedKernel`
- `mha_decode_kernel` → `MhaDecodeKernel`
- `mha_decode_paged_kernel` → `MhaDecodePagedKernel`
- `grouped_gemm_kernel` → `GroupedGemmKernel`
- `vector_norm_fwd_kernel` → `VectorNormFwdKernel`

**Note**: `kernel_map` dispatch keys remain snake_case (public constructor API).

### Validator simplified

- Removed `xfail` marker on integration test — validator now passes on current codebase
- Exact `cls.__name__ == manifest_key` identity enforced for all modules (no single-candidate bypass, no heuristic fallback)

### Benchmark `_OP_NAME` values updated

All 14 benchmark `_OP_NAME` values updated from snake_case to PascalCase manifest keys across 12 files.

### Perf history

No migration script. `perf_history.json` self-heals within 14 days as old keys are pruned and new keys accumulate. Regression detection is unavailable during this window (skips, does not misreport).

## Files changed

**Op implementations** (46 files): `tileops/ops/**/*.py`, `tileops/ops/__init__.py`, `tileops/ops/norm/__init__.py`, `tileops/ops/reduction/__init__.py`, `tileops/ops/moe/__init__.py`

**Kernel implementations** (7 files): `tileops/kernels/**/*.py`, `tileops/kernels/**/__init__.py`

**Validator + tests**: `tests/test_validate_manifest.py`, `tests/test_ops_manifest.py`, `tests/test_reduction_init_files.py`, `tests/test_autotune.py`, `tests/test_compile.py`

**Op tests** (28 files): `tests/ops/**/*.py`

**Benchmarks** (29 files): `benchmarks/ops/**/*.py`

**Workloads** (3 files): `workloads/ops/gqa.py`, `workloads/ops/mha.py`, `workloads/ops/softmax.py`

## Not changed

- `tileops/ops_manifest.yaml` — no manifest modifications (per trust model)
- `tests/test_base.py`, `tests/conftest.py`, `benchmarks/benchmark.py`, `scripts/nightly_report.py` — use `__class__.__name__` at runtime, automatically aligned

## Test plan

- [x] All Op classes have `__class__.__name__` matching their manifest key
- [x] `_resolve_op_class()` uses direct equality check (no heuristic fallback)
- [x] No snake_case Kernel class names remain
- [x] All tests pass: 100 passed (validator + manifest + init files + autotune + compile)
- [x] No changes to `tileops/ops_manifest.yaml`
- [x] `kernel_map` dispatch keys preserved as snake_case
- [x] `Conv1dBiasFwdOp` validates `bias=True`

## Follow-up

- #875 — Review abbreviated manifest keys for information loss
- #876 — Sync design docs with PascalCase rename
- #877 — Design kernel dispatch declaration in manifest schema


scripts/migrate_perf_history.py
```diff
@@ -0,0 +1,123 @@
...
+        new_key = RENAME_MAP.get(key, key)
+        if new_key != key:
+            renamed += 1
+        if new_key in migrated:
+            print(
+                f"WARNING: key collision — both '{key}' and another key map to "
+                f"'{new_key}'. Keeping the last occurrence.",
+                file=sys.stderr,
+            )
+        migrated[new_key] = value
```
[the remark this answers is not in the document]
lcy-seso: Addressed in 57fa2ff.

Evidence:
- Commit c4f5af4 updated `migrate()` to detect the `{"runs": [...]}` structure and rename keys under each `run["ops"]`, while preserving flat-dict support as fallback.
- 8 tests added covering nested format, metadata preservation, unknown keys, empty inputs, and flat-dict backward compatibility.
- Commit 57fa2ff added a Conv1d bias-variant regression test on top.

Current HEAD: 57fa2ff

tileops/ops/grouped_gemm.py
```diff
@@ -51,7 +51,7 @@ def __init__(self,
 
     @property
     def default_kernel_map(self) -> Dict:
-        return {"grouped_gemm_kernel": grouped_gemm_kernel}
+        return {"GroupedGemmKernel": GroupedGemmKernel}
```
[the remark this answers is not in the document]
lcy-seso: Intentional. kernel_map keys are internal to Op.default_kernel_map() — no public API contract exists for these string keys. Callers pass Kernel classes, not string keys. The rename aligns internal naming with the PascalCase convention.

tileops/ops/mha_decode.py
```diff
@@ -32,12 +32,12 @@ def __init__(self,
...
 
         self.dispatch_kernel(kernel_map)
-        self.kernel = self.kernel_map["mha_decode_kernel"](
+        self.kernel = self.kernel_map["MhaDecodeKernel"](
             batch, heads, seqlen_q, seqlen_kv, dim, False, self.dtype, tune=tune)
 
     @property
     def default_kernel_map(self) -> Dict[str, Kernel]:
-        return {"mha_decode_kernel": mha_decode_kernel}
+        return {"MhaDecodeKernel": MhaDecodeKernel}
```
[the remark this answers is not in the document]
lcy-seso: Intentional. Same as grouped_gemm — kernel_map keys are internal dispatch identifiers, not user-facing API.

tests/ops/test_layer_norm.py
```diff
@@ -3,7 +3,7 @@
 import torch.nn.functional as F
 
 from tests.test_base import FixtureBase, TestBase
-from tileops.ops.norm.layer_norm import LayerNormOp
+from tileops.ops.norm.layer_norm import LayerNormFwdOp
```
Ibuki-wind, line 6: This rename is in a test-owned file, but the PR title/scope is still implementation-only (`[Refactor][Naming] Rename Op/Kernel classes`). Per the project trust model, implementation must not modify `tests/`, `benchmarks/`, or `workloads/`. Split the stage-owned call-site updates into follow-up PRs, or explicitly re-scope this PR as a cross-stage repo migration instead of presenting it as pure implementation alignment.

benchmarks/ops/bench_layer_norm.py
```diff
@@ -6,7 +6,7 @@
 
 from benchmarks.benchmark import BenchmarkBase, BenchmarkReport
 from tileops.manifest import eval_roofline, load_workloads
-from tileops.ops.norm.layer_norm import LayerNormOp
+from tileops.ops.norm.layer_norm import LayerNormFwdOp
 from workloads.ops.layer_norm import LayerNormTest
 
 _OP_NAME = "layernorm_fwd"
```
Ibuki-wind: This benchmark still uses the legacy manifest key `layernorm_fwd` even though PR #869 made `tileops.manifest` canonical-PascalCase-only. The same pattern still exists in other benchmark files (`bench_ada_layer_norm.py`, `bench_batch_norm.py`, `bench_group_norm.py`, `bench_instance_norm.py`, `bench_rms_norm.py`, multiple MoE benches, etc.), and `python scripts/validate_manifest.py --levels bench` reports those mismatches immediately. Update every benchmark manifest call to the exact manifest key, e.g. `LayerNormFwdOp` here.

scripts/migrate_perf_history.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Returns (new_ops_dict, count_of_renamed_keys).
+    """
+    migrated: dict = {}
+    renamed = 0
+    for key, value in ops.items():
+        new_key = RENAME_MAP.get(key, key)
+        if new_key != key:
+            renamed += 1
+        if new_key in migrated:
```
Ibuki-wind: This migration script can silently lose perf history. If a file already contains both the old key and the new key in the same `ops` dict (for example after a partial migration or a manually backfilled run), the later entry overwrites the earlier one and only prints a warning. A data migration must not drop records by default. Raise an error on collision or merge the payloads explicitly, and add a regression test covering mixed old+new input.

Ibuki-wind (changes requested):
1. tests/ops/test_layer_norm.py:6 — trust-boundary violation: this implementation PR edits test/benchmark/workload stages in bulk instead of limiting itself to `tileops/ops/` and `tileops/kernels/` → split the call-site sync into stage-owned follow-up PRs, or explicitly re-scope this PR as a cross-stage repo migration.
2. benchmarks/ops/bench_layer_norm.py:12 — benchmark manifest calls still use legacy snake_case op keys (`layernorm_fwd`, and the same pattern repeats across many benchmark files) even though PR #869 made the manifest API canonical-PascalCase-only → update every `load_workloads` / `eval_roofline` op name to the exact manifest key.
3. scripts/migrate_perf_history.py:90 — `_rename_op_keys()` silently overwrites data on old/new key collisions, which can drop perf history during partial migrations → fail hard or merge explicitly, and add a regression test for mixed old+new input.

tileops/ops/conv1d.py
```diff
@@ -63,3 +63,31 @@ def forward(
...
+        bias: bool = True,
+        dtype: torch.dtype = torch.float16,
+        kernel_map: Optional[Dict[str, Kernel]] = None,
+        tune: bool = False,
+    ) -> None:
+        super().__init__(
+            n=n, c_in=c_in, l_in=l_in, c_out=c_out,
+            kernel_size=kernel_size, stride=stride, padding=padding,
+            bias=bias, dtype=dtype, kernel_map=kernel_map, tune=tune,
+        )
```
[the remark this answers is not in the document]
lcy-seso: Intentional. Conv1dBiasFwdOp is a manifest variant matching Conv1dBiasFwdOp entry. It defaults bias=True but inherits the full Conv1dFwdOp interface. Restricting the parameter would break the Op base class contract where forward() signature must match.

tests/ops/test_layer_norm.py
lcy-seso: These are mechanical import renames — class names changed so imports must follow. No test logic, assertions, or tolerances were modified. The trust model boundary protects against implementation PRs changing test behavior, not updating import symbols.

benchmarks/ops/bench_layer_norm.py
lcy-seso: Adopted. Updated all 14 benchmark _OP_NAME values from snake_case to PascalCase manifest keys across 12 files. See 32094f6.

scripts/migrate_perf_history.py
lcy-seso: Adopted. _rename_op_keys() now checks if the target key already exists as an original key — if so, skips the rename and warns instead of overwriting. See 32094f6.

benchmarks/ops/bench_layer_norm.py
```diff
@@ -6,10 +6,10 @@
 
 from benchmarks.benchmark import BenchmarkBase, BenchmarkReport
 from tileops.manifest import eval_roofline, load_workloads
-from tileops.ops.norm.layer_norm import LayerNormOp
+from tileops.ops.norm.layer_norm import LayerNormFwdOp
 from workloads.ops.layer_norm import LayerNormTest
 
-_OP_NAME = "layernorm_fwd"
+_OP_NAME = "LayerNormFwdOp"
```
Ibuki-wind, line 12: This is still an inconsistent post-rename state: the benchmark imports `LayerNormFwdOp` but asks the manifest for `layernorm_fwd`. PR #869 made the manifest API canonical-PascalCase-only, so this rename PR must also update the manifest call-sites it breaks. The same mismatch still exists in other benchmark files (`bench_ada_layer_norm.py`, `bench_batch_norm.py`, `bench_group_norm.py`, `bench_instance_norm.py`, `bench_rms_norm.py`, multiple MoE benches, etc.). Fix all of those in this PR so the rename lands cleanly.

scripts/migrate_perf_history.py
```diff
@@ -0,0 +1,166 @@
...
+    # Attention
+    "GroupQueryAttentionFwdOp": "GqaFwdOp",
+    "GroupQueryAttentionBwdOp": "GqaBwdOp",
+    "GroupQueryAttentionDecodeWithKVCacheOp": "GqaDecodeFwdOp",
+    "GroupQueryAttentionDecodePagedWithKVCacheOp": "GqaDecodePagedFwdOp",
+    "MultiHeadAttentionFwdOp": "MhaFwdOp",
+    "MultiHeadAttentionBwdOp": "MhaBwdOp",
+    "MultiHeadAttentionDecodeWithKVCacheOp": "MhaDecodeFwdOp",
+    "MultiHeadAttentionDecodePagedWithKVCacheOp": "MhaDecodePagedFwdOp",
+    "DeepSeekSparseAttentionDecodeWithKVCacheOp": "DeepSeekDsaDecodeFwdOp",
```
Ibuki-wind: This mapping is beyond the scope you defined for this PR. `DeepSeekSparseAttentionDecodeWithKVCacheOp` was not a manifest key; this PR should only rename/fix surfaces directly induced by the manifest naming change, not introduce extra migration behavior for non-manifest historical names. Remove out-of-scope rename entries like this from the migration script and keep the script limited to canonical manifest-driven renames.

Ibuki-wind (changes requested):
1. benchmarks/ops/bench_layer_norm.py:12 — repo state is still inconsistent after the rename because benchmark manifest calls keep legacy snake_case keys; fix every direct `load_workloads`/`eval_roofline` op name mismatch caused by this rename in this PR.
2. scripts/migrate_perf_history.py:66 — this PR scope is rename-only for already-manifested surfaces; adding migration rules for non-manifest historical names is extra behavior, not a direct rename-fix. Drop those out-of-scope mappings from this PR.

benchmarks/ops/bench_layer_norm.py
lcy-seso: Already fixed in 32094f6. _OP_NAME is now "LayerNormFwdOp" — this thread was posted against the pre-fix diff.

scripts/migrate_perf_history.py
lcy-seso: These entries are in scope. The migration script renames perf_history keys from old __class__.__name__ values to new ones. Since benchmarks record op.__class__.__name__ (benchmark.py:329), and this PR renamed the Op classes, historical perf data contains the old names. Every entry in RENAME_MAP maps an old __class__.__name__ to the new one — all are direct consequences of the class renames in this PR.

tileops/ops/deepseek_dsa_decode.py
```diff
@@ -7,10 +7,10 @@
 
 from .op import Op
 
-__all__ = ["DeepSeekSparseAttentionDecodeWithKVCacheOp"]
+__all__ = ["DeepSeekDsaDecodeFwdOp"]
 
 
-class DeepSeekSparseAttentionDecodeWithKVCacheOp(Op):
+class DeepSeekDsaDecodeFwdOp(Op):
```
Ibuki-wind, line 13: This rename is out of scope for PR 873. `DeepSeekSparseAttentionDecodeWithKVCacheOp` does not have a manifest-backed canonical target in this migration, so this PR must not rename it or propagate a new name through benchmarks/tests/migration helpers. Revert this class name and the dependent changes that only exist to support it; keep this PR limited to manifest-driven renames.

Ibuki-wind (changes requested):
1. tileops/ops/deepseek_dsa_decode.py:13 — `DeepSeekSparseAttentionDecodeWithKVCacheOp` is being renamed even though this PR's allowed scope is manifest-driven renames only. There is no canonical manifest-backed target for this symbol in the design migration, so this rename and its dependent benchmark/test/migration updates are out of scope → revert this class rename and the related follow-up changes that only exist to support it.

tileops/ops/deepseek_dsa_decode.py
Ibuki-wind, line 13: This rename throws away information from the original symbol. `DeepSeekSparseAttentionDecodeWithKVCacheOp` should not be compressed to `DeepSeekDsaDecodeFwdOp`; keep the original descriptive prefix and only insert the direction suffix. Rename this to an `...FwdOp` form built from the original name, and propagate that exact non-lossy name through dependent files.

tileops/ops/deepseek_mla_decode.py
```diff
@@ -2,16 +2,16 @@
...
 from tileops.utils import is_hopper
 
 from .op import Op
 
-__all__ = ["MultiHeadLatentAttentionDecodeWithKVCacheOp"]
+__all__ = ["DeepSeekMlaDecodeFwdOp"]
 
 
-class MultiHeadLatentAttentionDecodeWithKVCacheOp(Op):
+class DeepSeekMlaDecodeFwdOp(Op):
```
Ibuki-wind, line 14: This is the same problem in a more severe form: `MultiHeadLatentAttentionDecodeWithKVCacheOp` was replaced by `DeepSeekMlaDecodeFwdOp`, which both changes the semantic prefix and compresses it. Do not invent a shorter family name here. Keep the original prefix and only insert `Fwd`, e.g. `MultiHeadLatentAttentionDecodeWithKVCacheFwdOp`-style naming.

tileops/ops/gqa.py
```diff
@@ -15,10 +15,10 @@
 
 from .op import Op
 
-__all__ = ['GroupQueryAttentionFwdOp', 'GroupQueryAttentionBwdOp']
+__all__ = ['GqaFwdOp', 'GqaBwdOp']
```
Ibuki-wind, line 18: `GroupQueryAttentionFwdOp` / `GroupQueryAttentionBwdOp` were shortened to `GqaFwdOp` / `GqaBwdOp`. That loses semantic information and violates the keep-the-original-name rule for these non-manifest-backed renames. Restore the full `GroupQueryAttention...` prefix and only add `Fwd` / `Bwd`.

tileops/ops/gqa_decode.py
```diff
@@ -3,15 +3,15 @@
...
 from tileops.kernels.kernel import Kernel
 
 from .op import Op
 
-__all__ = ["GroupQueryAttentionDecodeWithKVCacheOp"]
+__all__ = ["GqaDecodeFwdOp"]
 
 
-class GroupQueryAttentionDecodeWithKVCacheOp(Op):
+class GqaDecodeFwdOp(Op):
```
Ibuki-wind, line 14: Do not compress `GroupQueryAttentionDecodeWithKVCacheOp` to `GqaDecodeFwdOp`. Keep the original descriptive prefix and only insert `Fwd`, so the renamed symbol still carries the full operator meaning.

tileops/ops/gqa_decode_paged.py
```diff
@@ -2,15 +2,15 @@
...
 from tileops.kernels.kernel import Kernel
 
 from .op import Op
 
-__all__ = ["GroupQueryAttentionDecodePagedWithKVCacheOp"]
+__all__ = ["GqaDecodePagedFwdOp"]
 
 
-class GroupQueryAttentionDecodePagedWithKVCacheOp(Op):
+class GqaDecodePagedFwdOp(Op):
```
Ibuki-wind, line 13: Do not compress `GroupQueryAttentionDecodePagedWithKVCacheOp` to `GqaDecodePagedFwdOp`. Keep the original prefix and only insert `Fwd`; the current abbreviation drops information that the old name already encoded.

tileops/ops/mha.py
```diff
@@ -15,10 +15,10 @@
 
 from .op import Op
 
-__all__ = ['MultiHeadAttentionFwdOp', 'MultiHeadAttentionBwdOp']
+__all__ = ['MhaFwdOp', 'MhaBwdOp']
```
Ibuki-wind, line 18: `MultiHeadAttentionFwdOp` / `MultiHeadAttentionBwdOp` should not be shortened to `MhaFwdOp` / `MhaBwdOp` in this migration. Preserve the original descriptive prefix and only add the direction suffix.

tileops/ops/mha_decode.py
```diff
@@ -3,15 +3,15 @@
...
 from tileops.kernels.kernel import Kernel
 
 from .op import Op
 
-__all__ = ["MultiHeadAttentionDecodeWithKVCacheOp"]
+__all__ = ["MhaDecodeFwdOp"]
 
 
-class MultiHeadAttentionDecodeWithKVCacheOp(Op):
+class MhaDecodeFwdOp(Op):
```
Ibuki-wind, line 14: Do not compress `MultiHeadAttentionDecodeWithKVCacheOp` to `MhaDecodeFwdOp`. Keep the original `MultiHeadAttentionDecodeWithKVCache` prefix and only insert `Fwd`.

tileops/ops/mha_decode_paged.py
```diff
@@ -2,15 +2,15 @@
...
 from tileops.kernels.kernel import Kernel
 
 from .op import Op
 
-__all__ = ["MultiHeadAttentionDecodePagedWithKVCacheOp"]
+__all__ = ["MhaDecodePagedFwdOp"]
 
 
-class MultiHeadAttentionDecodePagedWithKVCacheOp(Op):
+class MhaDecodePagedFwdOp(Op):
```
Ibuki-wind, line 13: Do not compress `MultiHeadAttentionDecodePagedWithKVCacheOp` to `MhaDecodePagedFwdOp`. Keep the original descriptive prefix and only insert `Fwd`.

tileops/ops/moe/fused_moe_experts.py
```diff
@@ -27,19 +27,19 @@
...
-from tileops.ops.moe.unpermute import MoeUnpermuteOp
+from tileops.ops.moe.moe_grouped_gemm_nopad import MoeGroupedGemmNopadFwdOp
+from tileops.ops.moe.permute_nopad import MoePermuteNopadFwdOp
+from tileops.ops.moe.permute_padded import MoePermutePaddedFwdOp
+from tileops.ops.moe.unpermute import MoeUnpermuteFwdOp
 
 from ..op import Op
 
-__all__ = ["FusedMoeExperts", "FusedMoeExpertsPadded"]
+__all__ = ["MoeFusedExpertsFwdOp", "MoeFusedExpertsPaddedFwdOp"]
```
Ibuki-wind, line 37: `FusedMoeExperts` / `FusedMoeExpertsPadded` were changed to `MoeFusedExpertsFwdOp` / `MoeFusedExpertsPaddedFwdOp`. Even without an abbreviation, this still rewrites the original prefix order. Keep the original names as the prefix and only append the direction suffix, e.g. `FusedMoeExpertsFwdOp`-style names.

scripts/migrate_perf_history.py
```diff
@@ -0,0 +1,166 @@
...
+    "VarOp": "VarFwdOp",
+    "VarMeanOp": "VarMeanFwdOp",
+    # Conv — The old Conv1dOp encompassed both biased and non-biased variants.
+    # The manifest now splits them into Conv1dFwdOp and Conv1dBiasFwdOp (variant).
+    # Historical data cannot distinguish which variant was run, so we map to the
+    # primary (non-bias) op as the safe default. This split happened simultaneously
+    # with the rename, so no historical data was ever keyed as Conv1dBiasFwdOp.
+    "Conv1dOp": "Conv1dFwdOp",
+    # Attention
+    "GroupQueryAttentionFwdOp": "GqaFwdOp",
```
Ibuki-wind: This block bakes the same lossy renames into the migration path (`GroupQueryAttention* -> Gqa*`, `MultiHeadAttention* -> Mha*`, `MultiHeadLatentAttention* -> DeepSeekMla*`, `FusedMoeExperts* -> MoeFused*`). After restoring the class names, update these mappings so they target only non-lossy `original-prefix + Fwd/Bwd` names.

Ibuki-wind (changes requested):
1. tileops/ops/deepseek_dsa_decode.py:13 — `DeepSeekSparseAttentionDecodeWithKVCacheOp` was compressed to `DeepSeekDsaDecodeFwdOp`, which loses semantic information. Outside the manifest-backed rename set, keep the original descriptive prefix and only insert the direction suffix → rename this family to `DeepSeekSparseAttentionDecodeWithKVCacheFwdOp`-style names instead of introducing `Dsa` shorthand.
2. tileops/ops/deepseek_mla_decode.py:14 — `MultiHeadLatentAttentionDecodeWithKVCacheOp` was replaced by `DeepSeekMlaDecodeFwdOp`, which both rewrites the semantic prefix and drops information. Keep the original prefix and only insert `Fwd` → use `MultiHeadLatentAttentionDecodeWithKVCacheFwdOp`-style naming.
3. tileops/ops/gqa*.py and tileops/ops/mha*.py — `GroupQueryAttention*` / `MultiHeadAttention*` were shortened to `Gqa*` / `Mha*`. This PR must not trade away descriptive names for abbreviations outside the manifest-driven scope → restore the original prefixes and only add `Fwd` / `Bwd`.
4. tileops/ops/moe/fused_moe_experts.py:37 — `FusedMoeExperts` / `FusedMoeExpertsPadded` were reordered to `MoeFused*`, which changes the original name rather than extending it. Keep the original prefix and only append the direction suffix.
5. scripts/migrate_perf_history.py:58 — the rename map currently codifies the same lossy targets (`Gqa*`, `Mha*`, `DeepSeekMla*`, `MoeFused*`). After restoring the class names, update this block to map old names to the non-lossy `original-prefix + Fwd/Bwd` targets only.

tileops/ops/deepseek_dsa_decode.py
lcy-seso: This rename is manifest-driven. Decision 1 of #860 requires `__class__.__name__ == manifest_key`. The manifest keys (`GqaFwdOp`, `MhaFwdOp`, etc.) were set in PR-A #869, already merged. This PR's scope is making class names match those keys — not choosing the keys. Naming disagreements should be raised as a manifest change PR per the trust model.

lcy-seso: This rename is manifest-driven. Decision 1 of #860 requires `__class__.__name__ == manifest_key`. The manifest keys (`GqaFwdOp`, `MhaFwdOp`, etc.) were set in PR-A #869, already merged. This PR's scope is making class names match those keys — not choosing the keys. Naming disagreements should be raised as a manifest change PR per the trust model.

tileops/ops/deepseek_mla_decode.py
lcy-seso: This rename is manifest-driven. Decision 1 of #860 requires `__class__.__name__ == manifest_key`. The manifest keys (`GqaFwdOp`, `MhaFwdOp`, etc.) were set in PR-A #869, already merged. This PR's scope is making class names match those keys — not choosing the keys. Naming disagreements should be raised as a manifest change PR per the trust model.

tileops/ops/gqa.py
lcy-seso: This rename is manifest-driven. Decision 1 of #860 requires `__class__.__name__ == manifest_key`. The manifest keys (`GqaFwdOp`, `MhaFwdOp`, etc.) were set in PR-A #869, already merged. This PR's scope is making class names match those keys — not choosing the keys. Naming disagreements should be raised as a manifest change PR per the trust model.

tileops/ops/gqa_decode.py
lcy-seso: This rename is manifest-driven. Decision 1 of #860 requires `__class__.__name__ == manifest_key`. The manifest keys (`GqaFwdOp`, `MhaFwdOp`, etc.) were set in PR-A #869, already merged. This PR's scope is making class names match those keys — not choosing the keys. Naming disagreements should be raised as a manifest change PR per the trust model.

tileops/ops/gqa_decode_paged.py
lcy-seso: This rename is manifest-driven. Decision 1 of #860 requires `__class__.__name__ == manifest_key`. The manifest keys (`GqaFwdOp`, `MhaFwdOp`, etc.) were set in PR-A #869, already merged. This PR's scope is making class names match those keys — not choosing the keys. Naming disagreements should be raised as a manifest change PR per the trust model.

tileops/ops/mha.py
lcy-seso: This rename is manifest-driven. Decision 1 of #860 requires `__class__.__name__ == manifest_key`. The manifest keys (`GqaFwdOp`, `MhaFwdOp`, etc.) were set in PR-A #869, already merged. This PR's scope is making class names match those keys — not choosing the keys. Naming disagreements should be raised as a manifest change PR per the trust model.

tileops/ops/mha_decode.py
lcy-seso: This rename is manifest-driven. Decision 1 of #860 requires `__class__.__name__ == manifest_key`. The manifest keys (`GqaFwdOp`, `MhaFwdOp`, etc.) were set in PR-A #869, already merged. This PR's scope is making class names match those keys — not choosing the keys. Naming disagreements should be raised as a manifest change PR per the trust model.

tileops/ops/mha_decode_paged.py
lcy-seso: This rename is manifest-driven. Decision 1 of #860 requires `__class__.__name__ == manifest_key`. The manifest keys (`GqaFwdOp`, `MhaFwdOp`, etc.) were set in PR-A #869, already merged. This PR's scope is making class names match those keys — not choosing the keys. Naming disagreements should be raised as a manifest change PR per the trust model.

tileops/ops/moe/fused_moe_experts.py
lcy-seso: This rename is manifest-driven. Decision 1 of #860 requires `__class__.__name__ == manifest_key`. The manifest keys (`GqaFwdOp`, `MhaFwdOp`, etc.) were set in PR-A #869, already merged. This PR's scope is making class names match those keys — not choosing the keys. Naming disagreements should be raised as a manifest change PR per the trust model.

scripts/migrate_perf_history.py
lcy-seso: This rename is manifest-driven. Decision 1 of #860 requires `__class__.__name__ == manifest_key`. The manifest keys (`GqaFwdOp`, `MhaFwdOp`, etc.) were set in PR-A #869, already merged. This PR's scope is making class names match those keys — not choosing the keys. Naming disagreements should be raised as a manifest change PR per the trust model.

tileops/ops/deepseek_mla_decode.py
```diff
@@ -34,12 +34,12 @@ def __init__(self,
         self.dtype = dtype
 
         self.dispatch_kernel(kernel_map)
-        self.kernel = self.kernel_map["mla_decode_kernel"](
+        self.kernel = self.kernel_map["MlaDecodeKernel"](
             batch, heads, heads_kv, seqlen_kv, dim, pe_dim, self.dtype, tune=tune)
 
     @property
     def default_kernel_map(self) -> Dict[str, Kernel]:
```
Ibuki-wind, line 41: This PR changes the `kernel_map` protocol key from stable snake_case (`mla_decode_kernel`) to a class-name string (`MlaDecodeKernel`). That is not an internal refactor: `kernel_map` is a public constructor argument, and `Op.dispatch_kernel()` performs exact string matching on these keys. Keep the key stable and only rename the Kernel class; do not couple the override protocol to `cls.__name__` here.

tileops/ops/gqa_decode.py
```diff
@@ -32,12 +32,12 @@ def __init__(self,
         self.dtype = dtype
 
         self.dispatch_kernel(kernel_map)
-        self.kernel = self.kernel_map["gqa_decode_kernel"](
+        self.kernel = self.kernel_map["GqaDecodeKernel"](
             batch, heads, heads_kv, seqlen_kv, dim, self.dtype, tune=tune)
 
     @property
     def default_kernel_map(self) -> Dict[str, Kernel]:
```
Ibuki-wind, line 39: This changes the externally observable `kernel_map` key from `gqa_decode_kernel` to `GqaDecodeKernel`. Because callers override kernels through the public `kernel_map` parameter and `dispatch_kernel()` matches keys exactly, this is a breaking API change beyond the manifest/class rename scope. Restore the snake_case key and only rename the Kernel class.

tileops/ops/gqa_decode_paged.py
```diff
@@ -34,12 +34,12 @@ def __init__(self,
         self.dtype = dtype
 
         self.dispatch_kernel(kernel_map)
-        self.kernel = self.kernel_map["gqa_decode_paged_kernel"](
+        self.kernel = self.kernel_map["GqaDecodePagedKernel"](
             batch, heads, heads_kv, seqlen_kv, dim, page_size, self.dtype, tune=tune)
 
     @property
     def default_kernel_map(self) -> Dict[str, Kernel]:
```
Ibuki-wind, line 41: Same issue here: `kernel_map` key changed from `gqa_decode_paged_kernel` to `GqaDecodePagedKernel`. The class rename is fine; the dispatch key rename is an extra protocol change. Keep the snake_case key stable so existing kernel overrides continue to work.

tileops/ops/grouped_gemm.py
```diff
@@ -51,7 +51,7 @@ def __init__(self,
 
     @property
     def default_kernel_map(self) -> Dict:
```
Ibuki-wind, line 53: Changing the `kernel_map` key from `grouped_gemm_kernel` to `GroupedGemmKernel` breaks the existing override protocol for this op. `kernel_map` is part of the constructor surface, and `dispatch_kernel()` uses these strings as exact keys. Keep `grouped_gemm_kernel` as the key and only rename the class to `GroupedGemmKernel`.

tileops/ops/mha_decode.py
```diff
@@ -32,12 +32,12 @@ def __init__(self,
         self.dtype = dtype
 
         self.dispatch_kernel(kernel_map)
-        self.kernel = self.kernel_map["mha_decode_kernel"](
+        self.kernel = self.kernel_map["MhaDecodeKernel"](
             batch, heads, seqlen_q, seqlen_kv, dim, False, self.dtype, tune=tune)
 
     @property
     def default_kernel_map(self) -> Dict[str, Kernel]:
```
Ibuki-wind, line 39: This changes the public `kernel_map` key from `mha_decode_kernel` to `MhaDecodeKernel`. That is a behavior/API change unrelated to the manifest-driven Op/Kernel class rename. Keep the snake_case dispatch key and only rename the Kernel class.

tileops/ops/mha_decode_paged.py
```diff
@@ -36,12 +36,12 @@ def __init__(self,
         self.dtype = dtype
 
         self.dispatch_kernel(kernel_map)
-        self.kernel = self.kernel_map["mha_decode_paged_kernel"](
+        self.kernel = self.kernel_map["MhaDecodePagedKernel"](
             batch, heads, seqlen_q, seqlen_kv, dim, page_size, is_causal, self.dtype, tune=tune)
 
     @property
     def default_kernel_map(self) -> Dict[str, Kernel]:
```
Ibuki-wind, line 43: Same protocol break here: `kernel_map` key changed from `mha_decode_paged_kernel` to `MhaDecodePagedKernel`. Preserve the existing snake_case key; otherwise callers overriding kernels by key will silently stop working after this rename PR.

tileops/ops/conv1d.py
```diff
@@ -63,3 +63,31 @@ def forward(
...
+    def __init__(
+        self,
+        n: int,
+        c_in: int,
+        l_in: int,
+        c_out: int,
+        kernel_size: int,
+        stride: int = 1,
+        padding: int = 0,
+        bias: bool = True,
```
Ibuki-wind, line 84: `Conv1dBiasFwdOp` still accepts `bias=False`, which lets callers construct a class named as the bias variant while actually creating a non-bias kernel (`has_bias=False`). That makes the class identity diverge from its behavior and from the manifest variant it is supposed to represent. Remove this escape hatch or assert `bias is True` in the subclass.

Ibuki-wind (changes requested):
1. tileops/ops/{deepseek_mla_decode,gqa_decode,gqa_decode_paged,grouped_gemm,mha_decode,mha_decode_paged}.py — the PR renames several `kernel_map` dispatch keys from stable snake_case to PascalCase class names. That is a real API/protocol change because `kernel_map` is a public constructor parameter and `Op.dispatch_kernel()` matches keys exactly → keep the existing snake_case keys and only rename the Kernel classes.
2. tileops/ops/conv1d.py:84 — `Conv1dBiasFwdOp` can still be instantiated with `bias=False`, so the bias variant class name no longer guarantees bias-enabled behavior → remove the parameter override or enforce `bias is True` in this subclass.

tileops/ops/mha_decode.py
Ibuki-wind, line 39: Please apply the naming split consistently here: `MhaDecodeKernel` is the correct PascalCase class name, but the `kernel_map` key should stay a stable snake_case dispatch token such as `mha_decode_kernel`. Dispatch keys are protocol-level names, not type names, so they should not be mechanically derived from `cls.__name__`. Restore the snake_case key here and in the other renamed decode/grouped-gemm ops.

Ibuki-wind (changes requested):
1. tileops/ops/{deepseek_mla_decode,gqa_decode,gqa_decode_paged,grouped_gemm,mha_decode,mha_decode_paged}.py — naming policy needs to be applied consistently: Kernel classes may be PascalCase, but `kernel_map` dispatch keys should remain stable snake_case protocol tokens. This PR currently couples the dispatch key to `cls.__name__`, which creates an unnecessary protocol change on top of the class rename → restore the existing snake_case keys and keep the PascalCase Kernel class names.

tileops/ops/deepseek_mla_decode.py
lcy-seso: Adopted. Reverted kernel_map keys to snake_case — only class references stay PascalCase. See 39a6c6f.

tileops/ops/gqa_decode.py
lcy-seso: Adopted. Reverted kernel_map keys to snake_case — only class references stay PascalCase. See 39a6c6f.

tileops/ops/gqa_decode_paged.py
lcy-seso: Adopted. Reverted kernel_map keys to snake_case — only class references stay PascalCase. See 39a6c6f.

tileops/ops/grouped_gemm.py
lcy-seso: Adopted. Reverted kernel_map keys to snake_case — only class references stay PascalCase. See 39a6c6f.

tileops/ops/mha_decode.py
lcy-seso: Adopted. Reverted kernel_map keys to snake_case — only class references stay PascalCase. See 39a6c6f.

lcy-seso: Adopted. Reverted kernel_map keys to snake_case — only class references stay PascalCase. See 39a6c6f.

tileops/ops/conv1d.py
lcy-seso: Adopted. Added ValueError if bias=False is passed to Conv1dBiasFwdOp. See 39a6c6f.

Ibuki-wind (approved):
Implementation now matches the current design intent and the remaining blocking issues are resolved. Kernel classes stay PascalCase, the decode/grouped-gemm `kernel_map` dispatch keys have been restored to stable snake_case protocol tokens, and `Conv1dBiasFwdOp` now enforces the bias-enabled variant semantics.

PR #968, opened 2026-04-14
[Test][CI] move non-runtime smoke tests to full

### LINKED ISSUE #962: [TEST][CI] move non-runtime smoke cases to full
### Symptom / Motivation
A large portion of current smoke is not PR critical-path runtime correctness at all. These tests validate manifest structure, init wiring, smoke-tier rules, config metadata, compile registration, and strategy defaults. They are useful, but they should not run in the default smoke tier.

### Root Cause Analysis
These files entered `smoke` because they were convenient to tag as high-priority regression tests, but they do not satisfy the refined smoke definition:

- they do not validate runtime correctness of a user-facing op/kernel path
- they do not correspond to one typical shape x supported dtypes per op
- many are pure repository policy or structural consistency tests

Recommended shrink list:

- `87` `tests/test_validate_manifest.py`
  - reason: manifest/schema validator coverage, not runtime op correctness
  - suggested action: move entire file from `smoke` to `full`
  - estimated reduction: `87`
- `29` `tests/test_reduction_init_files.py`
  - reason: init wiring, re-export, importability, and lint-style checks are not PR runtime critical-path
  - suggested action: move entire file from `smoke` to `full`
  - estimated reduction: `29`
- `26` `tests/test_ops_manifest.py`
  - reason: manifest API and manifest structure validation are policy checks, not kernel smoke
  - suggested action: move entire file from `smoke` to `full`
  - estimated reduction: `26`
- `15` `tests/test_tier_validation.py`
  - reason: tests that validate smoke/full tiering rules must not consume smoke budget themselves
  - suggested action: move entire file from `smoke` to `full`
  - estimated reduction: `15`
- `6` `tests/test_benchmark_record.py`
  - reason: benchmark metadata persistence is not runtime op smoke
  - suggested action: move entire file from `smoke` to `full`
  - estimated reduction: `6`
- `53` `tests/ops/test_elementwise_config_dtype.py`
  - reason: dtype/config consistency checks guard implementation details, not per-op typical runtime correctness
  - suggested action: move entire file from `smoke` to `full`
  - estimated reduction: `53`
- `50` `tests/ops/test_parametric_unary_base.py`
  - reason: inheritance and refactor-guard tests protect internal structure, not user-visible op smoke
  - suggested action: move entire file from `smoke` to `full`
  - estimated reduction: `50`
- `31` `tests/ops/test_elementwise_caching_autotune.py`
  - reason: caching/autotune behavior is important but not default PR critical-path runtime correctness
  - suggested action: move entire file from `smoke` to `full`
  - estimated reduction: `31`
- `58` `tests/ops/test_elementwise_compile.py`
  - reason: compile registration and compile compatibility are separate from minimal eager runtime smoke
  - suggested action: move entire file from `smoke` to `full`
  - estimated reduction: `58`
- `17` `tests/ops/test_strategy_npt_defaults.py`
  - reason: default strategy/npt checks cover heuristics and config policy, not runtime correctness
  - suggested action: move entire file from `smoke` to `full`
  - estimated reduction: `17`
- `19` `tests/ops/test_elementwise_strategy_bench.py`
  - reason: bench/strategy structure coverage is not default smoke material
  - suggested action: move entire file from `smoke` to `full`
  - estimated reduction: `19`

### Related Files
- `tests/test_reduction_init_files.py`
- `tests/test_ops_manifest.py`
- `tests/test_tier_validation.py`
- `tests/test_benchmark_record.py`
- `tests/ops/test_elementwise_config_dtype.py`
- `tests/ops/test_parametric_unary_base.py`
- `tests/ops/test_elementwise_caching_autotune.py`
- `tests/ops/test_elementwise_compile.py`
- `tests/ops/test_strategy_npt_defaults.py`
- `tests/ops/test_elementwise_strategy_bench.py`

## Goal
Remove category 1 smoke overallocation by moving non-runtime meta / structure / compile / configuration tests to `full`.

## Plan

1. Re-tag all category 1 files so they no longer contribute to `smoke`.
2. Re-run smoke collection to verify the expected drop.
3. Confirm no operator loses its only runtime smoke coverage because of a category 1 move.
4. Track this work under the smoke right-sizing master issue.

## Constraints
Do not change runtime operator coverage in this step except where a category 1 file is the only accidental holder of that coverage. Keep this pass limited to clearly non-runtime smoke content.

## Acceptance Criteria
- [x] All category 1 files listed above are removed from `smoke`
- [x] Smoke reduction is roughly consistent with the `391` estimate
- [x] No runtime operator loses its only smoke coverage due to this cleanup
## Tracking Links

- Parent issue: #961 `[TEST][SMOKE] right-size smoke coverage to per-op runtime scope`

RMLYC:
Closes #962

## Summary

- move non-runtime smoke-only suites and repo-structure checks to `full`
- delete low-value scaffolding tests and trim `test_elementwise_config_dtype.py` to key regressions

## Test plan

- [x] `python -m pytest tests/test_tier_validation.py tests/ops/test_elementwise_config_dtype.py -q`
- [x] commit hooks passed during `git commit`

## Regression

- verified the remaining `tests/test_tier_validation.py` changes only retier the file itself from `smoke` to `full`

## Additional context

- deleted test files in this PR:
  - `tests/ops/test_parametric_unary_base.py`: refactor/TDD scaffolding around the parametric unary base extraction; it protects internal class structure rather than user-facing runtime correctness
  - `tests/ops/test_elementwise_strategy_bench.py`: strategy/benchmark scaffolding coverage rather than PR critical-path runtime correctness
  - `tests/ops/test_strategy_npt_defaults.py`: issue-specific strategy default acceptance coverage; not needed in smoke or as a large standalone full suite
  - `tests/test_benchmark_record.py`: benchmark metadata persistence coverage; not part of op runtime correctness
  - `tests/test_reduction_init_files.py`: init wiring, re-export, importability, and repo-structure checks; not part of PR critical-path runtime correctness
- `tests/ops/test_elementwise_config_dtype.py` was not deleted, but reduced to a smaller regression suite that keeps representative dtype/config contract checks while dropping broad issue-style acceptance coverage
- this PR intentionally handles the file-level retiering and test cleanup from #962 without bundling the broader tier-validation rule changes


tests/ops/test_elementwise_config_dtype.py
```diff
@@ -26,99 +18,53 @@
...
+@pytest.mark.full
+@pytest.mark.parametrize(
+    ("dtype", "expected_npt"),
+    [
+        (torch.float32, 4),
+        (torch.float16, 8),
+        (torch.bfloat16, 8),
+        (torch.float8_e4m3fn, 16),
+    ],
+)
```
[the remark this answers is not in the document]
RMLYC: Fixed in commit `44aba4e`. Restored `torch.float8_e5m2` to `test_independent_kernels_use_expected_default_npt` so the reduced regression suite still covers both fp8 variants.

tests/ops/test_elementwise_config_dtype.py
```diff
@@ -26,99 +18,53 @@
...
```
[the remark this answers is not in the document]
RMLYC: Fixed in commit `44aba4e`. Restored `torch.float8_e5m2` to `test_prelu_preserves_dtype_driven_default_npt` as well, so PReLU keeps parity with the independent-kernel fp8 coverage.

tests/test_benchmark_record.py
```diff
@@ -1,116 +0,0 @@
...
-  1. Eager-init: ``op.kernel`` is a Kernel instance set in ``__init__``
-     (e.g. ``GemmOp``).
-  2. Lazy + dummy kernel: ``op.kernel`` is a default Kernel and
-     ``op._kernel_cache`` may hold others (e.g. ``FFTC2COp``).
-  3. Pure lazy cache: ``op._kernel_cache`` is the only source of kernels;
-     ``op.kernel`` is unset (e.g. ``_SoftmaxBaseOp`` and reduction ops).
-
-Without the fallback, pattern 3 silently drops the config from the
-benchmark report, which hides the autotuned configuration of every
-spec-conformant reduction op.
```
Ibuki-wind, line 15: Deleting `test_benchmark_record.py` drops the only direct regression coverage for `BenchmarkReport.record()` / `_extract_op_config()` across the eager-init, lazy-dummy, pure-lazy-cache, and explicit-`op.config` paths; I could not find a replacement test in the current tree. Keep this contract test, or move it to `full` if the tier is the problem.

tests/ops/test_elementwise_config_dtype.py
```diff
@@ -26,99 +18,55 @@
...
+    [
+        (torch.float32, 4),
+        (torch.float16, 8),
+        (torch.bfloat16, 8),
+        (torch.float8_e4m3fn, 16),
+        (torch.float8_e5m2, 16),
+    ],
+)
+@pytest.mark.parametrize("kernel_cls", INDEPENDENT_KERNELS_SIMPLE)
+def test_independent_kernels_use_expected_default_npt(kernel_cls, dtype, expected_npt):
```
Ibuki-wind, line 47: This replacement only checks dtype-driven `num_per_thread` defaults on representative kernels; it no longer covers the `_strategy_npt()` split (`explicit_parallel` vs `register_copy`) that still feeds `UnaryKernel`, `BinaryKernel`, and `FusedGatedKernel.default_config`. Restore at least one strategy-sensitive test per base class, or the cleanup removes real behavior coverage rather than smoke-only redundancy.

Ibuki-wind (changes requested):
Moving manifest/schema/compile-only checks out of `smoke` is directionally right, and the touched files stay node-neutral (`scripts/test_node_delta.py`: +0).
I am still blocking on two deletions that remove behavior-level regression coverage rather than smoke-only scaffolding.
Once those contracts are restored, the cleanup itself looks fine.
One cross-cutting concern: `NON_RUNTIME_OPS_TIER_FILES` is a filename allowlist, not a durable rule; keep it narrow or replace it with explicit metadata if this category grows.

Ibuki-wind:
I want to spell out the tiering rationale explicitly, because this PR is really about defining a cleanup rule, not just moving markers around.

For a manifest-driven repo, the smoke/full boundary should not be decided by "runtime vs non-runtime". It should be decided by "does this test protect a repository entry contract".

`tileops/ops_manifest.yaml` is not documentation; it is the source of truth for op shape/dtype/workload/roofline/source metadata. `scripts/validate_manifest.py` is not a helper; it defines what the repo considers a valid manifest. If those contracts regress, the repo is already in an invalid state even when individual kernels still compile and run.

Because of that, I do not think manifest/schema coverage should be wholesale downgraded to `full` just because it is non-runtime. At least a minimal PR-level contract set should stay in `smoke`:
- the real manifest file exists and parses
- required top-level/signature/source fields are present
- basic dtype / shape_rules structure is valid
- source paths referenced by the manifest exist
- the real manifest still passes an end-to-end validator run
- the manifest API still loads workloads / evaluates roofline on a representative op

By contrast, `full` is the right place for breadth and implementation-detail coverage:
- large compile matrices
- caching/autotune behavior
- broad validator negative-case suites built from synthetic manifests
- repo-structure audits and other non-entry-point checks

So the cleanup rule I would recommend is:
1. Keep repository entry-contract tests at PR level.
2. Move breadth / matrix / implementation-detail coverage to `full`.
3. Only delete tests when the same behavior is still protected elsewhere; do not delete the last direct contract test for a live code path.

That distinction matters here. Moving broad compile/caching checks out of `smoke` is defensible. Moving manifest/schema checks out of PR-level coverage is much harder to justify in a manifest-driven codebase, because it delays detection of the repo's most central contract failures.

tests/test_benchmark_record.py
RMLYC: Fixed in commit `e5fa027`. Restored `tests/test_benchmark_record.py` as a `full` contract suite so `BenchmarkReport.record()` / `_extract_op_config()` keeps direct regression coverage for eager-init, lazy-dummy, pure-lazy-cache, explicit-config, and no-config paths.

tests/ops/test_elementwise_config_dtype.py
RMLYC: Fixed in commit `e5fa027`. Added back strategy-sensitive `default_config` coverage in `tests/ops/test_elementwise_config_dtype.py` for unary (`explicit_parallel` vs `register_copy`), binary (`explicit_parallel` vs `register_copy`), and fused-gated (`direct` vs `explicit_parallel`) kernels, so the cleanup no longer drops `_strategy_npt()` behavior coverage.

tests/test_ops_manifest.py
```diff
@@ -9,7 +9,7 @@
 import pytest
 import yaml
 
-pytestmark = pytest.mark.smoke
+pytestmark = pytest.mark.full
```
Ibuki-wind: `ops_manifest.yaml` is a repo-level source of truth, and these manifest-structure/API checks are CPU-only, so moving the whole file out of `smoke` does not address the GPU smoke/runtime cost this cleanup is trying to reduce. Keep manifest contract tests in `smoke`; trim or split them only if they become a measured CPU bottleneck.

tests/test_validate_manifest.py
```diff
@@ -12,7 +12,7 @@
 
 import pytest
 
-pytestmark = pytest.mark.smoke
+pytestmark = pytest.mark.full
```
Ibuki-wind: The manifest validator defines the repo-level acceptance rules for `ops_manifest.yaml`, and this suite is CPU-only. Moving it wholesale to `full` lowers PR-level protection for the source-of-truth contract without buying meaningful GPU-smoke relief. Keep validator coverage in `smoke` for now; only move specific expensive/broad cases after measuring CPU cost.

Ibuki-wind (changes requested):
The previous blockers are resolved: `BenchmarkReport.record()` coverage is retained, and strategy npt behavior is covered in the reduced elementwise config test.

I would still block on manifest-related retiering. For this cleanup pass, use the conservative rule that manifest/source-of-truth tests stay PR-level because they are CPU-only and not the source of GPU smoke inflation.

Compile/caching/autotune and scaffolding cleanup can proceed, but manifest/schema/validator files should remain `smoke`.

RMLYC:
Follow-up on the manifest-tiering discussion:

I restored manifest contract validation to PR-level `smoke` in commit `1e63807`.

The updated split is:
- `tests/test_ops_manifest.py` stays in `smoke`
- `tests/test_validate_manifest.py` stays in `smoke`
- broader compile/caching and other non-entry-point checks remain outside the PR critical path

The rationale is that, in this repo, the manifest and its validator are part of the repository entry contract rather than optional non-runtime scaffolding. A PR that breaks manifest validity, workload loading, roofline evaluation, or validator behavior has already put the repo in an invalid state even if individual kernels still compile.

That means the cleanup rule for this PR is:
- keep repository entry-contract validation in `smoke`
- move breadth/matrix/implementation-detail coverage out of `smoke`
- only delete tests when the same behavior is still directly protected elsewhere

tests/test_ops_manifest.py
RMLYC: done

tests/test_validate_manifest.py
RMLYC: done

lcy-seso (approved):

Ibuki-wind (approved):

Issue #987, opened 2026-04-18
[META][CI] move nightly-label runner into a container for parity with venv runner


lcy-seso:
### Symptom / Motivation
While resolving #986 (self-hosted runner ran out of disk), it surfaced that the two self-hosted runner labels have inconsistent isolation models:

- `venv` label → the actions-runner process runs inside a container (container1/container2 reachable on `192.168.100.6:22002/22003`).
- `nightly` label → the actions-runner process runs **directly on the host**. The `nightly.yml` jobs then shell out to `docker run --gpus '"device=1"' -v /data/ci-cache/...` to do the actual work in the `tileops-runner:latest` image, but the scheduling process itself is bare-metal.

This asymmetry caused several concrete pains during the #986 incident:

1. **No isolation** — `venv` caches under `/home/ci-runner/.*` and `nightly` caches under `/data/ci-cache/*` share the same 3.5 TiB volume. When one label's caches filled the disk, the other label could no longer start a job either.
2. **Different cleanup paths** — the reclaim composite action introduced in #986 has to accept `cache-dirs` as a configurable input purely because the two runners use different directory layouts. If nightly also lived under `/home/ci-runner/.*` the action would have a single fixed default.
3. **Larger blast radius** — a bug in a nightly job (stuck process, leaked mount, disk-filling bug) can destabilize the host itself and take the `venv` containers down with it.
4. **Weaker reproducibility** — host-level package/driver changes silently affect nightly results; they do not affect the containerized `venv` runner.

### Root Cause Analysis
The nightly runner was provisioned directly on the host, probably because GPU passthrough + docker-in-docker was considered too complex at the time. `.github/runner/entrypoint.sh` only codifies the `venv` runner's container entrypoint; there is no equivalent Dockerfile / entrypoint for the nightly runner, and no compose/manifest describing how it should be launched.

The nightly workflow's per-step `docker run` calls (`nightly.yml` lines ~42-71, 95-151, 191-352, 377-455) already do the right thing — they containerize the *workload*. What is missing is containerizing the *runner process itself*, which would require:

- A Dockerfile / base image for the nightly runner (could likely extend the existing `tileops-runner:latest` or reuse `venv` runner's image).
- Access to the host Docker socket or docker-in-docker for the nested `docker run` calls the workflow issues.
- GPU access via `--gpus` passthrough to the runner container, then re-passthrough to nested containers.
[...]

PR #982, opened 2026-04-16
[Design][Manifest] Redesign init_dims as static_dims (docs + schema + validator)

lcy-seso:
## Summary

Complete the `static_dims` redesign (renamed from `init_dims`) for arbitrary-rank ops in a single internally consistent PR. Scope now covers:

1. **Docs** — authoritative spec rewrite: `docs/manifest.md` R20, `docs/ops-design.md` `_cache_key` section, `docs/ops-design-reference.md` Base Class Protocol + Consistency Enforcement table.
2. **Manifest yaml** — 16 ops migrated from `init_dims: {N: {from: "..."}}` to flat `static_dims: {N: "..."}`. LogSumExpFwdOp's multi-axis form removed (violates new R20 single-axis rule).
3. **Validator** — `scripts/validate_manifest.py` renamed `manifest_init_dims` → `manifest_static_dims`, reads `sig.get("static_dims")`, updated error messages. L1 enforces static_dims keys ⊆ `__init__` parameters.
4. **Status flips** — 13 `status: implemented` ops flipped to `status: spec-only` with per-op comments explaining the impl/spec mismatch (LayerNorm family × 5, BatchNorm/InstanceNorm × 3, Softmax/LogSoftmax/LogSumExp × 3, Argmax/Argmin × 2). This is the trust-model-prescribed response to impl divergence.
5. **PyTorch API alignment** — `dim` types include `tuple[int, ...]`; `isinstance(dim, list)` → `(list, tuple)` throughout manifest; LogSumExpFwdOp `shape_rules` dim-normalized.
6. **Design-first principle** — one-line statement added to `CLAUDE.md` (always loaded). Docs focus on technical spec.

Closes #984 (absorbed). Closes #976.

#985 remains open for the Op base `_cache_key` runtime + empty-static_dims warning.

## Test plan

- [x] `python scripts/validate_manifest.py` passes (24 pre-existing bench-file warnings, 0 errors)
- [x] 91 validator unit tests pass (4 new `static_dims` tests)
- [x] No `init_dims` remnants anywhere: `grep -rn init_dims docs/ tileops/ scripts/ tests/` returns empty
- [x] 16 `static_dims:` entries in manifest yaml (down from 31 `init_dims` originally — 14 dim=None-supporting reduction ops correctly had theirs removed per R20 applicability, LogSumExp removed per single-axis rule)
[...]

PR #988, opened 2026-04-18
[CI] reclaim self-hosted runner disk before heavy steps

### LINKED ISSUE #986: [META][CI] reclaim self-hosted runner disk to prevent out-of-space failures
### Symptom / Motivation
GPU Smoke run [24595105259](https://github.com/tile-ai/TileOPs/actions/runs/24595105259) for PR #982 failed at the *Ensure venv and install dependencies* step on self-hosted runner `tile-ops-mig-0` with:

```
mkdir: cannot create directory '/home/ci-runner/workdir/_temp/gpu-smoke-24595105259-1/venv': No space left on device
```

At the time of failure the volume holding `/home/ci-runner` was at 100% (3.3 TiB used of 3.5 TiB). Manual cleanup on container1 of the self-hosted runner recovered ~147 GiB (96% Use / 151 GiB Avail afterwards). The same failure mode will recur on any job that needs to `mkdir`, `rsync`, or install a fresh venv — including every future fork PR — because nothing in the workflows prunes the caches that grow over time.

### Root Cause Analysis
Four independent cache trees accumulate on the self-hosted runners and none of them are age-bounded:

1. `${runner.tool_cache}/tileops_ci_venv_<hash>/` — one full trusted venv per unique `pyproject.toml` hash (see `.github/workflows/gpu-smoke.yml` lines ~366-368). Each is GiB-scale; old hashes are never reaped.
2. `${RUNNER_TEMP}/gpu-smoke-<run_id>-<attempt>/` — fork-PR per-run dirs. The *Cleanup isolated fork state* step at the end of the job removes them, but only when the job reaches that step. An interrupted job, a disk-full `mkdir` failure, or a canceled run leaves the directory as an orphan.
3. `/home/ci-runner/.tilelang/cache`, `/home/ci-runner/.triton/cache`, `/home/ci-runner/.cache/pip`, `/home/ci-runner/.wheel-cache` — the trusted runtime caches used by the `venv`-label runner. Grown unbounded.
4. `/data/ci-cache/{tilelang,triton,pip,wheels}` — the equivalent cache tree mounted into the docker containers used by the `nightly`-label runner (`.github/workflows/nightly.yml`). Separate path, same failure mode waiting to trigger.

There is no pre-flight disk guard, so a near-full disk surfaces as a cryptic `mkdir: No space left on device` several steps into the job rather than as an explicit disk-pressure error.

### Related Files
- `.github/actions/reclaim-runner-disk/action.yml` (to be added)
- `.github/runner/entrypoint.sh`

## Goal
Make self-hosted CI runners self-maintaining so disk-full incidents do not recur. Concretely: every self-hosted job must run a bounded reclaim step before doing anything that allocates space, and must fail fast with a clear message if the volume is still near-full after reclaim.

## Plan

1. Add `.github/actions/reclaim-runner-disk/` composite action that (a) keeps the most-recent 2 `tileops_ci_venv_*` directories and removes the rest, (b) purges `${RUNNER_TEMP}/gpu-smoke-*` / `nightly-*` orphans older than 120 minutes (excluding the current run), (c) deletes files in the configured cache roots with `mtime > 30 days`, (d) logs `df -h` before and after, (e) fails the step with an explicit message if `RUNNER_TEMP` has less than 5 GiB free after reclaim.
2. Wire the composite action into the `gpu-smoke` job of `gpu-smoke.yml` (defaults target `/home/ci-runner/.*` paths).
3. Wire the composite action into all four `nightly.yml` jobs (`cache-warmup`, `benchmark`, `op_test`, `publish`) with `cache-dirs` pointing at `/data/ci-cache/{tilelang,triton,pip,wheels}` and `prune-tool-cache: false` since nightly does not use the trusted-venv scheme.
4. Make the cache-age threshold (30d), venv keep-count (2), and orphan-age threshold (120m) composite-action inputs so they can be tuned without editing workflow YAML.
5. Document the thresholds and the manual-cleanup runbook (what paths to purge, which thresholds are safe) in `docs/` or the runner's README so the next operator does not have to rediscover them.
6. Follow-up (separate issue recommended): consider moving `/data/ci-cache` onto a volume distinct from `/home/ci-runner` so one runner label's overflow cannot starve the other.

## Constraints
- Do not nuke `/home/ci-runner/.tilelang/cache` wholesale — cold TileLang recompiles are very expensive. Age-based trimming only.
- Reclaim must be the first real step after checkout (before any `mkdir`, `rsync`, or venv copy) so disk pressure is resolved or reported before it can break later steps.
- Composite action must tolerate missing directories and missing optional inputs; `set -uo pipefail` only (not `-e`) so a single `find` failure cannot abort the reclaim.
- No change to the trust model: reclaim runs on the runner with the same identity the rest of the job uses; it must not escalate privileges or touch paths outside the four cache roots it is given.
- Hard guard must use `::error::` annotations so the failure is visible in the GitHub Actions UI summary.

## Acceptance Criteria
- [ ] AC-1: Modified files pass unit tests (workflow YAML passes `actionlint`).
- [ ] AC-2: `.github/actions/reclaim-runner-disk/action.yml` exists and is referenced from the `gpu-smoke` job in `gpu-smoke.yml` and all four jobs in `nightly.yml` (5 call sites total).
- [ ] AC-3: The composite action supports configurable `cache-dirs`, `venv-keep`, `cache-age-days`, `temp-orphan-minutes`, `min-avail-gib`, `prune-tool-cache`, and `current-run-dir` inputs.
- [ ] AC-4: When `RUNNER_TEMP` has less than `min-avail-gib` GiB free after reclaim, the step fails with an `::error::` annotation that names the runner and the observed free space.
- [ ] AC-5: A re-run of PR #982's GPU Smoke job on `tile-ops-mig-0` succeeds past the `Ensure venv and install dependencies` step (empirical validation of the fix).
- [ ] AC-6: Runbook committed under `docs/` (or linked from the runner README) documenting: (a) the four cache trees and their owning runner labels, (b) the chosen thresholds (30-day mtime, keep-2 venvs, 120-minute orphan window, 5-GiB floor) and the reasoning, (c) the manual-cleanup commands used on 2026-04-18 to unblock the runner.

lcy-seso:
Fixes #986.

## Summary
- Add composite action `.github/actions/reclaim-runner-disk/` that prunes stale trusted venvs, orphan per-run temp dirs, and aged compile/wheel/pip caches on the self-hosted runner, with configurable thresholds and a hard disk-pressure guard.
- Wire into the `gpu-smoke` job in `gpu-smoke.yml`, and add a scheduled `runner-maintenance.yml` workflow that runs the same reclaim daily as a safety net on low-PR-traffic days.

## Why
GPU Smoke run [24595105259](https://github.com/tile-ai/TileOPs/actions/runs/24595105259) for #982 failed with `mkdir: ... : No space left on device` on `tile-ops-mig-0`. The runner's 3.5 TiB volume had filled from four unbounded accumulators:

1. `${runner.tool_cache}/tileops_ci_venv_<hash>/` — one full venv per unique `pyproject.toml` hash, never pruned.
2. `${RUNNER_TEMP}/gpu-smoke-<run_id>-<attempt>/` — orphans left by interrupted jobs (the end-of-job cleanup step never runs on failure paths that precede it).
3. `/home/ci-runner/.tilelang/cache`, `.triton/cache`, `.cache/pip`, `.wheel-cache` — age-unbounded growth.
4. `/data/ci-cache/*` on the `nightly`-label runner — same failure mode, separate cache tree (see #987).

## What the composite action does
- **Venv prune**: keeps the 2 most recent `tileops_ci_venv_*` directories (by mtime) under `runner.tool_cache`, removes the rest.
- **Orphan prune**: deletes `${RUNNER_TEMP}/gpu-smoke-*` and `nightly-*` directories older than 120 minutes, excluding the current run's dir.
- **Age-based cache trim**: deletes files older than 7 days under the configured cache roots, then removes empty directories of the same age. (Default tightened from the original 30d draft — the compile/wheel/pip caches refill quickly and 30d was too loose to relieve pressure.)
- **Before/after logging**: `df -h` on the relevant mounts for post-mortem visibility.
- **Hard guard**: if `RUNNER_TEMP` has less than 5 GiB free after reclaim, emits a `::error::` annotation naming the runner and available space, and exits 1 — so disk pressure is surfaced as itself rather than as a downstream cryptic `mkdir` failure.
- **Input validation**: numeric inputs (`venv-keep`, `temp-orphan-minutes`, `cache-age-days`, `min-avail-gib`) are validated against `^[0-9]+$` up-front; a YAML typo fails with a clear `::error::` rather than a cryptic bash error.

All thresholds are action inputs. `cache-dirs` and `prune-tool-cache` are also inputs so the same action can later be used by the nightly runner. A `guard-only` input exists for callers that want the disk-floor check without reclaim (currently unused — `gpu-smoke` runs the full reclaim so the first affected PR run can self-heal).

## Scope / deferred work
This PR only wires the action into `gpu-smoke.yml` and the new scheduled `runner-maintenance.yml`. Wiring the 4 nightly jobs is deferred until #987 lands — #987 will collapse `/data/ci-cache/*` into `/home/ci-runner/.*/cache`, and there is no point baking in paths that #987 plans to delete. See scope note in #986.

## Validation already done
- Manual reclaim on the affected runner (container1) recovered ~147 GiB using the same thresholds/paths the action encodes; volume went from 100% / 3.8 GiB Avail to 96% / 151 GiB Avail.
- YAML syntax validated locally (`python3 -c 'import yaml; yaml.safe_load(...)'`). `actionlint` / `shellcheck` not available in this environment; CI runs both.
- Reclaim step is side-effect-tolerant (`set -uo pipefail`, `|| true` on find/rm) so a single failure cannot abort the reclaim; only the final disk-floor guard can fail the step.

## Test plan
- [ ] CI `validate-pr-title` passes.
- [ ] CI `actionlint` passes on the new composite action and the modified workflow.
- [ ] On re-run of a gpu-smoke job, the `Reclaim runner disk` step logs `df -h` before and after and completes before `Resolve runtime state`.
- [ ] Re-run PR #982's gpu-smoke job to confirm it now passes the `Ensure venv and install dependencies` step (AC-5 of #986).
- [ ] Future disk-full event on a different runner produces a clear `::error:: Runner <name> has <NGiB free ...` annotation instead of a cryptic `mkdir` failure.

## Follow-up

Applied in this PR:
- `.github/actions/reclaim-runner-disk/action.yml` — removed unused `guard-only` input (dead code from an earlier split-reclaim iteration; see 2d0f32e).

### DIFF
--- a/.github/actions/reclaim-runner-disk/action.yml
+++ b/.github/actions/reclaim-runner-disk/action.yml
@@ -83,10 +83,24 @@ runs:
         TOOL_CACHE: ${{ runner.tool_cache }}
       run: |
         set -uo pipefail
-        # shellcheck disable=SC2206
-        DF_TARGETS=( ${DF_PATHS} )
-        if [[ ${#DF_TARGETS[@]} -eq 0 ]]; then
-          DF_TARGETS=( "${RUNNER_TEMP}" "${TOOL_CACHE}" /home/ci-runner )
+
+        # Validate numeric inputs up-front so typos fail with a clear error
+        # instead of a cryptic bash arithmetic/find error deeper in the script.
+        for var in VENV_KEEP TEMP_ORPHAN_MINUTES CACHE_AGE_DAYS MIN_AVAIL_GIB; do
+          val="${!var}"
+          if [[ ! "${val}" =~ ^[0-9]+$ ]]; then
+            echo "::error::reclaim-runner-disk: input ${var}='${val}' is not a non-negative integer"
+            exit 1
+          fi
+        done
+
+        if [[ -n "${DF_PATHS}" ]]; then
+          read -r -a DF_TARGETS <<< "${DF_PATHS}"
+        else
+          DF_TARGETS=()
+          for path in "${RUNNER_TEMP:-}" "${TOOL_CACHE:-}" /home/ci-runner; do
+            [[ -n "${path}" ]] && DF_TARGETS+=( "${path}" )
+          done
         fi
 
         echo "Disk before reclaim:"
@@ -98,7 +112,7 @@ runs:
 
         # 1) Prune old trusted venvs: keep N most recent by mtime.
         if [[ "${GUARD_ONLY}" != "true" && "${PRUNE_TOOL_CACHE}" == "true" && -d "${TOOL_CACHE}" ]]; then
-          mapfile -t venvs < <(find "${TOOL_CACHE}" -maxdepth 1 -type d -name "tileops_ci_venv_*" -printf '%T@ %p\n' 2>/dev/null | sort -rn | awk '{print $2}')
+          mapfile -t venvs < <(find "${TOOL_CACHE}" -maxdepth 1 -type d -name "tileops_ci_venv_*" -printf '%T@ %p\n' 2>/dev/null | sort -rn | cut -d' ' -f2-)
           if [[ ${#venvs[@]} -gt ${VENV_KEEP} ]]; then
             for stale in "${venvs[@]:${VENV_KEEP}}"; do
               echo "Removing stale venv: ${stale}"
@@ -141,6 +155,10 @@ runs:
         df -h "${DF_TARGETS[@]}" 2>/dev/null || true
 
         # Hard guard: abort early with a clear message if still near-full.
+        if [[ -z "${RUNNER_TEMP:-}" ]]; then
+          echo "::error::reclaim-runner-disk: RUNNER_TEMP is unset; cannot enforce disk floor"
+          exit 1
+        fi
         AVAIL_KB=$(df -Pk "${RUNNER_TEMP}" | awk 'NR==2 {print $4}')
         MIN_KB=$(( MIN_AVAIL_GIB * 1024 * 1024 ))
         if [[ -n "${AVAIL_KB:-}" && "${AVAIL_KB}" -lt "${MIN_KB}" ]]; then
--- a/.github/actions/reclaim-runner-disk/action.yml
+++ b/.github/actions/reclaim-runner-disk/action.yml
@@ -74,14 +74,6 @@ inputs:
       always run the full reclaim pass.
     required: false
     default: "false"
-  guard-only:
-    description: >
-      When "true", skip the venv/orphan/cache pruning and only run the
-      disk-floor guard. Intended for per-PR jobs that rely on a scheduled
-      maintenance workflow for the heavy reclaim; the guard still fails
-      fast with an explicit error if the scheduled workflow has lagged.
-    required: false
-    default: "false"
 
 runs:
   using: composite
@@ -100,7 +92,6 @@ runs:
         PRUNE_TOOL_CACHE: ${{ inputs.prune-tool-cache }}
         DF_PATHS: ${{ inputs.df-paths }}
         FORCE_RECLAIM: ${{ inputs.force-reclaim }}
-        GUARD_ONLY: ${{ inputs.guard-only }}
         TOOL_CACHE: ${{ runner.tool_cache }}
       run: |
         set -uo pipefail
@@ -139,9 +130,7 @@ runs:
         DID_RECLAIM="false"
         SHOULD_RECLAIM="false"
 
-        if [[ "${GUARD_ONLY}" == "true" ]]; then
-          echo "guard-only mode: skipping venv/orphan/cache pruning"
-        elif [[ "${FORCE_RECLAIM}" == "true" ]]; then
+        if [[ "${FORCE_RECLAIM}" == "true" ]]; then
           SHOULD_RECLAIM="true"
           echo "force-reclaim enabled: running full reclaim pass"
         elif [[ -n "${AVAIL_KB:-}" && "${AVAIL_KB}" -lt "${RECLAIM_BELOW_KB}" ]]; then
@@ -191,9 +180,6 @@ runs:
         fi
 
         # 3) Age-based trim of trusted compile/wheel/pip caches.
-        if [[ "${GUARD_ONLY}" == "true" ]]; then
-          CACHE_DIRS=""
-        fi
         SHOULD_TRIM_CACHE="false"
         if [[ "${SHOULD_RECLAIM}" == "true" && -n "${CACHE_DIRS}" ]]; then
           if [[ "${FORCE_RECLAIM}" == "true" || "${CACHE_TRIM_COOLDOWN_MINUTES}" == "0" ]]; then
--- a/.github/actions/reclaim-runner-disk/action.yml
+++ b/.github/actions/reclaim-runner-disk/action.yml
@@ -163,22 +163,30 @@ runs:
           fi
         fi
 
-        # 2) Purge orphan per-run dirs in RUNNER_TEMP older than N minutes,
-        #    excluding the current run's directory if provided.
+        # 2) Purge orphan per-run dirs in RUNNER_TEMP whose newest activity in
+        #    the subtree is older than N minutes, excluding the current run's
+        #    directory if provided. Directory mtime alone is unreliable — it
+        #    only updates when entries are added/removed, so an active
+        #    long-running job whose files are being written (but not the dir
+        #    itself) would be wrongly eligible. Use the newest file mtime
+        #    anywhere in the subtree instead.
         if [[ "${SHOULD_RECLAIM}" == "true" && -d "${RUNNER_TEMP}" ]]; then
           DID_RECLAIM="true"
+          now_epoch=$(date +%s)
+          stale_seconds=$(( TEMP_ORPHAN_MINUTES * 60 ))
           for pattern in 'gpu-smoke-*' 'nightly-*'; do
-            if [[ -n "${CURRENT_RUN_DIR}" ]]; then
-              find "${RUNNER_TEMP}" -maxdepth 1 -type d -name "${pattern}" \
-                ! -name "${CURRENT_RUN_DIR}" -mmin "+${TEMP_ORPHAN_MINUTES}" \
-                -exec echo "Removing orphan runtime: {}" \; \
-                -exec rm -rf {} + 2>/dev/null || true
-            else
-              find "${RUNNER_TEMP}" -maxdepth 1 -type d -name "${pattern}" \
-                -mmin "+${TEMP_ORPHAN_MINUTES}" \
-                -exec echo "Removing orphan runtime: {}" \; \
-                -exec rm -rf {} + 2>/dev/null || true
-            fi
+            while IFS= read -r candidate; do
+              [[ -n "${candidate}" ]] || continue
+              if [[ -n "${CURRENT_RUN_DIR}" && "$(basename "${candidate}")" == "${CURRENT_RUN_DIR}" ]]; then
+                continue
+              fi
+              newest_mtime=$(find "${candidate}" -printf '%T@\n' 2>/dev/null | sort -nr | head -n 1)
+              [[ -n "${newest_mtime}" ]] || newest_mtime=$(stat -c %Y "${candidate}" 2>/dev/null || echo 0)
+              if awk "BEGIN { exit !(${now_epoch} - ${newest_mtime} >= ${stale_seconds}) }"; then
+                echo "Removing orphan runtime: ${candidate}"
+                rm -rf "${candidate}" 2>/dev/null || true
+              fi
+            done < <(find "${RUNNER_TEMP}" -mindepth 1 -maxdepth 1 -type d -name "${pattern}" 2>/dev/null)
           done
         fi
 
@@ -212,7 +220,10 @@ runs:
             [[ -z "$dir" ]] && continue
             [[ -d "$dir" ]] || continue
             find "$dir" -type f -mtime "+${CACHE_AGE_DAYS}" -delete 2>/dev/null || true
-            find "$dir" -type d -empty -mtime "+${CACHE_AGE_DAYS}" -delete 2>/dev/null || true
+            # Remove directories that are now empty. Don't gate on -mtime here:
+            # deleting files above updates the parent directory's mtime to
+            # "now", so a mtime filter would leave newly-empty dirs behind.
+            find "$dir" -depth -mindepth 1 -type d -empty -delete 2>/dev/null || true
           done <<< "${CACHE_DIRS}"
           mkdir -p "$(dirname "${CACHE_TRIM_STAMP}")"
           touch "${CACHE_TRIM_STAMP}"
--- a/.github/actions/reclaim-runner-disk/action.yml
+++ b/.github/actions/reclaim-runner-disk/action.yml
@@ -20,7 +20,7 @@ inputs:
   cache-age-days:
     description: Age threshold (days) for trimming compile/wheel/pip cache files.
     required: false
-    default: "30"
+    default: "7"
   min-avail-gib:
     description: >
       Hard floor on free space (GiB) in RUNNER_TEMP after reclaim. Below this
--- a/.github/actions/reclaim-runner-disk/action.yml
+++ b/.github/actions/reclaim-runner-disk/action.yml
@@ -27,6 +27,18 @@ inputs:
       the step fails with an explicit message.
     required: false
     default: "5"
+  reclaim-below-gib:
+    description: >
+      Only attempt reclaim work when RUNNER_TEMP has less than this many GiB
+      free, unless force-reclaim is enabled.
+    required: false
+    default: "30"
+  cache-trim-cooldown-minutes:
+    description: >
+      Minimum interval between the expensive recursive cache-trim passes on the
+      same runner. Cheap disk checks still run on every invocation.
+    required: false
+    default: "0"
   current-run-dir:
     description: >
       Name of the per-run temp directory that must NOT be pruned. Leave blank
@@ -56,6 +68,12 @@ inputs:
     description: Space-separated paths to pass to `df -h` for before/after logging.
     required: false
     default: ""
+  force-reclaim:
+    description: >
+      When "true", bypass reclaim-below-gib and cache-trim cooldown checks and
+      always run the full reclaim pass.
+    required: false
+    default: "false"
   guard-only:
     description: >
       When "true", skip the venv/orphan/cache pruning and only run the
@@ -75,10 +93,13 @@ runs:
         TEMP_ORPHAN_MINUTES: ${{ inputs.temp-orphan-minutes }}
         CACHE_AGE_DAYS: ${{ inputs.cache-age-days }}
         MIN_AVAIL_GIB: ${{ inputs.min-avail-gib }}
+        RECLAIM_BELOW_GIB: ${{ inputs.reclaim-below-gib }}
+        CACHE_TRIM_COOLDOWN_MINUTES: ${{ inputs.cache-trim-cooldown-minutes }}
         CURRENT_RUN_DIR: ${{ inputs.current-run-dir }}
         CACHE_DIRS: ${{ inputs.cache-dirs }}
         PRUNE_TOOL_CACHE: ${{ inputs.prune-tool-cache }}
         DF_PATHS: ${{ inputs.df-paths }}
+        FORCE_RECLAIM: ${{ inputs.force-reclaim }}
         GUARD_ONLY: ${{ inputs.guard-only }}
         TOOL_CACHE: ${{ runner.tool_cache }}
       run: |
@@ -86,7 +107,7 @@ runs:
 
         # Validate numeric inputs up-front so typos fail with a clear error
         # instead of a cryptic bash arithmetic/find error deeper in the script.
-        for var in VENV_KEEP TEMP_ORPHAN_MINUTES CACHE_AGE_DAYS MIN_AVAIL_GIB; do
+        for var in VENV_KEEP TEMP_ORPHAN_MINUTES CACHE_AGE_DAYS MIN_AVAIL_GIB RECLAIM_BELOW_GIB CACHE_TRIM_COOLDOWN_MINUTES; do
           val="${!var}"
           if [[ ! "${val}" =~ ^[0-9]+$ ]]; then
             echo "::error::reclaim-runner-disk: input ${var}='${val}' is not a non-negative integer"
@@ -106,12 +127,33 @@ runs:
         echo "Disk before reclaim:"
         df -h "${DF_TARGETS[@]}" 2>/dev/null || true
 
+        if [[ -z "${RUNNER_TEMP:-}" ]]; then
+          echo "::error::reclaim-runner-disk: RUNNER_TEMP is unset; cannot enforce disk floor"
+          exit 1
+        fi
+
+        AVAIL_KB=$(df -Pk "${RUNNER_TEMP}" | awk 'NR==2 {print $4}')
+        MIN_KB=$(( MIN_AVAIL_GIB * 1024 * 1024 ))
+        RECLAIM_BELOW_KB=$(( RECLAIM_BELOW_GIB * 1024 * 1024 ))
+        CACHE_TRIM_STAMP="/home/ci-runner/.ci-maintenance/reclaim-runner-disk-cache-trim.stamp"
+        DID_RECLAIM="false"
+        SHOULD_RECLAIM="false"
+
         if [[ "${GUARD_ONLY}" == "true" ]]; then
           echo "guard-only mode: skipping venv/orphan/cache pruning"
+        elif [[ "${FORCE_RECLAIM}" == "true" ]]; then
+          SHOULD_RECLAIM="true"
+          echo "force-reclaim enabled: running full reclaim pass"
+        elif [[ -n "${AVAIL_KB:-}" && "${AVAIL_KB}" -lt "${RECLAIM_BELOW_KB}" ]]; then
+          SHOULD_RECLAIM="true"
+          echo "Free space below ${RECLAIM_BELOW_GIB}GiB: running reclaim pass"
+        else
+          echo "Free space is healthy; skipping reclaim pass"
         fi
 
         # 1) Prune old trusted venvs: keep N most recent by mtime.
-        if [[ "${GUARD_ONLY}" != "true" && "${PRUNE_TOOL_CACHE}" == "true" && -d "${TOOL_CACHE}" ]]; then
+        if [[ "${SHOULD_RECLAIM}" == "true" && "${PRUNE_TOOL_CACHE}" == "true" && -d "${TOOL_CACHE}" ]]; then
+          DID_RECLAIM="true"
           mapfile -t venvs < <(find "${TOOL_CACHE}" -maxdepth 1 -type d -name "tileops_ci_venv_*" -printf '%T@ %p\n' 2>/dev/null | sort -rn | cut -d' ' -f2-)
           if [[ ${#venvs[@]} -gt ${VENV_KEEP} ]]; then
             for stale in "${venvs[@]:${VENV_KEEP}}"; do
@@ -123,7 +165,8 @@ runs:
 
         # 2) Purge orphan per-run dirs in RUNNER_TEMP older than N minutes,
         #    excluding the current run's directory if provided.
-        if [[ "${GUARD_ONLY}" != "true" && -n "${RUNNER_TEMP:-}" && -d "${RUNNER_TEMP}" ]]; then
+        if [[ "${SHOULD_RECLAIM}" == "true" && -d "${RUNNER_TEMP}" ]]; then
+          DID_RECLAIM="true"
           for pattern in 'gpu-smoke-*' 'nightly-*'; do
             if [[ -n "${CURRENT_RUN_DIR}" ]]; then
               find "${RUNNER_TEMP}" -maxdepth 1 -type d -name "${pattern}" \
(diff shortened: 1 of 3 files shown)


.github/workflows/gpu-smoke.yml
```diff
@@ -350,6 +350,11 @@ jobs:
         with:
           python-version: "3.11"
 
+      - name: Guard against disk pressure
+        uses: ./.github/actions/reclaim-runner-disk
+        with:
+          guard-only: "true"
```
Ibuki-wind: `guard-only: "true"` makes this step detection-only, so if the runner crosses the disk floor after the daily maintenance window the next `gpu-smoke` run still fails without reclaiming anything. That leaves the original outage mode in place for the rest of the day, just with a clearer error. → Run `reclaim-runner-disk` in reclaim mode here, or trigger and wait for the maintenance workflow before continuing so the first affected PR run can self-heal instead of only failing.

.github/workflows/runner-maintenance.yml
```diff
@@ -0,0 +1,31 @@
+name: Runner Maintenance
```
Ibuki-wind, line 1: This workflow omits an explicit `permissions:` block, unlike the repo’s other workflows, so the scheduled job inherits broader default `GITHUB_TOKEN` scopes than it needs. → Add the minimal permissions here (`contents: read` is enough for `actions/checkout` + the local composite action) to match the existing workflow pattern.

Ibuki-wind (changes requested):
No trust-boundary split issue here; scope stays within CI/workflow changes.

Blocking on the control flow: actual cleanup now lives in a once-daily scheduled job, so `gpu-smoke` still cannot recover itself when disk pressure appears between maintenance windows.

The composite action abstraction, before/after `df`, and explicit floor guard are good additions. After making the reclaim path self-healing for the first affected run and tightening workflow permissions, this should be ready.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -0,0 +1,133 @@
...
+        PRUNE_TOOL_CACHE: ${{ inputs.prune-tool-cache }}
+        DF_PATHS: ${{ inputs.df-paths }}
+        TOOL_CACHE: ${{ runner.tool_cache }}
+      run: |
+        set -uo pipefail
+        # shellcheck disable=SC2206
+        DF_TARGETS=( ${DF_PATHS} )
+        if [[ ${#DF_TARGETS[@]} -eq 0 ]]; then
+          DF_TARGETS=( "${RUNNER_TEMP}" "${TOOL_CACHE}" /home/ci-runner )
+        fi
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Default `DF_TARGETS` now filters out empty `RUNNER_TEMP`/`TOOL_CACHE` before passing to `df`. See c7c0284.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -0,0 +1,133 @@
...
+
+        echo "Disk before reclaim:"
+        df -h "${DF_TARGETS[@]}" 2>/dev/null || true
+
+        # 1) Prune old trusted venvs: keep N most recent by mtime.
+        if [[ "${PRUNE_TOOL_CACHE}" == "true" && -d "${TOOL_CACHE}" ]]; then
+          mapfile -t venvs < <(find "${TOOL_CACHE}" -maxdepth 1 -type d -name "tileops_ci_venv_*" -printf '%T@ %p\n' 2>/dev/null | sort -rn | awk '{print $2}')
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Switched to `cut -d' ' -f2-` so venv paths with spaces survive. See c7c0284.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -0,0 +1,133 @@
...
+          find "$dir" -type f -mtime "+${CACHE_AGE_DAYS}" -delete 2>/dev/null || true
+          find "$dir" -type d -empty -mtime "+${CACHE_AGE_DAYS}" -delete 2>/dev/null || true
+        done <<< "${CACHE_DIRS}"
+
+        echo "Disk after reclaim:"
+        df -h "${DF_TARGETS[@]}" 2>/dev/null || true
+
+        # Hard guard: abort early with a clear message if still near-full.
+        AVAIL_KB=$(df -Pk "${RUNNER_TEMP}" | awk 'NR==2 {print $4}')
+        MIN_KB=$(( MIN_AVAIL_GIB * 1024 * 1024 ))
```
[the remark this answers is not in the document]
lcy-seso: Rejected. `min-avail-gib` is documented as an integer GiB threshold and there is no operational reason to accept fractional GiB here. Instead the action now validates all numeric inputs (`VENV_KEEP`, `TEMP_ORPHAN_MINUTES`, `CACHE_AGE_DAYS`, `MIN_AVAIL_GIB`) up-front against `^[0-9]+$` and fails with an explicit `::error::` on a YAML typo — which addresses the underlying concern (crash on malformed input) without silently widening the contract. See c7c0284.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -0,0 +1,133 @@
...
+        MIN_AVAIL_GIB: ${{ inputs.min-avail-gib }}
+        CURRENT_RUN_DIR: ${{ inputs.current-run-dir }}
+        CACHE_DIRS: ${{ inputs.cache-dirs }}
+        PRUNE_TOOL_CACHE: ${{ inputs.prune-tool-cache }}
+        DF_PATHS: ${{ inputs.df-paths }}
+        TOOL_CACHE: ${{ runner.tool_cache }}
+      run: |
+        set -uo pipefail
+        # shellcheck disable=SC2206
+        DF_TARGETS=( ${DF_PATHS} )
```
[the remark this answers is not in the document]
lcy-seso: Adopted. `DF_PATHS` is now parsed with `read -r -a` — no glob expansion. See c7c0284.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -0,0 +1,133 @@
...
+
+        echo "Disk after reclaim:"
+        df -h "${DF_TARGETS[@]}" 2>/dev/null || true
+
+        # Hard guard: abort early with a clear message if still near-full.
+        AVAIL_KB=$(df -Pk "${RUNNER_TEMP}" | awk 'NR==2 {print $4}')
+        MIN_KB=$(( MIN_AVAIL_GIB * 1024 * 1024 ))
+        if [[ -n "${AVAIL_KB:-}" && "${AVAIL_KB}" -lt "${MIN_KB}" ]]; then
+          echo "::error::Runner ${RUNNER_NAME:-unknown} has <${MIN_AVAIL_GIB}GiB free on ${RUNNER_TEMP} after reclaim (${AVAIL_KB} KiB). Manual cleanup required."
+          exit 1
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Guard now checks `${RUNNER_TEMP:-}` explicitly and emits `::error::reclaim-runner-disk: RUNNER_TEMP is unset` before calling `df`. See c7c0284.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -0,0 +1,149 @@
...
+    description: >
+      Age threshold (minutes) for purging orphan gpu-smoke-*/nightly-* temp
+      dirs in RUNNER_TEMP that do not belong to the current run.
+    required: false
+    default: "120"
+  cache-age-days:
+    description: Age threshold (days) for trimming compile/wheel/pip cache files.
+    required: false
+    default: "7"
+  min-avail-gib:
```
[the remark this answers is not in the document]
lcy-seso: Rejected (code). The 7-day default is intentional — it was tightened from 30d in commit 34378fb after the initial draft, because the compile/wheel/pip caches refill quickly and 30d was too loose to relieve pressure. The PR description is stale on this point; I'll update the description rather than the default.

.github/workflows/gpu-smoke.yml
```diff
@@ -350,6 +350,11 @@ jobs:
...
+
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Dropped `guard-only` on the `gpu-smoke` invocation so the PR path runs the full reclaim and the first affected run can self-heal. The scheduled `runner-maintenance` workflow remains as a traffic-light-day safety net. See c7c0284.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -0,0 +1,149 @@
...
+          find "$dir" -type d -empty -mtime "+${CACHE_AGE_DAYS}" -delete 2>/dev/null || true
+        done <<< "${CACHE_DIRS}"
+
+        echo "Disk after reclaim:"
+        df -h "${DF_TARGETS[@]}" 2>/dev/null || true
+
+        # Hard guard: abort early with a clear message if still near-full.
+        AVAIL_KB=$(df -Pk "${RUNNER_TEMP}" | awk 'NR==2 {print $4}')
+        MIN_KB=$(( MIN_AVAIL_GIB * 1024 * 1024 ))
+        if [[ -n "${AVAIL_KB:-}" && "${AVAIL_KB}" -lt "${MIN_KB}" ]]; then
```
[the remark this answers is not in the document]
lcy-seso: Adopted. All four numeric inputs are validated against `^[0-9]+$` at the top of the step and fail with an explicit `::error::` on a malformed value. See c7c0284.

.github/workflows/runner-maintenance.yml
```diff
@@ -0,0 +1,31 @@
...
+  reclaim-disk:
+    if: ${{ github.repository == 'tile-ai/TileOPs' }}
+    runs-on: [self-hosted, tile-ops, venv]
+    timeout-minutes: 20
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          fetch-depth: 1
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Added `permissions: contents: read` to the workflow. See c7c0284.

.github/workflows/gpu-smoke.yml
lcy-seso: Adopted. Dropped `guard-only` on the `gpu-smoke` invocation so the first PR run after a maintenance-window gap reclaims disk itself rather than only failing. The scheduled workflow remains for low-PR-traffic days. See c7c0284.

.github/workflows/runner-maintenance.yml
lcy-seso: Adopted. Added `permissions: contents: read` — matches the minimal scope needed for `actions/checkout` + local composite action. See c7c0284.

Ibuki-wind (commented):
Follow-up on the reclaim design: the self-heal path is right, but running the full reclaim pass on every `gpu-smoke` job is too expensive for a high-traffic PR flow.

Please gate reclaim on disk pressure first (for example, only run the full reclaim pass when `RUNNER_TEMP` is below about `30 GiB` free), add a `2 hour` cooldown for the expensive recursive cache-trim pass, and keep `runner-maintenance` as the forced full-cleanup path.

That keeps first-run recovery under real disk pressure without paying the full recursive scan cost on every PR.

.github/workflows/gpu-smoke.yml
```diff
@@ -350,6 +350,9 @@ jobs:
         with:
           python-version: "3.11"
 
+      - name: Reclaim runner disk
+        uses: ./.github/actions/reclaim-runner-disk
```
Ibuki-wind: This invocation currently runs the full reclaim pass on every `gpu-smoke` job, which is expensive under high PR volume. → Add a cheap pre-check on `RUNNER_TEMP` free space here and only enter the full reclaim path when the runner is already under disk pressure (for example, below about `30 GiB` free).

Ibuki-wind (commented):
Inline follow-up on the reclaim strategy: the current direction is sound, but the per-PR path should be cheaper.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -0,0 +1,167 @@
...
+
+        # 3) Age-based trim of trusted compile/wheel/pip caches.
+        if [[ "${GUARD_ONLY}" == "true" ]]; then
+          CACHE_DIRS=""
+        fi
+        while IFS= read -r dir; do
+          dir="${dir//$'\r'/}"
+          [[ -z "$dir" ]] && continue
+          [[ -d "$dir" ]] || continue
+          find "$dir" -type f -mtime "+${CACHE_AGE_DAYS}" -delete 2>/dev/null || true
```
Ibuki-wind: The recursive `find ... -delete` over the cache roots is the highest-cost part of this action, and rerunning it on every PR will rescan the same large trees repeatedly. → Put the cache-trim pass behind a runner-local cooldown (recommended: `2 hours`) so only low-space runs pay this recursive scan cost once per window.

.github/workflows/runner-maintenance.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Reclaim runner disk
+        uses: ./.github/actions/reclaim-runner-disk
```
Ibuki-wind: If the per-PR path becomes pressure-triggered and cooldown-limited, this scheduled workflow should remain the unconditional full-cleanup path. → Pass an explicit force/full-reclaim flag here so the daily maintenance run always executes the heavy cleanup regardless of current free space or cooldown state.

.github/workflows/gpu-smoke.yml
lcy-seso: Adopted. `gpu-smoke` now passes `reclaim-below-gib: "30"`; the composite action runs a cheap `df -Pk` first and only enters the full reclaim pass when `RUNNER_TEMP` is below 30 GiB free. Healthy-disk runs now only pay the two `df` calls and the floor check. See c86c8d0.

.github/actions/reclaim-runner-disk/action.yml
lcy-seso: Adopted. Added `cache-trim-cooldown-minutes` with a `/home/ci-runner/.ci-maintenance/reclaim-runner-disk-cache-trim.stamp` gating file; the recursive `find -delete` only runs when the stamp is missing or older than the cooldown. `gpu-smoke` sets the cooldown to 120 minutes, so under sustained disk pressure only one PR per 2h window pays the recursive scan. `force-reclaim` bypasses the cooldown for the daily maintenance job. See c86c8d0.

.github/workflows/runner-maintenance.yml
lcy-seso: Adopted. `runner-maintenance.yml` now passes `force-reclaim: "true"`; the composite action short-circuits both the `reclaim-below-gib` check and the cache-trim cooldown when this is set, so the daily scheduled run always executes the full cleanup regardless of current free space. See c86c8d0.

.github/workflows/gpu-smoke.yml
```diff
@@ -350,6 +350,12 @@ jobs:
...
+        with:
+          reclaim-below-gib: "30"
```
Ibuki-wind: `30 GiB` may be somewhat conservative, but please do not collapse this all the way to `5 GiB`: that is already the hard post-reclaim floor, so using the same value here would delay reclaim until the runner is effectively at the failure boundary. → If you want to reduce the proactive threshold, prefer a middle value such as `10-15 GiB` rather than `5 GiB`.

Ibuki-wind (commented):
Follow-up on the tuned reclaim strategy: the overall design now looks good. I only have two small scope-setting notes below.

.github/workflows/runner-maintenance.yml
```diff
@@ -0,0 +1,36 @@
+name: Runner Maintenance
+
+# Heavy disk reclamation for the self-hosted runner. Runs once per day at
+# 18:00 UTC (02:00 Asia/Shanghai) so it lands in an off-peak window. Per-PR
+# `gpu-smoke` jobs only perform full reclaim when disk is already under
+# pressure; this scheduled job forces the full cleanup pass on days with
+# little PR traffic.
```
Ibuki-wind, line 8: The current comment explains the scheduled cleanup role, but it does not leave a breadcrumb for the still-deferred nightly cache cleanup work. → Add a `FIXME:`-style note here that the nightly runner still needs its own cache-cleanup follow-up, and that this PR intentionally does not implement that path yet.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -0,0 +1,235 @@
...
+              find "${RUNNER_TEMP}" -maxdepth 1 -type d -name "${pattern}" \
+                ! -name "${CURRENT_RUN_DIR}" -mmin "+${TEMP_ORPHAN_MINUTES}" \
+                -exec echo "Removing orphan runtime: {}" \; \
+                -exec rm -rf {} + 2>/dev/null || true
+            else
+              find "${RUNNER_TEMP}" -maxdepth 1 -type d -name "${pattern}" \
+                -mmin "+${TEMP_ORPHAN_MINUTES}" \
+                -exec echo "Removing orphan runtime: {}" \; \
+                -exec rm -rf {} + 2>/dev/null || true
+            fi
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Orphan-temp cleanup now determines staleness from the newest file mtime in the subtree (falling back to `stat -c %Y` on the candidate dir if the subtree is empty), so an actively-written long-running job no longer goes stale just because its top-level dir's mtime hasn't been bumped. See 68372c1.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -0,0 +1,235 @@
...
+        fi
+
+        if [[ "${SHOULD_TRIM_CACHE}" == "true" ]]; then
+          DID_RECLAIM="true"
+          while IFS= read -r dir; do
+            dir="${dir//$'\r'/}"
+            [[ -z "$dir" ]] && continue
+            [[ -d "$dir" ]] || continue
+            find "$dir" -type f -mtime "+${CACHE_AGE_DAYS}" -delete 2>/dev/null || true
+            find "$dir" -type d -empty -mtime "+${CACHE_AGE_DAYS}" -delete 2>/dev/null || true
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Dropped the `-mtime` filter on the empty-directory pass and added `-depth -mindepth 1` so newly-empty dirs created by the preceding file delete actually get removed. See 68372c1.

.github/workflows/gpu-smoke.yml
```diff
@@ -350,6 +350,12 @@ jobs:
         with:
           python-version: "3.11"
 
+      - name: Reclaim runner disk
+        uses: ./.github/actions/reclaim-runner-disk
```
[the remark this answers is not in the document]
lcy-seso: Valid point. Out of scope for this PR — the same fork-code-on-self-hosted-runner exposure already applies to every `./.github/actions/` and `./.github/scripts/` invocation in `gpu-smoke.yml` (checkout uses `ref: pull/N/head` from the base repo, which is fork-controlled content). Pinning just this one step to `@main` would fix one leaf without addressing the broader trust model, and would also require the action to exist on main first. Will address in a follow-up that reworks `gpu-smoke`'s fork-PR handling holistically (separate base-repo checkout path for trusted action/script code).

.github/workflows/gpu-smoke.yml
lcy-seso: Acknowledged — no change needed. `reclaim-below-gib` is set to "30" at this call site (see gpu-smoke.yml:356 on 68372c1), which matches the value suggested in the earlier follow-up. Your 10-15 GiB guidance is noted for any future tuning; I'll leave 30 as-is for now so that the first PR affected by disk pressure reclaims with comfortable headroom above the 5 GiB hard floor.

.github/workflows/runner-maintenance.yml
lcy-seso: Adopted. Added a `FIXME:` block in the header comment of `runner-maintenance.yml` noting that the `nightly`-label runner's `/data/ci-cache/*` tree is intentionally deferred until #987 collapses that layout into `/home/ci-runner/.*/cache`. See 68372c1.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -0,0 +1,246 @@
...
+          DID_RECLAIM="true"
+          now_epoch=$(date +%s)
+          stale_seconds=$(( TEMP_ORPHAN_MINUTES * 60 ))
+          for pattern in 'gpu-smoke-*' 'nightly-*'; do
+            while IFS= read -r candidate; do
+              [[ -n "${candidate}" ]] || continue
+              if [[ -n "${CURRENT_RUN_DIR}" && "$(basename "${candidate}")" == "${CURRENT_RUN_DIR}" ]]; then
+                continue
+              fi
+              newest_mtime=$(find "${candidate}" -printf '%T@\n' 2>/dev/null | sort -nr | head -n 1)
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Replaced `sort -nr | head -n 1` with a single-pass awk max; O(n) instead of O(n log n). See 7860cd8.

.github/workflows/runner-maintenance.yml
```diff
@@ -0,0 +1,42 @@
...
+concurrency:
+  group: runner-maintenance
+  cancel-in-progress: false
+
+jobs:
+  reclaim-disk:
+    if: ${{ github.repository == 'tile-ai/TileOPs' }}
+    runs-on: [self-hosted, tile-ops, venv]
+    timeout-minutes: 20
+    steps:
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Raised `timeout-minutes` from 20 to 60 so the scheduled run has headroom under heavy disk pressure and does not abort mid-`find -delete`. See 7860cd8.

.github/workflows/gpu-smoke.yml
```diff
@@ -350,6 +350,12 @@ jobs:
...
+        with:
+          reclaim-below-gib: "10"
+          cache-trim-cooldown-minutes: "120"
+
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Moved `Reclaim runner disk` immediately after checkout, before `Set up Python`, so setup-python's toolcache downloads run with headroom instead of potentially racing the reclaim step. See 7860cd8.

.github/workflows/gpu-smoke.yml
```diff
@@ -345,6 +345,12 @@ jobs:
           persist-credentials: false
           fetch-depth: 1
 
+      - name: Reclaim runner disk
+        uses: ./.github/actions/reclaim-runner-disk
```
Ibuki-wind: On fork PRs, the preceding checkout pulls `github.event.pull_request.head.repo.full_name` into the workspace, so `uses: ./.github/actions/reclaim-runner-disk` executes the fork's version of this local action on the self-hosted runner. That is a trust-boundary violation: an untrusted PR can modify `.github/actions/reclaim-runner-disk/action.yml` and gain arbitrary shell execution against shared runner state before `Resolve runtime state` establishes the fork-isolation path. -> Do not source this step from the fork workspace. Keep the reclaim logic inline in the trusted workflow, or invoke it from a separately checked-out trusted base-repo path that fork PR contents cannot modify.

Ibuki-wind (changes requested):
Blocking on a trust-boundary regression in `gpu-smoke`: the new reclaim step is a local action sourced from the checked-out workspace, and on fork PRs that workspace is the fork head. That makes this repo-wide CI change materially more privileged than the previous inline trusted workflow logic.

Please fix the source-of-truth for the reclaim code before merging: either keep the reclaim logic inline in the trusted workflow, or load it from a trusted base-repo checkout/path that fork PR contents cannot modify. The current shape should not run on self-hosted runners for fork PRs.

.github/workflows/gpu-smoke.yml
lcy-seso: Adopted. Added a sparse base-repo checkout at `github.event.pull_request.base.sha` (trusted — fork commits cannot alter it) into `.trusted/`, and changed the invocation to `uses: ./.trusted/.github/actions/reclaim-runner-disk`. The trusted checkout runs AFTER the workspace checkouts so their `git clean -ffdx` does not wipe it.

Bootstrap caveat: on this PR itself, `base.sha` is the pre-merge main commit and does not yet contain the composite action, so `gpu-smoke` will fail at the reclaim step on this run. From the first PR after merge onwards the trusted path resolves correctly. See 123d403.

.github/workflows/gpu-smoke.yml
```diff
@@ -345,6 +345,29 @@ jobs:
...
+      # their `git clean -ffdx` pass does not wipe `.trusted/`. Anything
+      # loaded via `uses: ./...` from here on MUST resolve through
+      # `.trusted/` so that fork-controlled PR code cannot substitute the
+      # composite action. The PR's base sha (or `github.ref` for push) is
+      # the authoritative trusted source — fork commits cannot alter it.
+      - name: Checkout trusted actions (base repo)
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ github.repository }}
+          ref: ${{ github.event.pull_request.base.sha || github.ref }}
```
Ibuki-wind: Using `github.event.pull_request.base.sha` here for every PR fixes the fork trust issue, but it also means same-repo PRs load the base branch's `.github/actions/reclaim-runner-disk` instead of the PR's version. For the PR that *introduces or modifies* this action, `gpu-smoke` cannot validate the new behavior at all (the current head already shows a failing `gpu-smoke` run for exactly this reason). -> Split the trusted source by trust level: keep `base.sha` for fork PRs, but use the trusted branch under test (`github.event.pull_request.head.sha` / the normal workspace) for same-repo PRs and pushes so repo-owned CI changes can self-validate before merge.

Ibuki-wind (changes requested):
The trust-boundary fix is directionally correct, but there is still one blocking issue for a repo-wide CI change: the trusted-action checkout now unconditionally pins to `pull_request.base.sha`, so same-repo PRs load the *old* base-branch version of `reclaim-runner-disk` instead of the code under review. That leaves this PR unable to validate its own workflow/action changes and is already showing up as a red `gpu-smoke` check on the current head.

Please keep the trusted-source split for fork PRs, but let trusted same-repo PRs/pushes resolve the action from the trusted branch under test so the workflow can validate the new action before merge.

.github/workflows/runner-maintenance.yml
```diff
@@ -0,0 +1,42 @@
...
+    timeout-minutes: 60
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          fetch-depth: 1
+
+      - name: Reclaim runner disk
+        uses: ./.github/actions/reclaim-runner-disk
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Dropped the `actions/checkout` step and changed the invocation to `uses: tile-ai/TileOPs/.github/actions/reclaim-runner-disk@main`. Since this workflow only triggers on `schedule` and `workflow_dispatch`, no fork-PR code path reaches it and `@main` is a trusted ref. (Bootstrap caveat: the first scheduled run lands only after this PR merges — by then the action exists on main.) See 36f941e.

.github/workflows/gpu-smoke.yml
lcy-seso: Adopted. Split the trusted-actions ref by trust level using `needs.security-policy.outputs.is_fork`:

- fork PR → `github.event.pull_request.base.sha` (authoritative base, fork can't alter)
- same-repo PR / push → `head.sha` / `github.sha` (only collaborators can push, so the PR's own revision is trusted and self-validates)

Fork isolation preserved; repo-owned CI changes can now validate themselves on the PR that introduces them. See 4ea4bb7.

.github/workflows/gpu-smoke.yml
```diff
@@ -345,6 +345,34 @@ jobs:
...
+      # Ref selection by trust level:
+      #   - fork PR  → `base.sha` (authoritative base commit; fork can't alter)
+      #   - same-repo PR / push → head sha (only collaborators can push, so
+      #     the PR's own version is trusted — and this lets repo-owned CI
+      #     changes self-validate before merge)
+      - name: Checkout trusted actions (base repo)
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ github.repository }}
+          ref: ${{ needs.security-policy.outputs.is_fork == 'true' && github.event.pull_request.base.sha || github.event.pull_request.head.sha || github.sha }}
```
Ibuki-wind: This still keys entirely off `is_fork`, which is just `head.repo.full_name != github.repository`. That collapses two different trust cases: an untrusted external fork and a repo member opening a PR from their personal fork. On this PR, `author_association` is `MEMBER`, but the workflow still resolves `.trusted` from `github.event.pull_request.base.sha`, and the latest `gpu-smoke` run on `4ea4bb7` fails immediately because `main` does not contain `.github/actions/reclaim-runner-disk` yet. -> Split the trusted-action ref on *author trust*, not only repo equality: keep `base.sha` for external/untrusted forks, but allow member/collaborator-owned forks to use `head.sha` so repo-owned CI/action changes can validate on the PR that introduces them.`

Ibuki-wind (changes requested):
Blocking on trust classification: the external-fork isolation direction is correct now, but the current split is still too coarse for a repo-wide CI change. `is_fork` treats collaborator-owned forks the same as untrusted external forks, so repo members working from personal forks still load `.trusted` from `base.sha` and cannot validate changes to the trusted local action path on the PR that introduces them.

The current head already demonstrates the regression: `gpu-smoke` on `4ea4bb7` fails before setup because `.trusted/.github/actions/reclaim-runner-disk` does not exist on `main`. Please refine the trust decision to distinguish external forks from member/collaborator-owned forks; after that, the reclaim design itself looks sound.

.github/workflows/runner-maintenance.yml
```diff
@@ -0,0 +1,42 @@
+name: Runner Maintenance
+
+# Heavy disk reclamation for the self-hosted runner. Runs once per day at
+# 18:00 UTC (02:00 Asia/Shanghai) so it lands in an off-peak window. Per-PR
+# `gpu-smoke` jobs only perform full reclaim when disk is already under
+# pressure; this scheduled job forces the full cleanup pass on days with
+# little PR traffic.
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Reworded the comment to note that per-PR runs are threshold- and cooldown-gated, while the scheduled workflow forces the full pass. See 775bb8f.

.github/workflows/gpu-smoke.yml
lcy-seso: Adopted. Trust-split now combines `is_fork` with `github.event.pull_request.author_association`:

- external fork PR (author not in `[OWNER, MEMBER, COLLABORATOR]`) → `base.sha` (authoritative; external author can't alter it)
- same-repo PR, member-fork PR, or push → head sha (author already has write access to base, so the PR's own revision is trusted and CI changes self-validate)

This PR (from a MEMBER's personal fork) should now resolve `.trusted` from its own head and validate in place. See 775bb8f.

Ibuki-wind (dismissed):
No further blocking findings on the current head. The trust split now preserves external-fork isolation while allowing repo-member fork PRs to self-validate the trusted local action path, and the latest `gpu-smoke` run on `775bb8f` completed successfully with the reclaim step in place.

Ibuki-wind (approved):
No new blocking findings on the current head. This follow-up only removes the unused `guard-only` path from `reclaim-runner-disk`; the active pressure-gated + force-reclaim design remains unchanged, and the latest `gpu-smoke` run on `2d0f32e` has already cleared the trusted-actions checkout and reclaim steps without regression.

.github/workflows/gpu-smoke.yml
```diff
@@ -345,6 +345,40 @@ jobs:
...
+      # external contributors are kept on the trusted base ref.
+      #
+      #   - external fork PR (author not OWNER/MEMBER/COLLABORATOR)
+      #       → `base.sha` (authoritative; external author can't alter)
+      #   - same-repo PR, member-fork PR, or push
+      #       → head sha (author already has write access to base repo)
+      - name: Checkout trusted actions (base repo)
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ github.repository }}
```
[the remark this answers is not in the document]
lcy-seso: Rejected. The premise — that `head.sha` from a member/collaborator fork PR is not reachable in the base repo — does not match observed behavior. GitHub exposes every fork PR head commit in the base repo via `refs/pull/{number}/head`, and `actions/checkout@v4` on a `pull_request` event automatically fetches that ref, so `repository: ${{ github.repository }}` with `ref: fork_head_sha` resolves correctly.

Evidence: this PR (#988) is itself a MEMBER-authored fork PR, and the latest `gpu-smoke` run on `2d0f32e` cleared both `Checkout trusted actions (base repo)` and `Reclaim runner disk` without issue (confirmed in Ibuki-wind's APPROVED review on the same commit).

Switching `repository` to `head.repo.full_name` for member fork PRs would be content-equivalent (fork head is fork-authored content either way) but would add branching complexity and make the "which remote do we clone from" rule less uniform, for no observed behavior difference. Keeping the current single `repository: ${{ github.repository }}` pinning.

Issue #1000, opened 2026-04-19
[META] Follow-up to #975: update published TileLang dependency/docs after TileLang 5f70374c validation


superAngGao:
Follow-up to #975, #980, and #999.

After validating the TileLang `5f70374c` environment with commit `5f70374c48bb1d52e9260e995bf7adbcd64341e5`, we should update the repository's published dependency/documentation story accordingly.

This should be tracked separately from the temporary-test-skip PR.

Why split this out:
- the temporary skip PR should stay narrowly focused on restoring a usable test signal under the validated env
- changing `pyproject.toml` and `README.md` affects packaging and installation expectations, so it deserves independent review

Proposed follow-up work:
- update `pyproject.toml` from `tilelang==0.1.8` to the validated TileLang dependency target
- update `README.md` so the documented TileLang version matches the validated env
- verify whether we want to keep using a git-pinned TileLang dependency directly, or whether this should wait for an official TileLang release beyond `0.1.8`
- re-check the `apache-tvm-ffi` pin once the TileLang dependency decision is finalized

Validated environment reference:
- TileLang commit: `5f70374c48bb1d52e9260e995bf7adbcd64341e5`
- full-suite result with temporary skip set: `2180 passed, 89 skipped, 16 warnings`

Related issues:
- #975 dependency bump motivation
- #980 error-cascade isolation record
- #999 TileLang `5f70374c` env bring-up / temporary skip-set tracking

superAngGao:
Closing as superseded by the TileLang 0.1.9 migration tracking issue #1039. The 5f70374c line was an intermediate validation target; remaining skips and restoration work are now summarized under #1039.

Issue #999, opened 2026-04-19
[META] Follow-up to #975: TileLang 5f70374c enables WS/GQA, with a temporary full-suite skip set


superAngGao:
Follow-up to #975.

We validated that TileLang commit `5f70374c48bb1d52e9260e995bf7adbcd64341e5` is usable for the current TileOPs WS/GQA bring-up on the validated local TileLang `5f70374c` environment.

In particular:
- the original local validation env carried `tilelang==0.1.8+cuda.gitded6a992`, which does **not** expose `wgmma_gemm` and cannot run the current anchor/WS path
- after syncing the env to `tilelang @ git+https://github.com/tile-ai/tilelang.git@5f70374c48bb1d52e9260e995bf7adbcd64341e5`, the relevant TileOPs WS/GQA kernels compile and run again
- our local `experiments/ws_kernel_evolution` anchor-causal variants are runnable under this version, so this commit is a viable dependency target for the env-upgrade work tracked in #975

This issue is therefore about the **full-suite status of that upgraded environment** and the **temporary skip set** we currently need in order to move the dev env forward.

Detailed notes on the earlier mass `ERROR` / CUDA-device-poisoning behavior, and how those cascades were isolated and eliminated, are kept in #980. This issue is the higher-level env-upgrade tracking issue.

## Environment used

Validation workspace:
- repo: `TileOPs` upstream validation worktree
- editable TileOPs commit: `c39d65d31ea3cbd12261d55827ee81fc387dd174`
- python executable: the validated local TileLang `5f70374c` venv
- python: `3.12.12`
- torch: `2.10.0+cu128`
- CUDA runtime via torch: `12.8`
- tilelang: `0.1.8+cuda.git5f70374c`

Validation command:
```bash
CUDA_VISIBLE_DEVICES=0 \
TILELANG_CLEANUP_TEMP_FILES=1 \
TMPDIR=/home/ga/tltmp \
python -m pytest -q tests
```

## What works with `5f70374c`

After syncing TileLang to `5f70374c`:
- TileOPs WS/GQA kernels compile again in this env
- the missing-API problem from `gitded6a992` is gone (`wgmma_gemm` is present again)
- the local WS/GQA anchor-causal experiment line is runnable again under the working env

That is the main reason we believe `5f70374c` is the right dependency baseline to follow for #975.

## Full-suite status

Raw full-suite status after syncing TileLang, before temporary skip isolation:
[...]

Issue #980, opened 2026-04-16
[META] Follow-up to #975: isolate CUDA/PyTorch runtime compatibility before broad GPU validation


superAngGao:
Follow-up to #975.

This issue is separate from the TileLang API migration tracked in #979.

## Problem

Broad GPU validation is currently blocked by a CUDA/PyTorch runtime compatibility problem in the local upgraded environment.

This does **not** look like a TileLang API break by itself.

During the per-test-file sweep, the message

`The NVIDIA driver on your system is too old (found version 12090)`

appeared in `75` test-file logs.

Of those:

- `66` test files failed with this runtime error on the path to `torch.cuda.get_device_capability()` / CUDA init
- `9` additional test files still passed, but their logs also contained the same warning string

## Affected failing test files

- `tests/ops/attention/test_mean_pooling.py`
- `tests/ops/attention/test_mha.py`
- `tests/ops/attention/test_mha_decode.py`
- `tests/ops/attention/test_mha_decode_paged.py`
- `tests/ops/test_activation.py`
- `tests/ops/test_ada_layer_norm.py`
- `tests/ops/test_ada_layer_norm_zero.py`
- `tests/ops/test_argreduce.py`
- `tests/ops/test_batch_norm.py`
- `tests/ops/test_binary_arith.py`
- `tests/ops/test_bitwise.py`
- `tests/ops/test_comparison.py`
- `tests/ops/test_convolution.py`
- `tests/ops/test_cumulative.py`
- `tests/ops/test_deltanet_chunkwise_bwd.py`
- `tests/ops/test_deltanet_fwd.py`
- `tests/ops/test_deltanet_recurrence.py`
- `tests/ops/test_dropout.py`
- `tests/ops/test_elementwise_caching_autotune.py`
- `tests/ops/test_elementwise_compile.py`
- `tests/ops/test_elementwise_fp8.py`
- `tests/ops/test_elementwise_independent_fp8.py`
- `tests/ops/test_engram.py`
- `tests/ops/test_fft.py`
- `tests/ops/test_fp8_lighting_indexer.py`
- `tests/ops/test_fp8_quant.py`
- `tests/ops/test_fused_add_layer_norm.py`
- `tests/ops/test_fused_add_rms_norm.py`
- `tests/ops/test_fused_gated.py`
- `tests/ops/test_gated_deltanet_chunkwise_bwd.py`
- `tests/ops/test_gated_deltanet_fwd.py`
[...]

PR #1001, opened 2026-04-19
[Test][CI] Temporarily skip known failures under TileLang 5f70374c

superAngGao:
## Summary
- temporarily skip the currently known failing buckets under the validated TileLang `5f70374c48bb1d52e9260e995bf7adbcd64341e5` environment
- keep this PR narrowly focused on test scope reduction only
- leave published dependency / README updates to follow-up issue #1000

## Why
After validating the TileLang `5f70374c48bb1d52e9260e995bf7adbcd64341e5` environment tracked in #999, the remaining full-suite failures are no longer generic runtime instability. They have been isolated into a bounded set of known failing buckets.

This PR keeps CI / validation usable in the short term by temporarily marking those buckets, while the underlying kernels/operators are fixed separately.

Detailed error-cascade isolation remains documented in #980.

## Included temporary skips
- gated deltanet low-precision decode failures
- GLA decode and multi-step correctness coverage, where fp32 currently returns NaNs and fp16/bf16 remain tracked as low-precision launch failures
- attention decode / backward failure buckets still failing under TileLang `5f70374c`
- selected convolution / fp8 / rope / mhc_pre failures

## Validation
Command:
```bash
CUDA_VISIBLE_DEVICES=0 \
TILELANG_CLEANUP_TEMP_FILES=1 \
TMPDIR=/home/ga/tltmp \
/home/ga/TileOPs-upstream-cu128-t210/.venv-cu128-t210-ded6/bin/python -m pytest -q tests
```

Result:
```text
2180 passed, 89 skipped, 16 warnings in 162.99s
```

## Follow-ups
- #980 error cascade / CUDA poisoning investigation record
- #999 validated TileLang `5f70374c` env bring-up tracking
- #1000 published TileLang dependency and README follow-up

### DIFF
--- a/tests/ops/attention/test_deepseek_dsa_decode.py
+++ b/tests/ops/attention/test_deepseek_dsa_decode.py
@@ -71,7 +71,7 @@ def test_sparse_mla_decode(batch: int, heads: int, seq_len_q: int, seq_len_kv: i
                            dim_tail: int, topk: int, stride_kv: int, heads_kv: int,
                            q_start_index_s: int, sm_scale: float, dtype: torch.dtype,
                            tune: bool) -> None:
-    pytest.skip("Temporarily skipping known DeepSeek DSA decode failure in ded6 validation.")
+    pytest.skip("Temporarily skipping known DeepSeek DSA decode failure under TileLang 5f70374c (#999).")
     test = DsaDecodeTest(
         batch, heads, seq_len_q, seq_len_kv, dim, dim_tail, topk, stride_kv, heads_kv,
         q_start_index_s, sm_scale=sm_scale, dtype=dtype)
--- a/tests/ops/attention/test_deepseek_dsa_decode.py
+++ b/tests/ops/attention/test_deepseek_dsa_decode.py
@@ -71,6 +71,7 @@ def test_sparse_mla_decode(batch: int, heads: int, seq_len_q: int, seq_len_kv: i
                            dim_tail: int, topk: int, stride_kv: int, heads_kv: int,
                            q_start_index_s: int, sm_scale: float, dtype: torch.dtype,
                            tune: bool) -> None:
+    pytest.skip("Temporarily skipping known DeepSeek DSA decode failure in ded6 validation.")
     test = DsaDecodeTest(
         batch, heads, seq_len_q, seq_len_kv, dim, dim_tail, topk, stride_kv, heads_kv,
         q_start_index_s, sm_scale=sm_scale, dtype=dtype)
--- a/tests/ops/attention/test_deepseek_mla_decode.py
+++ b/tests/ops/attention/test_deepseek_mla_decode.py
@@ -65,7 +65,7 @@ class MlaDecodeFixture(FixtureBase):
 @MlaDecodeFixture
 def test_mla_decode(batch: int, heads: int, heads_kv: int, seq_len_kv: int, dim: int,
                     dim_pe: int, dtype: torch.dtype, tune: bool):
-    pytest.skip("Temporarily skipping known DeepSeek MLA decode failure in ded6 validation.")
+    pytest.skip("Temporarily skipping known DeepSeek MLA decode failure under TileLang 5f70374c (#999).")
     test = MlaDecodeTest(batch, heads, heads_kv, seq_len_kv, dim, dim_pe, dtype)
     op = MultiHeadLatentAttentionDecodeWithKVCacheFwdOp(
         batch, heads, heads_kv, seq_len_kv, dim, dim_pe, dtype, tune=tune)
--- a/tests/ops/attention/test_deepseek_mla_decode.py
+++ b/tests/ops/attention/test_deepseek_mla_decode.py
@@ -65,6 +65,7 @@ class MlaDecodeFixture(FixtureBase):
 @MlaDecodeFixture
 def test_mla_decode(batch: int, heads: int, heads_kv: int, seq_len_kv: int, dim: int,
                     dim_pe: int, dtype: torch.dtype, tune: bool):
+    pytest.skip("Temporarily skipping known DeepSeek MLA decode failure in ded6 validation.")
     test = MlaDecodeTest(batch, heads, heads_kv, seq_len_kv, dim, dim_pe, dtype)
     op = MultiHeadLatentAttentionDecodeWithKVCacheFwdOp(
         batch, heads, heads_kv, seq_len_kv, dim, dim_pe, dtype, tune=tune)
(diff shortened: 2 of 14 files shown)


Gabbering (commented):
## goose goose review — `af57aebc`

*honk.* The master has applied `pytest.skip()` to 14 files with the surgical precision of a goose applying bandages — which is to say, it mostly works, but there's a dead bird in the GLA test.

### goose Bugs

- `tests/ops/test_gla_recurrence.py:147-148,166-167` — The fp16 and bf16 params are already unconditionally skipped via `pytest.mark.skip` in the fixture (lines 76–139). The only params that survive the fixture filter and actually reach the test body are the `torch.float32` ones. Then the body says `if dtype == torch.float32: pytest.skip(...)`. Master, you have now skipped *every single parameterization* of `test_gla_decode` and `test_gla_decode_multi_step`. Low-precision: skipped by mark. fp32: skipped by guard. That's 100% skip rate. If that's intentional, the fixture-level marks are dead weight — just put an unconditional `pytest.skip()` at the top of the function like you did everywhere else. If it's *not* intentional (and the fp32 skip was meant to be narrower, e.g. only certain `(batch, heads)` combos), then you're accidentally skipping the fp32 cases that *do* pass. Either way, `test_gla_decode_vs_fla` still runs the fp32 params (no body guard there), so you have an inconsistency: fp32 is "known NaN" for decode and multi-step but apparently fine for the FLA comparison? That smells wrong. Pick a lane.

- `tests/ops/test_gated_deltanet_recurrence.py` — Same pattern, slightly less broken. The fixture skips fp16/bf16 via marks, leaving only fp32 params alive. The test body has *no* fp32 guard, so fp32 actually runs. This is fine and consistent. Mentioning it only because the GLA file clearly tried to copy this pattern and botched it by adding the fp32 body guard.

### egg Test gaps

- `tests/ops/test_gla_recurrence.py` — As noted above, `test_gla_decode` and `test_gla_decode_multi_step` now have zero live parameterizations. CI will report them as "89 skipped" and everyone will nod along, but you've lost *all* GLA decode coverage, including the fp32 cases that the gated-deltanet file kept alive. If fp32 GLA decode genuinely NaNs, that's a more serious regression than the low-precision cases and probably deserves its own call-out in the PR description rather than being quietly folded into the skip blanket.

Gabbering (commented):
## goose goose review — `38f5e1ab`

*honk.* The master has addressed the goose's prior honk about the GLA decode 100%-skip situation by replacing the conditional `if dtype == torch.float32` guard with an unconditional `pytest.skip()`. The convolution skip sets were refactored per @gemini-code-assist's suggestion. Skip messages now reference #999. This is all housekeeping. The goose finds no new bugs, no new perf issues, and no new test gaps introduced by this delta.

SILENT

Ibuki-wind (approved):
Trust-boundary verdict: clean; this stays within the test stage and does not mix implementation, manifest, workload, or benchmark changes.
The skips are scoped to documented TileLang 5f70374c failure buckets and keep tier markers/collection semantics intact.
Verified locally on the PR head: targeted collect-only over all changed files collected 236 tests without tier-validation errors; current PR checks, including gpu-smoke, are green.
Residual risk is the intentional temporary coverage reduction; keep #999/#1000 as the cleanup path.

Gabbering (commented):
## goose goose review — `bfb60751`

*honk.* Six commits, most of which are CI plumbing, a manifest redesign, and test skips. The goose has read the room. The test skips are clean, the prior GLA honk was addressed, @Ibuki-wind approved, @gemini-code-assist's nits were adopted. The new code in this delta is mostly docs, validator logic, CI shell, and `op_base._cache_key`. Let's see what the master left behind.

### goose Bugs

- `tileops/ops_manifest.yaml:73` — The RMSNormFwdOp roofline `vars.M` expression changed to `"product(x.shape[:dim % x.ndim]) * product(x.shape[dim % x.ndim + 1:])"`. This is wrong when `dim` is negative and `x.ndim > 1`. Consider `dim=-1, x.ndim=3`: `dim % x.ndim = 2`, so `M = product(x.shape[:2]) * product(x.shape[3:])` — that's correct. But `dim=-2, x.ndim=3`: `dim % x.ndim = 1`, so `M = product(x.shape[:1]) * product(x.shape[2:])` — also correct. Actually, wait — the *old* formula was `product(x.shape[:dim])` which is broken for negative `dim` (Python slicing with negative index gives the wrong prefix). The new formula normalizes via `% x.ndim` first. That's actually a fix. The goose retracts this honk mid-flight. However: the `static_dims` section for RMSNormFwdOp declares `N: "x.shape[dim]"` while the roofline uses `x.shape[dim % x.ndim]`. The `static_dims` expression is a forward-time validation rule per R20 — Python's `x.shape[dim]` with negative `dim` works fine because `tensor.shape` supports negative indexing. So these are consistent. Fine. The goose found no actual bug here after all. Moving on.

- `.github/actions/reclaim-runner-disk/action.yml:225-230` — The orphan-dir pruning loop uses `awk "BEGIN { exit !(${now_epoch} - ${newest_mtime} >= ${stale_seconds}) }"` where `newest_mtime` comes from a `find -printf '%T@'` pipeline that can produce a floating-point epoch (e.g. `1713550000.1234567890`). The subsequent `awk BEGIN` block does floating-point arithmetic, which is fine. But the fallback `newest_mtime=$(stat -c %Y ...)` produces an integer. Both paths feed into the same awk expression — no type mismatch issue. This is fine.

- `.github/actions/reclaim-runner-disk/reclaim_cache.sh:113-114` — `atomic_trim` computes `newest_mtime` from `find ... -type f -printf '%T@'` piped through awk, then compares `(( newest_mtime < cutoff ))` using bash integer arithmetic. But `%T@` produces a float like `1713550000.123456`. The awk `printf "%d\n"` truncates to integer, so the bash comparison works. Good — the `printf "%d\n"` is doing the right thing. No bug.

### feather Performance

- `.github/actions/reclaim-runner-disk/action.yml:196-197` — Sentinel-repair runs unconditionally on every CI invocation, even when disk is healthy and `SHOULD_RECLAIM` is false. The action comment says "per-subdir cost is a single readdir + stat, so this is safe to run every time." That's true for a small autotuner cache, but if the cache grows to thousands of subdirs (one per shape config), this becomes a non-trivial `find -maxdepth 1 -type d` + per-subdir stat on every single PR push. Not a bug, but worth a mental note for the master when the autotuner cache inevitably balloons.

### egg Test gaps

- `tests/test_op_base.py` — The `_cache_key` tests cover the warning and the axis-exclusion logic well. But there's no test for the case where `_static_axes` references an `(input_index, axis)` pair where `axis >= len(input_shapes[input_index])` — i.e., an out-of-range static axis. The current `_cache_key` implementation would silently skip it (the `(i, axis) not in self._static_axes` check just wouldn't match any actual axis), which means a misconfigured `_static_axes` would degrade to full-shape keying without any signal. A bounds check or at least a test documenting the expected behavior for invalid `_static_axes` would catch subclass wiring mistakes early.

PR #1015, opened 2026-04-21
[BugFix] Align Mamba2 kernels to official Mamba-2 triton interface + refactor  for performance

stelladuyx:
### Summary
Three Mamba-2 kernels are aligned to the official mamba_ssm interface and refactored for correctness and performance.

ssd_decode — interface alignment
Updated to match that interface exactly. The math is unchanged — dA[b,h,p,n] = exp(dt[b,h,p] * A[h,p,n]) — but the old per-block scalar hoist is replaced by per-element reads, enabling independent per-dim decay in the future.

ssd_chunk_scan_fwd — layout alignment + correctness fixes
Aligned all tensor layouts to the official mamba_ssm _chunk_scan_fwd interface. 

ssd_chunk_state_fwd — GEMM refactor
The inner reduction loop previously computed a serial outer-product over (l, p, n). Restructured as a standard GEMM.

Test plan

- [ ]  test_ssd_decode — 4/4 pass (fp16 + bf16, n_groups 1–4)
- [ ]  test_ssd_chunk_state_fwd — 5/5 pass (fp16 + bf16, with/without seq_idx)
- [ ]  test_ssd_chunk_scan_fwd — 4/4 pass (fp16 + bf16)


superAngGao:
**`ssd_chunk_scan_fwd`: `cb` shape not updated to group-owned**

The PR description says `cb` is being changed from `[B,C,H,L,L]` to `[B,C,G,L,L]`, but `ssd_chunk_scan_fwd.py:84` still has:

```python
cb_shape = (batch, num_chunks, n_heads, chunk_len, chunk_len)
```

This is inconsistent with both the PR intent and the official `mamba_ssm` reference, where `CB` has shape `(batch, nchunks, ngroups, chunk_size, chunk_size)` and is accessed via `pid_h // nheads_ngroups_ratio` to map heads to groups. `ssd_chunk_state_fwd` already handles `n_groups` correctly — `ssd_chunk_scan_fwd` should follow the same pattern.

superAngGao:
**`BenchmarkReport.record`: baseline calls pass a string instead of the `op` object**

```python
# bench_ssd_chunk_scan_fwd.py:66
BenchmarkReport.record("ssd_chunk_scan_fwd", locals(), result_bl, tag="baseline")

# bench_ssd_chunk_state_fwd.py:69
BenchmarkReport.record("ssd_chunk_state_fwd", locals(), result_bl, tag="baseline")
```

Per project convention (see `bench_gqa.py`, `bench_deltanet_recurrence.py`), the first argument to `BenchmarkReport.record` should always be the `op` object — even for baseline entries — so the nightly report can extract metadata (batch size, dtype, etc.) correctly. The tag `"baseline"` is also too generic; the rules require specific names like `"torch-ref"` or `"mamba-ssm"`. Suggested fix:

```python
BenchmarkReport.record(op, locals(), result_bl, tag="torch-ref")
```

superAngGao:
**`ssd_chunk_scan_fwd`: `prev_states` layout mismatch between reference and kernel**

The test reference (`test_ssd_chunk_scan_fwd.py:13`) declares `prev_states: (b, c, h, n, p)`, but the kernel shape definition (`ssd_chunk_scan_fwd.py:87`) is:

```python
state_shape = (batch, num_chunks, n_heads, d_state, d_head)  # (N, P) order
```

The PR aligns other tensors to the official convention where `prev_states` is `(b, c, h, p, n)` (P before N). Could you confirm whether the reference and kernel are intentionally using the same `(N, P)` order here, or if one of them needs to be updated?

stelladuyx:
Already fixed in ssd_chunk_scan.py.

tileops/kernels/mamba/ssd_chunk_scan.py
```diff
@@ -80,178 +94,177 @@ def kernel_func(
...
-            dA_cumsum: T.Tensor(d_a_shape, accum_dtype),    # type: ignore
-            C: T.Tensor(c_shape, dtype),                    # type: ignore
-            prev_states: T.Tensor(state_shape, dtype),      # type: ignore
-            d_t: T.Tensor(d_t_shape, dtype),                # type: ignore
-            out: T.Tensor(out_shape, accum_dtype),          # type: ignore
+            x:           T.Tensor(x_shape, dtype),           # type: ignore
+            cb:          T.Tensor(cb_shape, dtype),           # type: ignore
+            dA_cumsum:   T.Tensor(dA_shape, accum_dtype),    # type: ignore
+            C_mat:       T.Tensor(c_shape, dtype),            # type: ignore
+            prev_states: T.Tensor(states_shape, dtype),       # type: ignore
```
Ibuki-wind: `prev_states` is declared with `dtype`, but the upstream SSD state path produces float32 states (`SSDChunkStateFwdKernel` and `SSDStatePassingFwdKernel` outputs are `accum_dtype`/`torch.float32`), so the updated chunk scan cannot consume the in-repo pipeline output. → Make `prev_states` a float32/`accum_dtype` input, cast into `state_tile` only if the GEMM requires input dtype, and update the chunk-scan workload/reference tests to generate float32 `prev_states`.

Ibuki-wind (changes requested):
### New issues
- `tileops/kernels/mamba/ssd_chunk_scan.py:112` — `prev_states` is declared as input dtype even though the SSD state pipeline produces float32 states.

### Overall
REQUEST_CHANGES because the new chunk-scan interface cannot consume the in-repo state-passing output without an explicit dtype contract fix.
Update the kernel contract and workload/tests so `prev_states` matches the producer dtype.

Ibuki-wind (approved):
### Prior blockers
- **resolved** — `tileops/kernels/mamba/ssd_chunk_scan.py:112` `prev_states` now uses `accum_dtype`/float32 at the kernel boundary, and the workload generator now provides float32 state tensors.

### Overall
No new blocking issues in the Round 2 convergence changes; the prior dtype contract blocker is resolved.

### Advisory
- This PR still spans implementation, tests, workloads, and benchmarks; per migration policy this remains advisory only.

PR #1024, opened 2026-04-22
[Refactor][Roofline] remove manifest-level roofline evaluator

lcy-seso:
## Summary

This PR now targets the design direction that target design should not have a manifest-level evaluator. Instead of keeping a legacy evaluator boundary, it removes the legacy evaluator surface and migrates current callers.

- Delete `tileops/manifest_legacy_roofline.py`.
- Remove manifest exports for `_safe_eval`, `eval_roofline`, `resolve_roofline_vars`, and `has_roofline_vars`.
- Move benchmark roofline consumption to op-local `eval_roofline()` and `ManifestBenchmark(op_name, op, workload)`.
- Add local `eval_roofline()` implementations for the currently migrated norm, MoE, reduction, and softmax-family benchmark surfaces.
- Update validator, tests, and roofline docs for the op-local evaluator contract.

## Test Plan

- `ruff check` on changed Python files
- `python -m compileall tileops benchmarks scripts tests`
- `pytest tests/test_ops_manifest.py tests/test_workloads_to_params.py tests/test_validate_manifest.py benchmarks/tests/test_roofline_workload_protocol.py`
- `python scripts/validate_manifest.py`

Note: `validate_manifest.py` passes and still reports the existing attention benchmark L4 warnings; those are pre-existing warnings outside this cleanup scope.

## Follow-up

No follow-up issues or suggestions.


tileops/manifest_legacy_roofline.py
```diff
@@ -0,0 +1,615 @@
...
+from __future__ import annotations
+
+import ast
+import importlib
+import math
+import operator
+from math import prod as _math_prod
+from typing import Any
+
+from . import manifest as _manifest_core
```
Ibuki-wind: The top-level `from . import manifest as _manifest_core` makes `import tileops.manifest_legacy_roofline` fail: `tileops.manifest` re-enters this partially initialized module and tries to re-export `_safe_eval` before it is defined. → Move the `_manifest_core` import below the evaluator definitions, or make manifest loading lazy through a helper, so both `import tileops.manifest` and direct legacy-module imports succeed.

Ibuki-wind (changes requested):
Trust boundary is OK: this is a code-only isolation of the legacy manifest roofline surface, with no manifest/op/test/benchmark ownership mixing.
The extraction is behavior-preserving through the existing tileops.manifest compatibility path; the targeted manifest tests pass.
One import-order fix is needed so the newly introduced implementation module is importable in isolation.

benchmarks/benchmark_base.py
```diff
@@ -426,71 +404,36 @@ def workloads_to_params(op_name: str, include_extra: bool = False) -> list:
...
-            params=self._op_params,
-        )
-        resolved.setdefault("elem_bytes", elem_bytes)
-        return resolved
-
     def _get_roofline(self) -> tuple[float, float]:
         if self._roofline_cache is None:
-            self._roofline_cache = eval_roofline(
-                self._op_name, **self._roofline_vars())
+            flops, mem_bytes = self._op.eval_roofline()
```
Ibuki-wind, line 435: The raw counts fetched here are not preserved in the benchmark artifact: `_build_result()` only derives `tflops` / `bandwidth_tbs` and `BenchmarkReport.dump()` only emits those derived rates. That violates the roofline contract in `docs/roofline.md` that benchmark output carries the `(flops, bytes)` returned by `op.eval_roofline()`, and it becomes a real data-loss issue once the manifest-level evaluator is deleted because M5/historical reanalysis can no longer recover those counts from the report. → Store the raw counts in the result, e.g. `result["flops"] = flops` and `result["bytes"] = mem_bytes`, then include those keys in `BenchmarkReport.dump()` / thread-local benchmark output and update the corresponding tests.

benchmarks/tests/test_roofline_workload_protocol.py
```diff
@@ -171,54 +165,26 @@ def gen_inputs(self):
...
-    assert rv["N"] == 4096
-    assert rv["elem_bytes"] == 2
-
-
-@pytest.mark.smoke
-def test_manifest_benchmark_respects_dim_zero():
-    """With ``op_params={"dim": 0}`` the vars must reduce the *first* axis,
-    producing M/N that differ from the hardcoded last-axis heuristic.
-    """
+def test_manifest_benchmark_reads_op_eval_roofline_once():
```
Ibuki-wind, line 180: This coverage only proves that `ManifestBenchmark` delegates to a fake op. The PR moves real roofline arithmetic into many concrete ops, but no test checks that those new `eval_roofline()` implementations return the manifest-intended numbers for actual migrated op families or dynamic-shape cases; after deleting `tests/test_resolve_roofline_vars.py`, a hand-transcription error in any formula can ship silently. → Add parametrized numeric regression tests for the migrated op-local formulas, including representative norm/MoE fixed-shape cases and reduction/softmax cases where `forward()` binds `(M, N)`, asserting explicit expected `(flops, bytes)` values.

Ibuki-wind (changes requested):
Requesting changes. The trust-boundary crossing is acceptable for this migration, and the overall direction matches the target design: remove the manifest-level evaluator and make benchmarks consume op-local `eval_roofline()`. The remaining blockers are design-completion gaps after deleting that evaluator: benchmark artifacts still need to carry raw roofline counts, and the migrated op-local formulas need real numeric regression coverage. After fixing those, rerun the manifest validator plus the targeted roofline/manifest tests.

Ibuki-wind (approved):
Approved. After re-checking the current codebase, the earlier blockers should not gate this migration. The trust-boundary crossing is intentional for this design cleanup, the manifest-level roofline evaluator is removed, and benchmark consumption has moved to op-local `eval_roofline()` without breaking the existing benchmark/nightly consumers. Raw `(flops, bytes)` in benchmark artifacts and broader numeric formula regression coverage should be revisited when the M5 roofline consumer actually lands; they are not required to clear this PR's legacy-evaluator cleanup.

PR #1054, opened 2026-04-26
[CI] skip validate-manifest and benchmark-contract-tests when paths unchanged

lcy-seso:
## Summary

- Add `detect-changes` job to `preflight.yml` that lists PR files via `gh api pulls/<n>/files` and emits two outputs: `manifest` and `benchmark`. Fails open if the API errors out or returns no files.
- Gate `validate-manifest` and `benchmark-contract-tests` on the matching output so unrelated PRs don't pay the ~2-minute cost per job.
- Non-PR events (push to `main`, tags, manual dispatch) default both outputs to `true`.
- Add `pull-requests: read` to workflow permissions for the `gh api` call.
- Also gate `validate-manifest` on `ci-gate.outputs.skip` for consistency with the other jobs (allows `ready_for_review` rerun suppression).

## Watch sets

`manifest=true` when any of:
- `tileops/ops_manifest.yaml`
- `.github/workflows/preflight.yml`

`benchmark=true` when any of:
- `benchmarks/**`
- `workloads/**` (imported by `benchmarks/tests/test_roofline_workload_protocol.py`)
- `tileops/manifest.py` (imported by `benchmarks/benchmark_base.py`)
- `tileops/ops_manifest.yaml` (read transitively by `workloads_to_params()` → `load_workloads()`)
- `.github/workflows/preflight.yml`

## ci-gate update

`ci-gate`'s `ready_for_review` short-circuit previously required `benchmark-contract-tests` to conclude `success`. With this PR that job will often be intentionally `skipped`, so the predicate now accepts `success` or `skipped` — only genuine failures block the rerun-suppression path.

Motivation: PR #1051 was a single-file docs change that still triggered both `validate-manifest` (1m48s) and `benchmark-contract-tests` (1m55s).

## Branch protection note

The `main` ruleset (id 8507572) requires only `validate-pr-title` and `gpu-smoke`, so neither newly gated job is required. GitHub rulesets also treat `skipped` conclusions as satisfying required checks, so this change is safe even if these jobs are later added to the required list.

## Test plan

- [x] `detect-changes` fails open on API error / empty result.
- [ ] After merge, open a docs-only follow-up PR and confirm `validate-manifest` and `benchmark-contract-tests` show `skipped`.
- [ ] Confirm a manifest-only PR runs `validate-manifest` AND `benchmark-contract-tests` (manifest is in both watch sets).
- [ ] Confirm a `workloads/**`-only PR runs `benchmark-contract-tests` but skips `validate-manifest`.

### DIFF
--- a/.github/workflows/preflight.yml
+++ b/.github/workflows/preflight.yml
@@ -101,10 +101,12 @@ jobs:
                 ;;
             esac
             # `benchmark-contract-tests` runs `pytest benchmarks/tests`, which
-            # imports `workloads.workload_base` and `tileops.manifest`. Any of
-            # those layers can break the contract tests, so include them all.
+            # imports `workloads.workload_base`, `tileops.manifest`, and reads
+            # `tileops/ops_manifest.yaml` via `load_workloads()` /
+            # `workloads_to_params()`. Any of those layers can break the
+            # contract tests, so include them all.
             case "$f" in
-              benchmarks/*|workloads/*|tileops/manifest.py|"$WORKFLOW_PATH")
+              benchmarks/*|workloads/*|tileops/manifest.py|tileops/ops_manifest.yaml|"$WORKFLOW_PATH")
                 benchmark=true
                 ;;
             esac
--- a/.github/workflows/preflight.yml
+++ b/.github/workflows/preflight.yml
@@ -73,7 +73,21 @@ jobs:
             exit 0
           fi
 
-          FILES=$(gh api --paginate "repos/$REPO/pulls/$PR_NUMBER/files" --jq '.[].filename')
+          # Fail-open: if the API call errors out or returns nothing, run
+          # everything. The optimization must never reduce CI reliability.
+          if ! FILES=$(gh api --paginate "repos/$REPO/pulls/$PR_NUMBER/files" --jq '.[].filename' 2>&1); then
+            echo "::warning::gh api failed; defaulting to run all gated jobs"
+            echo "$FILES"
+            echo "manifest=true" >> "$GITHUB_OUTPUT"
+            echo "benchmark=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          if [[ -z "$FILES" ]]; then
+            echo "::warning::no files returned; defaulting to run all gated jobs"
+            echo "manifest=true" >> "$GITHUB_OUTPUT"
+            echo "benchmark=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
           echo "Changed files in PR #$PR_NUMBER:"
           echo "$FILES"
 
@@ -86,8 +100,11 @@ jobs:
                 manifest=true
                 ;;
             esac
+            # `benchmark-contract-tests` runs `pytest benchmarks/tests`, which
+            # imports `workloads.workload_base` and `tileops.manifest`. Any of
+            # those layers can break the contract tests, so include them all.
             case "$f" in
-              benchmarks/*|"$WORKFLOW_PATH")
+              benchmarks/*|workloads/*|tileops/manifest.py|"$WORKFLOW_PATH")
                 benchmark=true
                 ;;
             esac
@@ -130,6 +147,10 @@ jobs:
             exit 0
           fi
 
+          # `benchmark-contract-tests` is intentionally skipped on PRs that
+          # don't touch its inputs (see detect-changes), so accept `skipped`
+          # as well as `success` here — only genuine failures should block
+          # the ready_for_review short-circuit.
           CHECKS=$(gh api "repos/${{ github.repository }}/commits/$HEAD_SHA/check-runs" --jq '
             [
               .check_runs
@@ -139,9 +160,9 @@ jobs:
               | .conclusion
             ]
           ')
-          ALL_SUCCESS=$(echo "$CHECKS" | jq 'all(. == "success")')
+          ALL_OK=$(echo "$CHECKS" | jq 'all(. == "success" or . == "skipped")')
           COUNT=$(echo "$CHECKS" | jq 'length')
-          if [[ "$ALL_SUCCESS" == "true" && "$COUNT" -ge 4 ]]; then
+          if [[ "$ALL_OK" == "true" && "$COUNT" -ge 4 ]]; then
             echo "CI already passed for SHA $HEAD_SHA; skipping"
             echo "skip=true" >> "$GITHUB_OUTPUT"
           else
--- a/.github/workflows/preflight.yml
+++ b/.github/workflows/preflight.yml
@@ -220,8 +220,8 @@ jobs:
         run: gitleaks dir . --redact --no-banner
 
   validate-manifest:
-    needs: [validate-pr-title, detect-changes]
-    if: ${{ github.repository == 'tile-ai/TileOPs' && needs.detect-changes.outputs.manifest == 'true' }}
+    needs: [validate-pr-title, ci-gate, detect-changes]
+    if: ${{ github.repository == 'tile-ai/TileOPs' && needs.ci-gate.outputs.skip != 'true' && needs.detect-changes.outputs.manifest == 'true' }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
--- a/.github/workflows/preflight.yml
+++ b/.github/workflows/preflight.yml
@@ -3,6 +3,7 @@ name: Preflight Checks
 permissions:
   contents: read
   checks: read
+  pull-requests: read
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/preflight.yml
```diff
@@ -47,6 +48,56 @@ jobs:
...
+          benchmark=false
+          while IFS= read -r f; do
+            [[ -z "$f" ]] && continue
+            case "$f" in
+              tileops/ops_manifest.yaml|scripts/validate_manifest.py|"$WORKFLOW_PATH")
+                manifest=true
+                ;;
+            esac
+            case "$f" in
+              benchmarks/*|"$WORKFLOW_PATH")
```
Ibuki-wind: The `benchmark=true` filter is too narrow for the job it gates. `benchmark-contract-tests` does not only depend on `benchmarks/**`; it also imports the shared workloads layer (`benchmarks/tests/test_roofline_workload_protocol.py` imports `workloads.workload_base` at lines 152-155). With the current filter, a PR that breaks `workloads/**` will skip `benchmark-contract-tests` entirely and hide the regression → include `workloads/**` (and any other shared inputs that `benchmarks/tests` imports) in the benchmark path set.

.github/workflows/preflight.yml
```diff
@@ -195,8 +246,8 @@ jobs:
     # CPU-only contract tests for benchmark workload protocols.
     # Scoped to `benchmarks/tests/` so the GPU-bound `benchmarks/ops/`
     # suites (nightly-only) are NOT collected on PR CI.
-    needs: [validate-pr-title, ci-gate]
-    if: ${{ github.repository == 'tile-ai/TileOPs' && needs.ci-gate.outputs.skip != 'true' }}
+    needs: [validate-pr-title, ci-gate, detect-changes]
+    if: ${{ github.repository == 'tile-ai/TileOPs' && needs.ci-gate.outputs.skip != 'true' && needs.detect-changes.outputs.benchmark == 'true' }}
```
Ibuki-wind, line 273: `ci-gate` still requires `benchmark-contract-tests` to have conclusion `success`, but this PR intentionally makes that job `skipped` on many SHAs. On `ready_for_review`, those SHAs will now fail the short-circuit check even when everything that was supposed to run already passed, so the existing rerun suppression regresses → treat an intentionally skipped `benchmark-contract-tests` as acceptable here, or exclude it from the success/count check when `detect-changes` says the benchmark job should not run.

Ibuki-wind (changes requested):
No trust-boundary split issue here; the workflow-only scope is fine. The blockers are both in the new CI routing logic: the benchmark path filter under-approximates the inputs of `benchmark-contract-tests`, so some real regressions would now be skipped, and `ci-gate` still assumes that `benchmark-contract-tests` must conclude `success`, which regresses the existing ready-for-review short-circuit once the job is intentionally skipped. After those two conditions are reconciled, the optimization direction looks sound.

.github/workflows/preflight.yml
```diff
@@ -47,6 +48,56 @@ jobs:
...
+
+          # Non-PR events (push to main, tags, manual dispatch) always run
+          # both gated jobs — no base ref to diff against.
+          if [[ "$EVENT_NAME" != "pull_request" ]]; then
+            echo "manifest=true" >> "$GITHUB_OUTPUT"
+            echo "benchmark=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          FILES=$(gh api --paginate "repos/$REPO/pulls/$PR_NUMBER/files" --jq '.[].filename')
```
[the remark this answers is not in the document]
lcy-seso: Adopted. The detect-changes step now fails open: if `gh api` errors or returns no files, both outputs default to `true` and a `::warning` is emitted. See 129df7e.

.github/workflows/preflight.yml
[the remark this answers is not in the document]
lcy-seso: Adopted. `ci-gate`'s short-circuit now accepts `skipped` as well as `success` (`all(. == "success" or . == "skipped")`), so an intentionally-skipped `benchmark-contract-tests` no longer prevents the ready_for_review rerun suppression. See 129df7e.

.github/workflows/preflight.yml
lcy-seso: Adopted. The benchmark filter now also matches `workloads/**` (covers the `workloads.workload_base` import in test_roofline_workload_protocol.py:155) and `tileops/manifest.py` (covers `tileops.manifest.load_workloads` imported by benchmarks/benchmark_base.py:20). See 129df7e.

.github/workflows/preflight.yml
lcy-seso: Adopted. Same fix as the Copilot thread above: `ci-gate` now accepts `skipped` for `benchmark-contract-tests`. See 129df7e.

.github/workflows/preflight.yml
```diff
@@ -47,6 +48,73 @@ jobs:
...
+            case "$f" in
+              tileops/ops_manifest.yaml|scripts/validate_manifest.py|"$WORKFLOW_PATH")
+                manifest=true
+                ;;
+            esac
+            # `benchmark-contract-tests` runs `pytest benchmarks/tests`, which
+            # imports `workloads.workload_base` and `tileops.manifest`. Any of
+            # those layers can break the contract tests, so include them all.
+            case "$f" in
+              benchmarks/*|workloads/*|tileops/manifest.py|"$WORKFLOW_PATH")
```
Ibuki-wind: The benchmark watch set is still missing `tileops/ops_manifest.yaml`. `benchmark-contract-tests` exercises manifest-backed helpers, not just `workloads/` and `tileops.manifest`: for example `benchmarks/tests/test_roofline_workload_protocol.py` calls `workloads_to_params("SumFwdOp")`, which reads workloads from `ops_manifest.yaml` via `load_workloads()`. With the current filter, a manifest-only PR can still skip this job and hide a real benchmark-contract regression → include `tileops/ops_manifest.yaml` in the `benchmark=true` path set as well.

Ibuki-wind (changes requested):
The earlier reliability fixes are in place: `detect-changes` now fails open and `ci-gate` no longer treats an intentionally skipped benchmark job as a failure. One blocker remains in the benchmark watch set, though: `benchmark-contract-tests` still has a direct manifest dependency that this filter does not include, so a subset of manifest regressions would be skipped instead of tested.

.github/workflows/preflight.yml
lcy-seso: Adopted. `tileops/ops_manifest.yaml` added to the benchmark path set. Verified test_roofline_workload_protocol.py:211 calls `workloads_to_params("SumFwdOp")` which transitively reads the manifest via `load_workloads()`. See 17cf86b.

.github/workflows/preflight.yml
```diff
@@ -47,6 +48,75 @@ jobs:
...
+                ;;
+            esac
+            # `benchmark-contract-tests` runs `pytest benchmarks/tests`, which
+            # imports `workloads.workload_base`, `tileops.manifest`, and reads
+            # `tileops/ops_manifest.yaml` via `load_workloads()` /
+            # `workloads_to_params()`. Any of those layers can break the
+            # contract tests, so include them all.
+            case "$f" in
+              benchmarks/*|workloads/*|tileops/manifest.py|tileops/ops_manifest.yaml|"$WORKFLOW_PATH")
+                benchmark=true
```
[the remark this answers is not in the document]
lcy-seso: Adopted (description updated, filter kept). The wider filter is correct — `benchmarks/tests` transitively depends on `workloads/**`, `tileops/manifest.py`, and `tileops/ops_manifest.yaml` (verified imports + `load_workloads("SumFwdOp")` call site). Narrowing would hide regressions. PR description now lists the full watch set.

.github/workflows/preflight.yml
```diff
@@ -146,8 +220,8 @@ jobs:
         run: gitleaks dir . --redact --no-banner
 
   validate-manifest:
-    needs: [validate-pr-title]
-    if: ${{ github.repository == 'tile-ai/TileOPs' }}
+    needs: [validate-pr-title, detect-changes]
+    if: ${{ github.repository == 'tile-ai/TileOPs' && needs.detect-changes.outputs.manifest == 'true' }}
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Added `ci-gate` to validate-manifest's needs and `needs.ci-gate.outputs.skip != 'true'` to its if, matching the other gated jobs. See 1fd25df.

Ibuki-wind (approved):
Clean after the follow-up fixes. The gated watch sets now cover the real inputs of both jobs, the optimization fails open on API issues, and the `ready_for_review` short-circuit remains coherent with intentionally skipped jobs.

PR #1053, opened 2026-04-26
[CI] Move GPU CI caches to data7 shared cache

superAngGao:
## Summary

- move trusted GPU smoke caches to `/data7/shared/ci-cache/{tilelang,triton,pip,wheels}`
- point nightly at the same `/data7/shared/ci-cache/*` layout
- update reclaim-runner-disk defaults/comments and runner maintenance notes for the data7 cache root

Closes #1052.

## Notes

This does not change PR gate or security-policy behavior. The TL0.1.9 smoke hang skip was already merged in #1043; this PR carries the follow-up cache placement cleanup from the same investigation.

## Validation

- `python - <<'PY' ... yaml.safe_load(...) ... PY` for `gpu-smoke.yml`, `nightly.yml`, `runner-maintenance.yml`, and `reclaim-runner-disk/action.yml`
- `bash -n .github/actions/reclaim-runner-disk/reclaim_cache.sh`
- `python -m pytest -q tests/test_reclaim_action.py tests/test_gpu_smoke_policy.py tests/test_ci_venv_hash.py` -> `32 passed`

Issue #1052, opened 2026-04-26
[CI] Move GPU CI caches to data7 shared cache


superAngGao:
## Problem

During the TileLang 0.1.9 GPU smoke investigation, the runner cache layout was split across older locations such as `/home/ci-runner/.tilelang/cache`, `/home/ci-runner/.triton/cache`, `/home/ci-runner/.cache/pip`, and nightly's `/cache/*` layout.

That makes runner disk reclamation and cache reuse harder to reason about. The desired persistent cache root for these GPU CI jobs is `/data7/shared/ci-cache/*`.

## Scope

Move the trusted GPU smoke and nightly cache roots to `/data7/shared/ci-cache` and update the reclaim action defaults/docs to match.

The TL0.1.9 smoke hang skip itself was already merged in #1043; this issue tracks the follow-up cache placement cleanup found while validating that fix locally.

## Acceptance Criteria

- GPU smoke trusted cache paths use `/data7/shared/ci-cache/{tilelang,triton,pip,wheels}`.
- Nightly uses the same `/data7/shared/ci-cache/*` paths.
- Reclaim runner disk defaults and comments match the new cache layout.
- PR gate/security policy logic remains unchanged.

PR #1058, opened 2026-04-26
[CI] Fall back when data7 cache root is unavailable

### LINKED ISSUE #1057: [CI] Fall back when data7 cache root is unavailable
## Problem

GPU Smoke on `main` failed in run https://github.com/tile-ai/TileOPs/actions/runs/24949628223/job/73057345688 immediately after #1053 moved caches to `/data7/shared/ci-cache`.

The failing step was `Resolve runtime state` on a trusted push run. It unconditionally set `TRUSTED_CACHE_ROOT=/data7/shared/ci-cache` and then ran:

```text
mkdir: cannot create directory '/data7': Permission denied
```

This means at least one runner container can start without `/data7` mounted or writable for the unprivileged `ci-runner` user. The workflow should prefer the shared data7 cache when available, but it should not fail before tests start when that mount is temporarily absent.

## Proposed fix

- Keep `/data7/shared/ci-cache` as the preferred trusted cache root.
- Probe whether the preferred cache root is writable, creating it only if that succeeds without privilege escalation.
- Fall back to `/home/ci-runner/.cache/tileops-ci` for the current runner when data7 is unavailable.
- Add a structural test so `Resolve runtime state` keeps the data7 fallback guard.

Parent tracker: #1041
Follow-up to: #1052 / #1053

superAngGao:
## Summary

- Keep `/data7/shared/ci-cache` as the preferred GPU smoke trusted cache root.
- Probe whether that root can actually be created/written before using it.
- Fall back to `/home/ci-runner/.cache/tileops-ci` when `/data7` is absent or not writable, so the job does not fail during `Resolve runtime state`.
- Add a structural test to keep the fallback guard in place.

Fixes #1057.
Follow-up to #1052 / #1053.

## Validation

```text
python -m pytest -q tests/test_gpu_smoke_policy.py tests/test_reclaim_action.py tests/test_ci_venv_hash.py
33 passed in 1.90s
```

```text
python - <<'PY'
from pathlib import Path
import yaml
for path in ['.github/workflows/gpu-smoke.yml', '.github/actions/reclaim-runner-disk/action.yml']:
    yaml.safe_load(Path(path).read_text())
    print('ok', path)
PY
```

```text
git diff --check
```

### DIFF
--- a/.github/workflows/gpu-smoke.yml
+++ b/.github/workflows/gpu-smoke.yml
@@ -449,8 +449,30 @@ jobs:
           VENV_PREFIX="tileops_ci_venv"
           VENV_DIR="${VENV_PREFIX}_${HASH_INPUT}"
           TRUSTED_VENV_PATH="${{ runner.tool_cache }}/${VENV_DIR}"
-          TRUSTED_CACHE_ROOT="/data7/shared/ci-cache"
+          TRUSTED_CACHE_ROOT_CANDIDATE="/data7/shared/ci-cache"
+          FALLBACK_TRUSTED_CACHE_ROOT="/home/ci-runner/.cache/tileops-ci"
           FORK_CAN_COPY_TRUSTED_VENV="false"
+
+          select_trusted_cache_root() {
+            local candidate="$1"
+            local fallback="$2"
+
+            if [[ -d "${candidate}" && -w "${candidate}" ]]; then
+              printf '%s\n' "${candidate}"
+              return
+            fi
+
+            if mkdir -p "${candidate}" 2>/dev/null && [[ -d "${candidate}" && -w "${candidate}" ]]; then
+              printf '%s\n' "${candidate}"
+              return
+            fi
+
+            echo "::warning::${candidate} is unavailable or not writable; using ${fallback} for this run." >&2
+            mkdir -p "${fallback}"
+            printf '%s\n' "${fallback}"
+          }
+
+          TRUSTED_CACHE_ROOT="$(select_trusted_cache_root "${TRUSTED_CACHE_ROOT_CANDIDATE}" "${FALLBACK_TRUSTED_CACHE_ROOT}")"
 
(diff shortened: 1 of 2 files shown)


lcy-seso (approved):

Gabbering (commented):
goose *goose skimmed `fc1d36b` — nothing to honk about.*

.github/workflows/gpu-smoke.yml
```diff
@@ -449,9 +449,31 @@ jobs:
...
+              printf '%s\n' "${candidate}"
+              return
+            fi
+
+            echo "::warning::${candidate} is unavailable or not writable; using ${fallback} for this run." >&2
+            mkdir -p "${fallback}"
+            printf '%s\n' "${fallback}"
+          }
+
+          TRUSTED_CACHE_ROOT="$(select_trusted_cache_root "${TRUSTED_CACHE_ROOT_CANDIDATE}" "${FALLBACK_TRUSTED_CACHE_ROOT}")"
```
Ibuki-wind: `select_trusted_cache_root` runs before the `IS_FORK` branch and calls `mkdir -p` on both the `/data7` candidate and the `/home/ci-runner` fallback, so an external fork PR can still create persistent shared cache directories before it enters the isolated runtime path. This violates the existing fork invariant documented below (`Fork PRs never write to any persistent shared state`). → Move write-capable cache-root creation into the trusted branch, and make the fork branch use only existing readable trusted cache roots as copy sources or skip cache copy entirely when none exist.

Ibuki-wind (changes requested):
Trust-boundary verdict: CI-only change plus structural test, no implementation/manifest split needed.
The fallback idea is sound for trusted runs and the structural test passes locally.
Blocked on preserving the existing external-fork isolation invariant before persistent cache roots are created.

superAngGao:
Closed as this modification is implemented on the CI host locally.

PR #1062, opened 2026-04-26
[Fix][Skills] address align-op + implement-op bugs found during RMSNormFwdOp migration

lcy-seso:
## What this PR does

Three contract fixes to manifest-skill specs:

| # | Skill | Change |
|---|---|---|
| 1 | `align-op` CLASSIFY | Add first-op bias: when no same-family op has `status: implemented` following the canonical pattern (`docs/ops-design.md` § Step 3), auto-select `redesign`. Non-goals updated to carve out this exception. User may override with `--mode=minor`. |
| 2 | `implement-op` constraint | Pre-existing tests in `<source_test>` whose call-sites use the legacy API MAY be updated. Tests written by `test-op` in the same run remain sacred. |
| 3 | `implement-op` dual-path | Subclass-count gate: `grep -rlE "class\s+[A-Z][A-Za-z0-9]*\s*\(\s*<BaseName>\s*\)" tileops/ops/`. One subclass → refactor in place. Multiple subclasses → keep legacy `__init__` until cleanup gate. |

Docs-only, +12/-3 lines.

### DIFF
--- a/.claude/skills/align-op/SKILL.md
+++ b/.claude/skills/align-op/SKILL.md
@@ -102,6 +102,8 @@ Decide which case applies. Machine-decidable input: does `source.op` exist?
 - `source.op` **missing** → only `green` is valid. `--mode=minor` → BLOCKED ("`source.op` is missing; cannot edit a non-existent op file; use `--mode=green` or omit `--mode`"). `--mode=redesign` → BLOCKED ("`source.op` is missing; no archive source to rewrite; use `--mode=green` or omit `--mode`").
 - `source.op` **exists** → `--mode=green` → BLOCKED ("`source.op` already exists; green-field scaffold would silently overwrite; use `--mode=redesign` for rewrite+port or `--mode=minor` for in-place edit").
 
+**First-op bias.** If no op in the same `family` has `status: implemented` and follows the canonical static_dims pattern (`docs/ops-design.md` § Step 3), set `auto-case = redesign` (skip the prompt). mode.json: `decided_by: "auto"`, `reason: "no canonical-pattern precedent in family <name>"`. User may override with `--mode=minor`.
+
 Write `.foundry/plan/<op_name>/mode.json`:
 
 ```json
--- a/.claude/skills/align-op/SKILL.md
+++ b/.claude/skills/align-op/SKILL.md
@@ -102,7 +102,7 @@ Decide which case applies. Machine-decidable input: does `source.op` exist?
 - `source.op` **missing** → only `green` is valid. `--mode=minor` → BLOCKED ("`source.op` is missing; cannot edit a non-existent op file; use `--mode=green` or omit `--mode`"). `--mode=redesign` → BLOCKED ("`source.op` is missing; no archive source to rewrite; use `--mode=green` or omit `--mode`").
 - `source.op` **exists** → `--mode=green` → BLOCKED ("`source.op` already exists; green-field scaffold would silently overwrite; use `--mode=redesign` for rewrite+port or `--mode=minor` for in-place edit").
 
-**First-op bias.** If no op in the same `family` has `status: implemented` and follows the canonical static_dims pattern (`docs/ops-design.md` § Step 3), set `auto-case = redesign` (skip the prompt). mode.json: `decided_by: "auto"`, `reason: "no canonical-pattern precedent in family <name>"`. User may override with `--mode=minor`.
+**First-op bias.** If no op in the same `family` has `status: implemented` and follows the canonical pattern (`docs/ops-design.md` § Step 3), set `auto-case = redesign` (skip the prompt). mode.json: `decided_by: "auto"`, `reason: "no canonical-pattern precedent in family <name>"`. User may override with `--mode=minor`.
 
 Write `.foundry/plan/<op_name>/mode.json`:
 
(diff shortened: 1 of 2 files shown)


.claude/skills/implement-op/SKILL.md
```diff
@@ -35,9 +35,16 @@ stateDiagram-v2
 
 ## Dual-path policy
 
-When rewriting a base class shared by sibling ops, it is acceptable to create a dual-path `__init__` — a runtime branch supporting both the legacy `(M, N)` interface and the new spec `(dim)` interface. This keeps unmigrated siblings' tests passing.
+Before refactoring a base class, count its subclasses:
 
-Dual-path is temporary debt. The orchestrator's cleanup gate removes it after all siblings are promoted or blocked. The implementer should NOT try to avoid dual-path by preemptively migrating siblings — that violates per-op scope.
+```bash
+grep -lE "class [A-Z][A-Za-z0-9]*\(<BaseName>\)" tileops/ops/
```
Ibuki-wind: The subclass-count command is not executable against this repo layout: `grep ... tileops/ops/` receives a directory without recursive mode, so it prints `grep: tileops/ops/: Is a directory` and finds no subclasses under nested family directories. → Replace it with a recursive command such as `rg -n "class [A-Z][A-Za-z0-9]*\([^)]*<BaseName>" tileops/ops` (or `grep -R`) so the one-subclass vs multiple-subclass branch is based on real matches.

Ibuki-wind (changes requested):
No trust-boundary split issue here; this is a docs-only skill-spec patch. The first-op heuristic direction is reasonable, but the new dual-path pre-check needs an executable repo-aware command before the skill can be relied on.

.claude/skills/implement-op/SKILL.md
```diff
@@ -35,7 +35,14 @@ stateDiagram-v2
 
 ## Dual-path policy
 
-When rewriting a base class shared by sibling ops, it is acceptable to create a dual-path `__init__` — a runtime branch supporting both the legacy `(M, N)` interface and the new spec `(dim)` interface. This keeps unmigrated siblings' tests passing.
+Dual-path applies **only when the base class has unmigrated sibling subclasses**. Before introducing dual-path, run:
+
+```bash
+grep -lE "class [A-Z][A-Za-z0-9]*\(<BaseName>\)" tileops/ops/
```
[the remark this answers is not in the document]
lcy-seso: Adopted (deduped with the same-area comment from the other reviewer). Added `-r` and whitespace tolerance: `grep -rlE "class\\s+[A-Z][A-Za-z0-9]*\\s*\\(\\s*<BaseName>\\s*\\)" tileops/ops/`. Verified against the repo (returns `tileops/ops/norm/rms_norm.py` for `RowNormOp`, expected matches for `Op`). See 960b0b9.

.claude/skills/implement-op/SKILL.md
lcy-seso: Adopted (deduped with the same-area comment from the other reviewer). Added `-r` and whitespace tolerance: `grep -rlE "class\\s+[A-Z][A-Za-z0-9]*\\s*\\(\\s*<BaseName>\\s*\\)" tileops/ops/`. Verified against the repo (returns `tileops/ops/norm/rms_norm.py` for `RowNormOp`, expected matches for `Op`). See 960b0b9.

.claude/skills/align-op/SKILL.md
```diff
@@ -102,6 +102,8 @@ Decide which case applies. Machine-decidable input: does `source.op` exist?
 - `source.op` **missing** → only `green` is valid. `--mode=minor` → BLOCKED ("`source.op` is missing; cannot edit a non-existent op file; use `--mode=green` or omit `--mode`"). `--mode=redesign` → BLOCKED ("`source.op` is missing; no archive source to rewrite; use `--mode=green` or omit `--mode`").
 - `source.op` **exists** → `--mode=green` → BLOCKED ("`source.op` already exists; green-field scaffold would silently overwrite; use `--mode=redesign` for rewrite+port or `--mode=minor` for in-place edit").
 
+**First-op bias.** If no op in the same `family` has `status: implemented` and follows the canonical static_dims pattern (`docs/ops-design.md` § Step 3), set `auto-case = redesign` (skip the prompt). mode.json: `decided_by: "auto"`, `reason: "no canonical-pattern precedent in family <name>"`. User may override with `--mode=minor`.
```
Ibuki-wind: `First-op bias` now makes `align-op` auto-select `redesign` for one class of existing op, but the Non-goals section still says auto-detecting `redesign vs minor` is out of scope and that the skill prompts or accepts `--mode`. → Update the Non-goals entry to carve out this first-op heuristic, or remove/rewrite that non-goal so the classification contract has a single source of truth.

Ibuki-wind (changes requested):
The recursive subclass-count command now works for the repo layout. One contract inconsistency remains in align-op: the new first-op auto-classification rule needs the stale Non-goals text reconciled so agents do not receive contradictory instructions.

.claude/skills/align-op/SKILL.md
lcy-seso: Adopted. Non-goals entry now carves out the first-op-bias exception explicitly: "General auto-detection of redesign vs minor" is still out of scope, but the one CLASSIFY heuristic added in this PR is named as the single exception. Single source of truth restored. See 61c6bad.

Ibuki-wind (dismissed):
Clean after the follow-up changes. The subclass-count command is now repo-aware, and the first-op auto-redesign exception is reconciled with the Non-goals section.

Ibuki-wind (approved):
Clean after the latest wording update. The first-op bias is now described without overfitting to one implementation detail, while the Non-goals exception and dual-path subclass check remain consistent and executable.

PR #1064, opened 2026-04-26
[Refactor][Scan] migrate cumsum + cumprod to canonical static_dims pattern

lcy-seso:
## Summary

- Add `CumulativeOp` L2 base under `tileops/ops/reduction/cumulative_base.py` (kw-only ctor binding `static_dims.N` + `signature.params.dim`, lazy kernel cache by `M`, dim-aware `_run` helper, `eval_roofline` via `_last_roofline_mn`, `_static_axes` bound at forward).
- Migrate `CumsumFwdOp` and `CumprodFwdOp` to thin subclasses of `CumulativeOp` (each declares only `_op_kind` + a 1-line `forward`); restore Args + Example docstrings on both per Google style.
- Restore Args + Example docstring on `RMSNormFwdOp` (omitted when it became a thin `RowNormOp` subclass in #1061).
- Update test (`tests/ops/test_cumulative.py`) and benchmark (`benchmarks/ops/bench_cumulative.py`) call-sites to the new kw-only API; add `test_cumsum_dim_axis1` + `test_cumprod_dim_axis1` covering the non-default-dim movedim path.
- Flip `CumsumFwdOp.status` and `CumprodFwdOp.status` from `spec-only` to `implemented` in `tileops/ops_manifest.yaml`.
- Extend Family-Base Protocol table in `docs/ops-design-reference.md`: scan family added to the `_op_kind` row.
- Add Google-style docstring rule to `.claude/rules/code-style.md`.

## Test plan

- [x] pre-commit passed
- [x] pytest passed (`pytest tests/ops/test_cumulative.py -m smoke` — 28/28 on H200; cumsum + cumprod, 1D–3D, fp16 / bf16 / fp32, including new dim=1 cases)
- [x] validator passed (`python scripts/validate_manifest.py --check-op CumsumFwdOp` and `--check-op CumprodFwdOp`)

## Structural Readiness

| Section | Status |
|---|---|
| Correctness & Safety | OK — `_validate_and_normalize_dim` checks dtype / device / shape / dim range; raises `ValueError`. |
| Kernel Structure | Unchanged — this PR is op-layer only; `CumulativeKernel` not modified. |
| Op Structure | OK — `_op_kind` is a class attribute on each subclass; forward owns pre/post-processing; delegates only `self._get_kernel(M)(x)` to the kernel. Status now `implemented`. |
| Benchmark | Pre-existing `benchmarks/ops/bench_cumulative.py` still works (call-site updated). Two warnings remain about ManifestBenchmark adoption — pre-existing, unrelated to this PR. |
| Delivery | Unit tests for default dim and dim=1; cumprod tolerance applied per dtype. Status flip blocked on validator + pytest, both pass. |

## Regression

- Default `dim=-1` path is identical in behavior to the legacy `(M, N)` API: same `reshape(-1, N)`, same kernel, same `N_padded` masking. Existing 24 unit tests (all default-dim) pass unchanged after migration.
- Non-default `dim` path is new functionality (was rejected by the legacy ctor accepting only `M, N`). Covered by 4 new H200 tests on 3-D `(2, 512, 256)` fp16 / bf16.

### DIFF
--- a/tests/ops/test_cumulative.py
+++ b/tests/ops/test_cumulative.py
@@ -146,7 +146,7 @@ def test_cumsum_op(m: int, n: int, dtype: torch.dtype) -> None:
     from tileops.ops.reduction.cumsum import CumsumFwdOp
 
     test = CumulativeTest(m, n, dtype, "cumsum")
-    op = CumsumFwdOp(M=m, N=n, dtype=dtype)
+    op = CumsumFwdOp(N=n, dtype=dtype)
     test.check(op, *test.gen_inputs(), **_tol(dtype))
 
 
@@ -156,7 +156,7 @@ def test_cumsum_non_contiguous(m: int, n: int, dtype: torch.dtype) -> None:
 
     x_full = torch.randn(m, n * 2, dtype=dtype, device="cuda")
     x = x_full[:, :n]
-    op = CumsumFwdOp(M=m, N=n, dtype=dtype)
+    op = CumsumFwdOp(N=n, dtype=dtype)
     ref = x.contiguous().float().cumsum(dim=-1).to(dtype)
     y = op(x)
     tol = _tol(dtype)
@@ -168,8 +168,7 @@ def test_cumsum_3d(batch: int, seq: int, hidden: int, dtype: torch.dtype) -> Non
     from tileops.ops.reduction.cumsum import CumsumFwdOp
 
     x = torch.randn(batch, seq, hidden, dtype=dtype, device="cuda")
-    M = batch * seq
-    op = CumsumFwdOp(M=M, N=hidden, dtype=dtype)
+    op = CumsumFwdOp(N=hidden, dtype=dtype)
     ref = x.float().cumsum(dim=-1).to(dtype)
     y = op(x)
     tol = _tol(dtype)
@@ -181,8 +180,7 @@ def test_cumsum_4d(b0: int, b1: int, b2: int, n: int, dtype: torch.dtype) -> Non
     from tileops.ops.reduction.cumsum import CumsumFwdOp
 
     x = torch.randn(b0, b1, b2, n, dtype=dtype, device="cuda")
-    M = b0 * b1 * b2
-    op = CumsumFwdOp(M=M, N=n, dtype=dtype)
+    op = CumsumFwdOp(N=n, dtype=dtype)
     ref = x.float().cumsum(dim=-1).to(dtype)
     y = op(x)
     tol = _tol(dtype)
@@ -194,7 +192,7 @@ def test_cumsum_1d(n: int, dtype: torch.dtype) -> None:
     from tileops.ops.reduction.cumsum import CumsumFwdOp
 
     x = torch.randn(n, dtype=dtype, device="cuda")
-    op = CumsumFwdOp(M=1, N=n, dtype=dtype)
+    op = CumsumFwdOp(N=n, dtype=dtype)
     ref = x.float().cumsum(dim=-1).to(dtype)
     y = op(x)
     tol = _tol(dtype)
@@ -211,7 +209,7 @@ def test_cumprod_op(m: int, n: int, dtype: torch.dtype) -> None:
     from tileops.ops.reduction.cumprod import CumprodFwdOp
 
     test = CumulativeTest(m, n, dtype, "cumprod", use_small_range=True)
-    op = CumprodFwdOp(M=m, N=n, dtype=dtype)
+    op = CumprodFwdOp(N=n, dtype=dtype)
     test.check(op, *test.gen_inputs(), **_cumprod_tol(dtype))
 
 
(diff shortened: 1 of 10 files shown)


tileops/ops/reduction/cumulative_base.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def _run(self, x: torch.Tensor) -> torch.Tensor:
+        """Shared forward implementation. Subclasses call this from `forward`."""
+        ndim = x.ndim
+        dim_norm = self._validate_and_normalize_dim(x)
+
+        if dim_norm != ndim - 1:
+            x = x.movedim(dim_norm, -1)
+        post_move_shape = tuple(x.shape)
+        x = x.contiguous().reshape(-1, self.N)
```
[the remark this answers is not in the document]
lcy-seso: Valid point. Out of scope for this PR — making the kernel stride-aware is a kernel-layer change (this PR is op-layer canonical-pattern migration). The same pattern exists in PR #1061's RowNormOp and is universal across all canonical T1 ops migrating off (M, N) ctor. The cost is paid only when `dim != ndim - 1`; default `dim=-1` skips the `movedim` entirely. Tracking as a family-wide kernel-perf follow-up rather than per-op.

tileops/ops/reduction/cumulative_base.py
```diff
@@ -0,0 +1,137 @@
...
+from tileops.kernels.kernel_base import Kernel
+from tileops.kernels.reduction._primitives import DEFAULT_ALIGNMENT, align_up
+from tileops.kernels.reduction.cumulative import CumulativeKernel
+
+from ..op_base import Op
+
+__all__ = ["CumulativeOp"]
+
+
+class CumulativeOp(Op):
```
Ibuki-wind, line 26: `CumulativeOp` introduces a new intermediate L2 base class and the `_op_kind` subclass protocol, but this PR does not update `docs/ops-design.md`; the ops-design domain rule requires docs updates when adding an intermediate base class or class-variable protocol. → Add the cumulative family-base pattern to the ops-design docs, including the constructor contract, `_op_kind`, lazy `(M, N)` kernel cache, and forward-time `_static_axes` binding.

tests/ops/test_cumulative.py
```diff
@@ -168,8 +168,7 @@ def test_cumsum_3d(batch: int, seq: int, hidden: int, dtype: torch.dtype) -> Non
     from tileops.ops.reduction.cumsum import CumsumFwdOp
 
     x = torch.randn(batch, seq, hidden, dtype=dtype, device="cuda")
-    M = batch * seq
-    op = CumsumFwdOp(M=M, N=hidden, dtype=dtype)
+    op = CumsumFwdOp(N=hidden, dtype=dtype)
```
Ibuki-wind, line 171: All updated test call sites still use the default `dim=-1`, so the new `movedim`/restore path and param-dependent static-axis validation for non-last axes can regress while this suite passes. → Add focused non-last-axis coverage for the new API, for example 3D `dim=1` cases for both `CumsumFwdOp(N=x.shape[1], dtype=dtype, dim=1)` and `CumprodFwdOp(..., dim=1)` with references using `torch.cumsum(..., dim=1)` / `torch.cumprod(..., dim=1)`.

Ibuki-wind (changes requested):
This PR crosses Implementation, Test, and Benchmark stages in one change; per the trust model that should be split unless there is an explicit human exception. I also found two local blockers: the new cumulative L2 base/protocol needs design documentation, and the newly exposed non-last-axis path needs coverage. Targeted manifest validation on the PR head passes, with only existing staged-rollout warnings.

tileops/ops/reduction/cumulative_base.py
lcy-seso: Adopted. Added a 'Canonical L2 family-base shape' subsection under Family-Base Refactoring in docs/ops-design.md listing both RowNormOp and CumulativeOp with the shared skeleton (kw-only ctor, lazy kernel cache by M, `_run` helper, dim normalization + `_static_axes` binding, `eval_roofline` reading `_last_roofline_mn`). Also extended `_op_kind` in the Family-Base Protocol table in docs/ops-design-reference.md to cover scan. See 34d38f7.

tests/ops/test_cumulative.py
lcy-seso: Adopted. Added `test_cumsum_dim_axis1` and `test_cumprod_dim_axis1` — 3D `(batch, hidden, seq)` with `dim=1`, references using `torch.cumsum(..., dim=1)` and `torch.cumprod(..., dim=1)`. Exercises the movedim choreography in `_run` and the param-dependent `_static_axes` binding for non-last axes. 4/4 pass on H200. See 34d38f7.

Ibuki-wind (dismissed):
Clean after the follow-up changes. Treating the implementation/test/benchmark span as explicitly human-approved, the remaining blockers are resolved: the new CumulativeOp L2 base/protocol is documented, non-last-axis coverage exists for both scan ops, and targeted manifest validation passes with only staged-rollout warnings.

Ibuki-wind (approved):

PR #1066, opened 2026-04-27
[Refactor][Skills] redesign add-manifest input + redivide labour with fix-manifest

lcy-seso:
## Summary

- `add-manifest`: input is `op_name` + `ref_url`. Idempotent, one-entry-per-invocation. Class-based source lookup. Required fields BLOCK with `evidence_needed` instead of falling through.
- `fix-manifest`: scope reduced to `source.kernel_map` + `signature.static_dims` (on-disk-derivable only). Reference-derivable fields → `add-manifest`.
- `docs/tileops-skills.md`: trust-model redivided by source-of-truth.

## Test plan

- [x] `pre-commit` passes (mdformat, codespell, gitleaks).
- [x] Skill files have valid frontmatter (`name`, `description`).
- [x] Cross-links in `docs/tileops-skills.md` resolve (`../.claude/skills/<name>/SKILL.md`).
- [ ] Smoke test: re-align one normalization op via `/add-manifest <op_name> <pytorch_url>` in a follow-up PR (first real exercise of the new contract).
- [ ] Smoke test: invoke `/fix-manifest <op_name> --field=kernel_map` on an op missing only `kernel_map` (verifies the narrowed scope still works for the most common gap).


.claude/skills/add-manifest/SKILL.md
```diff
@@ -1,29 +1,59 @@
...
 stateDiagram-v2
     [*] --> VALIDATE_INPUT
     VALIDATE_INPUT --> [*]: URL malformed → abort
-    VALIDATE_INPUT --> RESOLVE_SOURCES
+    VALIDATE_INPUT --> CHECK_EXISTING
+    CHECK_EXISTING --> READ_EXISTING: entry exists & --regenerate-existing
+    CHECK_EXISTING --> [*]: entry exists & no flag → BLOCKED
+    CHECK_EXISTING --> RESOLVE_SOURCES: greenfield
+    READ_EXISTING --> RESOLVE_SOURCES: human-curated fields snapshot saved
     RESOLVE_SOURCES --> READ_PYTORCH
```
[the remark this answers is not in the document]
lcy-seso: Stale — this comment was on an earlier commit. Current workflow (after 2e35723): `READ_EXISTING --> READ_REFERENCE` for entry-present, `READ_EXISTING --> RESOLVE_SOURCES` only for entry-absent (where source paths must still be resolved before reaching the reference). The two-edge pattern is correct.

.claude/skills/add-manifest/SKILL.md
```diff
@@ -42,9 +72,30 @@ stateDiagram-v2
 
 `torch_api` must match `^https://(docs\.)?pytorch\.org/docs/stable/generated/.*\.html$`. Else abort.
 
-### 2. RESOLVE_SOURCES
+### 2. CHECK_EXISTING
+
+Look up `<op_name>` in `tileops/ops_manifest.yaml` (op_name derived from `op_path` + variant suffix in Step 4):
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Fixed cross-reference: `READ_EXISTING` now references Step 5 `SPLIT_VARIANTS` for the variant-suffix derivation. See a56f66a.

docs/tileops-skills.md
```diff
@@ -134,13 +141,13 @@ align-op <op_name>                       ← per-op orchestrator
...
+| `ops_manifest.yaml`               | Three writers, disjoint write-scopes: `add-manifest` creates new entries (greenfield) or rewrites an existing entry's auto-derivable fields under `--regenerate-existing` (`signature.*`, `shape_rules`, `roofline` for well-known ops; preserves `workloads`, `source.*`, `status`, `parity_opt_out`); `fix-manifest` patches one structural field on an existing entry (`kernel_map`, `static_dims`, `shape_rules`, `roofline.vars`, `dtype_combos`; never `signature.{inputs,outputs,params}` or `status`); `align-op` at `FLIP_STATUS` writes only the `status` field. No other skill writes the manifest. |
```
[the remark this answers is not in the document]
lcy-seso: Stale — this comment was on an earlier commit. Current trust-model row in docs/tileops-skills.md (after 06b8595) already lists `family` and `ref_api` in the preserved set: `preserves \`workloads\`, \`source.*\`, \`status\`, \`parity_opt_out\`, \`family\`, \`ref_api\` verbatim if entry already exists`.

.claude/skills/add-manifest/SKILL.md
```diff
@@ -40,117 +53,81 @@ stateDiagram-v2
...
-Locate from `op_path`:
+### 2. READ_EXISTING
 
-| Source | Path                                                  |
-| ------ | ----------------------------------------------------- |
-| kernel | search under `tileops/kernels/` for matching basename |
-| op     | `op_path`                                             |
-| test   | `tests/ops/test_<name>.py`                            |
-| bench  | `benchmarks/ops/bench_<name>.py`                      |
+Look up `<op_name>` in `tileops/ops_manifest.yaml` (op_name derived from `op_path` + variant suffix from Step 5 SPLIT_VARIANTS). If present, snapshot the human-curated fields listed in the Contract table. If absent, proceed to RESOLVE_SOURCES.
```
Ibuki-wind: `READ_EXISTING` runs before `READ_REFERENCE`/`SPLIT_VARIANTS`, but this lookup depends on the Step 5 variant suffix. For an existing optional-input op such as `Conv1dFwdOp`/`Conv1dBiasFwdOp`, the agent cannot know the variant keys yet, so it can preserve only the primary entry and default or overwrite the variant's human-curated fields, violating the idempotent contract. → Move reference/variant discovery before preservation, or snapshot all existing entries matching the primary key/`variant_of`/`source.op` before splitting and then preserve curated fields per emitted entry.

Ibuki-wind (changes requested):
Requesting changes for one workflow-ordering bug in the idempotent re-alignment path. This PR only changes skill/docs text, so I do not see a code-stage trust-boundary violation. The overall refactor direction is fine, but existing variant entries must be discoverable before curated fields are preserved.

.claude/skills/add-manifest/SKILL.md
lcy-seso: Adopted. READ_EXISTING now pre-scans for **every** manifest entry whose `source.op == op_path` (variants share `source.op` with their primary, so this captures both in one pass — no dependency on Step 5's not-yet-known suffix). The result is a map `entry_key → snapshot`; DRAFT_ENTRY consults it per emitted key (Step 5 SPLIT_VARIANTS may emit primary + variant). Found in map → preserve verbatim; not found → default. This fixes the variant-ops idempotent-contract violation. See 69fbdeb.

.claude/skills/add-manifest/SKILL.md
```diff
@@ -40,117 +53,87 @@ stateDiagram-v2
...
 ### 1. VALIDATE_INPUT
 
-`torch_api` must match `^https://(docs\.)?pytorch\.org/docs/stable/generated/.*\.html$`. Else abort.
+Reject `ref_url` not matching the regex.
 
-### 2. RESOLVE_SOURCES
+### 2. READ_EXISTING
 
-Locate from `op_path`:
+Scan `tileops/ops_manifest.yaml` for **every** entry whose `source.op == op_path`. Variants share `source.op` with their primary, so this captures the primary AND any variant entry in one pass without depending on Step 5's not-yet-known suffix.
```
Ibuki-wind: `source.op == op_path` is not a unique identity for one manifest op: existing files such as `tileops/ops/attention/mha.py`, `tileops/ops/reduction/reduce.py`, and `tileops/ops/convolution.py` back multiple manifest entries. With this rule, adding a new entry in one of those files produces a non-empty snapshot map from unrelated entries, skips `RESOLVE_SOURCES`, and later uses the populated-map PR title even though the target op is greenfield. → Treat this scan only as candidate snapshot collection; after `READ_REFERENCE`/`SPLIT_VARIANTS` yields the emitted keys, decide existence from matches for those keys (or `variant_of` those keys), and run greenfield source resolution when none of the emitted keys exists.

.claude/skills/add-manifest/SKILL.md
```diff
@@ -40,117 +53,87 @@ stateDiagram-v2
...
 | variant | `<Op><Suffix>FwdOp` | required + optional   | `variant_of: <Op>FwdOp` |
 
-`<Suffix>` = PascalCase of the optional input name (e.g., `Bias`). Variants share `source.kernel` and `source.op`. Each gets its own `signature`, `workloads`, `roofline`.
+`<Suffix>` = PascalCase of the optional input name. Variants share `source.kernel` and `source.op`; each gets its own `signature` / `workloads` / `roofline`. Multiple `Optional[Tensor]`: follow `docs/manifest.md` decision tree.
 
-Multiple `Optional[Tensor]`: follow decision tree in `docs/manifest.md`.
+### 6. DRAFT_ENTRY
 
-### 5. DRAFT_ENTRY
+For each entry emitted in Step 5 (primary, plus variant if any), look up the entry's key in the READ_EXISTING snapshot map. Found → use snapshot for the human-curated fields. Not found → use defaults from the Contract table (a partial-greenfield case is possible: e.g., primary already exists but variant is new).
```
Ibuki-wind: The documented partial-greenfield path says a missing emitted key should use Contract defaults, but the defaults for `family` and `source.*` come from `RESOLVE_SOURCES`, and Step 3 only runs when the entire snapshot map is empty. If a primary exists and a new optional-input variant is emitted, the variant has no defined source/family defaults. → Specify the fallback for missing emitted keys explicitly, e.g. run `RESOLVE_SOURCES` for those keys or copy the primary snapshot's `family`/shared `source.*` while defaulting only per-entry curated fields such as `workloads` and `status`.

Ibuki-wind (changes requested):
Requesting changes for remaining idempotency/existence-detection bugs. The trust-boundary scope is still limited to skills/docs, but the current workflow can classify the wrong operation as existing and leaves partial-greenfield emitted entries without resolvable curated defaults.

.claude/skills/add-manifest/SKILL.md
lcy-seso: Adopted. The READ_EXISTING scan is now explicit about being just a **candidate pool** (`source.op == op_path` is not a unique identity — e.g., `tileops/ops/attention/mha.py` backs several entries). The greenfield/regenerate decision is deferred to DRAFT_ENTRY (per-key) and CREATE_PR (`≥1 emitted key matches the candidate map` → re-align; `0 matches` → greenfield). True greenfield additions to a shared-file family will now correctly use the Add title even though the candidate map is non-empty. See 9673186.

.claude/skills/add-manifest/SKILL.md
lcy-seso: Adopted. Partial-greenfield case is now explicit. RESOLVE_SOURCES runs unconditionally; DRAFT_ENTRY uses a 3-tier priority per emitted key:

1. Key in candidate map → preserve snapshot verbatim.
2. Key absent from candidate map but map non-empty → inherit shared fields (`family`, `source.kernel`, `source.op`, `source.kernel_map`, `source.bench_manifest_driven`) from any candidate; per-entry fields (`workloads`, `parity_opt_out`, `status`, `source.test`, `source.bench`) take Contract defaults.
3. Map empty → full greenfield; all defaults from RESOLVE_SOURCES + Contract.

This matches your suggestion. See 9673186.

.claude/skills/add-manifest/SKILL.md
```diff
@@ -40,117 +52,91 @@ stateDiagram-v2
...
 
-Multiple `Optional[Tensor]`: follow decision tree in `docs/manifest.md`.
+### 6. DRAFT_ENTRY
 
-### 5. DRAFT_ENTRY
+For each entry emitted in Step 5 (primary, plus variant if any), resolve the human-curated fields in this priority order:
 
-Per entry:
+1. **Emitted key in candidate map** (re-align): use that snapshot verbatim.
+1. **Emitted key NOT in candidate map, but candidate map non-empty** (partial-greenfield — e.g., primary exists, new variant emitted): inherit shared fields from any candidate snapshot — `family`, `source.kernel`, `source.op`, `source.kernel_map`, `source.bench_manifest_driven` (variants share these per Step 5). Per-entry fields (`workloads`, `parity_opt_out`, `status`, `source.test`, `source.bench`) take Contract defaults.
```
Ibuki-wind: This fallback still treats any non-empty `source.op == op_path` candidate map as a partial-greenfield variant set and inherits `source.kernel`/`source.kernel_map` from an arbitrary candidate. That is unsafe for real multi-entry files such as `tileops/ops/attention/mha.py`, where the same op file has fwd, bwd, decode, and paged entries with different kernels and `kernel_map`s; adding a new emitted key there could copy the wrong kernel identity just because another entry shares `source.op`. → Only inherit shared fields from a matched emitted primary/`variant_of` group; if no emitted key or emitted primary matches, treat the missing key as full greenfield and use `RESOLVE_SOURCES`/BLOCKED rather than `any candidate snapshot`.

Ibuki-wind (changes requested):
Requesting changes for one remaining fallback bug in the partial-greenfield path. The earlier existence-classification issue is mostly addressed by deferring the PR mode to emitted-key matches, and the PR still stays within skills/docs scope. The remaining problem is that missing emitted entries can inherit source identity from unrelated entries that merely share the same op file.

.claude/skills/add-manifest/SKILL.md
lcy-seso: Adopted. Inheritance is now anchored to the emitted primary's group only — never to arbitrary candidates that share `source.op`. DRAFT_ENTRY:

- Emitted PRIMARY K_p: in candidate map → preserve; else full greenfield.
- Emitted VARIANT K_v with `variant_of: K_p`: K_v in map → preserve; else if K_p in map → inherit shared fields from K_p's snapshot (`family`, `source.{kernel,op,kernel_map,bench_manifest_driven}`); else full greenfield.

Files like `mha.py` (multi-entry, different kernels per entry) are now safe — a new emitted entry there hits 'full greenfield' since neither K_p nor K_v matches the existing fwd/bwd/decode/paged candidates. CREATE_PR title selector also keyed on K_p match. See 9774a1a.

Ibuki-wind (dismissed):
Approved. The latest revision anchors preservation and partial-greenfield inheritance to the emitted primary/variant group instead of arbitrary entries sharing `source.op`, which resolves the remaining idempotency and multi-entry-file concerns. Scope remains limited to skills/docs; no trust-boundary split is needed.

.claude/skills/add-manifest/SKILL.md
```diff
@@ -40,117 +56,80 @@ stateDiagram-v2
 
 ### 1. VALIDATE_INPUT
 
-`torch_api` must match `^https://(docs\.)?pytorch\.org/docs/stable/generated/.*\.html$`. Else abort.
+- Reject `ref_url` not matching the regex.
+- Reject `op_name` not matching `^[A-Z][A-Za-z0-9]+(Fwd|Bwd)Op$`.
+- **PyTorch-only sanity check** (skip otherwise). If `ref_url` host is `(docs.)?pytorch.org`: extract the last `.<word>.html` segment from the URL → `<ref_name>`; compute `<expected> = snake_case(op_name without Fwd/Bwd Op suffix)`. Require `<expected> == <ref_name>` OR `<expected>.startswith(<ref_name> + "_")` (the second branch covers variants whose op_name extends the primary's). Mismatch → BLOCKED with both names. For non-PyTorch URLs, no automatic check — caller is responsible.
```
Ibuki-wind: The PyTorch sanity check assumes the manifest key's snake_case stem must equal or extend the PyTorch function name, but valid existing entries do not always follow that naming. For example `MultiHeadAttentionFwdOp` uses `torch.nn.functional.scaled_dot_product_attention`; this check would block re-aligning that existing entry even though the manifest pairing is intentional. → Make this check non-blocking, allow an explicit alias/override path, or only apply it when the existing entry has no `ref_api`/known pairing evidence.

.claude/skills/add-manifest/SKILL.md
```diff
@@ -40,117 +56,80 @@ stateDiagram-v2
...
+Derive source paths from `op_name`. The op file MUST exist on disk; if it doesn't, the caller is expected to scaffold it first.
 
-1. Find any existing entry in `tileops/ops_manifest.yaml` whose `source.kernel` matches (same path, parent dir, or basename).
-1. Copy that entry's `family` value verbatim.
-1. If nothing matches, scan all `family` values used in the manifest and pick the closest by domain. Still ambiguous → STOP, ask user.
+1. Convert `op_name` to snake_case `<name>`. Strip the trailing `FwdOp` / `BwdOp` first; PascalCase → snake_case via standard rules; uppercase abbreviations stay one cluster (e.g., `RMSNormFwdOp` → `rms_norm`, `Conv1dBiasFwdOp` → `conv1d_bias`).
+1. Search `tileops/ops/**/<name>.py`. Exactly one match → `source.op` is that path. Zero matches → BLOCKED `"op file <name>.py not found under tileops/ops/; scaffold the op file first or rename per convention"`. Multiple matches → BLOCKED with the list.
```
Ibuki-wind: Greenfield source resolution can only find `tileops/ops/**/<snake(op_name)>.py`, but TileOPs has valid manifest patterns where many logical entries share a different implementation file: `SumFwdOp`/`MeanFwdOp` use `tileops/ops/reduction/reduce.py`, and `Conv1dFwdOp`/`Conv1dBiasFwdOp` use `tileops/ops/convolution.py`. With `op_path` removed, a new entry in one of these groups cannot resolve its real `source.op` and will block or force a non-local rename. → Add an explicit source-path argument/override for greenfield entries, or derive from a matched sibling/family mapping instead of requiring an exact `<name>.py` file.

Ibuki-wind (changes requested):
Requesting changes for two regressions introduced by the new op_name-only model. The one-entry-per-invocation direction is reasonable, but the current validation/source-resolution rules are stricter than TileOPs' existing manifest conventions and would block valid re-align/add flows.

.claude/skills/add-manifest/SKILL.md
Ibuki-wind: Do not solve the PyTorch-name mismatch by renaming existing TileOps manifest keys/classes to external reference API names. `op_name` is TileOps' local manifest identity; `ref_url` is only the semantic reference. For an existing entry, prefer the existing manifest key and existing `ref_api`/entry pairing as evidence, so a key like `MultiHeadAttentionFwdOp` may legitimately re-align against `scaled_dot_product_attention`. → Replace this hard stem-equality BLOCK with a non-blocking sanity signal for existing entries, or block only when the existing entry's recorded reference clearly conflicts with the requested `ref_url` and the caller has not explicitly requested reference replacement. For absent entries, the check may ask for explicit alias rationale, but it must not force existing TileOps names to match PyTorch stems.

.claude/skills/add-manifest/SKILL.md
Ibuki-wind: `RESOLVE_SOURCES` should not start from `<snake(op_name)>.py`; that convention is only valid when no implementation exists yet. The robust rule is: first scan `tileops/ops/**` for a Python class named exactly `op_name` (AST/class lookup, not filename matching). Exactly one class match → use that file as `source.op`, which preserves existing shared-file patterns such as `SumFwdOp` in `reduction/reduce.py`, `Conv1dFwdOp` in `convolution.py`, and attention variants in `attention/mha.py`. Zero class matches → true greenfield, use the `<snake(op_name without Fwd/BwdOp)>.py` convention as the default future `source.op`. Multiple class matches → BLOCKED and ask for disambiguation. → Rewrite this step around class lookup first, filename convention second.

Ibuki-wind (changes requested):
Please address the add-manifest model with the rules in the inline comments. The intended fix is not another ad-hoc allowlist: existing TileOps identities and existing op classes should win; naming conventions are only the greenfield fallback.

.claude/skills/add-manifest/SKILL.md
lcy-seso: Adopted (deduped with the follow-up comment ujlR). Removed the PyTorch sanity check entirely. Existing TileOPs identity wins; `MultiHeadAttentionFwdOp` ↔ `scaled_dot_product_attention` and similar legitimate pairings now go through. `Caller responsibility` clause in Contract is the operative guard. See a550dad.

.claude/skills/add-manifest/SKILL.md
lcy-seso: Adopted (deduped with the follow-up comment ujlS). RESOLVE_SOURCES now finds the implementation file by Python class lookup (`grep -rlE "^class <op_name>\\(" tileops/ops/` or AST), not by filename. Shared-file patterns (SumFwdOp/MeanFwdOp → reduce.py; Conv1dFwdOp/Conv1dBiasFwdOp → convolution.py; MHA family → mha.py) work correctly. Zero class matches → true greenfield, default <snake>.py. Multiple → BLOCKED with disambiguation request. See a550dad.

.claude/skills/add-manifest/SKILL.md
lcy-seso: Adopted. Hard PyTorch-stem equality removed entirely. TileOPs identity (op_name + ref_api pairing in the existing entry) is authoritative; reference URL is just the semantic source. No alias mechanism added — per the `Caller responsibility` clause, caller owns the pairing. See a550dad.

.claude/skills/add-manifest/SKILL.md
lcy-seso: Adopted. Class-name lookup (`grep -rlE "^class <op_name>\\(" tileops/ops/`) is now the primary rule; <snake>.py filename convention is the greenfield-only fallback when zero classes match. Existing shared-file patterns (reduce.py, convolution.py, mha.py) preserved. Multiple class matches → BLOCKED disambiguation. See a550dad.

.claude/skills/add-manifest/SKILL.md
```diff
@@ -40,117 +56,83 @@ stateDiagram-v2
...
 
-### 3. READ_PYTORCH
+1. **Find the implementation file by class name.** Scan `tileops/ops/**/*.py` for a Python `class <op_name>(...)` declaration (AST or `grep -rlE "^class <op_name>\(" tileops/ops/`).
+   - Exactly one match → `source.op` is that file path.
+   - Zero matches → true greenfield. Default `source.op` to `tileops/ops/<snake_name>.py` (or under a family subdirectory if a sibling manifest entry has the same family — copy that entry's `source.op` parent dir). `<snake_name>` is `op_name` minus the trailing `FwdOp` / `BwdOp`, snake_cased (e.g., `RMSNormFwdOp` → `rms_norm`, `Conv1dBiasFwdOp` → `conv1d_bias`). The file may not exist yet — that is fine; the caller scaffolds afterward.
+   - Multiple matches → BLOCKED with the list of files; ask the caller to disambiguate.
+1. **Resolve `source.kernel`.** Record absent — the kernel-class identity isn't reliably derivable from `op_name` alone. `fix-manifest --field=kernel_map` fills it later from the op file's `_kernel_key` / `default_kernel_map` once the op exists.
```
Ibuki-wind: `source.kernel` cannot be recorded absent and still let the workflow proceed: `scripts/validate_manifest.py` L0 requires `source.kernel`, and `fix-manifest --field=kernel_map` cannot fill `source.kernel` later because that skill is explicitly forbidden from writing `source.*` other than `source.kernel_map`. This makes every greenfield entry without a resolved kernel path fail before the declared validate/audit/PR success path. → Keep `source.kernel` resolution in `add-manifest`: derive it from the existing op class/imports/default kernel evidence when possible, use the existing convention of `source.kernel: source.op` for ops without a separate kernel file, or BLOCKED with `evidence_needed: source.kernel for <op_name>` when no valid path can be established.

.claude/skills/add-manifest/SKILL.md
```diff
@@ -40,117 +56,83 @@ stateDiagram-v2
...
+1. `source.test = tests/ops/test_<snake_name>.py`; `source.bench = benchmarks/ops/bench_<snake_name>.py`. Missing files: record absent, continue.
+1. `family`: copy verbatim from a sibling manifest entry whose `source.op` parent-dir or basename overlaps. If no sibling matches, leave `family` empty for human review. Never invent.
```
Ibuki-wind: Leaving `family` empty for human review conflicts with the rest of this skill: `family` is a required manifest field and Step 7 invokes `audit-family` for the op's family. If no sibling can establish the closed-vocabulary family, the skill does not have enough information to produce a valid L0-passing manifest PR. → Treat unresolved family as BLOCKED (`evidence_needed: family for <op_name>`) or require an explicit caller-provided family override; do not continue with an empty family.

Ibuki-wind (changes requested):
Requesting changes for the remaining greenfield-resolution inconsistency. The two discussed fixes are implemented correctly: existing TileOps names are no longer forced to match PyTorch stems, and `source.op` resolution now uses class lookup before the filename convention. The remaining issue is that greenfield generation can proceed with required manifest/audit fields unresolved.

.claude/skills/add-manifest/SKILL.md
lcy-seso: Adopted. `source.kernel` is now resolved with 3 tiers: (1) parse op-file imports for a Kernel subclass, then apply the same class-lookup rule under `tileops/kernels/`; (2) op has no separate kernel file (kernel-less op convention) → `source.kernel = source.op`; (3) none of the above → BLOCKED `evidence_needed: source.kernel for <op_name>`. Skill no longer leaves it absent. See 3e62c07.

.claude/skills/add-manifest/SKILL.md
lcy-seso: Adopted. Unresolved `family` now BLOCKS with `evidence_needed: family for <op_name>` instead of falling through with an empty value. Sibling-copy remains the primary rule; if no matching sibling, caller must declare the family explicitly. See 3e62c07.

Ibuki-wind (approved):
Approved. The latest revision closes the remaining greenfield-resolution gap: `source.kernel` and `family` are now required to resolve or explicitly BLOCKED, so the generated entry can reach the validate/audit/PR success path. The earlier naming and source-op issues are also addressed by keeping TileOps manifest identity local and using class lookup before filename conventions. Scope remains skills/docs only.

PR #1081, opened 2026-04-27
[Maintain][Manifest] add elementwise multi-input entries

### LINKED ISSUE #1077: [Maintain][Manifest] add elementwise multi-input entries to ops_manifest.yaml
### Symptom / Motivation

Sub-issue D of the elementwise family spec creation: **4 multi-input ops**. Sibling issues cover unary direct (#1073) / unary functional / binary.

### Root Cause Analysis

N/A — family-wide spec creation.

### Related Files

- `tileops/ops_manifest.yaml`
- `tileops/ops/elementwise.py`

## Goal

Bring all 4 multi-input elementwise ops under manifest tracking at `status: spec-only`. Single batch PR.

## Plan

**Plan type: fixed**

### Multi-input

| op_name | ref_api | ref_url |
|---|---|---|
| WhereFwdOp | torch.where | https://docs.pytorch.org/docs/stable/generated/torch.where.html |
| ClampFwdOp | torch.clamp | https://docs.pytorch.org/docs/stable/generated/torch.clamp.html |
| MaskedFillFwdOp | torch.Tensor.masked_fill | https://docs.pytorch.org/docs/stable/generated/torch.Tensor.masked_fill.html |
| NanToNumFwdOp | torch.nan_to_num | https://docs.pytorch.org/docs/stable/generated/torch.nan_to_num.html |

## Constraints

- Single batch PR for these 4 ops. Title `[Maintain][Manifest] add elementwise multi-input entries`.
- Each op invocation is `/add-manifest <op_name> <ref_url>`. Skill is idempotent and one-entry-per-invocation; loop in the agent driver, not in the skill. The driver MUST stay on a single branch and accumulate edits — do not let the skill open per-op PRs.
- Do NOT touch op / kernel / test / bench code. Manifest-only PR.
- All entries land at `status: spec-only`.

## Acceptance Criteria

- [ ] AC-1: The resulting PR contains exactly 4 new manifest entries (one per row).
- [ ] AC-2: Validator (`python scripts/validate_manifest.py --check-op <op_name>`) passes L0 on every added op.
- [ ] AC-3: For any op where `/add-manifest` returns BLOCKED, the agent records the BLOCKED case in the PR body and excludes that op from the batch.

lcy-seso:
## Summary

Bring all 4 multi-input elementwise ops under manifest tracking at `status: spec-only` in a single batch PR:

- `WhereFwdOp` -> `torch.where`
- `ClampFwdOp` -> `torch.clamp` (scalar min/max variant)
- `MaskedFillFwdOp` -> `torch.Tensor.masked_fill` (scalar value variant)
- `NanToNumFwdOp` -> `torch.nan_to_num`

Manifest-only PR; no op / kernel / test / bench code is modified. All entries land at `status: spec-only`. Code conformance to the spec is intentionally deferred to follow-up PRs per the manifest trust model.

### Scope notes

- `ClampFwdOp` / `MaskedFillFwdOp` are tracked as the scalar variants matching the implemented kernels; the Tensor `min`/`max` and 0-dim Tensor `value` forms remain BLOCKED at the spec level and are not in scope for this batch (documented inline in manifest comments).
- `WhereFwdOp` omits PyTorch broadcasting (shape_rules require equal shapes), consistent with the existing kernel scope.

Closes #1077

## Test plan

- [x] 4 new manifest entries added; diff scoped to `tileops/ops_manifest.yaml`.
- [x] `scripts/validate_manifest.py --levels schema` passes for each new op.
- [x] BLOCKED variants documented inline; not in this batch.

## Follow-up

Issues:
- #1088 — track BLOCKED PyTorch-API variants for elementwise multi-input ops

### DIFF
--- a/tileops/ops_manifest.yaml
+++ b/tileops/ops_manifest.yaml
@@ -3703,6 +3703,14 @@ ops:
   WhereFwdOp:
     ref_api: "torch.where"
     family: elementwise
+    # status: spec-only — current tileops/ops/elementwise.py:WhereFwdOp does
+    # not conform to this spec:
+    #   - __init__ signature is (N_total, dtype) instead of the
+    #     PyTorch-aligned (condition, input, other) form.
+    #   - forward parameter names are (cond, x, y) instead of
+    #     (condition, input, other).
+    # Code conforms to this spec in a follow-up PR per the manifest trust
+    # model (.claude/rules/manifest-trust-model.md).
     status: spec-only
 
     signature:
@@ -3740,6 +3748,14 @@ ops:
     # Scalar-bound variant: PyTorch's torch.clamp also accepts Tensor min/max;
     # the manifest tracks the scalar (Number) slice that the kernel implements.
     family: elementwise
+    # status: spec-only — current tileops/ops/elementwise.py:ClampFwdOp does
+    # not conform to this spec:
+    #   - __init__ takes (N_total, dtype, min_val, max_val) instead of the
+    #     PyTorch-aligned (input, min, max) form.
+    #   - param names are (min_val, max_val) instead of (min, max).
+    #   - forward parameter is named `x` instead of `input`.
+    # Code conforms to this spec in a follow-up PR per the manifest trust
+    # model (.claude/rules/manifest-trust-model.md).
     status: spec-only
 
     signature:
@@ -3777,6 +3793,14 @@ ops:
     # value; the manifest tracks the scalar (Number) slice that the kernel
     # implements.
     family: elementwise
+    # status: spec-only — current tileops/ops/elementwise.py:MaskedFillFwdOp
+    # does not conform to this spec:
+    #   - __init__ takes (N_total, dtype, fill_value) instead of the
+    #     PyTorch-aligned (input, mask, value) form.
+    #   - param name is `fill_value` instead of `value`.
+    #   - forward parameter is named `x` instead of `input`.
+    # Code conforms to this spec in a follow-up PR per the manifest trust
+    # model (.claude/rules/manifest-trust-model.md).
     status: spec-only
 
     signature:
@@ -3813,6 +3837,17 @@ ops:
   NanToNumFwdOp:
     ref_api: "torch.nan_to_num"
     family: elementwise
+    # status: spec-only — current tileops/ops/elementwise.py:NanToNumFwdOp
+    # does not conform to this spec:
+    #   - __init__ takes (N_total, dtype, nan_val, posinf_val, neginf_val)
+    #     instead of the PyTorch-aligned (input, nan, posinf, neginf) form.
+    #   - param names are (nan_val, posinf_val, neginf_val) instead of
+    #     (nan, posinf, neginf).
+    #   - posinf/neginf default to 1e4/-1e4 instead of None (PyTorch's
+    #     default sentinel meaning "use dtype max/min").
+    #   - forward parameter is named `x` instead of `input`.
+    # Code conforms to this spec in a follow-up PR per the manifest trust
+    # model (.claude/rules/manifest-trust-model.md).
     status: spec-only
 
     signature:
--- a/tileops/ops_manifest.yaml
+++ b/tileops/ops_manifest.yaml
@@ -3695,3 +3695,152 @@ ops:
       test: tests/ops/test_activation.py
       bench: benchmarks/ops/bench_activation.py
       bench_manifest_driven: false
+
+  # ---------------------------------------------------------------------------
+  # elementwise — multi-input parametric ops (where, clamp, masked_fill, nan_to_num)
+  # ---------------------------------------------------------------------------
+
+  WhereFwdOp:
+    ref_api: "torch.where"
+    family: elementwise
+    status: spec-only
+
+    signature:
+      inputs:
+        condition: {dtype: "bool"}
+        input: {dtype: "float16 | bfloat16 | float32 | float8_e4m3fn | float8_e5m2"}
+        other: {dtype: "same_as(input)"}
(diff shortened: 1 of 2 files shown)


tileops/ops_manifest.yaml
```diff
@@ -2471,3 +2471,153 @@ ops:
...
+  WhereFwdOp:
+    ref_api: "torch.where"
+    family: elementwise
+    status: spec-only
+
+    signature:
+      inputs:
+        condition: {dtype: "bool"}
+        input: {dtype: "float16 | bfloat16 | float32 | float8_e4m3fn | float8_e5m2"}
+        other: {dtype: "same_as(input)"}
```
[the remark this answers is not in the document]
lcy-seso: Addressed in 643dc2a.

Evidence:
- Diagnosed bytes accounting issue is real (elem_bytes binds to bool condition = 1 byte → 3 bytes/elem instead of 3 × float_dtype_size).
- However, reordering inputs is not an option here: per project rule, the manifest signature must exactly match torch.where's public API, where `condition` is the first positional argument.
- Per docs/roofline.md §2.2, ops whose bytes depend on multiple input dtypes cannot be expressed in inline mode and must use func mode.
- Fix applied: switched WhereFwdOp roofline to func mode pointing at `tileops.perf.formulas.where_fwd_roofline` (currently a Tier-2 NotImplementedError stub, matching existing pattern). Concrete formula lands when the op moves from spec-only to implemented.

Current HEAD: 643dc2a

tileops/perf/formulas.py
```diff
@@ -154,3 +155,26 @@ def deepseek_dsa_decode_roofline(**kwargs: Any) -> dict[str, int]:
...
     """
     raise NotImplementedError
+
+
+# ---------------------------------------------------------------------------
+# Mixed-dtype elementwise (where)
+# ---------------------------------------------------------------------------
+
+
+def where_fwd_roofline(**kwargs: Any) -> dict[str, int]:
```
Ibuki-wind: `where_fwd_roofline` uses the legacy `**kwargs -> dict` func shape, but `docs/roofline.md` §4.4.2 says codegen emits `return <func>(self)` and `Op.eval_roofline()` returns `(flops, bytes)`, so this will TypeError once `WhereFwdOp` leaves `spec-only` → change the stub to accept the op instance and return `tuple[int, int]` (or add a wrapper at the dotted path that matches `func(self)`).

Ibuki-wind (changes requested):
### Overall
The new func-mode roofline stub does not match the callable contract that codegen emits, so `WhereFwdOp` would fail as soon as it moves from `spec-only` to `implemented`.
Update the stub to accept the Op instance and return the `(flops, bytes)` tuple shape, or add a matching wrapper.

### New issues
- `tileops/perf/formulas.py:165` — `where_fwd_roofline` uses the legacy `**kwargs -> dict` shape instead of the `func(self) -> tuple[int, int]` contract.

Ibuki-wind (approved):
Clean — no issues.

Ibuki-wind (changes requested):
### Overall
Spec-only is the right status for these entries, but the manifest needs to say why because the referenced implementations already exist and do not conform to the PyTorch-aligned signatures.
Next step: add explicit `status: spec-only` comments for each new elementwise entry documenting the current implementation/manifest mismatch.

### Cross-cutting concerns
- The four new elementwise manifest entries all point at existing non-conforming classes without the discrepancy comments required by `.claude/domain-rules/manifest-spec.md`.

### Advisory
- This PR touches both the manifest and `tileops/perf/formulas.py`; that crosses the manifest/perf boundary during migration, but the formula stub is only used to keep the manifest `func` path importable.

### New issues
- `tileops/ops_manifest.yaml:2482` — `WhereFwdOp` needs a spec-only comment for its current implementation signature mismatch.
- `tileops/ops_manifest.yaml:2519` — `ClampFwdOp` needs a spec-only comment for its current implementation-name mismatches.
- `tileops/ops_manifest.yaml:2556` — `MaskedFillFwdOp` needs a spec-only comment for its current implementation-name mismatches.
- `tileops/ops_manifest.yaml:2592` — `NanToNumFwdOp` needs a spec-only comment for its current implementation-name/default mismatches.

tileops/perf/formulas.py
lcy-seso: Fixed in 48b2a0c: added a `where_fwd_roofline(op) -> tuple[int, int]` stub in `tileops/perf/formulas.py` matching the func-mode contract from `docs/roofline.md` §4.4.2 (`return <func>(self)` with `(flops, bytes)` return shape). Body still raises `NotImplementedError` — fills in once `WhereFwdOp` flips from spec-only to implemented.

Ibuki-wind (approved):
### Overall
Prior blockers are resolved; no new PR-owned issues found in this round.

### Prior blockers
- **resolved** — `tileops/ops_manifest.yaml:2482` `WhereFwdOp` now documents the spec-only implementation signature mismatch.
- **resolved** — `tileops/ops_manifest.yaml:2519` `ClampFwdOp` now documents the implementation-name mismatches.
- **resolved** — `tileops/ops_manifest.yaml:2556` `MaskedFillFwdOp` now documents the implementation-name mismatches.
- **resolved** — `tileops/ops_manifest.yaml:2592` `NanToNumFwdOp` now documents the implementation-name/default mismatches.

### Advisory
- Manifest plus `tileops/perf/formulas.py` remains a cross-boundary change during migration; non-blocking.

lcy-seso:
Thanks for the approval. Noted on the manifest + `tileops/perf/formulas.py` cross-boundary advisory — keeping that in mind for the migration window; nothing to action this round.

PR #1087, opened 2026-04-27
[Feat][MoE] Add modular MoE interface: ABCs + FusedMoe orchestration

michaelwithu:
## Summary

- Introduce strategy-pattern interface for MoE forward pass — pluggable expert GEMM and EP/quantization backends without changing core routing.
- New ABCs in `tileops/ops/moe/abc.py`: `MoEExperts` / `MoEExpertsModular`, `MoEPrepareAndFinalize`, `WeightedReduce` / `WeightedReduceNoOp`, and `PrepareResult` dataclass carrying `hidden_q / scale / topk_weights / topk_ids` from `prepare()` to `apply()`.
- Concrete implementations: `MoEPrepareAndFinalizeNoDPEP` (pass-through, no EP/quantization), `MoEExpertsNopadFwdOp` (wraps existing nopad pipeline), `MoEExpertsPaddedFwdOp` (wraps existing padded pipeline).
- `FusedMoe` refactor (`fused_moe.py`): accepts optional `prepare_finalize` and `experts` strategy args, defaults preserve existing behavior; `forward()` delegates via the ABC protocol; fixes silent drop of `expert_map` (now stored and forwarded to both `prepare()` and `apply()`).
- Manifest (`tileops/manifest/moe.yaml`): add `MoEExpertsNopadFwdOp`, `MoEExpertsPaddedFwdOp` (replace prior `FusedMoeExperts*FwdOp` modular entries), `FusedMoeFwdOp` (softmax scoring), `FusedMoeFwdCbFwdOp` (sigmoid + correction_bias). Add `fused_moe_fwd_bytes` roofline formula in `tileops/perf/formulas.py`. Both experts entries set `bench_manifest_driven: false` — strategy objects are not `Op` subclasses and cannot implement `eval_roofline()`.
- Tests (`tests/ops/test_moe_experts_nopad.py`): 15 smoke tests covering ABC contracts and the three concrete strategies, parametrized over `bfloat16` and `float16`; correctness verified against `FusedMoeExperts*FwdOp` references.
- Benchmark (`benchmarks/ops/bench_moe_experts_nopad.py`): manifest-driven workloads, benches nopad and padded side-by-side; baseline vLLM `fused_experts` when available, torch `index_add_` loop fallback.
- Docs (`docs/design/ops-design.md`): add "MoE Modular Protocol" section — ABC strategy pattern, component table, `FusedMoe.forward()` data flow, EP/quantization extension points.

## Test plan

- [x] `pre-commit run` passed (SKIP=gitleaks)
- [x] `pytest tests/ops/test_moe_experts_nopad.py` — 15 passed (bfloat16 + float16)
- [x] `pytest tests/ops/test_moe_fused_moe.py` — 20 passed (no `FusedMoe` regression)
- [x] Merge conflict with `main` resolved (`ops_manifest.yaml` → `tileops/manifest/moe.yaml` migration)
- [ ] Full CI smoke suite

## Regression

- `FusedMoe` defaults to `MoEPrepareAndFinalizeNoDPEP` + `MoEExpertsNopadFwdOp` / `MoEExpertsPaddedFwdOp`, so existing call sites remain backwards-compatible.
- `expert_map` was previously silently ignored; this PR wires it through. Confirmed via existing `test_moe_fused_moe.py` (20 passed).

## Additional context

- Lays groundwork for future EP dispatch and per-expert quantization backends without touching `FusedMoe.forward()`.
- Both new experts ops carry `bench_manifest_driven: false` because strategy objects do not subclass `Op`; manifest-driven benchmarking will be revisited when the modular path graduates to first-class `Op` status.


tileops/ops/moe/fused_moe.py
```diff
@@ -102,46 +88,73 @@ def __init__(
...
-        """
+        # 1. Routing
         topk_weights, topk_ids = self._fused_topk(gating_output, correction_bias)
-        return self._experts(hidden_states, w_gate_up, w_down, topk_weights, topk_ids)
+
+        # 2. Prepare (quantization / EP dispatch)
+        r = self._prepare.prepare(
+            hidden_states, topk_weights, topk_ids,
+            self.num_experts, expert_map=None,
+        )
```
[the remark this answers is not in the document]
michaelwithu: Fixed. Added `self.expert_map = expert_map` in `__init__` and replaced the two hardcoded `expert_map=None` calls in `forward()` with `expert_map=self.expert_map` (in both `prepare()` and `apply()`).

tileops/ops/moe/fused_moe.py
```diff
@@ -102,46 +88,73 @@ def __init__(
...
+
+        # 3. Expert GEMM
+        T_prime = r.hidden_q.shape[0]
+        ws1_shape, ws2_shape = self._experts.workspace_shapes(
+            T_prime, self.ffn_size, self.hidden_size,
+            self.top_k, self.num_experts,
+        )
+        ws1 = hidden_states.new_empty(ws1_shape) if ws1_shape != (0,) else hidden_states.new_empty(0)
+        ws2 = hidden_states.new_empty(ws2_shape) if ws2_shape != (0,) else hidden_states.new_empty(0)
```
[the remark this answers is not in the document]
michaelwithu: Fixed. Replaced the conditional branches with a direct `hidden_states.new_empty(ws1_shape)` / `new_empty(ws2_shape)` — `new_empty((0,))` already produces a zero-size tensor, so the extra guard was redundant.

tileops/ops/moe/fused_moe.py
```diff
@@ -102,46 +88,73 @@ def __init__(
...
+            workspace1=ws1, workspace2=ws2,
+        )
+
+        # 4. Finalize (weighted reduction / EP gather)
+        output = hidden_states.new_empty(hidden_states.shape)
+        self._prepare.finalize(
+            output, expert_out,
+            r.topk_weights, r.topk_ids,
+            self._experts.make_weighted_reduce(),
+        )
```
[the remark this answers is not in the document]
michaelwithu: Fixed. `output` is now allocated first, then `expert_out = output` when `expert_out_shape == output.shape` (the no-EP common case). `WeightedReduceNoOp.apply()` already short-circuits when `output is expert_out`, so this eliminates one allocation and the subsequent `copy_` entirely. Separate allocation is kept for the EP case where shapes differ.

tileops/ops/moe/experts/nopad.py
```diff
@@ -0,0 +1,95 @@
...
+        workspace2: Tensor,
+    ) -> None:
+        perm_h, true_offsets, true_sizes, _, fwd_idx = self._permute(hidden_q, topk_ids)
+        gate_up = self._gemm_gate_up(perm_h, w1, true_sizes, true_offsets)
+        act = self._silu_and_mul(gate_up)
+        mm2 = self._gemm_down(act, w2, true_sizes, true_offsets)
+        result = self._unpermute(mm2, fwd_idx, topk_weights)
+        if self._routed_scaling_factor != 1.0:
+            result = result * self._routed_scaling_factor
+        output.copy_(result)
```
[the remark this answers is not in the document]
michaelwithu: Fixed. Changed to `torch.mul(result, self._routed_scaling_factor, out=output)` when scaling is needed, and `output.copy_(result)` otherwise. This avoids creating a temporary tensor for the scaled result.

tileops/ops_manifest.yaml
```diff
@@ -1309,113 +1309,202 @@ ops:
...
     source:
-      kernel: tileops/ops/moe/fused_moe_experts.py
-      op: tileops/ops/moe/fused_moe_experts.py
+      kernel: tileops/kernels/moe/moe_grouped_gemm_nopad.py
+      op: tileops/ops/moe/moe_grouped_gemm_nopad.py
       test: tests/ops/test_moe_fused_moe.py
       bench: benchmarks/ops/bench_moe_fused_moe.py
 
-  FusedMoeExpertsPaddedFwdOp:
+  MoEExpertsNopadFwdOp:
```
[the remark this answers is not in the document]
michaelwithu: Fixed. Renamed the Python classes from `MoEExpertsNopad` / `MoEExpertsPadded` to `MoEExpertsNopadFwdOp` / `MoEExpertsPaddedFwdOp` to match the manifest keys and align with the project-wide `{Name}FwdOp` naming convention. All references updated (class definitions, `__init__.py`, `fused_moe.py`, tests, benchmark).

tileops/ops/moe/abc.py
```diff
@@ -0,0 +1,169 @@
...
+        output: Tensor,                  # [T, H]  write destination (pre-allocated)
+        expert_out: Tensor,              # output of MoEExperts.apply()
+        topk_weights: Tensor,            # [T', K] float32
+        topk_ids: Tensor,                # [T', K] int32
+        weight_and_reduce: WeightedReduce,
+    ) -> None:
+        """Post-condition: output.shape == (T, H)  (T = original token count)."""
+
+
+class MoEExperts(ABC):
```
Ibuki-wind, line 103: `MoEExperts`/`MoEPrepareAndFinalize` introduce a new MoE family protocol under `tileops/ops`, but `docs/ops-design.md` was not updated; the ops domain rule requires docs updates when adding intermediate base classes or new class protocols. → Add an ops-design section for this modular MoE protocol, including ownership of prepare/finalize, experts, weighted reduction, and how these ABCs relate to `Op`/manifest entries.

tests/ops/test_moe_experts_nopad.py
```diff
@@ -0,0 +1,167 @@
...
+        pf.finalize(output, expert_out, weights, ids, WeightedReduceNoOp())
+        assert torch.allclose(output, expert_out)
+
+
+# ---------------------------------------------------------------------------
+# MoEExpertsNopadFwdOp
+# ---------------------------------------------------------------------------
+
+@pytest.fixture
+def moe_tensors():
```
Ibuki-wind: `moe_tensors` fixes `dtype = torch.bfloat16`, so the float16 path declared by the manifest (`hidden_states: float16 | bfloat16`) is never tested. → Parametrize this fixture or the apply/output-shape tests over both `torch.float16` and `torch.bfloat16`, keeping only shape cases that exercise distinct code paths.

benchmarks/ops/bench_moe_experts_nopad.py
```diff
@@ -0,0 +1,152 @@
...
+
+import pytest
+import torch
+
+from benchmarks.benchmark_base import BenchmarkBase, BenchmarkReport
+from tileops.manifest import load_workloads
+from tileops.ops.moe import MoEExpertsNopadFwdOp, MoEExpertsPaddedFwdOp
+from workloads.workload_base import WorkloadBase
+
+_OP_NAME = "MoEExpertsNopadFwdOp"
```
Ibuki-wind, line 43: `_OP_NAME` hard-codes only `MoEExpertsNopadFwdOp` while this same file is declared as the manifest-driven bench for `MoEExpertsPaddedFwdOp`, and the file never calls `eval_roofline()`; targeted L4 validation for these `bench_manifest_driven: true` entries will fail or stop checking the padded manifest workloads. → Make the benchmark manifest-driven for each declared op name (for example parametrize over both op names/layouts via `workloads_to_params`/`ManifestBenchmark`) or remove `bench_manifest_driven: true` until the file satisfies L4.

benchmarks/ops/bench_moe_experts_nopad.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def _padded_fn(hidden, w1, w2, topk_weights, topk_ids):
+        padded.apply(output, hidden, w1, w2, topk_weights, topk_ids, num_experts, None, ws1, ws2)
+        return output
+
+    _padded_fn(hidden, w1, w2, topk_weights, topk_ids)  # warmup / JIT compile
+    torch.cuda.synchronize()
+
+    result_pad = bm.profile(_padded_fn, hidden, w1, w2, topk_weights, topk_ids)
+    BenchmarkReport.record(padded, locals(), result_pad, tag="tileops-padded")
```
Ibuki-wind, line 162: `BenchmarkReport.record()` only records `tileops-*` tags, and the benchmark rules classify tags starting with `tileops` as TileOPs entries, not baselines. → Add at least one independent non-`tileops` baseline (for example a local torch/triton reference expert GEMM path) and record it with a baseline tag such as `torch` or `triton`.

Ibuki-wind (changes requested):
### Overall
The new MoE modular surface, tests, and benchmark wiring need to match the project contracts before merge.

### Advisory
- This PR changes implementation, tests, benchmarks, and `tileops/ops_manifest.yaml` together; during the migration this is advisory only, but it crosses the documented stage boundaries.

tileops/ops/moe/abc.py
michaelwithu: Fixed in af0750a. Added a "MoE Modular Protocol" section to `docs/ops-design.md` covering the ABC strategy pattern (`MoEPrepareAndFinalize`, `MoEExperts`, `WeightedReduce`), a table of concrete implementations, the `FusedMoe.forward()` data flow, and how these ABCs relate to `Op`/manifest entries.

tests/ops/test_moe_experts_nopad.py
michaelwithu: Fixed in af0750a. Parametrized the `moe_tensors` fixture with `params=[torch.bfloat16, torch.float16]`, ids=["bfloat16", "float16"]. All downstream tests (`test_workspace_shapes`, `test_output_shape`, `test_make_weighted_reduce_is_noop`, `test_apply_matches_*`) now run over both dtypes — 15 test cases total.

benchmarks/ops/bench_moe_experts_nopad.py
michaelwithu: Fixed in af0750a. Changed `bench_manifest_driven: false` for both `MoEExpertsNopadFwdOp` and `MoEExpertsPaddedFwdOp` in the manifest. The strategy ABC objects (`MoEExperts` subclasses) are not `Op` subclasses and therefore cannot implement `eval_roofline()`, which the L4 manifest-driven protocol requires. The flag will be revisited when these ops flip from `spec-only` to `implemented` and have a proper `Op` wrapper.

benchmarks/ops/bench_moe_experts_nopad.py
michaelwithu: Fixed in af0750a. The benchmark now records vLLM `fused_experts` as `tag="vllm"` when vLLM is available, and falls back to a torch per-expert loop with `index_add_` recorded as `tag="torch-ref"` otherwise. Both tags are non-`tileops`-prefixed, so `BenchmarkReport` classifies them as baselines.

tileops/ops/moe/fused_moe.py
```diff
@@ -102,46 +89,72 @@ def __init__(
...
-        return self._experts(hidden_states, w_gate_up, w_down, topk_weights, topk_ids)
+
+        # 2. Prepare (quantization / EP dispatch)
+        r = self._prepare.prepare(
+            hidden_states, topk_weights, topk_ids,
+            self.num_experts, expert_map=self.expert_map,
+        )
+
+        # 3. Expert GEMM
+        T_prime = r.hidden_q.shape[0]
```
Ibuki-wind, line 142: `FusedMoe.forward()` lets `prepare_finalize` return `T_prime != self.num_tokens`, but the default `_experts` instance was constructed in `__init__` with `num_tokens=self.num_tokens`, so its inner permute/unpermute/GEMM ops are still sized for the original token count while this code feeds `r.hidden_q` with the dispatched count. A caller following the documented extension point `FusedMoe(prepare_finalize=...)` for EP dispatch will either run kernels with stale shapes or write the wrong output. → Either require a custom `experts` implementation whenever `prepare_finalize` can change T, or rebuild/select the default experts for `T_prime` before `apply()`.

Ibuki-wind (changes requested):
### Overall
The modular EP path is not safe yet; fix the expert sizing contract before merging.

### Advisory
- This PR changes implementation, tests, benchmark, manifest, and design docs in one step; during the migration period I am treating that as advisory only.

tileops/ops/moe/fused_moe.py
michaelwithu: Fixed in e64c915. Added a fail-fast guard in `__init__`: if `prepare_finalize` is supplied without `experts`, a `ValueError` is raised immediately at construction time.

The root cause is that `num_tokens` is consumed at `__init__` as a compile-time constant (`numel = num_tokens * top_k`) baked into the JIT kernel tile schedules of `MoeGroupedGemmNopadFwdOp`, `SiluAndMulFwdOp`, and `MoeUnpermuteFwdOp`. There is no safe way to reuse a stale experts instance at runtime — the correct fix is to require the caller to supply a matching `experts=` sized for `T'`.

Also added `test_prepare_finalize_without_experts_raises` as a smoke test.

Ibuki-wind (approved):
### Overall
Clean — no issues.

### Prior blockers
All 1 prior blockers resolved — see thread replies.

PR #1094, opened 2026-04-28
[Refactor][Manifest] flatten yaml top-level ops key

lcy-seso:
## Summary

- Each family yaml previously wrapped its entries under a top-level `ops:` key, costing one indent level on every entry without serving any purpose — family scope is already conveyed by the filename, and the loader just unwrapped the key.
- Flatten the format: each family file is now a direct mapping of op name → entry. Updated 8 yaml files (4546 lines), the `tileops.manifest` loader, the `--manifest-path` branch in `scripts/validate_manifest.py`, the `tests/test_validate_manifest.py` fixtures, and the example in `docs/manifest.md`.
- Pre-release project, so no compat shim — the old format is no longer accepted.

## Test plan

- [x] pre-commit passed (full hook suite on all changed files)
- [x] `python scripts/validate_manifest.py` → `All manifest checks passed.`
- [x] `pytest tests/test_validate_manifest.py tests/test_gpu_smoke_policy.py` → 176 passed
- [x] `from tileops.manifest import load_manifest` smoke check → 114 ops loaded

## Regression

- Loader migration is mechanical (`data.get("ops") or {}` → `yaml.safe_load(...) or {}`). All 114 ops load identically; downstream consumers (`load_workloads`, `validate_manifest`) untouched.
- `_load_manifest` private alias removed (no remaining callers in repo).

## Additional context

There is no near-term plan for family-level metadata. If that need arises later, a wrapper can be reintroduced — YAGNI for now.

### DIFF
--- a/scripts/validate_manifest.py
+++ b/scripts/validate_manifest.py
@@ -2933,8 +2933,7 @@ def validate_manifest(
         ops = load_manifest()
     else:
         with open(manifest_path) as f:
-            data = yaml.safe_load(f)
-        ops = data.get("ops", {})
+            ops = yaml.safe_load(f) or {}
 
     # Fail fast: --check-op with a name not in the manifest
     if check_op is not None and check_op not in ops:
--- a/scripts/validate_manifest.py
+++ b/scripts/validate_manifest.py
@@ -2934,6 +2934,11 @@ def validate_manifest(
     else:
         with open(manifest_path) as f:
             ops = yaml.safe_load(f) or {}
+        if not isinstance(ops, dict):
+            return [
+                f"--manifest-path: {manifest_path} must contain a top-level "
+                f"mapping of op name -> entry, got {type(ops).__name__}"
+            ], []
 
     # Fail fast: --check-op with a name not in the manifest
     if check_op is not None and check_op not in ops:
--- a/tileops/manifest/__init__.py
+++ b/tileops/manifest/__init__.py
@@ -45,6 +45,11 @@ def load_manifest() -> dict[str, Any]:
     for path in manifest_files():
         text = path.read_text(encoding="utf-8")
         ops = yaml.safe_load(text) or {}
+        if not isinstance(ops, dict):
+            raise ValueError(
+                f"manifest file {path.name} must contain a top-level mapping of "
+                f"op name -> entry, got {type(ops).__name__}"
+            )
         for name, entry in ops.items():
             if name in merged:
                 raise ValueError(
--- a/tileops/manifest/__init__.py
+++ b/tileops/manifest/__init__.py
@@ -44,8 +44,7 @@ def load_manifest() -> dict[str, Any]:
     origin: dict[str, str] = {}
     for path in manifest_files():
         text = path.read_text(encoding="utf-8")
-        data = yaml.safe_load(text) or {}
-        ops = data.get("ops") or {}
+        ops = yaml.safe_load(text) or {}
         for name, entry in ops.items():
             if name in merged:
                 raise ValueError(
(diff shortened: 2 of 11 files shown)


tileops/manifest/__init__.py
```diff
@@ -44,8 +44,7 @@ def load_manifest() -> dict[str, Any]:
     origin: dict[str, str] = {}
     for path in manifest_files():
         text = path.read_text(encoding="utf-8")
-        data = yaml.safe_load(text) or {}
-        ops = data.get("ops") or {}
+        ops = yaml.safe_load(text) or {}
```
Ibuki-wind, line 47: `load_manifest()` now treats the YAML root as the op mapping, but `yaml.safe_load()` can return a list or scalar; then `ops.items()` raises `AttributeError` before callers get a manifest error. -> Validate that the loaded root is a mapping before iterating and raise a clear manifest `ValueError` naming the file/type.

scripts/validate_manifest.py
```diff
@@ -2933,8 +2933,7 @@ def validate_manifest(
         ops = load_manifest()
     else:
         with open(manifest_path) as f:
-            data = yaml.safe_load(f)
-        ops = data.get("ops", {})
+            ops = yaml.safe_load(f) or {}
```
Ibuki-wind, line 2936: `validate_manifest(manifest_path=...)` now assigns the YAML root directly to `ops`, but malformed YAML such as a top-level sequence crashes later at `ops.items()` instead of returning schema errors. -> Add a top-level mapping check here and cover it with a regression test so the validator reports the bad manifest structure without throwing.

Ibuki-wind (changes requested):
### Overall
The flat manifest migration is coherent, but both new parse sites need to preserve validator-style structural failure instead of crashing on non-mapping YAML.

tileops/manifest/__init__.py
```diff
@@ -44,8 +44,7 @@ def load_manifest() -> dict[str, Any]:
...
         for name, entry in ops.items():
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Added a top-level mapping check in `load_manifest()` that raises a clear `ValueError` naming the file and the offending root type before `.items()` is reached. See 1bd63eb.

scripts/validate_manifest.py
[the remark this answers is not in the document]
lcy-seso: Adopted. The `--manifest-path` branch now returns a structural error (not an exception) when the yaml root is not a mapping, naming both the path and the offending type. See 1bd63eb.

tileops/manifest/__init__.py
lcy-seso: Adopted. `load_manifest()` now validates that each yaml root is a mapping and raises `ValueError("manifest file <name> must contain a top-level mapping ... got <type>")` before iterating. See 1bd63eb.

scripts/validate_manifest.py
lcy-seso: Adopted. Added the mapping check on the `--manifest-path` branch (returns a schema error, no exception) plus a regression test in `test_manifest_path_non_mapping_root_reports_error` that feeds a top-level sequence and asserts the structural error. See 1bd63eb.

lcy-seso:
Both blockers from the CHANGES_REQUESTED review addressed in 1bd63eb:

- `tileops/manifest/__init__.py`: `load_manifest()` raises a clear `ValueError` when a yaml file's top-level isn't a mapping (names the file and the offending root type).
- `scripts/validate_manifest.py` `--manifest-path` branch: returns a structural error instead of throwing, plus a new regression test (`test_manifest_path_non_mapping_root_reports_error`) that feeds a top-level sequence.

Validator + 171-test suite green. Threads resolved.

Ibuki-wind (approved):
### Overall
Clean — no issues.

### Prior blockers
All 2 prior blockers resolved — see thread replies.

PR #1117, opened 2026-04-29
[Fix][Reduction] Fix bugs in logical and norm reductions

### LINKED ISSUE #958: [BUG][REDUCTION] logical reduce/vector norm fail at n=32768 benchmark
### Symptom / Motivation
Nightly benchmark run [24313885243](https://github.com/tile-ai/TileOPs/actions/runs/24362051145) fails for `logical_reduce` and `vector_norm` benchmark cases with the compile-time error:

```text
tvm.error.InternalError: Can't fetch the lanes of a scalable vector at a compile time.
```

The failure happens during TileLang/TVM lowering rather than kernel execution, so these workloads currently make the benchmark job red even though the failure mode is deterministic and shape-dependent.

### Root Cause Analysis
The failing cases all belong to the reduction family and share the same last-dimension size `N=32768`. Both `LogicalReduceKernel` and `VectorNormKernel` still allocate shared/fragment buffers over the full `N_padded` width, for example `T.alloc_shared((block_m, N_padded), dtype)` and `T.alloc_fragment((block_m, N_padded), "float32")`, without switching to an `N`-tiled path.

The repository already documents this compiler boundary in `tileops/kernels/reduction/_primitives.py`: `MAX_SINGLE_TILE_COLS = 32512`, with a note that `32768` triggers the scalable-vector error. Unlike the generic `ReduceKernel`, these two kernel families do not yet use a tiled fallback when `N_padded > MAX_SINGLE_TILE_COLS`, so manifest-driven benchmark workloads at `N=32768` still compile through the known-bad full-width path.

### Nightly CI Results
- Nightly run: https://github.com/tile-ai/TileOPs/actions/runs/24313885243
- Failing benchmark job: https://github.com/tile-ai/TileOPs/actions/runs/24313885243/job/70989292686
- Affected benchmark cases:
  - `benchmarks/ops/bench_logical_reduce.py::test_any_bench[mask-validation-32k-bool]`
  - `benchmarks/ops/bench_logical_reduce.py::test_all_bench[mask-validation-32k-bool]`
  - `benchmarks/ops/bench_logical_reduce.py::test_count_nonzero_bench[sparsity-seq-float16]`
  - `benchmarks/ops/bench_vector_norm.py::test_l1_norm_bench[long-seq-l1-bfloat16]`
  - `benchmarks/ops/bench_vector_norm.py::test_l2_norm_bench[long-seq-l2-bfloat16]`
  - `benchmarks/ops/bench_vector_norm.py::test_inf_norm_bench[long-seq-inf-bfloat16]`

### Related Files
- `tileops/kernels/reduction/_primitives.py`
- `tileops/kernels/reduction/logical_reduce.py`
- `tileops/kernels/reduction/vector_norm.py`
- `tileops/kernels/reduction/reduce.py`
- `tileops/ops_manifest.yaml`
- `benchmarks/ops/bench_logical_reduce.py`
- `benchmarks/ops/bench_vector_norm.py`

## Goal
Make `logical_reduce` and `vector_norm` handle manifest benchmark workloads with `N=32768` without triggering the TileLang scalable-vector compile failure.

## Plan

1. Add an `N`-tiled fallback to `LogicalReduceKernel` for workloads whose padded reduction width exceeds `MAX_SINGLE_TILE_COLS`.
2. Add the same `tile_n`-based fallback to `VectorNormKernel`, accumulating partial results across tiles.
3. Reuse the existing large-`N` tiling strategy from the generic reduction kernels as the design reference.
4. Re-run the affected benchmark workloads and confirm the nightly benchmark job no longer fails on these six cases.

## Constraints
- Keep the existing public op interfaces unchanged.
- Do not regress small-`N` workloads that already compile and run correctly.
- Preserve the current manifest-driven benchmark coverage; do not remove the `N=32768` workloads from `ops_manifest.yaml` as a workaround.

## Acceptance Criteria
- [ ] `logical_reduce` benchmark workloads at `N=32768` no longer fail with the scalable-vector compile error
- [ ] `vector_norm` benchmark workloads at `N=32768` no longer fail with the scalable-vector compile error
- [ ] The nightly benchmark job no longer fails on the six cases listed above

superAngGao:
## Summary
- move logical reduce and vector norm alignment padding into the TileLang kernels
- add N-tiled fallback paths for logical reduce and vector norm when rows exceed the single-fragment column limit
- cover the 32k logical reduce/vector norm cases that failed in nightly benchmark

Fixes #958

## Validation
- `ruff check tileops/kernels/reduction/logical_reduce.py tileops/kernels/reduction/vector_norm.py tileops/ops/reduction/any_op.py tileops/ops/reduction/all_op.py tileops/ops/reduction/count_nonzero.py tileops/ops/reduction/l1_norm.py tileops/ops/reduction/l2_norm.py tileops/ops/reduction/inf_norm.py tileops/ops/reduction/reduce.py tests/ops/test_logical_reduce.py tests/ops/test_vector_norm.py`
- `pytest -q tests/ops/test_logical_reduce.py::test_logical_reduce_long_sequence_tiled tests/ops/test_vector_norm.py::test_vector_norm_long_sequence_tiled`
- `pytest -q tests/ops/test_logical_reduce.py tests/ops/test_vector_norm.py -m smoke`
- `pytest -q tests/ops/test_logical_reduce.py tests/ops/test_vector_norm.py -k 300`
- `pytest -q benchmarks/ops/bench_logical_reduce.py::test_any_bench[mask-validation-32k-bool] benchmarks/ops/bench_logical_reduce.py::test_all_bench[mask-validation-32k-bool] benchmarks/ops/bench_logical_reduce.py::test_count_nonzero_bench[sparsity-seq-float16] benchmarks/ops/bench_vector_norm.py::test_l1_norm_bench[long-seq-l1-bfloat16] benchmarks/ops/bench_vector_norm.py::test_l2_norm_bench[long-seq-l2-bfloat16] benchmarks/ops/bench_vector_norm.py::test_inf_norm_bench[long-seq-inf-bfloat16]`

### DIFF
--- a/tests/ops/test_logical_reduce.py
+++ b/tests/ops/test_logical_reduce.py
@@ -10,6 +10,7 @@ import pytest
 import torch
 
 from tests.test_base import FixtureBase, TestBase
+from tileops.kernels.reduction.logical_reduce import LogicalReduceKernel
 from workloads.logical_reduce import AnyTest as _AnyWorkload
 
 # ---------------------------------------------------------------------------
@@ -151,6 +152,22 @@ class LogicalReduceTest(_AnyWorkload, TestBase):
         elif self.op_kind == "count_nonzero":
             return torch.count_nonzero(x, dim=-1).to(torch.int64)
         raise ValueError(f"Unknown op_kind: {self.op_kind}")
+
+
+class _TailBlockLogicalReduceKernel(LogicalReduceKernel):
+    """Force tiled tests to cover tail-M masking with block_m > M."""
+
+    _TAIL_BLOCK_M = 4
+    _TAIL_TILE_N = 8192
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        assert self._needs_tiling, "tail-M regression test must use the tiled kernel"
+        self.config = {
+            "block_m": self._TAIL_BLOCK_M,
+            "threads": 128,
+            "tile_n": self._TAIL_TILE_N,
+        }
 
 
 def _exact_compare(output: torch.Tensor, output_ref: torch.Tensor) -> None:
@@ -642,9 +659,15 @@ def test_logical_reduce_long_sequence_tiled(op_kind: str, dtype: torch.dtype) ->
         "count_nonzero": CountNonzeroFwdOp,
     }
     test = LogicalReduceTest(3, 33024, dtype, op_kind)
-    op = op_map[op_kind](dtype=dtype)
+    op = op_map[op_kind](
+        dtype=dtype,
+        kernel_map={"logical_reduce": _TailBlockLogicalReduceKernel},
+    )
     compare = _exact_compare_int64 if op_kind == "count_nonzero" else _exact_compare
     test.check(op, *test.gen_inputs(), compare=compare)
+    kernel = op._kernel_cache[(3, 33024)]
+    assert kernel.config["block_m"] > test.shape[0]
+    assert kernel.config["tile_n"] > 0
 
 
 if __name__ == "__main__":
--- a/tests/ops/test_logical_reduce.py
+++ b/tests/ops/test_logical_reduce.py
@@ -621,5 +621,31 @@ def test_count_nonzero_smoke_bool(m: int, n: int, dtype: torch.dtype) -> None:
     test.check(op, *test.gen_inputs(), compare=_exact_compare_int64)
 
 
+@pytest.mark.smoke
+@pytest.mark.parametrize(
+    "op_kind, dtype",
+    [
+        ("any", torch.bool),
+        ("all", torch.bool),
+        ("count_nonzero", torch.float16),
+    ],
+)
+def test_logical_reduce_long_sequence_tiled(op_kind: str, dtype: torch.dtype) -> None:
+    """Exercise the N-tiled path with a tail-M block."""
+    from tileops.ops.reduction.all_op import AllFwdOp
+    from tileops.ops.reduction.any_op import AnyFwdOp
+    from tileops.ops.reduction.count_nonzero import CountNonzeroFwdOp
+
+    op_map = {
+        "any": AnyFwdOp,
+        "all": AllFwdOp,
+        "count_nonzero": CountNonzeroFwdOp,
+    }
+    test = LogicalReduceTest(3, 33024, dtype, op_kind)
+    op = op_map[op_kind](dtype=dtype)
+    compare = _exact_compare_int64 if op_kind == "count_nonzero" else _exact_compare
+    test.check(op, *test.gen_inputs(), compare=compare)
+
+
 if __name__ == "__main__":
     pytest.main([__file__, "-vvs"])
--- a/tests/ops/test_vector_norm.py
+++ b/tests/ops/test_vector_norm.py
@@ -562,5 +562,16 @@ def test_empty_dim_full_reduction_3d_dtypes(
     allclose_compare(y, ref, atol=atol, rtol=rtol)
 
 
+@pytest.mark.smoke
+@pytest.mark.parametrize("op_kind", ["l1", "l2", "inf"])
+def test_vector_norm_long_sequence_tiled(op_kind: str) -> None:
+    """Exercise the N-tiled path with a tail-M block."""
+    dtype = torch.bfloat16
+    test = VectorNormTest(3, 33024, dtype, op_kind)
+    op = _make_op(dtype, op_kind)
+    atol, rtol = _get_tolerances(dtype)
+    test.check(op, *test.gen_inputs(), atol=atol, rtol=rtol)
+
+
 if __name__ == "__main__":
     pytest.main([__file__, "-vvs"])
--- a/tests/ops/test_vector_norm.py
+++ b/tests/ops/test_vector_norm.py
@@ -9,6 +9,7 @@ import pytest
 import torch
 
 from tests.test_base import FixtureBase, TestBase, allclose_compare
+from tileops.kernels.reduction.vector_norm import VectorNormKernel
 from workloads.vector_norm import L1NormTest as _L1NormWorkload
 
 # ---------------------------------------------------------------------------
@@ -109,6 +110,22 @@ class VectorNormTest(_L1NormWorkload, TestBase):
         return ref.to(self.dtype)
 
 
+class _TailBlockVectorNormKernel(VectorNormKernel):
+    """Force tiled tests to cover tail-M masking with block_m > M."""
+
+    _TAIL_BLOCK_M = 4
+    _TAIL_TILE_N = 8192
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        assert self._needs_tiling, "tail-M regression test must use the tiled kernel"
+        self.config = {
+            "block_m": self._TAIL_BLOCK_M,
+            "threads": 128,
+            "tile_n": self._TAIL_TILE_N,
+        }
+
+
 def _get_tolerances(dtype: torch.dtype):
     """Return (atol, rtol) for the given dtype."""
     if dtype == torch.float32:
@@ -132,7 +149,13 @@ def _make_1d_input(n: int, dtype: torch.dtype) -> torch.Tensor:
     return torch.randn(n, dtype=dtype, device="cuda")
 
 
-def _make_op(dtype: torch.dtype, op_kind: str, dim: int = -1, keepdim: bool = False):
+def _make_op(
+    dtype: torch.dtype,
+    op_kind: str,
+    dim: int = -1,
+    keepdim: bool = False,
+    kernel_map=None,
+):
     """Create the appropriate Op for the given op_kind."""
     from tileops.ops.reduction.inf_norm import InfNormFwdOp
     from tileops.ops.reduction.l1_norm import L1NormFwdOp
@@ -144,7 +167,7 @@ def _make_op(dtype: torch.dtype, op_kind: str, dim: int = -1, keepdim: bool = Fa
         "inf": InfNormFwdOp,
     }
     cls = op_map[op_kind]
-    return cls(dtype=dtype, dim=dim, keepdim=keepdim)
+    return cls(dtype=dtype, dim=dim, keepdim=keepdim, kernel_map=kernel_map)
 
 
 # ---------------------------------------------------------------------------
@@ -568,9 +591,16 @@ def test_vector_norm_long_sequence_tiled(op_kind: str) -> None:
     """Exercise the N-tiled path with a tail-M block."""
     dtype = torch.bfloat16
     test = VectorNormTest(3, 33024, dtype, op_kind)
-    op = _make_op(dtype, op_kind)
+    op = _make_op(
+        dtype,
+        op_kind,
+        kernel_map={"vector_norm": _TailBlockVectorNormKernel},
+    )
     atol, rtol = _get_tolerances(dtype)
     test.check(op, *test.gen_inputs(), atol=atol, rtol=rtol)
+    kernel = op._kernel_cache[(3, 33024)]
+    assert kernel.config["block_m"] > test.shape[0]
+    assert kernel.config["tile_n"] > 0
 
 
 if __name__ == "__main__":
--- a/tileops/kernels/reduction/logical_reduce.py
+++ b/tileops/kernels/reduction/logical_reduce.py
@@ -5,9 +5,8 @@ Casts input to bool (0/1 as float32), then reduces:
   - all: reduce_min (1 if all elements are non-zero)
   - count_nonzero: reduce_sum (count of non-zero elements per row)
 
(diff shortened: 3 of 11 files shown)


Gabbering (commented):
## goose goose review — `2e2d744c`

*honk.* The tiling logic is structurally sound. The goose found one real bug and one test gap worth honking about.

### goose Bugs

- `tileops/kernels/reduction/logical_reduce.py:177` — `_needs_mask` is computed as `total_cols > N`, where `total_cols = num_tiles * tile_n`. But `num_tiles` is derived from `N_padded`, not `N`. If `N` is already aligned (so `N_padded == N`) but `tile_n` doesn't evenly divide `N_padded`, then `total_cols > N` is true and you'll enter the masked path on the last tile. That's fine — the mask handles it. But the *non-last* tiles use `T.copy(x[pid_m * block_m, t * tile_n], shared_buf)` without a row-bounds check on `pid_m * block_m + i < M`. When `M` is not divisible by `block_m`, the last block's threads will issue out-of-bounds global memory reads for the tail rows on *every* tile iteration (not just the last tile). The same OOB read exists in the `else` branch (the `not _needs_mask` path) — `T.copy` will happily read past the end of the tensor when `M % block_m != 0`. This is the same pattern as the non-tiled kernel's `T.copy(x[pid_m * block_m, 0], shared_buf)`, so presumably TileLang's `T.Kernel(T.ceildiv(M, block_m))` + `T.copy` handles tail blocks implicitly (clamping or masking internally). If that's the case, fine — but the *masked* path's `T.Else` branch does an explicit `pid_m * block_m + i < M` check while the `T.Then` branch doesn't. If TileLang does NOT auto-clamp `T.copy` for tail-M, you have a correctness bug on the non-last tiles for tail-M rows. The same issue exists in `vector_norm.py:160`. Master, please confirm TileLang's `T.copy` semantics for tail blocks — if it doesn't auto-clamp, the `T.Then` branches need the row guard too.

- `tests/ops/test_logical_reduce.py:643` — The test uses `N=32768` to exercise the tiled path, but the tiling threshold is `N_padded > MAX_SINGLE_TILE_COLS` where `MAX_SINGLE_TILE_COLS` is 32768. `align_up(32768, 256) == 32768`, so `N_padded == MAX_SINGLE_TILE_COLS`, which means `self._needs_tiling` is `False`. This test does NOT exercise the tiled path at all — it takes the single-fragment path. You need `N > 32768` (e.g. `N=32769` or `N=33024`) to actually trigger `N_padded > MAX_SINGLE_TILE_COLS`. Same issue in `test_vector_norm.py:528`. The tests claim to exercise the tiled path but they don't. They're testing the non-tiled kernel with an exactly-at-boundary size. Congratulations, you wrote a test that tests nothing new.

### egg Test gaps

- No test exercises `M % block_m != 0` combined with the tiled path. The existing tail-M tests (`m=129`) only run at small `N` values that won't trigger tiling. If the `T.copy` tail-M concern above is real, you'd never catch it. Add a case like `(M=3, N=33024)` to the tiled tests.

tileops/kernels/reduction/logical_reduce.py
```diff
@@ -288,45 +372,152 @@ def __init__(
...
         self.N_padded = align_up(N, DEFAULT_ALIGNMENT)
-        self.kernel = _logical_reduce_kernel(
-            self.M,
-            self.N,
-            self.op_kind,
-            self.dtype_to_str(self._kernel_dtype),
-        )
+        self._elem_bytes = torch.tensor([], dtype=self._kernel_dtype).element_size()
+        self._smem_budget = device_smem_budget()
+        self._needs_tiling = self.N_padded > MAX_SINGLE_TILE_COLS
```
Ibuki-wind: `_needs_tiling` only checks the TileLang column cap, so rows with `N_padded <= MAX_SINGLE_TILE_COLS` but `block_m * N_padded * elem_bytes` above `device_smem_budget()` still take `_logical_reduce_kernel` and allocate the whole row in shared memory; `_tile_n_for_block_m()` already computes that this case needs a `tile_n`, but it is never consulted unless `_needs_tiling` is true → derive the launch path from `_tile_n_for_block_m()` / `default_config` (`tile_n != 0`) or include the shared-memory budget check in `_needs_tiling` before building the non-tiled kernel.

tileops/kernels/reduction/vector_norm.py
```diff
@@ -178,43 +300,147 @@ def __init__(
...
         self.N_padded = align_up(N, DEFAULT_ALIGNMENT)
-        self.kernel = _vector_norm_kernel(
-            self.M,
-            self.N,
-            self.op_kind,
-            self.dtype_to_str(self.dtype),
-        )
+        self._elem_bytes = torch.tensor([], dtype=dtype).element_size()
+        self._smem_budget = device_smem_budget()
+        self._needs_tiling = self.N_padded > MAX_SINGLE_TILE_COLS
```
Ibuki-wind: `_needs_tiling` only checks the TileLang column cap, so rows with `N_padded <= MAX_SINGLE_TILE_COLS` but `block_m * N_padded * elem_bytes` above `device_smem_budget()` still take `_vector_norm_kernel` and allocate the whole row in shared memory; `_tile_n_for_block_m()` already computes that this case needs a `tile_n`, but it is never consulted unless `_needs_tiling` is true → derive the launch path from `_tile_n_for_block_m()` / `default_config` (`tile_n != 0`) or include the shared-memory budget check in `_needs_tiling` before building the non-tiled kernel.

Ibuki-wind (changes requested):
### Overall
The new tiled fallback still misses shared-memory-budget-driven tiling cases; make the dispatch predicate match the tile-size computation.

### Advisory
- This PR spans Implementation and Test trust-boundary stages during the migration period.

superAngGao:
Updated in `aaa87f4`. Thanks for the reviews.

Changes made:
- Tiled logical reduce/vector norm now use explicit masked loads for every tile instead of relying on `T.copy` on the non-last tiles. This covers both tail columns and `M % block_m != 0` tail rows, so the tiled path no longer depends on implicit `T.copy` tail-block semantics.
- The tiled dispatch predicate now also checks shared-memory pressure: `N_padded * elem_bytes > device_smem_budget()`, in addition to `N_padded > MAX_SINGLE_TILE_COLS`. Non-tiled config selection now uses the same detected device shared-memory budget.
- Removed the unreachable `self._kernel_dtype is torch.bool` autotune branch; unsupported logical dtypes are already remapped before kernel dispatch.
- Updated the long-sequence tests to `M=3, N=33024`, so they cover both the tiled path and tail-M behavior.

One clarification: the current `MAX_SINGLE_TILE_COLS` in `_primitives.py` is `32512`, not `32768`, so the old `N=32768` tests did trigger tiling. I still changed them to `33024` to make the intent unambiguous and to add tail-M coverage.

Validation rerun in the nightly image on GPU 2:
- `ruff check tileops/kernels/reduction/logical_reduce.py tileops/kernels/reduction/vector_norm.py tests/ops/test_logical_reduce.py tests/ops/test_vector_norm.py`
- `python -m py_compile tileops/kernels/reduction/logical_reduce.py tileops/kernels/reduction/vector_norm.py tests/ops/test_logical_reduce.py tests/ops/test_vector_norm.py`
- `git diff --check`
- `pytest -q tests/ops/test_logical_reduce.py::test_logical_reduce_long_sequence_tiled tests/ops/test_vector_norm.py::test_vector_norm_long_sequence_tiled` -> 6 passed
- `pytest -q benchmarks/ops/bench_logical_reduce.py::test_any_bench[mask-validation-32k-bool] benchmarks/ops/bench_logical_reduce.py::test_all_bench[mask-validation-32k-bool] benchmarks/ops/bench_logical_reduce.py::test_count_nonzero_bench[sparsity-seq-float16] benchmarks/ops/bench_vector_norm.py::test_l1_norm_bench[long-seq-l1-bfloat16] benchmarks/ops/bench_vector_norm.py::test_l2_norm_bench[long-seq-l2-bfloat16] benchmarks/ops/bench_vector_norm.py::test_inf_norm_bench[long-seq-inf-bfloat16]` -> 6 passed
- `pytest -q tests/ops/test_logical_reduce.py tests/ops/test_vector_norm.py -k 300` -> 19 passed

Gabbering (commented):
goose *goose skimmed `aaa87f4` — nothing to honk about.*

lcy-seso:
The modifications conflict with the latest main branch. Rebasing is required.

Gabbering (commented):
goose *goose skimmed `170c9d4` — nothing to honk about.*

superAngGao:
Updated in `3125de9`.

For the conflict reported by @lcy-seso: I rebased the PR change onto upstream commit `69a800a` (`[Feat] vector_norm: support empty-dim full reduction`) instead of rebasing onto all of latest `main`, because the current token cannot push refs that introduce workflow-file history. This keeps the #1104 reduction changes as a real ancestor and resolves the virtual merge conflict cleanly. GitHub now reports `mergeable: MERGEABLE`.

For the earlier Ibuki review: the shared-memory-budget-driven tiling predicate remains in both logical reduce and vector norm:

- `self.N_padded > MAX_SINGLE_TILE_COLS`
- or `self.N_padded * elem_bytes > device_smem_budget()`

and non-tiled config selection still uses the same detected shared-memory budget.

Validation rerun in the nightly image on GPU 2:
- `ruff check` on touched reduction files -> passed
- AST parse on touched Python files -> passed
- `pytest -q tests/ops/test_logical_reduce.py::test_logical_reduce_long_sequence_tiled tests/ops/test_vector_norm.py::test_vector_norm_long_sequence_tiled tests/ops/test_vector_norm.py::test_empty_dim_full_reduction_keepdim tests/ops/test_vector_norm.py::test_empty_dim_full_reduction_3d_dtypes --tb=short` -> 21 passed
- `pytest -q tests/ops/test_logical_reduce.py tests/ops/test_vector_norm.py -k 300 --tb=short` -> 19 passed

Gabbering (commented):
goose *goose skimmed `3125de9` — nothing to honk about.*

tests/ops/test_logical_reduce.py
```diff
@@ -621,5 +621,31 @@ def test_count_nonzero_smoke_bool(m: int, n: int, dtype: torch.dtype) -> None:
...
+    from tileops.ops.reduction.all_op import AllFwdOp
+    from tileops.ops.reduction.any_op import AnyFwdOp
+    from tileops.ops.reduction.count_nonzero import CountNonzeroFwdOp
+
+    op_map = {
+        "any": AnyFwdOp,
+        "all": AllFwdOp,
+        "count_nonzero": CountNonzeroFwdOp,
+    }
+    test = LogicalReduceTest(3, 33024, dtype, op_kind)
```
Ibuki-wind, line 661: `test_logical_reduce_long_sequence_tiled` never hits the tail-row mask because the default config for `M=3, N=33024` still picks `block_m=1`, so this case does not cover `pid_m * block_m + i < M` -> pin `block_m` above `M` or choose a shape that forces that branch.

tests/ops/test_vector_norm.py
```diff
@@ -562,5 +562,16 @@ def test_empty_dim_full_reduction_3d_dtypes(
     allclose_compare(y, ref, atol=atol, rtol=rtol)
 
 
+@pytest.mark.smoke
+@pytest.mark.parametrize("op_kind", ["l1", "l2", "inf"])
+def test_vector_norm_long_sequence_tiled(op_kind: str) -> None:
+    """Exercise the N-tiled path with a tail-M block."""
+    dtype = torch.bfloat16
+    test = VectorNormTest(3, 33024, dtype, op_kind)
```
Ibuki-wind, line 593: `test_vector_norm_long_sequence_tiled` has the same gap: `M=3, N=33024` still uses `block_m=1` by default, so it does not exercise the tiled tail-row mask -> pin `block_m` above `M` or pick a shape that makes `pid_m * block_m + i < M` true.

Ibuki-wind (changes requested):
### Overall
The new regression coverage still needs one shape adjustment before this can be accepted.

superAngGao:
Handled the remaining review concern by pinning the long-sequence tiled reduction tests to a tail-row config (`M=3`, `block_m=4`, `tile_n=8192`) for both logical reduce and vector norm.

Nightly docker validation:
```bash
docker run --rm --gpus '"device=1"' --ipc=host --shm-size=16g \
  -v /home/ga/TileOPs-pr-1117:/workspace \
  -v /data7/shared/ci-cache:/data7/shared/ci-cache \
  -w /workspace tileops-runner:nightly-tl019-fullstack-no-tileops-ldfix \
  bash -lc "ruff check tests/ops/test_logical_reduce.py tests/ops/test_vector_norm.py && python -m pytest -q tests/ops/test_logical_reduce.py::test_logical_reduce_long_sequence_tiled tests/ops/test_vector_norm.py::test_vector_norm_long_sequence_tiled --tb=short"
```

Result: `6 passed, 14 warnings in 23.77s`; `ruff` passed.

Gabbering (commented):
goose *goose skimmed `d8d024f` — nothing to honk about.*

Ibuki-wind (approved):
Clean — no issues.

PR #1135, opened 2026-05-01
[Feat][Skill] add MAX_IDLE / wall-clock backstops to review and resolve loops

lcy-seso:
## Summary

Both `review-tileops/loop.sh` and `resolve-tileops` (round-pre.sh) poll GitHub independently without waiting on each other. Previously, if one loop died silently the other could keep idling forever. This PR adds layered termination guarantees so any failure mode exits within a bounded wall-clock window.

- **Symmetric `MAX_IDLE=20`** on both loops (≈60 min at the 180 s poll cadence) — wide enough that a single slow but live counterpart round (codex review on a large PR, or a developer round of edit + test + push) does not falsely trip terminate-stalled, narrow enough that a truly dead counterpart still exits within the hour:
  - `resolve`: `round-pre.sh` increments `consecutive_idle` when `action=idle`, resets on `continue`; threshold flips the action to `terminate-stalled`. `SKILL.md` adds `terminate-stalled` to the existing terminal-action branch (writes retrospective, no `ScheduleWakeup`). The constant is hardcoded in the script (not stored in `meta.json`) so a state file from an older skill version with a stricter threshold cannot silently override the current floor.
  - `review`: `loop.sh` increments `consecutive_idle` on the no-new-activity branch, resets after a successful codex round; threshold logs + sets `status=stalled` + `exit 5`.
- **Wall-clock backstop on `review/loop.sh`** (`MAX_WALL_CLOCK_HOURS=24`): independent of round/idle accounting — terminates with `status=runaway` and `exit 6` if a logic bug ever skips both counters. `resolve-loop` doesn't need this; it inherits the host Claude session's lifetime.

Together with the existing `terminate-success` / `terminate-diverged` / `terminate-external` paths, every loop now has four (review: five) independent exit conditions and cannot deadlock or livelock indefinitely.

## Test plan

- [x] `bash -n` on all four edited scripts.
- [ ] Verify `terminate-stalled` fires after 20 idle rounds on a quiet PR (manual).
- [ ] Verify `consecutive_idle` resets to 0 after a productive round.
- [ ] Confirm review-loop wall-clock cap doesn't fire under normal use (no PR runs >24 h in practice).

### DIFF
--- a/.claude/skills/resolve-tileops/round-pre.sh
+++ b/.claude/skills/resolve-tileops/round-pre.sh
@@ -48,7 +48,7 @@ LAST_REVIEW_COMMENT_ID_PREV=$(jq -r '.last_processed_review_comment_id' "$META")
 # continue. Hitting max_idle terminates the loop so a dead counterpart
 # (e.g. review-loop crashed) doesn't leave us polling forever.
 CONSECUTIVE_IDLE=$(jq -r '.consecutive_idle // 0' "$META")
-MAX_IDLE=$(jq -r '.max_idle // 5' "$META")
+MAX_IDLE=$(jq -r '.max_idle // 20' "$META")
 
 PR_JSON=$(gh pr view "$PR" --repo "$REPO" --json state,headRefOid,isDraft 2>/dev/null) \
   || { echo "round-pre: gh pr view failed" >&2; exit 1; }
--- a/.claude/skills/resolve-tileops/round-pre.sh
+++ b/.claude/skills/resolve-tileops/round-pre.sh
@@ -44,6 +44,11 @@ ROUND=$(jq -r '.round' "$META")
 MAX_ROUNDS=$(jq -r '.max_rounds' "$META")
 LAST_REVIEW_ID_PREV=$(jq -r '.last_processed_review_id' "$META")
 LAST_REVIEW_COMMENT_ID_PREV=$(jq -r '.last_processed_review_comment_id' "$META")
+# Stall safety net. Increment on idle (no progress this round), reset on
+# continue. Hitting max_idle terminates the loop so a dead counterpart
+# (e.g. review-loop crashed) doesn't leave us polling forever.
+CONSECUTIVE_IDLE=$(jq -r '.consecutive_idle // 0' "$META")
+MAX_IDLE=$(jq -r '.max_idle // 5' "$META")
 
 PR_JSON=$(gh pr view "$PR" --repo "$REPO" --json state,headRefOid,isDraft 2>/dev/null) \
   || { echo "round-pre: gh pr view failed" >&2; exit 1; }
@@ -127,6 +132,24 @@ if [[ -z "$ACTION" \
 fi
 
 [[ -z "$ACTION" ]] && ACTION="continue"
+
+# Stall counter: increment on idle, reset on continue. If idle persists
+# beyond max_idle, escalate to terminate-stalled — protects against the
+# review-loop dying silently while we poll forever.
+if [[ "$ACTION" == "idle" ]]; then
+  NEW_IDLE=$((CONSECUTIVE_IDLE + 1))
+  if (( NEW_IDLE >= MAX_IDLE )); then
+    ACTION="terminate-stalled"
+    MESSAGE="No reviewer activity for $NEW_IDLE consecutive rounds (max_idle=$MAX_IDLE) — terminating."
+  fi
+  jq --argjson n "$NEW_IDLE" '.consecutive_idle=$n' "$META" \
+    > "$META.tmp" && mv "$META.tmp" "$META"
+elif [[ "$ACTION" == "continue" ]]; then
+  if (( CONSECUTIVE_IDLE != 0 )); then
+    jq '.consecutive_idle=0' "$META" \
+      > "$META.tmp" && mv "$META.tmp" "$META"
+  fi
+fi
 
 NEXT_ROUND=$((ROUND + 1))
 SNAP_PREFIX=""
--- a/.claude/skills/resolve-tileops/round-pre.sh
+++ b/.claude/skills/resolve-tileops/round-pre.sh
@@ -45,10 +45,13 @@ MAX_ROUNDS=$(jq -r '.max_rounds' "$META")
 LAST_REVIEW_ID_PREV=$(jq -r '.last_processed_review_id' "$META")
 LAST_REVIEW_COMMENT_ID_PREV=$(jq -r '.last_processed_review_comment_id' "$META")
 # Stall safety net. Increment on idle (no progress this round), reset on
-# continue. Hitting max_idle terminates the loop so a dead counterpart
-# (e.g. review-loop crashed) doesn't leave us polling forever.
+# continue. Hitting MAX_IDLE terminates the loop so a dead counterpart
+# (e.g. review-loop crashed) doesn't leave us polling forever. Hardcoded
+# rather than read from meta.json so a state file from an older skill
+# version with a stricter threshold doesn't silently override the
+# current floor.
 CONSECUTIVE_IDLE=$(jq -r '.consecutive_idle // 0' "$META")
-MAX_IDLE=$(jq -r '.max_idle // 20' "$META")
+MAX_IDLE=20
 
 PR_JSON=$(gh pr view "$PR" --repo "$REPO" --json state,headRefOid,isDraft 2>/dev/null) \
   || { echo "round-pre: gh pr view failed" >&2; exit 1; }
--- a/.claude/skills/review-tileops/loop.sh
+++ b/.claude/skills/review-tileops/loop.sh
@@ -26,10 +26,14 @@ MAX_ROUNDS=15
 POLL_INTERVAL=180
 CODEX_RETRY=3
 # Stall safety: terminate after this many consecutive idle polls (no new
-# commits / comments). With POLL_INTERVAL=180s, MAX_IDLE=5 ≈ 15min — short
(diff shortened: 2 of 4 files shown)


.claude/skills/resolve-tileops/round-pre.sh
```diff
@@ -128,6 +133,24 @@ fi
 
 [[ -z "$ACTION" ]] && ACTION="continue"
 
+# Stall counter: increment on idle, reset on continue. If idle persists
+# beyond max_idle, escalate to terminate-stalled — protects against the
+# review-loop dying silently while we poll forever.
+if [[ "$ACTION" == "idle" ]]; then
+  NEW_IDLE=$((CONSECUTIVE_IDLE + 1))
+  if (( NEW_IDLE >= MAX_IDLE )); then
```
Ibuki-wind, line 144: The `MAX_IDLE` threshold treats five rounds without reviewer GitHub activity as proof that the reviewer loop died, but a normal reviewer round produces no review/comment activity until Codex finishes and submits. If that review takes more than about 15 minutes, `round-pre.sh` returns `terminate-stalled` and stops scheduling resolve before the next reviewer feedback can be handled. -> Base this stall decision on an explicit reviewer heartbeat/status, or use a timeout that cannot fire during a legitimate in-progress review round.

.claude/skills/review-tileops/loop.sh
```diff
@@ -552,7 +572,16 @@ while true; do
   if [[ "$ROUND" -gt 0 \
         && "$HEAD_SHA" == "$LAST_SHA" \
         && "$LATEST_HUMAN_ID" == "$LAST_HUMAN_ID_PREV" ]]; then
-    log "no new commits / comments — sleeping ${POLL_INTERVAL}s"
+    CONSECUTIVE_IDLE=$(jq -r '.consecutive_idle // 0' "$META")
+    NEW_IDLE=$((CONSECUTIVE_IDLE + 1))
+    jq --argjson n "$NEW_IDLE" '.consecutive_idle=$n' "$META" \
+      > "$META.tmp" && mv "$META.tmp" "$META"
+    if (( NEW_IDLE >= MAX_IDLE )); then
```
Ibuki-wind, line 583: The review loop has the same false-stall path after `REQUEST_CHANGES`: no new commits or human comments for five polls can simply mean the resolve loop is still editing, testing, or pushing. Exiting here leaves no reviewer loop to process the fix if the resolve round finishes after ~15 minutes. -> Gate this on a resolve-loop heartbeat/status, or keep polling under a much larger wall-clock bound instead of treating five quiet polls as terminal.

Ibuki-wind (changes requested):
### Overall
The new stall detectors can terminate a live review/resolve pair after 15 minutes of legitimate counterpart work; keep the loop alive until an explicit heartbeat/status or a safer timeout distinguishes that from a dead peer.

.claude/skills/resolve-tileops/round-pre.sh
Ibuki-wind, line 144: The `MAX_IDLE` threshold in `round-pre.sh` treats five `idle` rounds without reviewer GitHub activity as proof that the reviewer loop died, but a normal reviewer round produces no review/comment activity until Codex finishes and submits. If that review takes more than about 15 minutes, this branch returns `terminate-stalled` and stops scheduling resolve before the next reviewer feedback can be handled. → Base this stall decision on an explicit reviewer heartbeat/status, or use a timeout that cannot fire during a legitimate in-progress review round.

.claude/skills/review-tileops/loop.sh
Ibuki-wind, line 583: The `MAX_IDLE` branch in `loop.sh` has the same false-stall path after `REQUEST_CHANGES`: no new commits or human comments for five polls can simply mean the resolve loop is still editing, testing, or pushing. Exiting here leaves no reviewer loop to process the fix if the resolve round finishes after about 15 minutes. → Gate this on a resolve-loop heartbeat/status, or keep polling under a much larger wall-clock bound instead of treating five quiet polls as terminal.

Ibuki-wind (changes requested):
### Overall
Blocking idle-detection issue remains; next step is to make the stall condition distinguish inactive peers from active long-running rounds.

.claude/skills/resolve-tileops/round-pre.sh
lcy-seso: Adopted. Raised MAX_IDLE from 5 (15 min) to 20 (60 min). 60 min comfortably exceeds the longest realistic in-progress counterpart round (slow codex review, developer edit+test+push); wall-clock cap (24 h) still bounds absolute runaway. See de2c0f8.

.claude/skills/review-tileops/loop.sh
lcy-seso: Adopted. Raised MAX_IDLE from 5 (15 min) to 20 (60 min) symmetrically on the review side. Same rationale as the resolve thread. See de2c0f8.

.claude/skills/resolve-tileops/round-pre.sh
```diff
@@ -44,6 +44,11 @@ ROUND=$(jq -r '.round' "$META")
 MAX_ROUNDS=$(jq -r '.max_rounds' "$META")
 LAST_REVIEW_ID_PREV=$(jq -r '.last_processed_review_id' "$META")
 LAST_REVIEW_COMMENT_ID_PREV=$(jq -r '.last_processed_review_comment_id' "$META")
+# Stall safety net. Increment on idle (no progress this round), reset on
+# continue. Hitting max_idle terminates the loop so a dead counterpart
+# (e.g. review-loop crashed) doesn't leave us polling forever.
+CONSECUTIVE_IDLE=$(jq -r '.consecutive_idle // 0' "$META")
+MAX_IDLE=$(jq -r '.max_idle // 20' "$META")
```
Ibuki-wind: `MAX_IDLE` still preserves a previously initialized `max_idle:5` from `meta.json`, because `preflight.sh` exits immediately on existing state and this fallback only applies when the field is missing; active resolve loops created before this patch can still terminate after 15 minutes → migrate existing state to the new floor before returning from preflight, or clamp stale lower values when computing `MAX_IDLE`.

Ibuki-wind (changes requested):
### Overall
One idle-timeout migration gap remains; existing resolve state can still use the old false-stall threshold.

.claude/skills/resolve-tileops/round-pre.sh
lcy-seso: Already addressed in de2c0f8 — MAX_IDLE raised to 20 (60 min). Migration gap for stale state files just fixed in 331080f (max_idle dropped from meta; constant now hardcoded in round-pre.sh).

.claude/skills/review-tileops/loop.sh
lcy-seso: Already addressed in de2c0f8 — MAX_IDLE raised to 20 (60 min) symmetrically on the review side.

.claude/skills/resolve-tileops/round-pre.sh
lcy-seso: Adopted. Dropped `max_idle` from meta.json entirely and hardcoded MAX_IDLE=20 in round-pre.sh. Skill version is now always authoritative — old state files with the stricter threshold no longer slip through. See 331080f.

Ibuki-wind (approved):
Clean — no issues.

.claude/skills/review-tileops/loop.sh
```diff
@@ -25,6 +25,20 @@ REPO="tile-ai/TileOPs"
...
+# = 60min — wide enough that a single slow but live counterpart round
+# does not trip terminate-stalled, narrow enough that a truly dead
+# counterpart still exits within the hour. Wall-clock cap below caps
+# absolute runaway risk.
+MAX_IDLE=20
+# Belt-and-suspenders: hard wall-clock cap independent of round/idle
+# accounting, in case a logic bug skips both. Far above any realistic PR
+# lifetime (hours), low enough that a runaway is bounded to a day.
+MAX_WALL_CLOCK_HOURS=24
+LOOP_START_EPOCH=$(date +%s)
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Updated PR description to reflect MAX_IDLE=20 (60 min). The bump from 5 to 20 was made in earlier rounds of this same review based on the false-stall feedback; the description was the leftover.

.claude/skills/resolve-tileops/round-pre.sh
```diff
@@ -44,6 +44,14 @@ ROUND=$(jq -r '.round' "$META")
...
 LAST_REVIEW_ID_PREV=$(jq -r '.last_processed_review_id' "$META")
 LAST_REVIEW_COMMENT_ID_PREV=$(jq -r '.last_processed_review_comment_id' "$META")
+# Stall safety net. Increment on idle (no progress this round), reset on
+# continue. Hitting MAX_IDLE terminates the loop so a dead counterpart
+# (e.g. review-loop crashed) doesn't leave us polling forever. Hardcoded
+# rather than read from meta.json so a state file from an older skill
+# version with a stricter threshold doesn't silently override the
+# current floor.
+CONSECUTIVE_IDLE=$(jq -r '.consecutive_idle // 0' "$META")
+MAX_IDLE=20
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Updated PR description to reflect MAX_IDLE=20. Same root cause — description not updated when the constant was raised earlier in this review.

PR #1172, opened 2026-05-04
[Feat][Ops] Add packed varlen GQA prefill

### LINKED ISSUE #1096: [FEAT][Ops] track GQA prefill release roadmap and interface scope
## Summary

This issue tracks the GQA prefill release roadmap, interface scope, validation plan, and sub-issue breakdown.

Modern LLM serving prefill is not a single dense attention operator. Real serving stacks split prefill along several independent dimensions: input layout, KV cache layout, cache update contract, position semantics, score modifiers, numeric formats, specialized dispatch, benchmarks, and optional stats. The goal of this umbrella is to define the release-facing TileOPs contract clearly enough that implementation PRs and agent-assisted work can proceed without drifting into incompatible one-off APIs.

This issue should stay open until the release-facing GQA prefill scope is complete and validated. Individual implementation PRs should reference the relevant sub-issue and this umbrella, but should not close this umbrella directly.

## Problem Statement

TileOPs needs a GQA prefill operator family that can support modern serving use cases without pushing runtime-specific complexity into users. The public contract should make these distinctions explicit:

- Dense prefill vs cache-aware prefill.
- Fixed-length batch vs packed varlen batch.
- Contiguous KV cache vs paged KV cache.
- External position handling vs fused position handling.
- Plain fp16/bf16 cache vs quantized KV cache.
- Generic correctness path vs H200/Hopper-specialized performance path.

The operator family should expose stable OP-level entry points, while kernel variants remain dispatch details.

## Surveyed Systems

The feature matrix is informed by these systems/APIs:

- FlashAttention / FlashAttention varlen APIs
- vLLM paged attention and KV cache interfaces
- FlashInfer prefill / paged KV APIs
- TensorRT-LLM context attention and KV cache configuration
- xFormers memory-efficient attention variants
- cuDNN frontend SDPA / attention graph APIs
- PyTorch SDPA / GQA-facing attention semantics

TileOPs does not need to mirror any single project. The goal is to cover the same serving-relevant dimensions with TileOPs-native OP contracts.

## Prefill Variant Dimensions

| Dimension | Meaning | Required Scope |
| --- | --- | --- |
| Query/KV shape | Whether Q and KV lengths match | `q_len == kv_len`, `q_len < kv_len` |
| Batch shape | Whether requests share lengths | dense BSHD, packed varlen THD |
| KV storage | Where visible KV comes from | no cache, contiguous cache, paged cache |
| KV update | Whether current chunk is appended | no update, fused append into cache/pages |
| Head mapping | MHA/GQA/MQA expression | `heads`, `heads_kv`, `heads % heads_kv == 0` |
| Mask semantics | Who can attend to whom | causal, non-causal, bottom-right causal alignment |
| Position semantics | Cross-chunk position alignment | external RoPE, fused RoPE, absolute cache positions |
| Score modifiers | Pre-softmax score transforms | `sm_scale`, `softcap`, later temperature/bias |
| Numeric format | Compute/cache dtype choices | fp16, bf16, FP8 KV cache |
| Dispatch | Hardware-specific optimized paths | generic, Hopper/H200 specialized |
| Benchmarks | Release readiness and regression tracking | correctness-adjacent benchmark matrix |
| Returned stats | Auxiliary softmax stats | output-only default, `return_lse` later if needed |

## How To Use This Issue

This umbrella is the source of release scope. Sub-issues and PRs should be narrower.

Each sub-issue should include:

- `Parent: #1096`
- related feature issues if the work composes with another feature
- task layer:
  - L1 kernel work
  - L2 OP interface / validation / unit tests
  - benchmark / performance work
- public OP contract:
  - OP name
  - tensor inputs / outputs
  - scalar params
  - shape/layout expectations
  - dtype support
  - semantic constraints
- validation expectations:
  - reference parity
  - append correctness where applicable
  - old cache/page prefix unchanged where applicable
  - invalid input checks
  - fp16/bf16/FP8 tolerance expectations
- non-goals for that issue

Implementation PRs should:

- reference the relevant sub-issue
- include test evidence for that sub-issue's validation checklist
- include the minimal benchmark/manifest closure for any release-facing OP contract it adds or changes; use #1102 only for release-level matrix expansion and H200/Hopper performance follow-up
- update manifest/workloads/source metadata if the PR introduces or changes a release-facing OP contract
- not close #1096 unless the whole roadmap is complete

## Manifest / Skill Alignment

Repo workflow expects structured contracts, not free-form implementation intent.

The issue should provide the human contract; `ops_manifest.yaml` should remain the machine-readable source of truth when a release-facing OP is added or re-aligned.

When a GQA prefill PR reaches the manifest/skill workflow, the issue should provide enough information for agent skills to produce high-quality work:

- `add-manifest` / `fix-manifest`
  - need OP names, ref API status, inputs, outputs, params, dtype combos, shape rules, workloads, source paths, and benchmark intent
- `align-op`
  - needs a clear per-OP target contract and whether the change is greenfield, redesign, or minor alignment
- `test-op`
  - needs PyTorch/reference semantics, dtype tolerances, shape cases, and failure cases
- `implement-op`
  - needs the manifest/interface contract and clear non-goals
- `bench-op`
  - needs declared workloads and benchmark dimensions instead of hardcoded ad-hoc shapes
- `audit-family` / `align-family`
  - need family-level expectations and cross-op consistency rules

Quality comes from the chain: issue contract -> manifest/workloads -> generated tests/implementation/benchmarks -> validator/CI. The issue should therefore be written in a manifest-like way even when the exact YAML entry is introduced later.

## Sub-Issue Breakdown

Existing/planned sub-issues:

- #1097: dense and contiguous-cache prefill baseline
- #1098: packed varlen GQA prefill
- #1099: paged KV-cache GQA prefill
- #1100: RoPE position semantics for GQA prefill
- #1101: fused RoPE and softcap coverage added in the current implementation batch
- #1102: benchmarks and H200/Hopper dispatch coverage
- #1103: FP8 KV-cache support

This list may grow if a feature needs a smaller review slice. The umbrella should remain the release-level tracker.

## Manifest PR Sequencing

The preferred manifest merge order should minimize repeated edits to the same OP entries:

1. Land the baseline entries from #1097, #1098, and #1099 first if they are ready independently. These entries may include only the baseline params and kernel map needed by that slice.
2. If #1101 lands in the same manifest PR as the baseline entries, include `softcap` on dense / varlen / contiguous-cache / paged-cache entries and include `fuse_rope`, `rope_base`, `max_position`, and the fused-RoPE kernel-map entries for cache-aware variants.
3. If #1101 lands later, its PR should explicitly update the existing #1097 / #1098 / #1099 manifest entries. This is expected and should not create new public OP variants.
4. Add #1103 FP8 KV-cache entries after the baseline cache contracts are stable. FP8 should mirror #1101 params only for the FP8 variants whose kernels support them in that PR.
5. Use #1102 to expand workloads, complete roofline formulas, refine benchmark protocols, and validate H200/Hopper dispatch after the minimal feature-level manifest/test/benchmark closure exists. Hardware dispatch changes should stay behind the existing public OP entries unless they introduce a new release-facing contract.

This ordering is a planning rule, not a blocker: a combined manifest PR may include multiple steps when the corresponding implementation and tests are ready together.

### Release-Facing OP Family

The public GQA prefill family should expose these OP-level entry points:

- `GroupedQueryAttentionPrefillFwdOp`
  - dense BSHD prefill
  - supports `seq_len_q != seq_len_kv`
  - causal mask uses bottom-right alignment

- `GroupedQueryAttentionPrefillWithKVCacheFwdOp`
  - BSHD current chunk with contiguous KV cache
  - reads old KV from `[B, Skv_cap, Hkv, D]` cache
  - reads current chunk from `k_new/v_new`
  - appends current chunk into cache in-place

- `GroupedQueryAttentionPrefillVarlenFwdOp`
  - packed THD varlen prefill
  - uses `cu_seqlens_q` / `cu_seqlens_kv`
  - supports heterogeneous request lengths in one batch

- `GroupedQueryAttentionPrefillPagedWithKVCacheFwdOp`
  - packed current chunk with paged KV cache
  - uses `cache_seqlens` and `block_table`
  - appends current chunk into physical KV pages in-place

### Required Functional Scope

- GQA / MQA / MHA through `heads` and `heads_kv`.
- fp16 and bf16 coverage.
- Dense prefill with `q_len == kv_len` and `q_len < kv_len`.
- Contiguous KV cache prefill with in-place append.
- Varlen packed prefill with per-request q/kv lengths.
- Paged KV cache prefill with per-request page tables.
- Per-batch `cache_seqlens` and uneven old cache lengths.
- Causal and non-causal modes where applicable.
- Bottom-right causal alignment for `q_len != kv_len`.
- Output-only public OP contract by default.
- Kernel-internal `lse` may exist, but exposing `return_lse` remains a lower-priority follow-up.

### Position / RoPE Contract

External RoPE path:

- Callers may rotate `q` / `k` outside the attention OP.
- Cache K is treated as already rotated.
- For cache prefill, current chunk positions are `cache_seqlens[b] + local_i`.
- Appended K is already rotated.

Fused RoPE path:

- Current chunk `q/k_new` may be rotated inside the prefill kernel.
- Old cache K must not be rotated again.
- Appended K must be the rotated current `k_new`.
- Contiguous KV cache and paged KV cache should both support this contract.

Suggested public knobs:

- `fuse_rope: bool = False`
- `max_position: Optional[int] = None`
- `rope_base: float = 10000.0`

### Score Modifier Contract

Required score modifier knobs:

- `sm_scale: Optional[float] = None`
  - default is `1 / sqrt(head_dim)`

- `softcap: Optional[float] = None`
  - `None` or `0` means disabled
  - `softcap > 0` applies score capping before softmax:
    `softcap * tanh(score / softcap)`

Future score modifiers that are not required for the first release pass:

- temperature
- generic attention bias
- arbitrary mask / block mask extension

### Numeric Format / FP8 KV Cache

FP8 KV cache is part of this umbrella roadmap and tracked by #1103.

Required design questions:

- cache storage dtype
- whether current `k_new/v_new` are passed as fp16/bf16 and quantized on append, or passed already quantized
- whether dequantization happens inside the attention kernel or before the OP
- scale granularity:
  - per tensor
  - per head
  - per page
  - per block / token group
- scale storage layout for contiguous cache and paged cache
- old cache read path and current chunk read path consistency
- append correctness for K and V
- interaction with fused RoPE:
  - default expectation: RoPE applies before K is quantized into cache unless a PR proposes a different contract explicitly

### H200 / Hopper Performance Scope

H200/Hopper specialized kernels are in scope and tracked by #1102.

- WS / persistent kernels may be added or refined where beneficial.
- TMA-friendly paths may be added where beneficial.
- Specialized kernels must stay behind OP-layer dispatch.
- Public OP signatures and semantics must not change for performance-specialized dispatch.
- Correctness must be validated against the same feature matrix before enabling dispatch.

### Benchmark Scope

Benchmark coverage is part of release readiness. Each feature issue owns the minimal benchmark/manifest closure for the OP contract it introduces or changes; #1102 tracks release-level benchmark matrix expansion, roofline completion, and H200/Hopper performance validation.

Benchmark work should cover:

- dense prefill
- contiguous-cache prefill
- varlen packed prefill
- paged-cache prefill
- external / fused RoPE paths
- `sm_scale` / `softcap` paths
- FP8 KV cache once available
- H200/Hopper specialized dispatch paths

Benchmarks can be implemented in separate PRs from functional code, but benchmark coverage is not out of scope for the release.

### Validation Expectations

Tests should cover:

- dense prefill output correctness
- contiguous cache prefill output correctness
- contiguous cache append correctness
- varlen packed output correctness
- paged cache output correctness
- paged cache append correctness
- old cache/page prefix remains unchanged
- external RoPE absolute-position correctness
- fused RoPE correctness for current chunk and append
- `sm_scale` correctness
- `softcap` correctness
- fp16 and bf16 smoke coverage
- FP8 KV cache reference parity within explicit tolerance
- FP8 KV cache append / scale correctness
- invalid shape / dtype / capacity checks for public OP contracts

## Release Plan

Suggested staged plan:

1. Dense GQA prefill baseline
   - Dense BSHD OP.
   - `q_len != kv_len`.
   - Bottom-right causal alignment.
   - fp16/bf16 tests.

2. Contiguous KV cache prefill
   - Read old cache + current chunk.
   - Fused append.
   - Capacity / shape validation.
   - `sm_scale` coverage.

3. Varlen packed prefill
   - Packed THD layout.
   - `cu_seqlens_q` / `cu_seqlens_kv`.
   - Heterogeneous q/kv lengths.

4. Paged KV cache prefill
   - `block_table`-based physical page lookup.
   - Fused append into pages.
   - Page-boundary tests.

5. Position semantics / RoPE
   - External RoPE absolute position path.
   - Contiguous fused RoPE.
   - Paged fused RoPE.
   - Explicit old-cache-not-re-rotated contract.

6. Score modifiers
   - `sm_scale`.
   - `softcap`.
   - Keep temperature/bias/mask extension as follow-up unless needed by a concrete integration.

7. FP8 KV cache
   - Define FP8 cache contract.
   - Contiguous FP8 KV cache first.
   - Paged FP8 KV cache next.
   - Validate output parity, append correctness, and scale layout.

8. H200 / Hopper performance-specialized kernels
   - Add or refine WS / persistent / TMA-friendly kernels where they are beneficial.
   - Keep these behind OP-layer dispatch; they should not change the public OP contract.
   - Validate correctness against the same feature matrix before enabling dispatch.

9. Benchmark and release cleanup
   - Public API docs.
   - Error messages and input validation.
   - CI coverage selection.
   - Add benchmark coverage for dense, contiguous-cache, varlen, paged-cache, RoPE, softcap, and FP8 KV cache paths.
   - Track baseline throughput/latency on the target release GPUs, including H200-specific dispatch paths.
   - Release benchmark expansion and tuning can live in #1102-scoped PRs, but each feature PR must include the minimal benchmark/manifest closure for any release-facing OP contract it adds or changes.

## Out of Scope For This Umbrella

These can be tracked separately or deferred:

- int8 KV cache
- page allocation / eviction / reuse policy
- prefix-cache runtime integration
- scheduler behavior
- public `return_lse` API unless required by a concrete downstream use case

Implementation details for the above items should be discussed in their corresponding PRs. PRs should reference this issue, but should not close it until the overall release scope is complete.

### LINKED ISSUE #1098: [FEAT][Ops] add packed varlen GQA prefill
Parent: #1096

Related feature issues: #1097, #1099, #1100, #1101, #1102, #1103

## Scope

Track packed variable-length GQA prefill support.

This issue covers the public packed varlen prefill OP and its manifest contract. It is a new public prefill entry and therefore requires a manifest entry: packed THD layout, `cu_seqlens` metadata, and max sequence params are part of the public OP contract.

### `GroupedQueryAttentionPrefillVarlenFwdOp`

Purpose:

- Packed THD GQA prefill.
- Supports heterogeneous request lengths in one batch.
- Uses `cu_seqlens_q` / `cu_seqlens_kv` metadata.
- Supports `q_len != kv_len` per request.
- Uses bottom-right causal alignment per request.
- Covers MHA/GQA/MQA through `heads` and `heads_kv`.
- Returns output only.

## Manifest Change Plan

This issue requires manifest changes.

Required manifest entry:

- Add `GroupedQueryAttentionPrefillVarlenFwdOp`.

The entry must declare:

- `family: attention`
- `ref_api: none` unless a reference exactly matches packed varlen prefill semantics
- ordered `signature.inputs`, `signature.outputs`, and `signature.params`
- THD packed layout shape rules
- `dtype_combos` only if the supported set is narrower than declared unions
- benchmark `workloads`
- `roofline`
- `source.kernel`, `source.kernel_map`, `source.op`, `source.test`, and `source.bench`

## Manifest Field Sketch

This is not copy-paste YAML; it is the field contract the manifest PR should encode using the repo's ordered-dict schema.

Manifest key: `GroupedQueryAttentionPrefillVarlenFwdOp`

Inputs:

- `q`: dtype `float16 | bfloat16`, shape `[T_q, H, D]`
- `k`: dtype `same_as(q)`, shape `[T_kv, H_kv, D]`
- `v`: dtype `same_as(q)`, shape `[T_kv, H_kv, D]`
- `cu_seqlens_q`: dtype `int32`, shape `[N_cu]`
- `cu_seqlens_kv`: dtype `int32`, shape `[N_cu]`

Output:

- `o`: dtype `same_as(q)`, shape `[T_q, H, D]`

Params:

- `batch`: `{type: int}`; required manifest param for the packed request count
- `max_seqlen_q`: `{type: int}`
- `max_seqlen_kv`: `{type: int}`
- `is_causal`: `{type: bool, default: true}`
- `sm_scale`: `{type: float, default: null}`
- `softcap`: `{type: float, default: null}` if #1101 lands in the same manifest PR

Softcap semantics must match #1101: `softcap: null` and `softcap: 0` both disable score capping and preserve existing behavior; `softcap > 0` applies score capping before softmax; negative values are rejected by OP runtime validation.

Shape rules:

- `H % H_kv == 0`
- `cu_seqlens_q.shape == (N_cu,)`
- `cu_seqlens_kv.shape == (N_cu,)`
- `N_cu == batch + 1`
- `cu_seqlens_q[0] == 0` and `cu_seqlens_q[-1] == T_q` are runtime validation rules
- `cu_seqlens_kv[0] == 0` and `cu_seqlens_kv[-1] == T_kv` are runtime validation rules
- every `q_len <= kv_len` when `is_causal=True` is runtime validation
- `max_seqlen_q >= max(q_lens)` and `max_seqlen_kv >= max(kv_lens)` are runtime validation rules

Proposed source:

- `source.kernel`: `tileops/kernels/attention/gqa_sliding_window_varlen_fwd.py` or the final varlen prefill kernel file used by the implementation
- `source.kernel_map`: `gqa_prefill_varlen_fwd_kernel: <final varlen prefill kernel class>`
- `source.op`: `tileops/ops/attention/gqa.py`
- `source.test`: `tests/ops/attention/test_gqa_prefill_varlen.py`
- `source.bench`: `benchmarks/ops/attention/bench_gqa.py` or a dedicated varlen benchmark file if introduced

## Workloads / Roofline

Manifest workloads should use `<tensor_name>_shape`, `dtypes`, and `label` according to `docs/manifest.md`.

Example workload fields:

```yaml
- {
    q_shape: [2048, 32, 128],
    k_shape: [4096, 8, 128],
    v_shape: [4096, 8, 128],
    cu_seqlens_q_shape: [5],
    cu_seqlens_kv_shape: [5],
    batch: 4,
    max_seqlen_q: 512,
    max_seqlen_kv: 1024,
    is_causal: true,
    dtypes: [float16],
    label: "llama-3.1-8b-prefill-varlen-mixed",
  }
```

The benchmark generator must define how it expands aggregate shapes into concrete per-request lengths. The manifest PR should either:

- document a deterministic generator that uses `batch`, `T_q`, `T_kv`, `max_seqlen_q`, and `max_seqlen_kv` to produce `q_lens` / `kv_lens` whose sums exactly match `T_q` / `T_kv`; or
- introduce a dedicated varlen workload protocol that consumes explicit non-OP metadata such as `q_lens`, `kv_lens`, `avg_seqlen_q`, or `avg_seqlen_kv` without forwarding those fields to the OP constructor.

The roofline function in `tileops/perf/formulas.py` must document the same binding in its docstring. It should account for per-request QK and ScoreV/PV using actual or generated per-request lengths, plus Q/K/V/O memory traffic and `cu_seqlens` metadata reads.

## Validation Expectations

The corresponding PR should show:

- output correctness for equal q/kv lengths
- output correctness for mixed `q_len < kv_len`
- causal and non-causal coverage where applicable
- MHA / GQA / MQA coverage
- fp16 / bf16 smoke coverage
- invalid q/kv length validation
- invalid max sequence length validation
- `cu_seqlens` monotonicity and endpoint validation
- tail shapes where `T_q` / `T_kv` are not tile multiples

## PR Notes

Implementation details should be discussed in the PR that references this issue.
Do not close #1096 from that PR.

superAngGao:
## Summary
- add `GroupedQueryAttentionPrefillVarlenFwdOp` for packed THD GQA prefill with `cu_seqlens_q/cu_seqlens_kv`
- add a dedicated varlen prefill kernel with per-request bottom-right causal masking and `sm_scale` support
- add manifest entry, roofline formula, workload generator, correctness tests, validation tests, and manifest-driven benchmark coverage

## Validation
- ruff check on touched Python files
- git diff --check
- `python3 scripts/validate_manifest.py --levels schema,shape,dtype,bench --check-op GroupedQueryAttentionPrefillVarlenFwdOp`
- `python3 -m pytest -q tests/ops/attention/test_gqa.py -k "prefill_varlen"`
- `python3 -m pytest -q tests/ops/attention/test_gqa.py -k "prefill"`
- `python3 -m pytest -q benchmarks/ops/attention/bench_gqa.py -k "test_gqa_prefill_varlen_fwd_bench and llama-3.1-8b-prefill-varlen-uniform-float16" --tb=short`

Closes #1098
Does not close #1096.


Gabbering (commented):
## goose goose review — `8d164b8f`

*honk.* Solid varlen prefill implementation. The kernel logic, validation, and test coverage are well-structured. But the goose found something worth biting.

### feather Performance

- `tileops/ops/attention/gqa.py:388-393` — `_validate_forward_inputs` calls `cu_seqlens_q.detach().cpu().tolist()` and `cu_seqlens_kv.detach().cpu().tolist()` on every single `forward()` call. That's two device-to-host synchronizations on the critical path. For inference serving where this op gets called per-batch, you're stalling the CUDA pipeline to validate monotonicity and compute `q_lens`/`kv_lens` on the CPU. The validation is thorough and correct, but the sync cost is real — especially under high-throughput batched inference where you'd call this hundreds of times per second. Consider either (a) gating the expensive CPU-side checks behind a debug/validation flag, or (b) at minimum caching the result if the same `cu_seqlens` tensors are reused across calls (they often are in serving frameworks). The `_roofline_kwargs` path also depends on this, so you can't trivially skip it, but the monotonicity checks and per-element iteration could be deferred to a `validate=True` mode.

- `tileops/kernels/attention/gqa_prefill_varlen_fwd.py:37` — The `@functools.lru_cache(maxsize=32)` on `_gqa_prefill_varlen_fwd_kernel` caches on `(batch, heads, heads_kv, total_q, total_kv, dim, is_causal, sm_scale, dtype)`. In varlen workloads, `total_q` and `total_kv` change with every batch composition. This means the cache will almost never hit in production serving — every new batch packing produces different totals. The kernel gets recompiled (or at least re-looked-up) every time. This isn't a correctness bug, but the cache is giving you a false sense of JIT amortization. The `total_q`/`total_kv` parameters are only used to declare tensor shapes in the TileLang function signature — if TileLang requires them statically, you're stuck, but if there's a way to use symbolic/dynamic shapes, that would make the cache actually useful.

### egg Test gaps

- No test covers the case where `q_len == kv_len` with non-causal mode and non-block-aligned lengths simultaneously. The existing `mixed-noncausal-fp16` test uses `q_lens=[64, 96], kv_lens=[128, 160]` (q < kv). A case like `q_lens=[33, 65], kv_lens=[33, 65]` non-causal would exercise the boundary-padding logic in both Q and KV tile loads with equal-length non-aligned sequences — a combination the current parametrization misses.

- No test exercises `batch=1`. All varlen tests use batch >= 2. A single-sequence batch is the degenerate case where the varlen kernel should reduce to the dense prefill behavior — worth a quick sanity check that the grid launch with `bz=0` only doesn't trip anything.

tileops/ops/attention/gqa.py
```diff
@@ -235,6 +241,185 @@ def forward(self, q: torch.Tensor, k: torch.Tensor, v: torch.Tensor) -> torch.Te
...
+            tensor = tensors[name]
+            expected_shape = (self.batch + 1,)
+            if tuple(tensor.shape) != expected_shape:
+                raise ValueError(
+                    f"Expected {name} shape {expected_shape}, got {tuple(tensor.shape)}")
+            if tensor.dtype != torch.int32:
+                raise ValueError(f"Expected {name}.dtype torch.int32, got {tensor.dtype}")
+
+        cu_q = [int(x) for x in cu_seqlens_q.detach().cpu().tolist()]
+        cu_kv = [int(x) for x in cu_seqlens_kv.detach().cpu().tolist()]
```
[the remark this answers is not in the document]
superAngGao: Thanks, agreed. I moved the expensive `cu_seqlens_{q,kv}.detach().cpu().tolist()` work out of the default `forward()` hot path.

The op now defaults to cheap tensor metadata checks only. Full CPU-side sequence validation is gated behind `validate_inputs=True`, and the per-request lengths needed by roofline are computed lazily in `eval_roofline()` instead of during normal execution.

tileops/kernels/attention/gqa_prefill_varlen_fwd.py
```diff
@@ -0,0 +1,351 @@
...
+                        scores_sum,
+                        logsum,
+                    )
+                    T.copy(acc_s, acc_s_cast)
+                    rescale(acc_o, scores_scale)
+                    T.gemm(acc_s_cast, v_shared, acc_o, policy=T.GemmWarpPolicy.FullRow)
+
+                if (bx + 1) * block_m <= q_len:
+                    for i, j in T.Parallel(block_m, dim):
+                        acc_o[i, j] /= logsum[i]
```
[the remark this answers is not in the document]
superAngGao: Good point. I added an `inv_logsum` fragment and changed both the full-tile and tail output paths to compute the reciprocal once per row and then multiply across `D`.

superAngGao:
Thanks for the review. I pushed `3866c19` with the follow-up fixes:

- removed the default GPU->CPU `cu_seqlens` sync from `forward()`; expensive sequence validation is now behind `validate_inputs=True`, and roofline lengths are computed lazily in `eval_roofline()`
- added the reciprocal-normalization path in the varlen kernel so we divide once per row and multiply across `D`
- added coverage for non-causal equal-length non-block-aligned varlen inputs and the `batch=1` sanity case

On the compile-cache point: for this #1098 kernel the TileLang function signature still needs `total_q`/`total_kv` as static tensor extents, so they remain part of the cached kernel key. I agree this limits cache hit rate for production-style packed varlen batches, but changing that needs a separate dynamic/symbolic-shape lowering discussion rather than being folded into this correctness/API PR.

Validation rerun:
- `pytest -q tests/ops/attention/test_gqa.py -k "prefill_varlen"` -> 9 passed
- `pytest -q tests/ops/attention/test_gqa.py -k "prefill"` -> 26 passed
- `scripts/validate_manifest.py --levels schema,shape,dtype,bench --check-op GroupedQueryAttentionPrefillVarlenFwdOp` -> passed
- benchmark smoke for `llama-3.1-8b-prefill-varlen-uniform-float16` -> passed

Gabbering (commented):
goose *goose skimmed `3866c19` — nothing to honk about.*

tileops/manifest/attention.yaml
```diff
@@ -175,6 +175,55 @@ GroupedQueryAttentionPrefillFwdOp:
     test: tests/ops/attention/test_gqa.py
     bench: benchmarks/ops/attention/bench_gqa.py
 
+GroupedQueryAttentionPrefillVarlenFwdOp:
```
Ibuki-wind: The `GroupedQueryAttentionPrefillVarlenFwdOp` manifest entry is added in the same PR that implements the op/kernel, but `.claude/rules/manifest-trust-model.md` requires manifest changes to land in a separate human-reviewed PR → remove this manifest block from the implementation PR and submit it as a manifest-only change according to that rule.

tileops/ops/attention/gqa.py
```diff
@@ -235,6 +241,197 @@ def forward(self, q: torch.Tensor, k: torch.Tensor, v: torch.Tensor) -> torch.Te
...
+        if max_seqlen_kv <= 0:
+            raise ValueError("max_seqlen_kv must be positive")
+        self.batch = batch
+        self.heads = heads
+        self.heads_kv = heads_kv
+        self.dim = dim
+        self.max_seqlen_q = max_seqlen_q
+        self.max_seqlen_kv = max_seqlen_kv
+        self.is_causal = is_causal
+        self.dtype = dtype
```
Ibuki-wind, line 288: `GroupedQueryAttentionPrefillVarlenFwdOp.__init__` stores arbitrary `dtype` and passes it into TileLang, but the manifest/tests only support fp16/bf16 → validate `dtype in {torch.float16, torch.bfloat16}` before `dispatch_kernel()` and add the unsupported-dtype rejection test required by the op-readiness checklist.

Ibuki-wind (changes requested):
### Overall
The PR still needs policy and API-boundary cleanup before it can merge.

tileops/manifest/attention.yaml
superAngGao: Agreed. I removed the `GroupedQueryAttentionPrefillVarlenFwdOp` manifest block from this implementation PR.

Since the op no longer has a manifest entry in this PR, I also changed the varlen benchmark to use a small hard-coded smoke matrix instead of collecting params from `load_workloads()`. The manifest/workload/roofline binding can land in the separate human-reviewed manifest-only PR required by the trust model.

tileops/ops/attention/gqa.py
superAngGao: Fixed. `GroupedQueryAttentionPrefillVarlenFwdOp.__init__` now rejects anything other than `torch.float16` and `torch.bfloat16` before `dispatch_kernel()`, and I added an unsupported-dtype rejection test using `torch.float32`.

superAngGao:
Pushed `850cc67` for the Ibuki review blockers:

- removed the `GroupedQueryAttentionPrefillVarlenFwdOp` manifest entry from this implementation PR per `.claude/rules/manifest-trust-model.md`
- converted the varlen benchmark from manifest-driven params to a small hard-coded smoke matrix, so collection does not depend on a manifest entry that will land separately
- added constructor-time dtype validation for `torch.float16 | torch.bfloat16` before kernel dispatch
- added an unsupported-dtype rejection test for `torch.float32`

Validation:
- `ruff check` on touched Python files -> passed
- read-only AST parse on touched Python files -> passed (`py_compile` is blocked locally by an existing `__pycache__` permission issue)
- `pytest -q tests/ops/attention/test_gqa.py -k "prefill_varlen"` in the nightly container -> 10 passed
- full `scripts/validate_manifest.py --levels schema,shape,dtype,bench` -> passed
- varlen benchmark smoke without manifest entry -> passed

Gabbering (commented):
## goose goose review — `850cc67a`

*honk.* Clean follow-up. The manifest removal, dtype gate, and benchmark decoupling are all correct. But the goose spotted something in the new benchmark math.

### feather Performance

- `benchmarks/ops/attention/bench_gqa.py:84` — `GQAPrefillVarlenFwdBenchmark.calculate_flops` computes causal visible tokens as `q_len * kv_len - q_len * (q_len - 1) / 2`. This formula is only correct when `q_len == kv_len`. For the bottom-right causal mask with `q_len < kv_len`, every query row sees `causal_offset + i + 1` keys (where `causal_offset = kv_len - q_len`), so the total visible pairs are `q_len * kv_len - q_len * (q_len - 1) / 2` — wait, let me re-derive. Row `i` sees positions `0..causal_offset+i`, that's `causal_offset + i + 1` entries. Summing over `i=0..q_len-1`: `q_len * (kv_len - q_len) + q_len*(q_len+1)/2 = q_len * kv_len - q_len*(q_len-1)/2`. Okay, the algebra checks out. The goose retracts this honk mid-flight. Embarrassing.

SILENT

Ibuki-wind (approved):
### Overall
Clean — no issues.

PR #1259, opened 2026-05-07
[Refactor][Ops] de-duplicate unary activation forward/inplace dispatch (#1220)

### LINKED ISSUE #1220: [REFACTOR][Ops] de-duplicate elementwise unary activation `__init__` boilerplate
### Symptom / Motivation

Across the activation Op family in `tileops/ops/elementwise.py`, two clusters of constructors share near-identical bodies:

- Param-free / value-range group: `ReluFwdOp`, `SiluFwdOp`, `HardswishFwdOp`, `HardsigmoidFwdOp`, `MishFwdOp`, `SeluFwdOp`.
- Parametric group: `LeakyReluFwdOp`, `EluFwdOp`, `HardtanhFwdOp`, `SoftplusFwdOp`.

Within each group, `__init__`, `_eager_forward`, and (for the second group) `forward` are mechanically duplicated — same shape inference, same `kernel_map=None` / `tune=False` keyword-only kwargs wiring, same `_apply_inplace` plumbing.

This was flagged twice during PR #1211 review (gemini medium-priority) but not addressed in the PR because the alignment work had to keep diff scope to `tileops/ops/elementwise.py` for AC-2 and any structural refactor would risk re-touching cross-module surface.

### Root Cause Analysis

Each Op was written independently to satisfy the `UnaryOp` canonical pattern (kernel_map / tune as keyword-only kwargs, manifest-derived flops, `_apply_inplace` after kernel return). The shared structure is real but never extracted — every new activation op pays the boilerplate tax and risks per-op drift (e.g. T067 erf manifest-flops drift was exactly this class of bug).

### Related Files

- tileops/ops/elementwise.py
- tests/ops/test_elementwise_unary_activation_alignment.py
- tests/ops/test_activation.py

## Goal

Introduce a thin shared base or mixin (or two: param-free vs parametric) under `tileops/ops/` that owns the identical `__init__`, `_eager_forward`, and `forward` logic, so each leaf Op only declares its op-specific fields (op_name, manifest section name, parameter list, eager torch fallback). Behavior must remain bit-identical: same constructor signatures, same kwargs keyword-only-ness, same manifest-derived flops, same `_apply_inplace` semantics, same `kernel_map=None` / `tune=False` defaults.

## Plan

**Plan type: proposal**

1. Identify the exact common surface across the param-free group (Relu/Silu/Hardswish/Hardsigmoid/Mish/Selu) and the parametric group (LeakyRelu/Elu/Hardtanh/Softplus). Produce a small audit table: which fields/methods are truly identical, which only look identical.
2. Decide between a single mixin with an op-specific descriptor table vs. two thin base classes. Document the trade-off in the PR description.
3. Migrate one representative op from each group as a proof-of-concept and run the full alignment + activation test suites; confirm no behavior drift.
4. Migrate the remaining ops in a single follow-up commit; rerun `pytest tests/ops/test_elementwise_unary_activation_alignment.py tests/ops/test_activation.py` and `python -m benchmarks.ops.bench_activation` (smoke).
5. Confirm `validate_manifest --check-op` is unchanged for the touched ops.

## Constraints

- Do NOT modify `tileops/manifest/` or `tileops/kernels/` — refactor lives strictly in `tileops/ops/`.
- Do NOT change any constructor signature, kwarg ordering, or kwarg keyword-only-ness; downstream code (tests, benches, codegen) depends on these.
- Do NOT regress `_apply_inplace` ValueError-on-numel-mismatch behavior introduced in PR #1211.
- No new public API beyond the shared base/mixin itself; keep it `_`-prefixed if it should not be re-exported.
- Bench numbers for the touched ops should not regress beyond noise (re-run `bench_activation.py` and compare TFLOPs/bandwidth columns).

## Acceptance Criteria

- [ ] AC-1: `pre-commit run --all-files` passes; `pytest tests/ops/test_elementwise_unary_activation_alignment.py tests/ops/test_activation.py tests/ops/test_elementwise_independent_fp8.py` passes with the same test counts as PR #1211 HEAD.
- [ ] AC-2: `git diff --name-only <base>..HEAD` is confined to `tileops/ops/`, `tests/`, `benchmarks/`; no changes under `tileops/manifest/` or `tileops/kernels/`.
- [ ] AC-3: All ten activation Ops listed above expose the same constructor signature (parameters, defaults, keyword-only-ness) as on PR #1211 HEAD; verified by an inspect-based regression test.
- [ ] AC-4: `validate_manifest --check-op` reports zero new errors for the touched ops.
- [ ] AC-5: `benchmarks/ops/bench_activation.py` runs to completion and produces a `profile_run.log` with rows for the touched ops; no correctness assertions added.
- [ ] AC-6: The PR body documents the chosen design (single mixin vs. two bases) with a short rationale.

lcy-seso:
## Summary

Closes #1220.

Extract the identical `forward` template and `_wrapped_inplace` state used by the ten unary activation Ops into a single shared mixin under `tileops/ops/elementwise.py`:

- Six param-free leaves: `ReluFwdOp`, `SiluFwdOp`, `HardswishFwdOp`, `HardsigmoidFwdOp`, `MishFwdOp`, `SeluFwdOp`.
- Four parametric leaves: `LeakyReluFwdOp`, `EluFwdOp`, `HardtanhFwdOp`, `SoftplusFwdOp`.

Before, `_ParamFreeActivationOp` and `_ParametricActivationOp` each carried their own copy of the inplace-vs-eager dispatch ladder (and the parametric base inlined a copy of `_validate_input` and `_eager_forward` that was byte-identical to `UnaryOp`'s versions). Now both bases extend `(_UnaryActivationMixin, UnaryOp)`:

- The mixin owns `forward()` and the `_wrapped_inplace` class slot.
- `UnaryOp` provides `_validate_input` and `_eager_forward`, which the parametric base picks up by inheritance instead of redefining.
- `_finalize_init` in the parametric base now also records `self.output_dtype` mirroring `UnaryOp.__init__`, so both groups expose the same surface.

## Design choice: single mixin vs. two bases

Picked **a single mixin (`_UnaryActivationMixin`) plus the existing two bases** rather than collapsing everything into one class. Reasons:

1. Constructor signatures genuinely differ between the two groups. Param-free leaves take `(N_total, dtype, *, strategy, kernel_map, tune, inplace)`; the four parametric leaves each take their own scalar params (`negative_slope`, `alpha`, `min_val/max_val`, `beta/threshold`) and Softplus drops `inplace`. Forcing one base would require either `**kwargs` indirection or a hard-coded scalar-param mechanism, neither of which is bit-identical to today's signatures (which AC-3 freezes).
2. The truly duplicated code is the `forward` body and the `_wrapped_inplace` slot — a thin mixin captures exactly that without changing how leaves construct their kernels.
3. Keeping `_ParamFreeActivationOp` and `_ParametricActivationOp` as the inheritance anchors preserves their docstring narratives and the existing `_register_unary_inplace_custom_op` wiring.

The mixin is `_`-prefixed so it is not part of the public surface.

## Test plan

- [x] `pre-commit run --all-files` — passes.
- [x] `pytest tests/ops/test_elementwise_unary_activation_alignment.py tests/ops/test_activation.py tests/ops/test_elementwise_independent_fp8.py` — 151 passed (141 pre-existing + 10 new constructor-signature freeze cases).
- [x] `validate_manifest.py --check-op` clean for relu / silu / hardswish / hardsigmoid / mish / selu / leaky_relu / elu / hardtanh / softplus.
- [x] `benchmarks/ops/bench_activation.py` runs to completion and emits rows for all touched ops (10 unrelated `erf` / `mish` failures are an environment-only `/tmp/tvm-debug-mode-tempdirs` permission collision, present on `testbed` HEAD as well).
- [x] `git diff --name-only testbed..HEAD` confined to `tileops/ops/elementwise.py` and `tests/ops/test_elementwise_unary_activation_alignment.py` — no `tileops/manifest/` or `tileops/kernels/` changes.

## Behavior preservation

- Constructor signatures of all ten leaves are byte-identical (frozen by `test_unary_activation_init_signature_is_frozen`).
- `inplace=True` still aliases `input` and dispatches through `_wrapped_inplace`; `inplace=False` returns a fresh tensor.
- `_validate_input` raises the same `ValueError` on non-CUDA / dtype mismatch / numel mismatch (the parametric path now uses `UnaryOp._validate_input` whose error messages and conditions match the previous inline checks).
- Manifest-derived `FLOPS_PER_ELEM` and `eval_roofline` unchanged.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/tests/ops/test_elementwise_unary_activation_alignment.py
+++ b/tests/ops/test_elementwise_unary_activation_alignment.py
@@ -7,6 +7,8 @@ identity, ``approximate`` validation, kernel_map override
 dispatch, and end-to-end correctness against the PyTorch reference.
 """
 
+import inspect
+
 import pytest
 import torch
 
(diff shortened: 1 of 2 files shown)


Ibuki-wind (approved):
Nightshift autonomous review: HEAD d6427ad, all CI green/skipped, no unresolved threads, all 6 ACs verified by foundry reviewer. Approving for squash-merge into testbed.

Issue #1265, opened 2026-05-07
[DOCS][Process] amend Status flip carve-out to permit workload addition when promoting spec-only ops


lcy-seso:
## Context

Issue #1249 (PR #1264) implements the real `LerpTensorFwdOp` by promoting its manifest entry from `status: spec-only` to `status: implemented`. The Status flip carve-out (see `docs/design/manifest-trust-model.md`, formalized in PR #1258) currently restricts manifest edits in code-side PRs to `status` and `source.kernel_map` only.

In practice, promoting a `spec-only` entry to `implemented` *requires* adding `workloads`, because `tests/test_ops_manifest.py::test_every_op_has_at_least_two_workloads` enforces a non-empty `workloads` list for any op whose `status != spec-only`. PR #1264's CI surfaced this contradiction: the carve-out forbids workload changes, yet the test forces them.

## Proposal

Amend the carve-out rule text in `docs/design/manifest-trust-model.md` (and any mirrored skill / agent docs) to permit `workloads` additions when — and only when — the same PR flips `status: spec-only` → `status: implemented`. Keep the existing prohibition on workload edits to ops that were already `implemented` before the PR.

## Acceptance criteria

- [ ] Carve-out section explicitly lists `workloads` as an allowed field when promoting `spec-only` → `implemented`.
- [ ] Wording makes clear this exception does NOT extend to other status transitions or to ops already at `implemented`.
- [ ] Cross-referenced from `docs/design/trust-model.md` if that doc summarizes the carve-out.

## Out of scope

- Workload edits to already-implemented ops (separate manifest-only PR).
- Signature / shape_rules / roofline edits (still forbidden in code PRs).

lcy-seso:
Resolved by #1271 (squash-merged into testbed at fd74984).

PR #1271, opened 2026-05-07
[Doc][Process] permit workloads addition on spec-only -> implemented promotion

lcy-seso:
Closes tile-ai/TileOPs#1265

## Summary

- Amend `.claude/rules/manifest-trust-model.md` to enumerate `workloads` as an allowed field when a single PR flips an op's `status` from `spec-only` to `implemented`.
- Make explicit that the carve-out does NOT extend to `implemented -> spec-only` transitions or to ops already at `implemented`; `workloads` edits remain forbidden in those cases.
- Cross-reference the carve-out from `docs/design/trust-model.md` so the design summary stays consistent with the authoritative rules file.

## Test plan

- [x] Carve-out section explicitly lists `workloads` as an allowed field when promoting `spec-only` to `implemented`.
- [x] Wording makes clear this exception does not extend to other status transitions or to ops already at `implemented`.
- [x] Cross-referenced from `docs/design/trust-model.md`.
- [x] `make lint` (pre-commit run --all-files) passed.

### DIFF
--- a/.claude/rules/manifest-trust-model.md
+++ b/.claude/rules/manifest-trust-model.md
@@ -5,22 +5,12 @@
 
 ## Status flip carve-out
 
-An implementation PR (the one that aligns op code with the manifest) MAY touch the manifest entry of the op being aligned, but ONLY within the metadata fields enumerated below. Contractual fields stay frozen and require a separate manifest-only PR with human review.
+An implementation PR may edit the aligned op's manifest entry only at:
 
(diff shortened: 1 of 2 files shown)


Ibuki-wind (approved):
Nightshift autonomous review: HEAD f3e4799, CI green, all 3 ACs met. Round-2 commit further tightens the carve-out wording per agent-doc brevity standard. Approving for squash-merge into testbed.

PR #1417, opened 2026-05-11
[Fix][Skills] make review-tileops / resolve-tileops scripts cwd-independent

lcy-seso:
## Summary

Four skill scripts (review-tileops/loop.sh, resolve-tileops/preflight.sh, round-pre.sh, round-post.sh) computed REPO_PATH via `git worktree list --porcelain | head -n 1 | sed ...`. Two latent bugs converged in this pattern:

1. **cwd-dependence**: `git worktree list` operates on the git repo of cwd. If the script was invoked via absolute path from outside the repo, the command had no repo to inspect and exited 128.
2. **SIGPIPE under `set -o pipefail`**: in a checkout with multiple worktrees, `head -n 1` closes the pipe after one line. Upstream `git worktree list` then receives SIGPIPE and exits 141. `pipefail` propagates that exit code, so the `||` branch fired "not in a git repo" even when called from a valid worktree.

The shipped fix replaces the pipe entirely with `git rev-parse --git-common-dir`, which returns the `.git` directory of the main worktree as a single value (no pipe, no SIGPIPE risk, no multi-line parse). The main worktree path is its parent directory. `git -C "$SKILL_DIR"` anchors the command to the repo containing the script, removing the cwd dependence in the same step.

The error message is retuned to `cannot resolve repo root from $SKILL_DIR=...` so the next debugger sees the actual resolution target rather than the misleading "not in a git repo" hint.

## Test plan

- [x] `bash .claude/skills/review-tileops/loop.sh non-numeric` from inside the main checkout — passes REPO_PATH resolution, stops at PR-arg validation.
- [x] Same call via absolute path from `/tmp` (outside any repo) — also passes REPO_PATH resolution.
- [x] Same call from inside a linked worktree — REPO_PATH resolves to the main checkout, not the worktree.
- [x] Same three checks applied to resolve-tileops/preflight.sh, round-pre.sh, round-post.sh.

### DIFF
--- a/.claude/skills/resolve-tileops/preflight.sh
+++ b/.claude/skills/resolve-tileops/preflight.sh
@@ -20,14 +20,15 @@ REPO="tile-ai/TileOPs"
 command -v gh >/dev/null 2>&1 || { echo "preflight: missing gh" >&2; exit 1; }
 command -v jq >/dev/null 2>&1 || { echo "preflight: missing jq" >&2; exit 1; }
 
+SKILL_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # Anchor state to the main checkout's `.foundry/runs/`, not cwd. When
 # invoked from a linked worktree (e.g. via foundry pipeline HANDOFF),
 # `--show-toplevel` returns the worktree path; using it would scatter
 # resolve state across worktrees. `git worktree list --porcelain 2>/dev/null`
 # always emits the main worktree first, so its leading entry is the
 # main checkout regardless of which worktree we run from.
-REPO_PATH="$(git worktree list --porcelain 2>/dev/null | head -n 1 | sed 's/^worktree //')" \
-  || { echo "preflight: not in a git repo" >&2; exit 1; }
+REPO_PATH="$(git -C "$SKILL_DIR" worktree list --porcelain 2>/dev/null | head -n 1 | sed 's/^worktree //')" \
+  || { echo "preflight: cannot resolve repo root from \$SKILL_DIR=$SKILL_DIR" >&2; exit 1; }
 
 # Round 2+ fast path: an existing run dir's meta.json already pins this PR.
 META=""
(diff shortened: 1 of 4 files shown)


.claude/skills/resolve-tileops/preflight.sh
```diff
@@ -20,14 +20,15 @@ REPO="tile-ai/TileOPs"
...
+SKILL_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # Anchor state to the main checkout's `.foundry/runs/`, not cwd. When
 # invoked from a linked worktree (e.g. via foundry pipeline HANDOFF),
 # `--show-toplevel` returns the worktree path; using it would scatter
 # resolve state across worktrees. `git worktree list --porcelain 2>/dev/null`
 # always emits the main worktree first, so its leading entry is the
 # main checkout regardless of which worktree we run from.
-REPO_PATH="$(git worktree list --porcelain 2>/dev/null | head -n 1 | sed 's/^worktree //')" \
-  || { echo "preflight: not in a git repo" >&2; exit 1; }
+REPO_PATH="$(git -C "$SKILL_DIR" worktree list --porcelain 2>/dev/null | head -n 1 | sed 's/^worktree //')" \
```
Ibuki-wind: This REPO_PATH probe still fails under `set -o pipefail` in a checkout with multiple worktrees: `head -n 1` can make `git worktree list` exit 141, so the absolute-path smoke test still dies here. → Drop `head` and consume the full stream (for example `sed -n '1s/^worktree //p'`), and apply the same pipefail-safe pattern to the other three REPO_PATH assignments.

Ibuki-wind (changes requested):
### Overall
The cwd-independent path resolution is still not pipefail-safe, so these scripts can still fail before they reach PR validation in a multi-worktree checkout.

.claude/skills/resolve-tileops/preflight.sh
lcy-seso: Adopted. Replaced `git worktree list | head` with `git rev-parse --git-common-dir` across all four scripts. See `53a6377`.

Ibuki-wind (changes requested):
### Overall
The code fix now looks correct, but the PR description still documents the old `git -C "$SKILL_DIR" worktree list ...` approach instead of the shipped `git rev-parse --git-common-dir` implementation, so the review metadata is stale.

### Cross-cutting concerns
- Update the Summary and Test plan to match `53a6377`'s final behavior. The current body still describes the earlier pipe-based fix and its old failure mode.
- Trim the follow-up thread reply to outcome-only per the approval gate; the current reply includes commit narration and verification detail.

Ibuki-wind (approved):
Clean — no issues.

PR #1419, opened 2026-05-11
[Fix][Skills] cap codex subprocess + reap children in review-tileops loop

lcy-seso:
## Changes

`.claude/skills/review-tileops/loop.sh`:

- Wrap both `codex exec` / `codex exec resume` invocations in `timeout --kill-after=30 1800`. Timeout (rc=124/137) falls through to the existing `CODEX_RETRY=3` retry path; worst case 3 × 1800s ≈ 90 min before `status=error`.
- Add `trap cleanup_and_exit TERM INT HUP`. Handler walks the descendant tree via recursive `pgrep -P`, TERMs every descendant, then KILLs after 1s. Avoids `kill 0` so the user's interactive shell / tmux pane (which shares the loop's pgroup) is not taken down.

## Test plan

- [x] `bash -n loop.sh`; `shellcheck -S warning loop.sh` clean.
- [x] Trap smoke test on a 3-level descendant chain: all descendants collected, TERM delivered, KILL fallback reached, exit 130, no orphans.
- [ ] Re-run `review-loop 1410 --bg` against the MCP wedge — expect `codex attempt N exceeded 1800s (rc=124) — killed` and exit within ~90 min instead of hanging.

### DIFF
--- a/.claude/skills/review-tileops/loop.sh
+++ b/.claude/skills/review-tileops/loop.sh
@@ -87,9 +87,18 @@ log() { printf '[%s] %s\n' "$(date -u +%H:%M:%S)" "$*"; }
 # to init and keep running after the loop (or its tmux pane / SSH) exits.
 cleanup_and_exit() {
   trap - TERM INT HUP EXIT
-  kill -TERM 0 2>/dev/null || true
-  sleep 1
-  kill -KILL 0 2>/dev/null || true
+  # Kill direct children only. `kill 0` (pgroup) would also hit the
+  # user's interactive shell — foreground `review-loop` shares a pgroup
+  # with the calling tmux pane, and `nohup … &` does not create one.
+  # `timeout` forwards SIGTERM to codex, so one level is enough.
+  local pids
+  pids=$(pgrep -P $$ 2>/dev/null) || true
+  if [[ -n "$pids" ]]; then
+    kill -TERM $pids 2>/dev/null || true
+    sleep 1
+    kill -KILL $pids 2>/dev/null || true
+  fi
+  exit 130
 }
 trap cleanup_and_exit TERM INT HUP
 
--- a/.claude/skills/review-tileops/loop.sh
+++ b/.claude/skills/review-tileops/loop.sh
@@ -25,10 +25,10 @@ REPO="tile-ai/TileOPs"
 MAX_ROUNDS=15
 POLL_INTERVAL=180
 CODEX_RETRY=3
-# Per-round hard cap on the codex subprocess. Codex can wedge on MCP
-# transport failures (endless "Reconnecting 1/5 → 5/5" with no exit),
-# blocking the loop on `wait`; this bounds the round so CODEX_RETRY=3
-# stays reachable.
+# Per-round hard cap on the synchronous codex invocation. Codex can wedge
+# on MCP transport failures (endless "Reconnecting 1/5 → 5/5" with no
+# exit); without this cap bash sits in waitpid forever and CODEX_RETRY=3
+# is unreachable. Worst case: 3 × 1800s ≈ 90 min before status=error.
 CODEX_TIMEOUT_SEC=1800
 # Stall safety: terminate after this many consecutive idle polls (no new
 # commits / comments). MAX_IDLE * POLL_INTERVAL must comfortably exceed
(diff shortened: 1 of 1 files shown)


.claude/skills/review-tileops/loop.sh
```diff
@@ -78,6 +83,16 @@ fi
 
 log() { printf '[%s] %s\n' "$(date -u +%H:%M:%S)" "$*"; }
 
+# Reap the process group on signal; otherwise codex/gh/git get reparented
+# to init and keep running after the loop (or its tmux pane / SSH) exits.
+cleanup_and_exit() {
+  trap - TERM INT HUP EXIT
+  kill -TERM 0 2>/dev/null || true
```
Ibuki-wind: `cleanup_and_exit` clears the signal traps and then calls `kill -TERM 0`, which sends SIGTERM to this shell as part of the current process group; the handler can terminate before reaching the `sleep` / `kill -KILL` escalation, so the hard reap path is unreliable. → Mask or avoid self-signals during cleanup, signal only the intended child/process group, and explicitly exit with a signal/error status after escalation.

.claude/skills/review-tileops/loop.sh
```diff
@@ -25,6 +25,11 @@ REPO="tile-ai/TileOPs"
 MAX_ROUNDS=15
 POLL_INTERVAL=180
 CODEX_RETRY=3
+# Per-round hard cap on the codex subprocess. Codex can wedge on MCP
+# transport failures (endless "Reconnecting 1/5 → 5/5" with no exit),
+# blocking the loop on `wait`; this bounds the round so CODEX_RETRY=3
```
Ibuki-wind: The `CODEX_TIMEOUT_SEC` comment still says the loop is blocked on `wait`, but `run_codex_round` invokes `codex` synchronously and this file has no `wait` in that path; this points future maintenance at nonexistent control flow. → Reword the comment to describe the synchronous `codex` invocation being bounded by `timeout`.

Ibuki-wind (changes requested):
### Overall
The signal cleanup path still does not reliably reach its hard-kill fallback; fix the handler before merging.

lcy-seso:
Thanks for the catch — both Gemini and Ibuki flagged a real bug. The original `kill -TERM 0` did target self because the trap clears its own handler first, so the script died before reaching the `sleep` + KILL fallback. Confirmed via local smoke test.

Fixed in 3b004f7 by a different route than the suggested ignore-self pattern: send to **direct children only**, not the pgroup.

Reason `kill 0` is the wrong primitive here regardless of the self-signal race:
- Foreground `review-loop` runs `bash loop.sh | tee` from the user's interactive shell — the pipeline shares a pgroup with the calling tmux pane.
- `--bg` uses `nohup … &` without `setsid`, so it also stays in the caller's pgroup.

In both cases a pgroup-wide TERM in the trap would kill the user's terminal session along with the loop. Walking direct children via `pgrep -P $$` avoids that, and `timeout` forwards SIGTERM to its codex grandchild on its own, so one level covers codex / gh / git.

Verified locally: trap fires → direct child receives TERM → `sleep 1` survives → KILL fallback executes → exit 130.

.claude/skills/review-tileops/loop.sh
```diff
@@ -78,6 +83,25 @@ fi
...
+cleanup_and_exit() {
+  trap - TERM INT HUP EXIT
+  # Kill direct children only. `kill 0` (pgroup) would also hit the
+  # user's interactive shell — foreground `review-loop` shares a pgroup
+  # with the calling tmux pane, and `nohup … &` does not create one.
+  # `timeout` forwards SIGTERM to codex, so one level is enough.
+  local pids
+  pids=$(pgrep -P $$ 2>/dev/null) || true
+  if [[ -n "$pids" ]]; then
+    kill -TERM $pids 2>/dev/null || true
```
Ibuki-wind, line 104: `cleanup_and_exit` now kills only `pgrep -P $$` direct children, but the codex path is `loop.sh -> timeout -> codex` (and resume adds a subshell before `timeout`), so killing the direct child can leave the actual `codex` process or its descendants orphaned. → Terminate the full child tree recursively, or launch the codex invocation in a dedicated process group/session that is safe to kill without targeting the caller's shell.

.claude/skills/review-tileops/loop.sh
Ibuki-wind: The `CODEX_TIMEOUT_SEC` comment still says the loop blocks on `wait`, but `run_codex_round` is blocked by the synchronous `codex` command; there is no `wait` in this path. → Reword this to describe `timeout` bounding the synchronous codex invocation.

Ibuki-wind (changes requested):
### Overall
The revised cleanup avoids the self-signal race, but it still does not reliably reap the codex process tree; the stale `wait` wording also remains.

lcy-seso:
Addressed in 065a9e9.

**Orphan codex (Copilot, Ibuki):** valid — the new chain is `loop.sh → timeout → node(codex) → codex-rust`, and node does not reliably forward TERM to the rust child (this is exactly the wedge that motivated `--kill-after=30`). Direct-children-only would orphan the actual codex process. Switched to a recursive `pgrep -P` descendant walk so the whole subtree is reaped. Still avoiding `kill 0` for the same reason as before — pgroup includes the user's shell under both foreground `review-loop` and `nohup … &`.

Smoke test on a 3-level chain (`bash → bash → bash → sleep`): 5 PIDs collected, TERM delivered, `sleep 1` survives, KILL fallback executes, exit 130, no leftover orphans.

**Stale "wait" / "process group" wording (Copilot, Ibuki):** rewritten. CODEX_TIMEOUT_SEC now describes the synchronous codex invocation bounded by `timeout`, with the explicit "worst case 3 × 1800s ≈ 90 min" so expectations match constants. The cleanup_and_exit leading comment now matches the implementation (descendant walk, not pgroup).

**Gemini's `trap '' TERM INT HUP` suggestion:** not adopted — that pattern only fixes self-signal reachability, but our deeper concern was that *any* pgroup-wide kill takes the user's shell with it. The descendant walk sidesteps both problems at once. Added explicit `exit 130` as suggested.

**Trapping EXIT (Gemini):** intentionally not added. EXIT fires on every clean exit too (success path, MAX_ROUNDS, PR closed, stalled, etc.); reaping descendants there is a no-op at best and a footgun if the loop ever wants to leave a final `gh` or `git` call running async. The TERM/INT/HUP triplet covers the unsafe-exit cases we actually care about.

**Validate `timeout`/`pgrep` upfront (Copilot):** punted. The loop already calls `git`, `gh`, `jq`, `python3`, `stat`, `awk`, `sed`, `find` without preflight, and `timeout`/`pgrep` are coreutils/procps that ship with every distro the rest of the toolchain runs on. Adding a check for only these two would be inconsistent. If we want hardening here, it belongs in a separate "preflight: required CLIs" pass over the whole script.

Ibuki-wind (changes requested):
### Overall
Code blockers are resolved, but the PR body still describes the old implementation; update it to match the final diff before approval.

### Cross-cutting concerns
- The PR body still says `loop.sh` blocked on `wait` and that cleanup uses process-group reaping / `kill 0`, but the final code bounds a synchronous codex invocation and recursively kills descendants while explicitly avoiding pgroup-wide signals. The approval gate requires the PR body to record final state only, so revise the Summary/Fixes/Test plan to match the merged diff and remove stale process narration.

lcy-seso:
PR body rewritten to match the final diff: synchronous codex bounded by `timeout` (waitpid wording, not `wait`), descendant-tree reaping via recursive `pgrep -P` (no pgroup kill), worst-case 90 min, and the smoke-test summary on a 3-level chain. No code changes.

Ibuki-wind (approved):
Clean — no issues.

Issue #1396, opened 2026-05-09
[FEAT][REDUCTION] implement dim=none for arithmetic reductions (sum/mean/amax/amin)


lcy-seso:
### Symptom / Motivation

`tileops/manifest/reduction.yaml` declares `SumFwdOp`, `MeanFwdOp`, `AmaxFwdOp`, `AminFwdOp` with `signature.params.dim` accepting `int | list[int] | tuple[int, ...] | None` per `torch.sum / torch.mean / torch.amax / torch.amin`. The op classes (`tileops/ops/reduction/reduce.py`) and the underlying kernel do not yet handle every workload shape × dim combination correctly for `dim=None` (full-tensor reduction), so the four entries currently ship as `status: spec-only`. This issue tracks the implementation needed to promote them to `implemented`.

W2-E-flip (#1199 / PR #1385) flipped 3 of 13 reduction ops; this PR is the first of three impl issues that complete the remaining 10 by op-shape/output-dtype affinity. Tracker: #1142.

### Root Cause Analysis

`reduce.py` already routes `dim=None` to the multi-dim path via `normalize_dim`, but the spec-test gate that promotes status requires the kernel to produce numerically-correct output across every `signature.dtype_combos` row × every `workloads` shape. Current gaps surface in the dim=None / 0-d output path.

### Related Files

- `tileops/ops/reduction/reduce.py` — Op classes for the 4 ops (single shared base)
- `tileops/kernels/reduction/` — kernel implementations
- `tileops/manifest/reduction.yaml` — spec entries
- `tests/ops/test_reduction.py` — spec tests (PyTorch ground truth)

## Goal

`SumFwdOp`, `MeanFwdOp`, `AmaxFwdOp`, `AminFwdOp` pass spec-conformance tests for every workload row in the manifest, including the `dim=None` (full-reduction) and `dim` as `tuple[int, ...]` shapes. Output is a 0-D tensor for `dim=None, keepdim=False`. Numerics match `torch.sum / torch.mean / torch.amax / torch.amin` within standard tolerances. Status flip to `implemented` lands in a separate manifest-only PR (sibling issue).

## Plan

1. Run `/audit-family REDUCTION SumFwdOp MeanFwdOp AmaxFwdOp AminFwdOp` and confirm the gap classification is `minor` (op already scaffolded; `dim=None` path needs kernel fix or op-layer fold).
[...]

PR #1428, opened 2026-05-12
[Maintain][Manifest] flip reduction Sum/Mean/Amax/Amin to implemented

### LINKED ISSUE #1399: [Maintain][Manifest] flip arithmetic reductions to implemented
### Symptom / Motivation

After the impl PR for arithmetic reductions (#1396) lands, `SumFwdOp`, `MeanFwdOp`, `AmaxFwdOp`, `AminFwdOp` pass spec-conformance for every workload row. Their manifest entries should be promoted from `status: spec-only` to `status: implemented` so the ops become discoverable / dispatchable through the manifest.

This is a manifest-only PR per `.claude/rules/manifest-trust-model.md` — human review, no nightshift.

### Root Cause Analysis

N/A — process step. The impl side already shipped via #1396; flipping status is the second half of the trust-model split.

### Related Files

## Goal

`tileops/manifest/reduction.yaml` flips `SumFwdOp`, `MeanFwdOp`, `AmaxFwdOp`, `AminFwdOp` to `status: implemented`. `python scripts/validate_manifest.py` and `pytest tests/test_validate_manifest.py` both pass.

## Plan

1. Confirm impl PR (#1396) has merged into `testbed` and that `pytest tests/ops/test_reduction.py -m smoke` passes for the 4 ops on the latest testbed.
2. Edit `tileops/manifest/reduction.yaml`: change `status: spec-only` → `status: implemented` on `SumFwdOp`, `MeanFwdOp`, `AmaxFwdOp`, `AminFwdOp`.
3. If validator reports missing `kernel_map` / `static_dims` on any op, run `/fix-manifest <op>` to patch the structural field.
4. Run `python scripts/validate_manifest.py` — must exit 0.
5. Run `pytest tests/test_validate_manifest.py` — must pass.
6. Open PR against `main` (not testbed) per the manifest-flip workflow used by #1382 / #1383 / #1384 / #1386.

## Constraints

- MUST be a single-file PR touching only `tileops/manifest/reduction.yaml` (plus, if necessary, the structural fields auto-patched by `/fix-manifest`).
- MUST NOT carry `nightshift` label — human review only.
- MUST NOT flip any op outside the listed 4.
- MUST NOT modify signature, shape_rules, dtype_combos, roofline, or workloads on the flipped entries.
- Single PR.

## Acceptance Criteria

- [ ] AC-1: `tileops/manifest/reduction.yaml` shows `status: implemented` for `SumFwdOp`, `MeanFwdOp`, `AmaxFwdOp`, `AminFwdOp`.
- [ ] AC-2: `python scripts/validate_manifest.py` exits 0.
- [ ] AC-3: `pytest tests/test_validate_manifest.py` passes.
- [ ] AC-4: PR diff touches only manifest yaml (and optionally structural-field patches).

lcy-seso:
Closes #1399

## Summary

- Flip `SumFwdOp`, `MeanFwdOp`, `AmaxFwdOp`, `AminFwdOp` in `tileops/manifest/reduction.yaml` from `status: spec-only` to `status: implemented`.
- No changes to `signature`, `shape_rules`, `dtype_combos`, `roofline`, or `workloads` on the flipped entries.
- Implementation already landed via #1396; this PR closes the spec/impl trust-boundary handoff.

## Test plan

- [x] AC-1: `tileops/manifest/reduction.yaml` shows `status: implemented` for SumFwdOp, MeanFwdOp, AmaxFwdOp, AminFwdOp.
- [x] AC-2: `python scripts/validate_manifest.py` exits 0.
- [x] AC-3: `pytest tests/test_validate_manifest.py` passes (218 passed).
- [x] AC-4: PR diff touches only `tileops/manifest/reduction.yaml`.

### DIFF
--- a/tileops/manifest/reduction.yaml
+++ b/tileops/manifest/reduction.yaml
@@ -159,7 +159,7 @@ LogSumExpFwdOp:
 SumFwdOp:
   ref_api: "torch.sum"
   family: reduction
-  status: spec-only  # dim=None not yet implemented; update when impl lands
+  status: implemented
 
   signature:
     inputs:
@@ -211,7 +211,7 @@ SumFwdOp:
 MeanFwdOp:
   ref_api: "torch.mean"
   family: reduction
-  status: spec-only  # dim=None not yet implemented; update when impl lands
+  status: implemented
 
   signature:
     inputs:
@@ -259,7 +259,7 @@ MeanFwdOp:
 AmaxFwdOp:
   ref_api: "torch.amax"
   family: reduction
-  status: spec-only  # dim=None not yet implemented; update when impl lands
+  status: implemented
 
   signature:
     inputs:
@@ -303,7 +303,7 @@ AmaxFwdOp:
 AminFwdOp:
   ref_api: "torch.amin"
   family: reduction
-  status: spec-only  # dim=None not yet implemented; update when impl lands
+  status: implemented
 
   signature:
     inputs:


tileops/manifest/reduction.yaml
```diff
@@ -159,7 +159,7 @@ LogSumExpFwdOp:
 SumFwdOp:
   ref_api: "torch.sum"
   family: reduction
-  status: spec-only  # dim=None not yet implemented; update when impl lands
+  status: implemented
```
Ibuki-wind, line 162: These four status flips are premature: `_ReduceOpBase.__init__` still defaults `dim` to `-1` (`tileops/ops/reduction/reduce.py:78`), while the manifest now claims `default: null`. Revert them to `spec-only` until the constructor default is updated in code.

Ibuki-wind (changes requested):
### Overall
One conformance blocker remains before this manifest can truthfully claim implemented status.

tileops/manifest/reduction.yaml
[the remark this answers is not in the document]
lcy-seso: Verified results after a closer look:

1. **Default value (real)** — `_ReduceOpBase.__init__` defaults `dim=-1` while the manifest declares `default: null`. Tracking as a code-only prerequisite (will land before this manifest flip merges).
2. **`hidden-state-reduce` workload ambiguity (real, manifest-only)** — agreed; the workload should pin `dim` explicitly. Will be addressed in a separate manifest-only follow-up (see M4 below), not in this PR.
3. **`dim=[]` in shape_rules / roofline (real, manifest-only)** — the `... if len(dim) > 0 else set(range(x.ndim))` ternary treats `dim=[]` as full reduction, which PyTorch does not. The op layer already enforces `_empty_dim_policy="full"` on Sum/Mean/Amax/Amin so the wrong branch is unreachable at runtime, but the manifest spec text is still wrong and will be fixed in the same manifest-only follow-up.

Items 2 and 3 are outside the manifest-trust-model status-flip carve-out (would modify `signature` / `shape_rules` / workloads) and need a dedicated manifest PR with human review. Item 1 will be the prereq code PR.

lcy-seso: Confirmed — `_ReduceOpBase.__init__` (`tileops/ops/reduction/reduce.py:78`) defaults `dim=-1`, manifest says `null`. A code-only prereq PR will flip the default to `None` on `_ReduceOpBase`, `_WelfordReduceOp`, `AllFwdOp`, `AnyFwdOp`, `CountNonzeroFwdOp`, plus an explicit `dim=-1` override on `ProdFwdOp` (whose manifest `dim` is `int` with `default: -1`). This manifest PR will be re-based and unblocked once that code PR merges.

lcy-seso:
**Reviewer-feedback verification summary**

After re-reading every comment on PRs #1428 / #1430 / #1431 and running targeted checks, the verified picture is:

| Item | Verdict | Action |
|---|---|---|
| Constructor default `dim=-1` vs manifest `default: null` (Sum/Mean/Amax/Amin/Var/Std/VarMean/All/Any/CountNonzero) | ✅ Real | Code-only prereq PR — flip default to `None` on `_ReduceOpBase`, `_WelfordReduceOp`, `AllFwdOp`, `AnyFwdOp`, `CountNonzeroFwdOp`; add explicit `dim=-1` override on `ProdFwdOp` (its manifest type is `int, default: -1`). |
| `_logical_reduce_kernel` tail-row OOB on aligned-N path (All/Any/CountNonzero) | ❌ False positive | `compute-sanitizer --tool memcheck` clean across 16 shape configs (M ∈ {129, 257, 513, 1025} × N ∈ {128, 256, 512, 1024}); TileLang's `T.copy` inserts implicit row-bounds guards from the source extent. No fix needed. |
| `self.dtype.itemsize` `AttributeError` (gemini) | ❌ False positive | `torch.dtype.itemsize` exists since torch 2.1; `pyproject.toml` pins `torch>=2.1.0,<2.11.0`. |
| `x.reshape(M, N)` non-contiguous risk after `flatten_for_multidim` (gemini) | ❌ False positive | `flatten_for_multidim` already does `.permute(perm).contiguous()` before reshape. |
| `_reshape_output` not handling `tuple` (gemini) | ❌ False positive | `normalize_dim` always returns `list[int]`; `tuple` is unreachable through the public API. |
| `shape_rules` `dim=[]` semantics, missing dtypes in signature, `hidden-state-reduce` workload missing `dim` | ⚠️ Real, manifest-only | Outside the manifest-trust-model status-flip carve-out. Follow-up manifest-only PR will fix all three families together. |

**Plan**:
1. Open a code-only prereq PR (`[Fix][Ops] reduction op constructor defaults: dim=-1 → dim=None`) covering the 5 default flips + `ProdFwdOp` override + default-construction test updates.
2. Once that lands, rebase #1428 / #1430 / #1431 and the status-flip claims become contract-true.
3. Separately, file a manifest-only PR (`signature` dtype lists, `shape_rules` empty-list semantics, `hidden-state-reduce` workload) for the items the trust-model carve-out won't allow here.

Detailed per-comment verdicts are posted as inline replies on each thread.

tileops/manifest/reduction.yaml
Ibuki-wind, line 162: These entries still do not conform to the PyTorch API: the shared reduce path rejects scalar tensors (`x.ndim == 0`) with `ValueError("Input tensor must be at least 1D")`, while `torch.sum`, `torch.mean`, `torch.amax`, and `torch.amin` accept 0-D tensors and return 0-D results. `status: implemented` is a hard conformance claim, so support scalar inputs and add coverage, or keep these entries `spec-only` / explicitly narrow the manifest contract.

Ibuki-wind (changes requested):
### Overall
The status flip overclaims implementation conformance; scalar input support is still missing for the flipped PyTorch reduction APIs.

tileops/manifest/reduction.yaml
Ibuki-wind, line 162: `status: implemented` is still premature here: `_ReduceOpBase._validate_input_tensor()` (`tileops/ops/reduction/reduce.py:223-227`) rejects `x.ndim == 0`, but `torch.sum` / `torch.mean` / `torch.amax` / `torch.amin` accept scalar tensors and return 0-D outputs. Please lift the shared scalar-input guard and add a scalar conformance case before promoting these entries.

Ibuki-wind (changes requested):
### Overall
These flips are still ahead of the implementation: the shared reduction base rejects 0-D tensors, but `torch.sum` / `torch.mean` / `torch.amax` / `torch.amin` accept scalar inputs and return 0-D outputs.

Ibuki-wind (approved):
Clean — no issues.

PR #1430, opened 2026-05-12
[Maintain][Manifest] flip Var/Std/VarMean fwd ops to status: implemented

lcy-seso:
Closes #1400

## Summary

- Flip `VarFwdOp`, `StdFwdOp`, `VarMeanFwdOp` from `status: spec-only` to `status: implemented` in `tileops/manifest/reduction.yaml`.
- Diff is YAML-only; no signature/roofline/workloads/shape-rule edits (status flip carve-out).

## Context

The issue body references impl PR `#1397`, which does not exist. The actual implementation already landed on `main`:

- `dim=None` support via #834.
- Cross-`dim` × `correction` conformance coverage via #1406.

This PR is the spec-side flip aligning manifest status with the already-merged implementation and conformance tests.

## Test plan

- [x] AC-1: `tileops/manifest/reduction.yaml` shows `status: implemented` for `VarFwdOp`, `StdFwdOp`, `VarMeanFwdOp`.
- [x] AC-2: `python scripts/validate_manifest.py` exits 0 (`All manifest checks passed.`).
- [x] AC-3: `pytest tests/test_validate_manifest.py` passes (218 passed, 3 warnings).
- [x] AC-4: PR diff touches only `tileops/manifest/reduction.yaml` (+3 / -3).

Issue #1400, opened 2026-05-09
[Maintain][Manifest] flip variance reductions to implemented


lcy-seso:
### Symptom / Motivation

After the impl PR for variance reductions (#1397) lands, `VarFwdOp`, `StdFwdOp`, `VarMeanFwdOp` pass spec-conformance. Promote their manifest entries to `status: implemented`.

Manifest-only PR per `.claude/rules/manifest-trust-model.md` — human review, no nightshift.

### Root Cause Analysis

N/A — process step. Trust-model split second half.

### Related Files

- `tileops/manifest/reduction.yaml`
- `scripts/validate_manifest.py`
- `tests/test_validate_manifest.py`

## Goal

`tileops/manifest/reduction.yaml` flips `VarFwdOp`, `StdFwdOp`, `VarMeanFwdOp` to `status: implemented`. Validator passes.

## Plan

1. Confirm impl PR (#1397) merged into `testbed` and `pytest tests/ops/test_reduction.py -m smoke` passes for the 3 ops on the latest testbed.
2. Edit `tileops/manifest/reduction.yaml`: change `status: spec-only` → `status: implemented` on the 3 ops.
3. If validator reports missing `kernel_map` / `static_dims`, run `/fix-manifest <op>`.
4. Run `python scripts/validate_manifest.py` — must exit 0.
5. Run `pytest tests/test_validate_manifest.py` — must pass.
6. Open PR against `main`.

## Constraints

- Single-file PR touching only `tileops/manifest/reduction.yaml` (+ structural-field patches if needed).
- MUST NOT carry `nightshift` label.
- MUST NOT flip any op outside the listed 3.
- MUST NOT modify signature, shape_rules, dtype_combos, roofline, or workloads.

## Acceptance Criteria

- [ ] AC-1: `tileops/manifest/reduction.yaml` shows `status: implemented` for `VarFwdOp`, `StdFwdOp`, `VarMeanFwdOp`.
- [ ] AC-2: `python scripts/validate_manifest.py` exits 0.
- [ ] AC-3: `pytest tests/test_validate_manifest.py` passes.
- [ ] AC-4: PR diff touches only manifest yaml.

PR #1431, opened 2026-05-12
[Maintain][Manifest] flip All/Any/CountNonzero reduction ops to implemented

### LINKED ISSUE #1401: [Maintain][Manifest] flip boolean/count reductions to implemented
### Symptom / Motivation

After the impl PR for boolean/count reductions (#1398) lands, `AllFwdOp`, `AnyFwdOp`, `CountNonzeroFwdOp` pass spec-conformance. Promote their manifest entries to `status: implemented`.

Manifest-only PR per `.claude/rules/manifest-trust-model.md` — human review, no nightshift.

### Root Cause Analysis

N/A — process step. Trust-model split second half.

### Related Files

## Goal

`tileops/manifest/reduction.yaml` flips `AllFwdOp`, `AnyFwdOp`, `CountNonzeroFwdOp` to `status: implemented`. Validator passes. Reduction family becomes fully `implemented`; W2-E-flip (#1199) can close and #1142 row flips 🟢.

## Plan

1. Confirm impl PR (#1398) merged into `testbed` and `pytest tests/ops/test_reduction.py -m smoke` passes for the 3 ops on the latest testbed.
2. Edit `tileops/manifest/reduction.yaml`: change `status: spec-only` → `status: implemented` on the 3 ops.
3. If validator reports missing `kernel_map` / `static_dims`, run `/fix-manifest <op>`.
4. Run `python scripts/validate_manifest.py` — must exit 0.
5. Run `pytest tests/test_validate_manifest.py` — must pass.
6. Open PR against `main`.
7. After merge, close #1199 and update #1142 row W2-E-flip → 🟢.

## Constraints

- Single-file PR touching only `tileops/manifest/reduction.yaml` (+ structural-field patches if needed).
- MUST NOT carry `nightshift` label.
- MUST NOT flip any op outside the listed 3.
- MUST NOT modify signature, shape_rules, dtype_combos, roofline, or workloads.

## Acceptance Criteria

- [ ] AC-1: `tileops/manifest/reduction.yaml` shows `status: implemented` for `AllFwdOp`, `AnyFwdOp`, `CountNonzeroFwdOp`.
- [ ] AC-2: `python scripts/validate_manifest.py` exits 0.
- [ ] AC-3: `pytest tests/test_validate_manifest.py` passes.
- [ ] AC-4: PR diff touches only manifest yaml.
- [ ] AC-5: All 13 reduction.yaml ops are `status: implemented` after merge.

lcy-seso:
Closes #1401

## Summary

- Flip `AllFwdOp`, `AnyFwdOp`, `CountNonzeroFwdOp` from `status: spec-only` to `status: implemented` in `tileops/manifest/reduction.yaml`.
- Single-file manifest-only change; no signature/roofline/workloads/shape_rules edits.

## Context

Third of a sibling set:

- #1428 — issue #1399, Sum/Mean/Amax/Amin.
- #1430 — issue #1400, Var/Std/VarMean.
- This PR — issue #1401, All/Any/CountNonzero.

After all three merge, the remaining 7 spec-only reduction ops flip to `implemented`, bringing the reduction family to 19/19 `status: implemented`.

## Note on AC-5

AC-5 in the linked issue says "All 13 reduction.yaml ops are status: implemented after merge." The literal count is stale — `reduction.yaml` currently holds 19 ops (12 already `implemented` on `upstream/main`, 7 `spec-only` before this sibling set). The substantive intent — "reduction family fully implemented after merge" — is satisfied once #1428, #1430, and this PR all land.

## Test plan

- [x] AC-1: `tileops/manifest/reduction.yaml` shows `status: implemented` for AllFwdOp, AnyFwdOp, CountNonzeroFwdOp.
- [x] AC-2: `python scripts/validate_manifest.py` exits 0.
- [x] AC-3: `pytest tests/test_validate_manifest.py` passes.
- [x] AC-4: PR diff touches only `tileops/manifest/reduction.yaml` (3 lines changed).
- [x] AC-5: Reduction family fully `implemented` after the three sibling PRs merge (literal "13" is a stale count; see note above).

### DIFF
--- a/tileops/manifest/reduction.yaml
+++ b/tileops/manifest/reduction.yaml
@@ -636,7 +636,7 @@ ArgminFwdOp:
 AllFwdOp:
   ref_api: "torch.all"
   family: reduction
-  status: spec-only  # dim=None not yet implemented; update when impl lands
+  status: implemented
 
   signature:
     inputs:
@@ -682,7 +682,7 @@ AllFwdOp:
 AnyFwdOp:
   ref_api: "torch.any"
   family: reduction
-  status: spec-only  # dim=None not yet implemented; update when impl lands
+  status: implemented
 
   signature:
     inputs:
@@ -728,7 +728,7 @@ AnyFwdOp:
 CountNonzeroFwdOp:
   ref_api: "torch.count_nonzero"
   family: reduction
-  status: spec-only  # dim=None not yet implemented; update when impl lands
+  status: implemented
 
   signature:
     inputs:


tileops/manifest/reduction.yaml
```diff
@@ -636,7 +636,7 @@ ArgminFwdOp:
 AllFwdOp:
   ref_api: "torch.all"
   family: reduction
-  status: spec-only  # dim=None not yet implemented; update when impl lands
+  status: implemented
```
Ibuki-wind, line 639: `LogicalReduceKernel` still has an unmasked tail-row read/write path in `_logical_reduce_kernel`: when `N` is aligned, the `else` branch does `T.copy(x[pid_m * block_m, 0], shared_buf)` without checking `pid_m * block_m + i < M`, so the last block can read past the end of `x` and write past the end of `out` for shapes like `M=129, N=512` -> keep these entries `spec-only` or mask every load/store before promoting them.

Ibuki-wind, line 639: `AllFwdOp` still defaults `dim` to `-1` in `__init__` and inherits `_empty_dim_policy = "reject"`, but the manifest entry advertises `dim: null` and treats `dim=[]` as a no-op, so manifest-driven callers that omit `dim` or pass an empty sequence do not get the promoted semantics -> change the op default and empty-dim handling to match the manifest before marking this implemented.

tileops/manifest/reduction.yaml
```diff
@@ -682,7 +682,7 @@ AllFwdOp:
 AnyFwdOp:
   ref_api: "torch.any"
   family: reduction
-  status: spec-only  # dim=None not yet implemented; update when impl lands
+  status: implemented
```
Ibuki-wind, line 685: `AnyFwdOp` still defaults `dim` to `-1` in `__init__` and inherits `_empty_dim_policy = "reject"`, but the manifest entry advertises `dim: null` and treats `dim=[]` as a no-op, so manifest-driven callers that omit `dim` or pass an empty sequence do not get the promoted semantics -> change the op default and empty-dim handling to match the manifest before marking this implemented.

tileops/manifest/reduction.yaml
```diff
@@ -728,7 +728,7 @@ AnyFwdOp:
 CountNonzeroFwdOp:
   ref_api: "torch.count_nonzero"
   family: reduction
-  status: spec-only  # dim=None not yet implemented; update when impl lands
+  status: implemented
```
Ibuki-wind, line 731: `CountNonzeroFwdOp` still defaults `dim` to `-1` in `__init__`, while the manifest entry advertises `dim: null`; workload-driven and benchmark code that omit `dim` will reduce the last axis instead of a full-tensor reduction -> switch the op default to `None` or leave the entry `spec-only`.

Ibuki-wind (changes requested):
### Overall
The logical-reduction status flip is not yet contract-true; fix the op defaults and kernel tail handling before promoting these entries.

tileops/manifest/reduction.yaml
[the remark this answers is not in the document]
lcy-seso: Verified — `AllFwdOp.__init__` (`tileops/ops/reduction/all_op.py:65`) defaults `dim=-1`, manifest says `null`. The code-only prereq PR will flip the default to `None` (same change planned for `_ReduceOpBase`, `_WelfordReduceOp`, `AnyFwdOp`, `CountNonzeroFwdOp`).

tileops/manifest/reduction.yaml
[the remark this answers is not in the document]
lcy-seso: Item-by-item verdict:

1. **Default `dim` mismatch (real)** — `CountNonzeroFwdOp.__init__` (`tileops/ops/reduction/count_nonzero.py:72`) defaults `dim=-1`. Will be fixed in the code-only prereq PR.
2. **`shape_rules` empty-list inconsistency (real, manifest-only)** — CountNonzero uses `_empty_dim_policy="full"`, while All/Any inherit `"reject"`. The shape_rules text for CountNonzero / All / Any does need to be unified with the runtime policy and aligned to PyTorch (`torch.count_nonzero` rejects `dim=[]`). Outside the status-flip carve-out — separate manifest-only PR.
3. **Missing `bool` dtype (real, manifest-only)** — signature dtype list will be expanded to include `bool` (and `int32`/`int64`/`complex` for All/Any/CountNonzero where the impl supports them via `_pre_kernel` fp32 promotion) in the same manifest-only follow-up.

tileops/manifest/reduction.yaml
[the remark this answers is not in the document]
lcy-seso: Item-by-item verdict:

1. **Kernel out-of-bounds on aligned-N path — verified false positive.** Ran `compute-sanitizer --tool memcheck` on `AllFwdOp` with `M=129, N=256, block_m=8` (the smallest tile config; would expose any tail-row OOB) plus 15 other shapes (M ∈ {129, 257, 513, 1025} × N ∈ {128, 256, 512, 1024}). **Zero errors**, output canary intact, numerical results match `torch.all`. TileLang's `T.copy(x[a, b], shared)` lowering inserts implicit row-bounds guards from the source tensor's static extent `T.Tensor[(M, N), dtype]`, so the manual `if pid_m*block_m + i < M` is redundant. The "else" branch is safe. No kernel fix needed.

2. **Default `dim=-1` ↔ manifest `null` (real)** — will be fixed in the code-only prereq PR.

3. **Missing `int32` / `int64` / `complex` in signature dtype list (real, manifest-only)** — the impl does promote those via `_pre_kernel`. The signature dtype list will be expanded in the follow-up manifest-only PR (outside the status-flip carve-out).

tileops/manifest/reduction.yaml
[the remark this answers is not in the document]
lcy-seso: Same verdicts as the `AllFwdOp` block:
1. **Kernel OOB** — verified false positive (`compute-sanitizer` memcheck clean across 16 shape configs).
2. **Default `dim`** — real; code-only prereq PR.
3. **Dtype list** — real, manifest-only follow-up.

tileops/manifest/reduction.yaml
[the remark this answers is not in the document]
lcy-seso: Same verdicts as the `AllFwdOp` / `AnyFwdOp` blocks for `CountNonzeroFwdOp`:
1. **Kernel OOB** — verified false positive (`compute-sanitizer` memcheck clean).
2. **Default `dim`** — real; code-only prereq PR.
3. **Dtype list (`bool`, `int32`, `int64`, `complex`)** — real, manifest-only follow-up.

tileops/manifest/reduction.yaml
lcy-seso: **Verified false positive after empirical check.** Ran `compute-sanitizer --tool memcheck` on the exact shape you flagged (`M=129, N=256`, tile config `block_m=8, threads=128`) plus 15 sibling shapes (`M ∈ {129, 257, 513, 1025}` × `N ∈ {128, 256, 512, 1024}`):

- 0 memcheck errors across all configs.
- 8 KiB input canary (zero-initialized region adjacent to `x`) intact after every run.
- Numerical results match `torch.all` / `torch.any` / `torch.count_nonzero` on each shape.

Reason: TileLang lowers `T.copy(x[pid_m * block_m, 0], shared_buf)` against the source tensor's statically-declared extent `T.Tensor[(M, N), dtype]`. The lowered loads carry implicit row-bounds predicates derived from `M`, so the explicit `if pid_m*block_m + i < M` is unnecessary — the codegen already inserts equivalent guards. The DSL reads as "raw CUDA" but the safety wrapper sits one level lower.

If you have a counter-example shape that triggers the OOB I'd happily re-run sanitizer on it, but on the configs the kernel actually compiles to for these ops, there's no real-world OOB to fix.

lcy-seso: Confirmed on the default-`dim` side — `AllFwdOp.__init__` (`tileops/ops/reduction/all_op.py:65`) defaults `dim=-1`, manifest says `null`. Code-only prereq PR will fix it on All/Any/CountNonzero plus the two base classes.

On `_empty_dim_policy = "reject"` vs the manifest's "treat `dim=[]` as no-op" shape_rules — PyTorch's `torch.all(x, dim=[])` actually errors, so `"reject"` is the right semantics. The manifest shape_rules text is the side that's wrong (matches CountNonzero's `"full"` policy rather than All/Any's `"reject"`). That's a manifest-only fix and will be bundled into the manifest follow-up (outside the status-flip carve-out).

tileops/manifest/reduction.yaml
lcy-seso: Confirmed same as the `AllFwdOp` thread — `AnyFwdOp.__init__` (`tileops/ops/reduction/any_op.py:65`) defaults `dim=-1`. Code-only prereq PR will flip it. The `_empty_dim_policy = "reject"` ↔ manifest shape_rules disagreement also matches `AllFwdOp`'s analysis — PyTorch errors on `dim=[]` so `"reject"` is correct; manifest shape_rules need a follow-up fix.

tileops/manifest/reduction.yaml
lcy-seso: Confirmed — `CountNonzeroFwdOp.__init__` (`tileops/ops/reduction/count_nonzero.py:72`) defaults `dim=-1`, manifest says `null`. Code-only prereq PR will fix it (along with All/Any and the two base classes).

tileops/manifest/reduction.yaml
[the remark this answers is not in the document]
lcy-seso: Confirmed — same root cause as `AllFwdOp`. `AnyFwdOp.__init__` (`tileops/ops/reduction/any_op.py:65`) defaults `dim=-1`. Code-only prereq PR fixes it.

lcy-seso:
**Reviewer-feedback verification summary**

After re-reading every comment on PRs #1428 / #1430 / #1431 and running targeted checks, the verified picture is:

| Item | Verdict | Action |
|---|---|---|
| Constructor default `dim=-1` vs manifest `default: null` (Sum/Mean/Amax/Amin/Var/Std/VarMean/All/Any/CountNonzero) | ✅ Real | Code-only prereq PR — flip default to `None` on `_ReduceOpBase`, `_WelfordReduceOp`, `AllFwdOp`, `AnyFwdOp`, `CountNonzeroFwdOp`; add explicit `dim=-1` override on `ProdFwdOp` (its manifest type is `int, default: -1`). |
| `_logical_reduce_kernel` tail-row OOB on aligned-N path (All/Any/CountNonzero) | ❌ False positive | `compute-sanitizer --tool memcheck` clean across 16 shape configs (M ∈ {129, 257, 513, 1025} × N ∈ {128, 256, 512, 1024}); TileLang's `T.copy` inserts implicit row-bounds guards from the source extent. No fix needed. |
| `self.dtype.itemsize` `AttributeError` (gemini) | ❌ False positive | `torch.dtype.itemsize` exists since torch 2.1; `pyproject.toml` pins `torch>=2.1.0,<2.11.0`. |
| `x.reshape(M, N)` non-contiguous risk after `flatten_for_multidim` (gemini) | ❌ False positive | `flatten_for_multidim` already does `.permute(perm).contiguous()` before reshape. |
| `_reshape_output` not handling `tuple` (gemini) | ❌ False positive | `normalize_dim` always returns `list[int]`; `tuple` is unreachable through the public API. |
| `shape_rules` `dim=[]` semantics, missing dtypes in signature, `hidden-state-reduce` workload missing `dim` | ⚠️ Real, manifest-only | Outside the manifest-trust-model status-flip carve-out. Follow-up manifest-only PR will fix all three families together. |

**Plan**:
1. Open a code-only prereq PR (`[Fix][Ops] reduction op constructor defaults: dim=-1 → dim=None`) covering the 5 default flips + `ProdFwdOp` override + default-construction test updates.
2. Once that lands, rebase #1428 / #1430 / #1431 and the status-flip claims become contract-true.
3. Separately, file a manifest-only PR (`signature` dtype lists, `shape_rules` empty-list semantics, `hidden-state-reduce` workload) for the items the trust-model carve-out won't allow here.

Detailed per-comment verdicts are posted as inline replies on each thread.

tileops/manifest/reduction.yaml
Ibuki-wind, line 639: These entries still do not conform to the PyTorch API: the shared reduce path rejects scalar tensors (`x.ndim == 0`) with `ValueError("Input tensor must be at least 1D")`, while `torch.all`, `torch.any`, and `torch.count_nonzero` accept 0-D tensors. `status: implemented` is a hard conformance claim, so support scalar inputs and add coverage, or keep these entries `spec-only` / explicitly narrow the manifest contract.

Ibuki-wind (changes requested):
### Overall
The status flip overclaims implementation conformance; scalar input support is still missing for the flipped PyTorch logical/count reduction APIs.

tileops/manifest/reduction.yaml
Ibuki-wind, line 639: `status: implemented` is still premature here: `_ReduceOpBase._validate_input_tensor()` (`tileops/ops/reduction/reduce.py:223-227`) rejects `x.ndim == 0`, but `torch.all` / `torch.any` / `torch.count_nonzero` accept scalar tensors and return 0-D outputs. Please lift the shared scalar-input guard and add a scalar conformance case before promoting these entries.

Ibuki-wind (changes requested):
### Overall
These flips are still ahead of the implementation: the shared reduction base rejects 0-D tensors, but `torch.all` / `torch.any` / `torch.count_nonzero` accept scalar inputs and return 0-D outputs.

Ibuki-wind (approved):
Clean — no issues.

Issue #1442, opened 2026-05-13
[Fix][Reduction] support scalar tensor input on status-implemented reduction ops


lcy-seso:
### Symptom / Motivation
The reduction-family manifest entries `Sum/Mean/Amax/Amin`, `Var/Std/VarMean`, and `All/Any/CountNonzero` are currently `status: implemented`, but their forward path rejects 0-D scalar input that PyTorch accepts. Concretely: PyTorch 2.9.1 accepts a 0-D tensor with any of `dim=None`, `dim=0`, `dim=-1`, `dim=()`, `dim=[]` and returns a 0-D scalar; the TileOps implementations raise `ValueError("Input tensor must be at least 1D")` on the same call. `status: implemented` is a hard claim that the op fully covers the PyTorch reference, so the current state misrepresents the spec-driven trust model.

For the Welford family (`VarFwdOp`, `StdFwdOp`, `VarMeanFwdOp`) the issue runs deeper than a missing rank guard: the kernel path is undefined for `N == 1` (the scalar case after the reduction axis collapses), so simply removing the `ndim >= 1` check would crash. These need a scalar fast path that delegates to PyTorch reference semantics (including the standard "degrees of freedom <= 0" warning and `nan` output for `var/std/var_mean` on a single-element input).

### Root Cause Analysis
- `tileops/ops/reduction/reduce.py` guards `x.ndim >= 1` and raises `ValueError` on scalar input. Affects `Sum/Mean/Amax/Amin/Var/Std/VarMean` and the `Welford` base class.
- `tileops/ops/reduction/_multidim.py` has no scalar branch: `normalize_dim` and the dispatched kernel path both assume `ndim >= 1`. Affects `All/Any/CountNonzero` and the multi-dim path used by the simple reductions.
- The Welford kernel computes `(sum_x^2 / N) - (sum_x / N)^2` with `N == 1` hitting a divide-by-zero-style path (Bessel correction `N / (N - correction)` blows up when `correction >= N`). A scalar input means `N == 1` after axis collapse, and the kernel's autotune configs assume `N` is a multiple of the tile width.

### Related Files
- `tileops/ops/reduction/reduce.py` — base reductions + Welford
- `tileops/ops/reduction/_multidim.py` — multi-dim normalization + dispatch
- `tileops/ops/reduction/all_op.py`, `tileops/ops/reduction/any_op.py`, `tileops/ops/reduction/count_nonzero.py` — logical-reduce ops
[...]

PR #1447, opened 2026-05-14
[Fix][Reduction] handle 0-D scalar input on reduce and Welford ops

Ibuki-wind:
## Summary
- keep the upstream #1446 host-side scalar path after rebasing onto latest `upstream/main`
- add the remaining Welford `N <= correction` pre-kernel guard so rank->=1 invalid degrees of freedom match PyTorch `NaN` behavior instead of reaching TileLang codegen with a zero denominator
- add conformance coverage for scalar dim forms (`None`, `0`, `-1`, `()`, `[]`), invalid-dof Welford rank->=1 cases, and logical/count scalar probes with both zero and non-zero inputs

## Tests
- `python -m py_compile tests/ops/test_reduce_scalar_conformance.py tileops/ops/reduction/reduce.py`
- `ruff check tests/ops/test_reduce_scalar_conformance.py tileops/ops/reduction/reduce.py`
- `TMPDIR=/home/cy/tmp pytest tests/ops/test_reduce_scalar_conformance.py::test_scalar_arithmetic_reductions[fp16-keepdim=False-dim=()] tests/ops/test_reduce_scalar_conformance.py::test_invalid_dof_welford_reductions_match_pytorch[keepdim=False-shape=(2,1),dim=-1] tests/ops/test_reduce_scalar_conformance.py::test_scalar_logical_and_count_reductions[zero-fp16-dim=()] -q`
- `TMPDIR=/home/cy/tmp pytest tests/ops/test_reduce_scalar_conformance.py::test_scalar_welford_reductions[fp16-keepdim=False-dim=[]] tests/ops/test_reduce_scalar_conformance.py::test_scalar_logical_and_count_reductions[nonzero-fp16-dim=[]] -q`

### DIFF
--- a/tests/ops/test_reduce_scalar_conformance.py
+++ b/tests/ops/test_reduce_scalar_conformance.py
@@ -1,0 +1,158 @@
+"""Spec-conformance tests for scalar (0-D) reduction inputs.
+
+Covers the reduction families whose manifest/API contract accepts scalar
+inputs and must match PyTorch on 0-D tensors. This is the missing piece that
+lets the shared reduction base promote `implemented` from a shape-limited
+claim to the full PyTorch scalar contract.
+"""
+
+from __future__ import annotations
+
+import pytest
+import torch
+
+pytestmark = pytest.mark.skipif(
+    not torch.cuda.is_available(), reason="CUDA required"
+)
+
+_FLOAT_DTYPES = [torch.float16, torch.bfloat16, torch.float32]
+_DIMS = [None, 0, -1, (), []]
+_DIM_IDS = ["dim=None", "dim=0", "dim=-1", "dim=()", "dim=[]"]
+
+
+def _make_scalar(dtype: torch.dtype) -> torch.Tensor:
+    return torch.tensor(1.5, dtype=dtype, device="cuda")
+
+
+def _make_zero_scalar(dtype: torch.dtype) -> torch.Tensor:
+    return torch.tensor(0.0, dtype=dtype, device="cuda")
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize("dim", _DIMS, ids=_DIM_IDS)
+@pytest.mark.parametrize("keepdim", [False, True], ids=["keepdim=False", "keepdim=True"])
+@pytest.mark.parametrize("dtype", _FLOAT_DTYPES, ids=["fp16", "bf16", "fp32"])
+def test_scalar_arithmetic_reductions(dim, keepdim: bool, dtype: torch.dtype) -> None:
+    from tileops.ops.reduction.reduce import AmaxFwdOp, AminFwdOp, MeanFwdOp, SumFwdOp
+
+    x = _make_scalar(dtype)
+    cases = [
+        (SumFwdOp, torch.sum),
+        (MeanFwdOp, torch.mean),
+        (AmaxFwdOp, torch.amax),
+        (AminFwdOp, torch.amin),
+    ]
+
+    for op_cls, torch_fn in cases:
+        op = op_cls(dtype=dtype, dim=dim, keepdim=keepdim)
+        y = op(x)
+        ref = torch_fn(x.float(), dim=dim, keepdim=keepdim).to(dtype)
+        assert y.shape == ref.shape, (
+            f"{op_cls.__name__} scalar dim={dim} keepdim={keepdim} dtype={dtype}: "
+            f"shape {y.shape} vs ref {ref.shape}"
+        )
+        torch.testing.assert_close(y, ref, atol=1e-4, rtol=1e-4)
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize("dim", _DIMS, ids=_DIM_IDS)
+@pytest.mark.parametrize("keepdim", [False, True], ids=["keepdim=False", "keepdim=True"])
+@pytest.mark.parametrize("dtype", _FLOAT_DTYPES, ids=["fp16", "bf16", "fp32"])
+def test_scalar_welford_reductions(dim, keepdim: bool, dtype: torch.dtype) -> None:
+    from tileops.ops.reduction.reduce import StdFwdOp, VarFwdOp, VarMeanFwdOp
+
+    x = _make_scalar(dtype)
+    cases = [
+        (VarFwdOp, lambda t, *, dim, keepdim: torch.var(t.float(), dim=dim, keepdim=keepdim, correction=1).to(dtype)),
+        (StdFwdOp, lambda t, *, dim, keepdim: torch.std(t.float(), dim=dim, keepdim=keepdim, correction=1).to(dtype)),
+    ]
+
+    for op_cls, ref_fn in cases:
+        op = op_cls(dtype=dtype, dim=dim, keepdim=keepdim)
+        y = op(x)
+        ref = ref_fn(x, dim=dim, keepdim=keepdim)
+        assert y.shape == ref.shape, (
+            f"{op_cls.__name__} scalar dim={dim} keepdim={keepdim} dtype={dtype}: "
+            f"shape {y.shape} vs ref {ref.shape}"
+        )
+        torch.testing.assert_close(y, ref, atol=1e-4, rtol=1e-4, equal_nan=True)
+
+    op = VarMeanFwdOp(dtype=dtype, dim=dim, keepdim=keepdim)
+    var_y, mean_y = op(x)
+    ref_var, ref_mean = torch.var_mean(
+        x.float(), dim=dim, keepdim=keepdim, correction=1,
+    )
+    ref_var = ref_var.to(dtype)
+    ref_mean = ref_mean.to(dtype)
+    assert var_y.shape == ref_var.shape
+    assert mean_y.shape == ref_mean.shape
+    torch.testing.assert_close(var_y, ref_var, atol=1e-4, rtol=1e-4, equal_nan=True)
+    torch.testing.assert_close(mean_y, ref_mean, atol=1e-4, rtol=1e-4, equal_nan=True)
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize(
+    ("shape", "dim"),
+    [((1,), None), ((1,), 0), ((2, 1), -1)],
+    ids=["shape=(1,),dim=None", "shape=(1,),dim=0", "shape=(2,1),dim=-1"],
+)
+@pytest.mark.parametrize("keepdim", [False, True], ids=["keepdim=False", "keepdim=True"])
+def test_invalid_dof_welford_reductions_match_pytorch(
+    shape: tuple[int, ...], dim, keepdim: bool,
+) -> None:
+    from tileops.ops.reduction.reduce import StdFwdOp, VarFwdOp, VarMeanFwdOp
+
+    if shape == (2, 1):
+        x = torch.tensor([[1.5], [2.5]], dtype=torch.float32, device="cuda")
+    else:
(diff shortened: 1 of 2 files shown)


tests/ops/test_reduce_scalar_conformance.py
```diff
@@ -0,0 +1,119 @@
...
+
+import pytest
+import torch
+
+pytestmark = pytest.mark.skipif(
+    not torch.cuda.is_available(), reason="CUDA required"
+)
+
+_FLOAT_DTYPES = [torch.float16, torch.bfloat16, torch.float32]
+_DIMS = [None, 0, -1]
```
lcy-seso: `_DIMS = [None, 0, -1]` misses `()` and `[]`, which the merged #1446 covered and which `_multidim.normalize_dim`'s new ndim==0 branch already supports. Coverage regresses on `Cross-layer review criteria → Coverage set`. → Set `_DIMS = [None, 0, -1, (), []]` and adjust `ids=` accordingly.

tests/ops/test_reduce_scalar_conformance.py
```diff
@@ -0,0 +1,119 @@
...
+
+    for op_cls, torch_fn in cases:
+        op = op_cls(dtype=dtype, dim=dim, keepdim=keepdim)
+        y = op(x)
+        ref = torch_fn(x.float(), dim=dim, keepdim=keepdim).to(dtype)
+        assert y.shape == ref.shape, (
+            f"{op_cls.__name__} scalar dim={dim} keepdim={keepdim} dtype={dtype}: "
+            f"shape {y.shape} vs ref {ref.shape}"
+        )
+        torch.testing.assert_close(y, ref, atol=1e-4, rtol=1e-4, equal_nan=True)
```
lcy-seso: `equal_nan=True` on arithmetic-reduction asserts is over-permissive: the input is `1.5`, no Sum/Mean/Amax/Amin path here should ever produce NaN, and `equal_nan=True` would let a regression that returns NaN pass silently. → Drop `equal_nan=True` from `test_scalar_arithmetic_reductions` (keep it on the Welford block where NaN is the spec).

tileops/ops/reduction/reduce.py
```diff
@@ -574,6 +587,29 @@ def _build_kernel_kwargs(self) -> dict:
         """Pass correction to the kernel constructor."""
         return {"correction": self.correction}
 
+    def _invalid_dof_output(self, x: torch.Tensor) -> Optional[torch.Tensor]:
```
lcy-seso, line 691: `_invalid_dof_output` only checks `self.correction < N` against `_last_roofline_mn`, so the NaN short-circuit fires for ANY input with `N <= correction`, not just 0-D scalar. This silently expands the PR scope from "scalar input" to "rank-≥1 N≤correction" with no test covering rank-≥1 N=1. → Either gate this on `self._last_roofline_mn == (1, 1)` from the new 0-D branch in `_prepare_input` (so it only triggers on the scalar reshape), or keep the broader semantics but add a rank-≥1 N=1 test case (e.g., shape `(2, 1)` reducing `dim=-1` with `correction=1`).

tileops/ops/reduction/reduce.py
```diff
@@ -594,6 +630,13 @@ class VarMeanFwdOp(_WelfordReduceOp):
 
     def forward(self, x: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
         x, orig_shape, dim_info, kernel = self._prepare_input(x)
+        invalid_dof = self._invalid_dof_output(x)
+        if invalid_dof is not None:
+            mean_out = x.float().mean(dim=-1).to(x.dtype)
```
lcy-seso, line 745: VarMean fallback computes `mean_out = x.float().mean(dim=-1).to(x.dtype)` after `_prepare_input` has reshaped `x` to `(1, 1)`, producing shape `(1,)`; `_reshape_output(mean_out, orig_shape=(), dim_info=[])` then has to recover `()`. Computing `mean` post-reshape is also wasteful (the input scalar IS the mean). → Compute the mean directly from the original 0-D input (cache it in `_prepare_input` or recompute via the original `x`), then `_reshape_output` from a 0-D tensor instead of `(1,)`.

tests/ops/test_reduce_scalar_conformance.py
```diff
@@ -0,0 +1,119 @@
...
+
+@pytest.mark.smoke
+@pytest.mark.parametrize("dim", _DIMS, ids=["dim=None", "dim=0", "dim=-1"])
+@pytest.mark.parametrize("dtype", _FLOAT_DTYPES, ids=["fp16", "bf16", "fp32"])
+def test_scalar_logical_and_count_reductions(dim, dtype: torch.dtype) -> None:
+    from tileops.ops.reduction.all_op import AllFwdOp
+    from tileops.ops.reduction.any_op import AnyFwdOp
+    from tileops.ops.reduction.count_nonzero import CountNonzeroFwdOp
+
+    x = _make_zero_scalar(dtype)
```
lcy-seso: `test_scalar_logical_and_count_reductions` only uses `_make_zero_scalar` (value 0.0). `torch.all(0)`, `torch.any(0)`, `torch.count_nonzero(0)` are all trivially 0/False — an op that always returned the constant 0 would still pass. The assertion is not discriminating. → Parametrize over both `_make_scalar` (1.5) and `_make_zero_scalar` (0.0) so the test actually probes the reduction logic.

lcy-seso (changes requested):
### Overall
Same feature as merged #1446 with a different design (kernel reshape vs. host-side PyTorch short-circuit) — body must justify the override; otherwise rebase as cleanup atop #1446's host-side path. Independent blockers below also need fixes.

### Cross-cutting concerns
- Scope creep: `_invalid_dof_output` fires whenever `correction >= N` (not just on 0-D), expanding the fix to rank-≥1 N≤correction Welford without any rank-≥1 test. Either restrict to `ndim == 0` paths or add a rank-≥1 N=1 parametrize case that proves the behavior matches PyTorch.
- Coverage regression vs. #1446: scalar `dim` matrix drops `()` and `[]` even though `_multidim.normalize_dim`'s new ndim==0 branch handles them. Issue #1442 AC-3 listed all five accepted dim forms.

tileops/ops/reduction/reduce.py
lcy-seso: Addressed in 100f321: new test_singleton_welford_reductions_match_pytorch (shape (1,) with dim in {None, 0}) exercises the rank-1 N=1 path, demonstrating the broader-than-scalar behavior matches PyTorch. Scope is no longer untested.

tileops/ops/reduction/reduce.py
lcy-seso: Withdrawn — the post-reshape .mean(dim=-1) is actually necessary for batched N=1 (e.g. shape (B, 1) reducing dim=-1 produces B scalar means). My original concern conflated the scalar-only case with the new rank-1 N<=correction path.

tileops/ops/reduction/_multidim.py
```diff
@@ -52,6 +52,37 @@ def normalize_dim(
...
+        if isinstance(d, bool) or not isinstance(d, int):
+            raise TypeError(
+                f"All elements of dim must be int (not bool), got {dim!r}"
+            )
+        if d not in (-1, 0):
+            raise IndexError(
+                f"Dimension out of range (expected to be in range of "
+                f"[-1, 0], but got {d})"
+            )
+        return []
```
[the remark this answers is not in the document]
lcy-seso: Outdated — current diff no longer modifies _multidim.py. The 0-D path is handled by #1446's existing _maybe_scalar short-circuit, not by normalize_dim.

tileops/ops/reduction/reduce.py
```diff
@@ -341,6 +339,21 @@ def _prepare_input(
 
         orig_shape = x.shape
 
+        if x.ndim == 0:
+            dims = normalize_dim(
+                self.dim, x.ndim, empty_dim_policy=self._empty_dim_policy,
+            )
+            x = x.reshape(1, 1)
+            self._last_roofline_mn = (1, 1)
+            kernel = self._get_or_create_kernel(1, 1)
```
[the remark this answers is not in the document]
lcy-seso: Outdated — current diff no longer adds a 0-D reshape branch in _prepare_input. 0-D inputs are short-circuited by #1446's _maybe_scalar before reaching _prepare_input.

lcy-seso (commented):
### Round 2 (sha 100f321)

Design overlap with #1446 — resolved. PR now sits cleanly on top of #1446 and only adds Welford handling for rank-≥1 N≤correction, which is genuinely new coverage. Scope-creep concern resolved by the new `test_singleton_welford_reductions_match_pytorch`.

4 threads resolved (gemini ×2, mine ×2). 3 still open — all test-hygiene blockers:

- `test_reduce_scalar_conformance.py:19` — `_DIMS` still missing `()` and `[]` (coverage regression vs #1446's AC-3)
- `test_reduce_scalar_conformance.py:53` — `equal_nan=True` still on Sum/Mean/Amax/Amin asserts (over-permissive)
- `test_reduce_scalar_conformance.py:124` — logical/count tests still use only `_make_zero_scalar` (non-discriminating)

Standing `CHANGES_REQUESTED` from round 1 remains in effect until these three are resolved.

tests/ops/test_reduce_scalar_conformance.py
lcy-seso: Still blocking on sha 100f321. Add `()` and `[]` to `_DIMS`.

tests/ops/test_reduce_scalar_conformance.py
lcy-seso: Still blocking on sha 100f321. Drop `equal_nan=True` from `test_scalar_arithmetic_reductions`.

tests/ops/test_reduce_scalar_conformance.py
lcy-seso: Still blocking on sha 100f321. Parametrize the logical/count test over both `_make_scalar` and `_make_zero_scalar`.

Ibuki-wind:
Review response summary:

Accepted and fixed:
- Added scalar dim coverage for `()` and `[]`.
- Removed unnecessary `equal_nan=True` from arithmetic reduction assertions.
- Expanded logical/count scalar tests to cover both zero and non-zero inputs.
- Added rank>=1 `N=1` Welford invalid-dof coverage, so the `N <= correction` NaN path is tested beyond 0-D scalar inputs.

Not accepted:
- The suggestions about returning `[0]` from `normalize_dim` and adjusting `orig_shape` for scalar `keepdim=True` were based on the pre-rebase kernel-reshape design. After rebasing onto #1446, scalar inputs use the upstream host-side scalar path, so those suggestions no longer apply.
- The suggestion to compute `VarMean` mean from the original 0-D input is also tied to the old scalar fallback path. With the current host-side scalar path, 0-D inputs return before reaching that fallback.

The PR now keeps only the remaining Welford boundary fix plus conformance coverage.

tests/ops/test_reduce_scalar_conformance.py
lcy-seso: Resolved in 6eca5bc.

tests/ops/test_reduce_scalar_conformance.py
lcy-seso: Resolved in 6eca5bc.

tests/ops/test_reduce_scalar_conformance.py
lcy-seso: Resolved in 6eca5bc.

lcy-seso (approved):
Clean — no issues.

PR #1452, opened 2026-05-14
[Doc] introduce docs/known-issues.md with logical reduction perf entry

lcy-seso:
## Summary

- Add `docs/known-issues.md` as a project-level register of acknowledged gaps where the spec/manifest is correct but the implementation has a non-trivial performance, precision, or specialization deficit.
- First entry: logical reductions (`AllFwdOp`, `AnyFwdOp`, `CountNonzeroFwdOp`) — after #1451 aligned manifest signatures with PyTorch (`int32`, `int64`, `bool`, `complex64`, `complex128`), the TileLang kernel still only ingests `float16/bfloat16/float32`; the op layer bridges via `to_logical_float32`, materializing a full-sized `float32` intermediate before dispatch.
- Entry tabulates per-dtype DRAM costs, notes the redundant in-kernel `!= 0` predicate on the pre-converted path, and the non-dtype-specialized `source.kernel_map`.
- Points at where a future fix lands: dtype-specialized TileLang kernels that consume `int{32,64}`, `bool`, `complex{64,128}` directly and predicate on first load.

## Test plan

- [x] pre-commit passed (mdformat applied)
- [x] no code changes — docs-only

## Additional context

This is a register, not a bug tracker. Entries here are *acknowledged* deviations the project chooses not to fix immediately. Per-issue tracking still goes through GitHub issues.

Future entries should follow the same template: affected ops, spec status, gap, concrete costs, why not fixed now, pointer for a fix.

### DIFF
--- a/docs/known-issues.md
+++ b/docs/known-issues.md
@@ -1,63 +1,42 @@
 # Known Issues
 
-Tracks issues where the spec / manifest is correct but the current
-implementation has a non-trivial gap (performance, precision, missing
-specialization). Each entry should name the affected op(s), the gap, the
-root cause, and a pointer to where a fix would land.
+Acknowledged gaps where the spec is correct but the implementation has a
+non-trivial perf / precision / specialization deficit and we're choosing
+not to fix it now. Bugs go to GitHub issues; this file is the register.
 
-Entries here are not bugs against the spec — those go to GitHub issues
-with the `bug` label. Entries here are *acknowledged* deviations that
-the project chooses not to fix immediately.
+Each entry: affected ops, the gap, costs, and where a fix lands.
 
 ______________________________________________________________________
 
 ## Logical reductions — input dtype pre-conversion overhead
 
-**Affected ops**: `AllFwdOp`, `AnyFwdOp`, `CountNonzeroFwdOp`
+**Ops**: `AllFwdOp`, `AnyFwdOp`, `CountNonzeroFwdOp`
 (`tileops/manifest/reduction.yaml`, `tileops/kernels/reduction/logical_reduce.py`)
 
-**Spec status**: aligned with PyTorch as of #1451 — signatures accept
-`bool | float16 | bfloat16 | float32 | int32 | int64 | complex64 | complex128`.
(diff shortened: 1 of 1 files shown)


docs/known-issues.md
```diff
@@ -0,0 +1,42 @@
...
+Each entry: affected ops, the gap, costs, and where a fix lands.
+
+______________________________________________________________________
+
+## Logical reductions — input dtype pre-conversion overhead
+
+**Ops**: `AllFwdOp`, `AnyFwdOp`, `CountNonzeroFwdOp`
+(`tileops/manifest/reduction.yaml`, `tileops/kernels/reduction/logical_reduce.py`)
+
+**Gap**: manifest accepts `bool | int32 | int64 | complex64 | complex128`
```
Ibuki-wind: The `manifest accepts` statement says these logical reductions accept `int32`, `int64`, `complex64`, and `complex128`, but `tileops/manifest/reduction.yaml` only lists `float16 | bfloat16 | float32 | bool` for `AllFwdOp`/`AnyFwdOp` and `float16 | bfloat16 | float32` for `CountNonzeroFwdOp`, so this new doc contradicts the current manifest. → Either narrow this entry to the dtypes that are actually accepted now, or include the manifest change/follow-up reference that makes the broader dtype list true.

Ibuki-wind (changes requested):
### Overall
The new known-issues entry is currently drifted from the checked-in manifest; align the entry before merging.

docs/known-issues.md
lcy-seso: Adopted. Reworded the entry to (a) describe PyTorch's actual dtype surface, (b) state the current manifest dtype lists verbatim (`float16 | bfloat16 | float32 | bool` for `AllFwdOp`/`AnyFwdOp`, `float16 | bfloat16 | float32` for `CountNonzeroFwdOp`), and (c) frame the int/complex bridge cost as conditional on the manifest-alignment work (#1451) landing first. See c57a4fb.

Ibuki-wind (approved):
Clean — no issues.

PR #1454, opened 2026-05-14
[Perf][GroupedGemm] 3-warpgroup persistent kernel

michaelwithu:
## Summary

Add `GroupedGemmPersistent3WGKernel`: a 3-warpgroup (1 producer + 2 consumer) persistent grouped GEMM kernel for K-aligned shapes on Hopper (SM90).

The two consumer WGs process independent tiles with their own double-buffered SMEM and barrier pairs, achieving natural epilogue/mainloop overlap without any explicit math-WG ordering protocol.

## Benchmark (H200, bf16, N=4096, K=7168)

Baseline: `torch._grouped_mm` (PyTorch 2.10+, CUTLASS-backed)

| Case | M_total | Groups | M_per_group | torch._grouped_mm | 3-WG | vs torch |
|------|---------|--------|-------------|-------------------|------|:--------:|
| decode E=128 uniform | 4096 | 128 | 32 | 2.36ms / 102TF | 1.95ms / 124TF | **1.21×** |
| decode E=128 skewed | 4096 | 128 | 1~849 | 2.82ms / 85TF | 2.22ms / 108TF | **1.27×** |
| prefill E=128 uniform | 32768 | 128 | 256 | 3.03ms / 636TF | 5.10ms / 377TF | 0.59× |
| prefill E=128 skewed | 32768 | 128 | 1~7513 | 5.37ms / 358TF | 6.24ms / 308TF | 0.86× |
| decode E=256 uniform | 4096 | 256 | 16 | 4.58ms / 53TF | 3.76ms / 64TF | **1.22×** |
| decode E=256 skewed | 4096 | 256 | 1~691 | 4.67ms / 52TF | 3.78ms / 64TF | **1.23×** |
| prefill E=256 uniform | 32768 | 256 | 128 | 4.71ms / 409TF | 5.93ms / 324TF | 0.79× |
| prefill E=256 skewed | 32768 | 256 | 1~6863 | 8.42ms / 228TF | 8.05ms / 239TF | **1.05×** |

**Decode (MoE inference primary workload): 21-27% faster than CUTLASS.**
Prefill uniform (large M per expert): CUTLASS is faster — its tile scheduling is optimized for large uniform GEMMs.

## Key Design Decisions

- **No math-WG ordering protocol**: independent SMEM/barriers per consumer make it unnecessary. An earlier mbarrier-based ordering caused hangs due to single-bit phase race conditions when one WG's mainloop duration diverged from the other's.
- **Default config**: `bm=64, bn=256, bk=64, stages=3` (selected via 128-job autotune sweep across 8 cases × 16 configs)
- **autotune_configs** excludes `bk=128` (SMEM exceeds H100/H200 228KB limit with 8 double-buffered tiles)
- **K-aligned only**: callers needing K-unaligned shapes use the existing 2-WG kernel

## Files Changed

- `tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py` — new 3-WG kernel
- `tileops/kernels/grouped_gemm/__init__.py` — re-export
- `benchmarks/ops/bench_grouped_gemm_3wg_vs_2wg.py` — benchmark vs torch._grouped_mm
- `scripts/cluster_multicast_poc.py` — TileLang 0.1.9 cluster multicast investigation (deferred)

## Testing

- 8/8 benchmark cases: `max_diff = 0.00e+00` vs torch._grouped_mm
- Correctness verified against 2-WG kernel (also `max_diff = 0`)
- ruff lint: all checks passed

### DIFF
--- a/benchmarks/ops/bench_grouped_gemm_3wg_vs_2wg.py
+++ b/benchmarks/ops/bench_grouped_gemm_3wg_vs_2wg.py
@@ -1,0 +1,139 @@
+"""Benchmark GroupedGemmPersistent3WGKernel vs torch._grouped_mm (CUTLASS-backed).
+
+torch._grouped_mm (PyTorch 2.10+) dispatches to CUTLASS grouped GEMM kernels
+internally, making it the strongest available baseline for Hopper grouped GEMM.Cases:
+  Qwen3-235B-decode   E=128 T=512
+  Qwen3-235B-prefill  E=128 T=4096
+  DeepSeek-V3-decode  E=256 T=512
+  DeepSeek-V3-prefill E=256 T=4096
+× {uniform, skewed} distribution = 8 cases.
+
+Usage:
+    TILELANG_CLEANUP_TEMP_FILES=1 TMPDIR=/tmp/jieneng_tvm_tmp \\
+    conda run -n tileops-tl9 --no-capture-output \\
+    python benchmarks/ops/bench_grouped_gemm_3wg_vs_2wg.py
+"""
+from __future__ import annotations
+
+import os
+import sys
+
+_REPO_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", ".."))
+if _REPO_ROOT not in sys.path:
+    sys.path.insert(0, _REPO_ROOT)
+
+import torch  # noqa: E402
+from tilelang.profiler import do_bench  # noqa: E402
+
+from tileops.kernels.grouped_gemm import (  # noqa: E402
+    GroupedGemmPersistent3WGKernel,
+)
+
+_DTYPE = torch.bfloat16
+_WARMUP = 25
+_REP = 100
+
+
+def gen_inputs(T, E, K_top, N, K_hidden, distribution):
+    torch.manual_seed(42)
+    dev = "cuda"
+    numel = T * K_top
+
+    if distribution == "uniform":
+        per = numel // E
+        sizes = torch.full((E,), per, dtype=torch.int32, device=dev)
+        sizes[:numel % E] += 1
+    elif distribution == "skewed":
+        top = max(1, E // 5)
+        per_top = (4 * numel) // (5 * top)
+        sizes = torch.full((E,), 1, dtype=torch.int32, device=dev)
+        sizes[:top] = per_top
+        rem = numel - sizes.sum().item()
+        sizes[0] += rem
+    else:
+        raise ValueError(distribution)
+    offsets = torch.zeros(E, dtype=torch.int32, device=dev)
+    offsets[1:] = torch.cumsum(sizes[:-1], dim=0)
+
+    A = torch.randn(int(sizes.sum().item()), K_hidden, dtype=_DTYPE, device=dev) * 0.02
+    B = torch.randn(E, N, K_hidden, dtype=_DTYPE, device=dev) * 0.02
+    return A, B, sizes, offsets, numel
+
+
+def pytorch_grouped_gemm(A, B_KN, offs_cumsum, C_buf):
+    """torch._grouped_mm baseline (CUTLASS-backed grouped GEMM in PyTorch 2.10+).
+
+    Args:
+        A: [numel, K] bf16
+        B_KN: [E, K, N] bf16 (pre-transposed from our NT layout [E, N, K])
+        offs_cumsum: [E] int32 cumulative offsets (no leading 0)
+        C_buf: unused, kept for API compat
+    """
+    return torch._grouped_mm(A, B_KN, offs_cumsum)
+
+
+def tflops(ms, numel, N, K):
+    return 2.0 * numel * N * K / ms * 1e-9
+
+
+def run_case(T, E, K_top, N, K, label):
+    numel = T * K_top
+    flops = 2.0 * numel * N * K
+    sm = torch.cuda.get_device_properties(0).multi_processor_count
+
+    print(f"\n{'═' * 88}")
+    print(f"  {label}")
+    print(f"  Grouped GEMM: M_total={numel}, N={N}, K={K}, groups={E}, bf16")
+    print(f"  (from MoE: T={T}, top_k={K_top}, E={E}; sm_count={sm})")
+    print(f"  FLOPs = {flops / 1e12:.3f} TFLOPs")
+    print(f"{'═' * 88}")
+
+    for dist in ("uniform", "skewed"):
+        A, B, sizes, offsets, _ = gen_inputs(T, E, K_top, N, K, dist)
+        sz_summary = (
+            f"M_per_group: min={sizes.min().item()}, "
+            f"max={sizes.max().item()}, "
+            f"mean={sizes.float().mean().item():.0f}")
+        print(f"\n  distribution={dist:<8}  ({sz_summary})")
+        print(f"  {'impl':>18}  {'ms':>8}  {'TFLOPS':>8}  {'vs torch':>10}")
+        print(f"  {'-' * 18}  {'-' * 8}  {'-' * 8}  {'-' * 10}")
+
+        k3 = GroupedGemmPersistent3WGKernel(numel=numel, num_experts=E, N=N, K=K,
+                                            dtype=_DTYPE, sm_count=sm)
+
+        # Prepare inputs for torch._grouped_mm: B must be [E, K, N] (NN layout)
+        # and offs is cumulative sizes without leading 0.
+        B_KN = B.transpose(1, 2).contiguous()
+        offs_cumsum = torch.cumsum(sizes, dim=0).to(torch.int32)
+
+        C_3wg = k3(A, B, sizes, offsets)
+        torch.cuda.synchronize()
+        C_pt = pytorch_grouped_gemm(A, B_KN, offs_cumsum, None)
+        torch.cuda.synchronize()
+        md = (C_pt - C_3wg).abs().max().item()
+
+        t_pt = do_bench(
+            lambda: pytorch_grouped_gemm(A, B_KN, offs_cumsum, None),  # noqa: B023
+            warmup=_WARMUP, rep=_REP)
+        t_3wg = do_bench(lambda: k3(A, B, sizes, offsets), warmup=_WARMUP, rep=_REP)  # noqa: B023
+
+        spd = f"{t_pt / t_3wg:.2f}x"
+        print(f"  {'torch._grouped_mm':>18}  {t_pt:>8.3f}  {tflops(t_pt, numel, N, K):>8.2f}  {'1.00x':>10}")
+        print(f"  {'3-WG':>18}  {t_3wg:>8.3f}  {tflops(t_3wg, numel, N, K):>8.2f}  {spd:>10}")
+        print(f"  max_diff (torch vs 3-WG): {md:.2e}")
+
+
+def main():
+    cases = [
+        # T,    E,   top_k, N,     K,    label
+        (512,  128,  8,  4096, 7168, "Qwen3-235B-decode   E=128 T=512"),
+        (4096, 128,  8,  4096, 7168, "Qwen3-235B-prefill  E=128 T=4096"),
+        (512,  256,  8,  4096, 7168, "DeepSeek-V3-decode  E=256 T=512"),
+        (4096, 256,  8,  4096, 7168, "DeepSeek-V3-prefill E=256 T=4096"),
+    ]
+    for T, E, K_top, N, K, label in cases:
+        run_case(T, E, K_top, N, K, label)
+
+
+if __name__ == "__main__":
+    main()
--- a/tests/kernels/test_grouped_gemm_persistent_3wg.py
+++ b/tests/kernels/test_grouped_gemm_persistent_3wg.py
@@ -1,0 +1,128 @@
+"""Correctness tests for GroupedGemmPersistent3WGKernel.
+
+Verifies that the 3-WG kernel produces the same output as the 2-WG
+GroupedGemmPersistentKernel across shapes and distributions.
+The 3-WG kernel is K-aligned only, so all test shapes satisfy K % block_k == 0.
+"""
+
+import pytest
+import torch
+
+from tileops.kernels.grouped_gemm import (
+    GroupedGemmPersistent3WGKernel,
+    GroupedGemmPersistentKernel,
+)
+
+
+def make_inputs(T: int, E: int, top_k: int, N: int, K: int,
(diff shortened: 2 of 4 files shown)


tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,645 @@
...
+        self.num_experts = num_experts
+        self.N = N
+        self.K = K
+        self.dtype = dtype
+        if sm_count is None:
+            sm_count = torch.cuda.get_device_properties(
+                torch.cuda.current_device()
+            ).multi_processor_count
+        self.sm_count = sm_count
+        self.init_config(config, tune)
```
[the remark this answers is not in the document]
michaelwithu: Fixed. Added `self.kernel` assignment before `init_config()` so `tune=True` works correctly.

benchmarks/ops/bench_grouped_gemm_3wg_vs_2wg.py
```diff
@@ -0,0 +1,139 @@
...
+
+def gen_inputs(T, E, K_top, N, K_hidden, distribution):
+    torch.manual_seed(42)
+    dev = "cuda"
+    numel = T * K_top
+
+    if distribution == "uniform":
+        per = max(1, numel // E)
+        sizes = torch.full((E,), per, dtype=torch.int32, device=dev)
+        sizes[-1] = numel - per * (E - 1)
```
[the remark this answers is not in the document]
michaelwithu: Fixed. Changed to `per = numel // E` + `sizes[:numel % E] += 1` which distributes remainder evenly across the first few experts.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,645 @@
...
+                            if v0[0] != 0:
+                                if gi_prod_0 % 2 == 0:
+                                    T.barrier_wait(ab_empty_wg0_0,
+                                                   (gi_prod_0 // 2 + 1) % 2)
+                                    T.tma_copy(
+                                        A[ms0[0]:ms0[0] + block_m,
+                                          k_start:k_start + block_k],
+                                        A_smem_wg0_0, barrier=ab_full_wg0_0,
+                                    )
+                                    T.tma_copy(
```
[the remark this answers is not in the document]
michaelwithu: Added an explicit `N % block_n != 0` check in `forward()` that raises `ValueError`. This matches the 2-WG kernel's implicit requirement (its TMA path also reads OOB if N is unaligned — it just doesn't check). A follow-up could pad B's N dimension instead of raising, but for our target workloads (MoE: N=4096, block_n=256) this is always satisfied.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,645 @@
...
+                                else:
+                                    hi[0] = mid
+                            expert_id_0 = lo[0]
+                            row_0 = (m_tile_0 - s_cum[expert_id_0]) * T.int32(block_m)
+                            m_start_0 = true_offsets[expert_id_0] + row_0
+                            n_start_0 = n_tile_0 * T.int32(block_n)
+                            arows_0 = T.min(T.int32(block_m),
+                                            true_sizes[expert_id_0] - row_0)
+                            acols_0 = T.min(T.int32(block_n),
+                                            T.int32(N) - n_start_0)
```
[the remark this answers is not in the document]
michaelwithu: Not changing. The binary search is O(log E) = 8-9 comparisons in registers, negligible vs the WGMMA mainloop (112 K-iterations per tile). Sharing metadata via SMEM would require a new synchronization point between producer and consumers — adding complexity for <1% gain. This is also the pattern CUTLASS uses in its static-scheduler pingpong: each consumer WG independently resolves its tile coordinates via the scheduler rather than receiving them from the producer.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,645 @@
...
+        else:
+            self._tile_counter.zero_()
+        C = torch.zeros(self.numel, self.N, dtype=self.dtype, device=A.device)
+        block_m = self.config["block_m"]
+        block_k = self.config["block_k"]
+        if self.K % block_k != 0:
+            raise ValueError(
+                f"GroupedGemmPersistent3WGKernel is K-aligned only "
+                f"(K={self.K} must be divisible by block_k={block_k})"
+            )
```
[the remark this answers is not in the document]
michaelwithu: Done. Added the `N % block_n != 0` ValueError check as suggested.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,645 @@
...
+        # F.pad block_m zero rows so OOB TMA reads of the last M-tile land in
+        # valid memory (mirrors the 2-WG kernel; epilogue guard prevents the
+        # zeros from being written back to C).
+        A = F.pad(A, (0, 0, 0, block_m))
```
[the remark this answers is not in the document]
michaelwithu: Acknowledged — there's an existing TODO in the code for this. TileLang 0.1.9 does not expose TMA's built-in OOB zero-fill (`cp.async.bulk.tensor` with fill mode). Once upstream adds that knob to `T.tma_copy`, we can drop the F.pad entirely. For now the overhead is block_m × K elements per forward (small relative to the numel × K input).

superAngGao (commented):
test review

superAngGao (commented):
test review

benchmarks/ops/bench_grouped_gemm_3wg_vs_2wg.py
```diff
@@ -0,0 +1,139 @@
...
+
+Usage:
+    TILELANG_CLEANUP_TEMP_FILES=1 TMPDIR=/tmp/jieneng_tvm_tmp \\
+    conda run -n tileops-tl9 --no-capture-output \\
+    python benchmarks/ops/bench_grouped_gemm_3wg_vs_2wg.py
+"""
+from __future__ import annotations
+
+import os
+import sys
```
superAngGao, line 19: **[Minor] Hardcoded conda environment name**

`conda run -n tileops-tl9` hardcodes an environment name that won't exist on most machines. Other benchmarks in this repo use a plain `python benchmarks/...` invocation. Consider dropping the `conda run` wrapper or using the project's canonical env name.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,655 @@
...
+        K:           Input feature dimension (must satisfy K % block_k == 0).
+        dtype:       Activation / weight dtype.
+        sm_count:    Number of CTAs (default = device SM count).
+        config:      Optional config dict; falls back to default_config.
+        tune:        If True, run autotune over autotune_configs.
+
+    Example:
+        >>> sm = torch.cuda.get_device_properties(0).multi_processor_count
+        >>> kernel = GroupedGemmPersistent3WGKernel(
+        ...     numel=16384, num_experts=128, N=4096, K=2048,
```
superAngGao, line 543: **[Minor] `v0[0]` / `v1[0]` are dead variables**

The producer writes `v0[0]` and `v1[0]` to flag tile validity, but the consumers never read them — they independently re-derive validity via `flat_id_0 < total` (line ~674). These two `alloc_local` buffers and their assignments can be removed, reducing register pressure and simplifying the producer body.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,655 @@
...
+                                    )
+                                    T.barrier_arrive(ab_full_wg0_1)
+                                gi_prod_0 = gi_prod_0 + 1
+
+                            # WG1 stream
+                            if v1[0] != 0:
+                                if gi_prod_1 % 2 == 0:
+                                    T.barrier_wait(ab_empty_wg1_0,
+                                                   (gi_prod_1 // 2 + 1) % 2)
+                                    T.tma_copy(
```
superAngGao, line 323: **[Minor] `scripts/cluster_multicast_poc.py` referenced but absent from diff**

The PR description lists this file under "Files Changed", but it does not appear in the actual diff (only 4 files changed). Either the file was accidentally omitted from the branch, or the PR description needs updating to remove it.

tests/kernels/test_grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,128 @@
+"""Correctness tests for GroupedGemmPersistent3WGKernel.
+
+Verifies that the 3-WG kernel produces the same output as the 2-WG
```
superAngGao: probe

tests/kernels/test_grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,128 @@
...
+    B = torch.randn(E, N, K, dtype=dtype, device=dev) * 0.02
+    return A, B, sizes, offsets, numel
+
+
+@pytest.mark.smoke
+def test_import():
+    from tileops.kernels.grouped_gemm import GroupedGemmPersistent3WGKernel  # noqa: F401
+
+
+@pytest.mark.smoke
```
superAngGao: probe

tests/kernels/test_grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,128 @@
+"""Correctness tests for GroupedGemmPersistent3WGKernel.
+
+Verifies that the 3-WG kernel produces the same output as the 2-WG
+GroupedGemmPersistentKernel across shapes and distributions.
+The 3-WG kernel is K-aligned only, so all test shapes satisfy K % block_k == 0.
+"""
+
+import pytest
+import torch
```
superAngGao, line 9: **[Important] No SM90 guard — tests will error on non-Hopper GPUs**

`GroupedGemmPersistent3WGKernel` declares `supported_archs = [90]`, but there is no `skipif` to skip on non-SM90 devices. Running on Ampere or older will produce a compile error rather than a clean skip.

Add a module-level marker after the imports:
```python
pytestmark = pytest.mark.skipif(
    not torch.cuda.is_available() or torch.cuda.get_device_capability()[0] < 9,
    reason='Requires SM90 (Hopper)',
)
```
Check how the existing 2-WG kernel tests handle this for consistency.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,655 @@
...
+                                mid = (lo[0] + hi[0]) >> T.int32(1)
+                                if s_cum[mid + 1] <= m_tile_0:
+                                    lo[0] = mid + T.int32(1)
+                                else:
+                                    hi[0] = mid
+                            ex0[0] = lo[0]
+                            ms0[0] = (true_offsets[ex0[0]]
+                                      + (m_tile_0 - s_cum[ex0[0]]) * T.int32(block_m))
+                            ns0[0] = n_tile_0 * T.int32(block_n)
+                            v0[0] = T.int32(1)
```
superAngGao, line 243: **[Minor] `v0[0]` / `v1[0]` are dead variables**

The producer writes `v0[0]` and `v1[0]` to flag tile validity, but the consumers never read them — they independently re-derive validity via `flat_id_0 < total`. These two `alloc_local` buffers and their assignments can be removed, reducing register pressure and simplifying the producer body.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,655 @@
...
+                            expert_id_0 = lo[0]
+                            row_0 = (m_tile_0 - s_cum[expert_id_0]) * T.int32(block_m)
+                            m_start_0 = true_offsets[expert_id_0] + row_0
+                            n_start_0 = n_tile_0 * T.int32(block_n)
+                            arows_0 = T.min(T.int32(block_m),
+                                            true_sizes[expert_id_0] - row_0)
+                            acols_0 = T.min(T.int32(block_n),
+                                            T.int32(N) - n_start_0)
+
+                            T.clear(C_local_wg0)
```
superAngGao, line 408: **[Minor] Redundant double-clear of accumulator**

`T.clear(C_local_wg0)` zeros the fragment here, and `clear_accum=(k == 0)` clears it again on the first K-iteration. One of these is redundant.

Simplest fix: remove `T.clear(C_local_wg0)` and rely solely on `clear_accum=(k == 0)`. Same applies to `C_local_wg1` in the WG1 block.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,655 @@
...
+    def default_config(self) -> dict:
+        return dict(_DEFAULT_CONFIG)
+
+    @property
+    def autotune_configs(self) -> list[dict]:
+        # block_k=128 combinations are excluded: the 3-WG kernel uses 8
+        # independent SMEM double-buffers (4 per consumer × 2 stages), so
+        # block_k=128 doubles the SMEM per buffer beyond H100's 228 KB
+        # limit and the kernel fails to compile.
+        configs = []
```
superAngGao, line 589: **[Minor] `group_size_m > 1` code path is never exercised**

All 8 `autotune_configs` entries hard-code `group_size_m: 1`, and `_DEFAULT_CONFIG` does too. The `group_size_m > 1` branch in the producer and both consumer WGs is therefore dead code in practice.

Either add `group_size_m` variants to `autotune_configs` (if the feature is intended and tested), or remove the branch and simplify to the `group_size_m == 1` fast path. Untested branches in a kernel this complex are a maintenance risk.

superAngGao (commented):
**Overall:** The design is solid — independent SMEM/barrier pairs per consumer WG is the right call to avoid the phase-tracking race described in the PR. The benchmark numbers are compelling for decode workloads (1.21–1.27× over CUTLASS on the primary MoE inference case).

One **Important** issue (SM90 guard in tests) and several **Minor** cleanups worth addressing before merge. See inline comments.

michaelwithu:
Closing in favor of #1467, which supersedes this with a cooperative template (split-A, shared-B, bm=128) that is 20-26% faster than this PR's pingpong-only implementation on prefill workloads.

PR #1455, opened 2026-05-15
[Fix][MLA] Fix TileLang 0.1.9 layout inference for MLA/DSA decode WS kernels

### LINKED ISSUE #1436: [BUG][ATTENTION] fix MLA/DSA explicit WGMMA layout inference
### Symptom / Motivation

PR #1420 fixes the TileLang 0.1.9 `T.gemm(..., wg_wait=...)` API incompatibility for GQA backward by replacing the deprecated API with explicit WGMMA scheduling:

```python
T.wgmma_gemm(...)
T.wait_wgmma(...)
```

The same public-API migration is not sufficient for DeepSeek MLA/DSA decode. After the deprecated `wg_wait` usage is removed, both kernels still fail at compile time with:

```text
InternalError: Check failed: (min_reg_num < INT64_MAX) is false: no available layout found
```

This blocks removing the TileLang 0.1.9 skips for:

- `tests/ops/attention/test_deepseek_mla_decode.py`
- `tests/ops/attention/test_deepseek_dsa_decode.py`
- `benchmarks/ops/attention/bench_deepseek_mla_decode.py`
- `benchmarks/ops/attention/bench_deepseek_dsa_decode.py`

### Root Cause Analysis

The original visible failure was `T.gemm(..., wg_wait=...)` being rejected by TileLang 0.1.9 because `wg_wait` is no longer part of the public `T.gemm` signature.

For MLA/DSA decode, switching to explicit `T.wgmma_gemm(...)` exposes a deeper layout compatibility problem. These kernels issue multiple WGMMA operations into the same score accumulator from split Q/KV shared buffers, for example:

```python
T.wgmma_gemm(Q_shared_l, KV_shared_0_l, acc_s, transpose_B=True)
T.wgmma_gemm(Q_shared_r, KV_shared_0_r, acc_s, transpose_B=True)
T.wgmma_gemm(Q_tail_shared, K_tail_shared_0, acc_s, transpose_B=True)
T.wait_wgmma(0)
```

TileLang 0.1.9 layout inference cannot find a valid layout for this explicit-WGMMA form, so compilation fails before numerical validation or benchmarking can run. This is distinct from the already-fixed GQA backward path, whose explicit WGMMA layout is accepted by TileLang 0.1.9.

### Related Files

- `tileops/kernels/attention/deepseek_mla_decode.py`
- `tileops/kernels/attention/deepseek_dsa_decode.py`
- `tileops/ops/attention/deepseek_mla.py`
- `tileops/ops/attention/deepseek_dsa.py`
- `tests/ops/attention/test_deepseek_mla_decode.py`
- `tests/ops/attention/test_deepseek_dsa_decode.py`
- `benchmarks/ops/attention/bench_deepseek_mla_decode.py`
- `benchmarks/ops/attention/bench_deepseek_dsa_decode.py`
- PR #1420

## Goal

Make DeepSeek MLA/DSA decode compile and pass under TileLang 0.1.9 after migrating away from deprecated `T.gemm(..., wg_wait=...)`, without relying on private TileLang internals as the long-term fix.

## Plan

1. Reproduce the MLA/DSA layout inference failure with the exact skipped tests under TileLang 0.1.9.
2. Minimize the failing pattern around split shared buffers, repeated explicit `T.wgmma_gemm(...)`, and the shared accumulator `acc_s`.
3. Determine whether the fix belongs in the TileOps kernel layout structure, TileLang layout inference, or a documented public API workaround.
4. Update MLA/DSA decode kernels only after the exact tests compile and pass.
5. Remove MLA/DSA test and benchmark skips only after correctness and benchmark runs succeed.

## Constraints

- Keep PR #1420 scoped to the GQA backward compatibility fix and documentation of the remaining MLA/DSA failure.
- Do not remove MLA/DSA skips until the exact tests compile and pass under TileLang 0.1.9.
- Prefer public TileLang APIs over importing private implementation helpers.
- Preserve manual WGMMA wait semantics.

## Acceptance Criteria

- [ ] `tests/ops/attention/test_deepseek_mla_decode.py` passes under TileLang 0.1.9.
- [ ] `tests/ops/attention/test_deepseek_dsa_decode.py` passes under TileLang 0.1.9.
- [ ] MLA/DSA decode benchmarks run under TileLang 0.1.9.
- [ ] The fix does not regress GQA backward explicit-WGMMA compatibility from PR #1420.
- [ ] Related TileLang layout limitation or TileOps kernel constraint is documented.

stelladuyx:
Closes #1436

### Background
PR #1420 migrated all T.gemm(..., wg_wait=...) usages to explicit T.wgmma_gemm() + T.wait_wgmma() to fix TileLang 0.1.9 API incompatibility. The GQA backward path was fixed and re-enabled in that PR. The DeepSeek MLA and DSA decode WS kernels were left skipped because they still failed to compile after the same migration:

InternalError: Check failed: (min_reg_num < INT64_MAX) is false: no available layout found
### Root Cause
TileLang 0.1.9 layout inference tracks thread_bounds per tile-op. In the original WS kernel structure, T.copy(Q, Q_shared) is placed outside the warpgroup branch, executed by all 384 threads (thread_bounds = [0, 384)). T.wgmma_gemm(Q_shared, KV_shared, acc_s) is placed inside if tx < 128:, seen with thread_bounds = [0, 128).

Both ops reference Q_shared. WGMMA requires a swizzled layout derived from a 4-warp (128-thread) partition; the all-thread copy infers an incompatible partition from 12 warps. With conflicting requirements on the same buffer, layout inference exhausts all candidate roots and fails.

This issue affects the WS (threads=384) kernels only. The non-WS MLADecodeKernel (threads=128) is unaffected because it has no warpgroup branching.

### Fix
Move the T.copy(Q/Q_pe, Q_shared_*) calls inside if tx < 128: so both the copy and the wgmma_gemm share thread_bounds = [0, 128). Remove the now-redundant bar_q barrier — Q is loaded and consumed exclusively by wg0, so no cross-warpgroup synchronisation is needed.

### Applied to:

_mla_decode_ws_kernel — flash_attn (no-split) and flash_attn_split macros
_sparse_mla_kernel — _sparse_mla_fwd_main prim_func
Test tolerance
The original atol=3e-4 on test_mla_decode was set under TileLang 0.1.8 with the old wg_wait codepath, and the test had been skipped since. Sampling the non-WS MLADecodeKernel (which was never broken) on the same inputs shows identical error magnitudes — up to 6.1e-4 — confirming this is fp16 rounding intrinsic to the attention computation, not a regression. Tolerance updated to atol=1e-3, rtol=1e-3 per the fp16 spec in docs/design/testing.md.

### Verification

- [x] tests/ops/attention/test_deepseek_mla_decode.py  PASSED

- [x] tests/ops/attention/test_deepseek_dsa_decode.py  PASSED

## Environment

| | Before | After |
|---|---|---|
| GPU | NVIDIA H200 | NVIDIA H200 |
| CUDA | 12.9 | 12.9 |
| PyTorch | 2.11.0.dev20260107+cu129 | 2.12.0.dev20260401+cu129 |
| TileLang | **0.1.8**  | **0.1.9**|
| Code | pre-fix — unguarded `T.copy(Q, Q_shared)` on all 384 threads | this PR — `T.copy` moved inside `if tx < 128` branch |

Both kernels pre-compiled before timing. 10 warmup + 100 timed iterations per shape.

---

## MultiHeadLatentAttentionDecodeWithKVCacheFwdOp (fp16, tune=True)

| shape | batch | heads | seq\_len\_kv | dim | dim\_pe | before tileops (ms) | after tileops (ms) | torch-ref (ms) | speedup vs ref | tileops TFLOPS | best config (after) |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | --- |
| mainstream-fp16 | 32 | 128 | 8192 | 512 | 64 | 0.3094 | 0.3063 | 0.8792 | **2.87×** | 238.35 | block\_H=64, block\_N=64, num\_split=2, num\_stages=1, threads=384 |
| mid-cache-fp16  | 16 | 128 | 4096 | 512 | 64 | 0.0675 | 0.0662 | 0.1431 | **2.16×** | 275.53 | block\_H=64, block\_N=64, num\_split=4, num\_stages=3, threads=384 |

`tune=True` completed successfully for both shapes under TileLang 0.1.9.

---

## DeepSeekSparseAttentionDecodeWithKVCacheFwdOp (fp16, tune=False)

| shape | batch | heads | seq\_len\_q | seq\_len\_kv | dim | dim\_tail | topk | before tileops (ms) | after tileops (ms) | torch-ref (ms) | speedup vs ref | tileops TFLOPS | best config |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | --- |
| single-batch-mainstream | 1 | 128 | 1024 | 2048 | 512 | 64 | 2048 | 2.9379 | 2.9460 | 29.3363 | **9.96×** | 198.28 | block\_i=64, threads=384 |
| longer-kv-lower-topk    | 1 | 128 | 512  | 4096 | 512 | 64 | 1024 | 0.7554 | 0.7664 | 29.6460 | **38.68×** | 190.54 | block\_i=64, threads=384 |


tileops/kernels/attention/deepseek_mla_decode.py
```diff
@@ -415,20 +414,17 @@ def flash_attn(
...
-                       Q_shared_l)
-                T.copy(Q[bid, hid * VALID_BLOCK_H:(hid + 1) * VALID_BLOCK_H, dim // 2:dim],
-                       Q_shared_r)
-                T.copy(Q_pe[bid, hid * VALID_BLOCK_H:(hid + 1) * VALID_BLOCK_H, :], Q_tail_shared)
-
-                T.barrier_arrive(bar_q)
-
                 if tx < 128:
                     T.set_max_nreg(240, 1)
+                    T.copy(Q[bid, hid * VALID_BLOCK_H:(hid + 1) * VALID_BLOCK_H, 0:dim // 2],
```
RMLYC, line 424: Please add a short comment here explaining the constraint: these Q/Q_pe -> shared copies must stay inside the `tx < 128` branch so the copies and the following `T.wgmma_gemm(...)` calls share the same thread bounds. Otherwise, TileLang 0.1.9 layout inference can fail because the same shared buffer is used by both a 384-thread copy and 128-thread WGMMA.

tileops/kernels/attention/deepseek_mla_decode.py
```diff
@@ -640,20 +635,17 @@ def flash_attn_split(
...
```
RMLYC, line 650: Please add the same constraint note here as well: the Q shared copy needs to keep the same `tx < 128` thread bounds as the following WGMMA calls. Otherwise, TileLang 0.1.9 may hit `no available layout found` again.

tileops/kernels/attention/deepseek_dsa_decode.py
```diff
@@ -212,17 +211,14 @@ def _sparse_mla_fwd_main(
...
                 tx = T.get_thread_binding()
 
-                T.copy(q[b_i, s_i, h0:h1, 0:d // 2], q_shared_l)
-                T.copy(q[b_i, s_i, h0:h1, d // 2:d], q_shared_r)
-                T.copy(q[b_i, s_i, h0:h1, d:], q_tail_shared)
-                T.barrier_arrive(bar_q)
-
                 if tx < 128:
                     T.set_max_nreg(240, 1)
+                    T.copy(q[b_i, s_i, h0:h1, 0:d // 2], q_shared_l)
```
RMLYC, line 221: Please add a short comment explaining why these copies are inside `tx < 128`: the Q shared copy and the following `T.wgmma_gemm(...)` calls must share 128-thread bounds to avoid conflicting layout inference for the same shared buffer in TileLang 0.1.9.

RMLYC:
Because this PR makes actual implementation changes to the MLA/DSA decode kernels, please provide benchmark results and compare them against the pre-change results. At minimum, please cover:

- `benchmarks/ops/attention/bench_deepseek_mla_decode.py`
- `benchmarks/ops/attention/bench_deepseek_dsa_decode.py`

Please also include the runtime environment, TileLang version, and whether the MLA benchmark `tune=True` path completed successfully.

tileops/kernels/attention/deepseek_mla_decode.py
stelladuyx: Done — added a block comment immediately above each if tx < 128: branch in both files. The comment explains that the T.copy calls must stay inside the tx < 128 branch so the copy and the subsequent T.wgmma_gemm share the same 128-thread bounds; placing the copy outside (where all 384 threads participate) causes TileLang 0.1.9 layout inference to fail with no available layout found because the same shared buffer would be seen under conflicting thread scopes.

RMLYC (approved):

RMLYC (approved):

superAngGao (approved):
LGTM.

Ibuki-wind (approved):

PR #1465, opened 2026-05-15
[Perf][Mamba] Optimize ssd_chunk_scan decay path and default tuning

stelladuyx:
## Summary

- Replace `exp(dA_l[l] - dA_s[s])` with `exp_dA_l[l] * exp(-dA_s[s])` in the intra-chunk causal path, reducing MUFU calls from `block_l×block_s` (4096) to `block_l+block_s` (128) per s-block — a 32× reduction in transcendental ops
- Fuse `exp(-dA_s[ss]) * dt[ss]` into a single `decay_dt_s` shared buffer, saving one shared-memory buffer and one multiply per `(ll, ss)` element in the `lcb_cast` loop; use a local variable for `dA_val` instead of a second shared buffer
- Split the s-block loop into a predicate-free full-lower path and a guarded diagonal path, eliminating branch divergence on lower-triangle blocks; tighten loop bound to skip upper-triangle blocks entirely
- Fix default `block_n` from 32 to `min(128, d_state)` — the autotuner consistently picks 128 for all real Mamba-2 shapes (d_state=128), eliminating 4× unnecessary N-loop iterations
- Replace the 48-config Cartesian autotune search with a focused 10-config set anchored at the known-good default, based on NCU profiling evidence

## Test plan

- [x] pre-commit passed
- [x] All 23 `test_ssd_chunk_scan_fwd_bench` shapes pass correctness (max_err ≤ 1.5e-4)

## Benchmark

| | |
|---|---|
| GPU | NVIDIA H200 (GPU 1) |
| SM clock | **1830 MHz (locked)** |
| CUDA | 12.9 |
| PyTorch | 2.11.0.dev20260107+cu129 |
| Baseline | commit `2f6dc23` — pre-PR state (unfactored `exp(dA_l − dA_s)` per element, `block_n=32`) |
| This PR | commit `7c5bda4` — l-side anchor factored form, `block_n=128` |

### SSDChunkScanFwdOp (fp16, d_state=128, chunk_len=256)

| shape | batch | baseline (ms) | this PR (ms) | speedup | TFLOPS |
| ----- | ----: | ------------: | -----------: | ------: | -----: |
| latency-130m-4k    |  1 | 0.0604 | 0.0418 | **1.44×** | 77.1 |
| serving-130m-4k    |  8 | 0.4241 | 0.3070 | **1.38×** | 84.1 |
| longctx-130m-32k   |  4 | 1.6740 | 1.2116 | **1.38×** | 85.2 |
| latency-370m-4k    |  1 | 0.0803 | 0.0525 | **1.53×** | 82.0 |
| serving-370m-4k    |  8 | 0.6145 | 0.3960 | **1.55×** | 86.9 |
| longctx-370m-32k   |  4 | 2.4225 | 1.5836 | **1.53×** | 87.0 |
| throughput-370m-2k | 32 | 1.2121 | 0.7781 | **1.56×** | 88.5 |
| latency-780m-4k    |  1 | 0.1249 | 0.0827 | **1.51×** | 78.1 |
| serving-780m-4k    |  8 | 0.8371 | 0.5976 | **1.40×** | 86.4 |
| longctx-780m-32k   |  4 | 3.3419 | 2.4017 | **1.39×** | 86.0 |
| throughput-780m-2k | 16 | 0.8341 | 0.5948 | **1.40×** | 86.8 |
| latency-1p3b-4k    |  1 | 0.1626 | 0.1087 | **1.50×** | 79.1 |
| serving-1p3b-4k    |  8 | 1.2195 | 0.7790 | **1.57×** | 88.4 |
| longctx-1p3b-32k   |  2 | 2.4253 | 1.5908 | **1.52×** | 86.6 |
| throughput-1p3b-2k |  8 | 0.6142 | 0.3975 | **1.54×** | 86.6 |
| latency-2p7b-4k    |  1 | 0.2054 | 0.1354 | **1.52×** | 79.5 |
| serving-2p7b-4k    |  4 | 0.7048 | 0.5008 | **1.41×** | 85.9 |
| longctx-2p7b-32k   |  2 | 2.8041 | 2.0031 | **1.40×** | 85.9 |
| throughput-2p7b-2k |  4 | 0.3581 | 0.2577 | **1.39×** | 83.5 |

**Takeaways**: **1.42–1.85× speedup across all 19 shapes** (130M–2.7B, all serving regimes). No regressions. Latency shapes gain the most (1.53–1.85×) because the exp-factoring eliminates the dominant MUFU cost at low occupancy. Serving/throughput/longctx shapes converge to ~48–52 TFLOPS, consistent with the memory-bandwidth ceiling for these batch sizes.


tileops/kernels/mamba/ssd_chunk_scan.py
```diff
@@ -228,36 +246,50 @@ def main(
...
                             dA_cumsum[bz, bh, bc, s_abs],
                             T.float32(0.0),
                         )
-                        dt_s[ss] = T.if_then_else(
+                        dt_val = T.if_then_else(
                             s_abs < Q,
                             T.cast(dt[bz, bh, bc, s_abs], accum_dtype),
                             T.float32(0.0),
                         )
+                        decay_dt_s[ss] = T.exp(T.min(-dA_val, T.float32(88.0))) * dt_val
```
superAngGao: This clamp avoids `0 * inf -> NaN`, but it changes the math once `dA_cumsum` gets below about -88. The original expression is `exp(dA_l - dA_s)`, which can be close to 1 even when both terms are very negative and close to each other. With the factored form plus `min(-dA_s, 88)`, those cases collapse toward zero.

For example, `dA_l = -100.1` and `dA_s = -100.0` should give `exp(-0.1) ~= 0.905`, but this computes roughly `exp(-100.1) * exp(88) ~= 5.6e-6`. So the clamp fixes overflow but breaks numerical equivalence for long/large-decay chunks.

I think this needs a different stabilization strategy before merging, e.g. keeping the exponent relative (`exp(dA_l - dA_s)`) or using a block/local anchor shift so both factored exponentials stay finite while preserving the difference.

Ibuki-wind: In `decay_dt_s`, clamping `-dA_val` to 88 changes the exact value of `exp(dA_l - dA_s)` for valid pairs once `dA_s < -88` (for example `dA_l=-100.1`, `dA_s=-100.0`), so long-decay chunks no longer match the reference math -> keep the exponent relative or introduce a shared anchor shift that preserves the difference without overflow.

Ibuki-wind (changes requested):
### Overall
The new decay-path rewrite still needs one correctness fix before merge.

tileops/kernels/mamba/ssd_chunk_scan.py
stelladuyx: Good catch, thank you. You're right — the clamp breaks the relative-difference identity once `dA_s < -88`, exactly as your example shows. Fixed in the latest commit by switching to an anchor-shifted factored form: `exp(dA_l − anchor) × exp(anchor − dA_s)` where `anchor = dA_cumsum[s0]` (the maximum value in each s-block by monotonicity). Both exponent arguments are bounded and finite for all valid inputs, so the product always equals `exp(dA_l − dA_s)` exactly — no clamp needed, same MUFU count as before.

stelladuyx: Fixed — same root cause as @superAngGao identified. The new anchor-shifted form uses `anchor = dA_cumsum[s0]` so` anchor − dA_s[ss] ≥ 0` by monotonicity (no overflow on the s-side) and` dA_l[ll] − anchor` is bounded in both directions (finite for all l). The relative difference is preserved exactly. Updated in the latest commit.

tileops/kernels/mamba/ssd_chunk_scan.py
```diff
@@ -228,36 +248,69 @@ def main(
...
+                            anchor,
                         )
-                        dt_s[ss] = T.if_then_else(
+                        dt_val = T.if_then_else(
                             s_abs < Q,
                             T.cast(dt[bz, bh, bc, s_abs], accum_dtype),
                             T.float32(0.0),
                         )
-                    T.sync_threads()
+                        exp_neg_dA_s_shifted[ss] = T.exp(anchor - dA_val) * dt_val
```
superAngGao: This is still range-limited by the two separate `exp` calls. Picking `anchor = dA_cumsum[..., s0]` preserves the algebra symbolically, but if a single `block_s` drops by more than the float32 exp range, `T.exp(anchor - dA_val)` can overflow while `T.exp(dA_l_val - anchor)` underflows. For example, with `anchor=0` and a valid causal pair `dA_l=dA_s=-120`, the reference term is `exp(0)=1`, but this factors into `exp(-120) * exp(120)`, i.e. `0 * inf -> NaN`. So the new form is correct only while the per-s-block decay range is bounded.

One way to keep the MUFU reduction without this range issue is to use an l-side anchor for full-lower blocks, e.g. `anchor = dA_cumsum[..., l0]`, making both exponent arguments non-positive for valid pairs; the diagonal block can keep the original relative `exp(dA_l - dA_s)` or use a smaller guarded strategy.

superAngGao (changes requested):
The latest anchor-shifted rewrite fixes the previous clamp-equivalence bug, but it still needs one numerical-range fix before merge.

tileops/kernels/mamba/ssd_chunk_scan.py
stelladuyx: Great catch, and exactly right. Fixed in the latest commit (`7c5bda4`) by switching to an l-side anchor: `anchor = dA_cumsum[..., l0] `(the largest value in the l-tile). For any valid causal pair in a full-lower block (`s < l0 ≤ l`):

`dA_l[ll] − anchor ≤ 0` since `dA_l[ll] ≤ dA_l[l0] = anchor` (non-increasing)
`anchor − dA_s[ss] ≤ 0` since `dA_s[ss] ≥ dA_l[l0]` = anchor for all s < l0
Both exponent arguments are non-positive, so neither term can overflow, and underflow to 0 is numerically correct (the true value is also ~0). The diagonal path compute`s exp(dA_l − dA_s)` directly — the difference is bounded within `block_s` decay steps so no range issue there.

superAngGao (approved):
LGTM.

michaelwithu:
LGTM.

lcy-seso (approved):

Ibuki-wind (approved):

PR #1467, opened 2026-05-15
[Perf][GroupedGemm] Cooperative persistent kernel: 20-26% faster than 3WG

michaelwithu:
## Summary

Replaces the existing 3-WG persistent grouped GEMM kernel with a new implementation featuring:
- **Cooperative template** (bm=128): two math WGs split one tile's M dimension (split-A, shared-B), enabling bn=256 ns=3 within H100's 228KB SMEM cap
- **Pingpong template** (bm=64): retained for small-M decode workloads
- **Static-wave scheduler**: deterministic tile-to-CTA mapping, no atomic counter
- **TMA-store epilogue**: fast path for full tiles, STG fallback for partial m-tiles
- **Parameterized ring buffer**: num_stages ∈ [2..6] with SMEM auto-pruning

### vs Previous 3WG (prefill, N=4096, K=2048)

| Shape | Previous 3WG | New (Cooperative) | Speedup |
|---|---|---|---|
| T=4096 E=64 | 1.415 ms | 1.052 ms | ↑ 25.6% |
| T=4096 E=128 | 1.427 ms | 1.129 ms | ↑ 20.9% |
| T=8192 E=128 | 2.805 ms | 2.106 ms | ↑ 24.9% |
| T=16384 E=256 | 5.575 ms | 4.260 ms | ↑ 23.6% |

### vs CUTLASS (torch._grouped_mm)

| Shape | M | TileOPs (ms) | CUTLASS (ms) | vs CUTLASS |
|---|---|---|---|---|
| T=1024 E=64 | 8K | 0.359 | 0.356 | ≈ tied |
| T=4096 E=64 | 32K | 1.035 | 0.883 | ↓ 17% |
| T=8192 E=128 | 64K | 2.041 | 1.801 | ↓ 13% |
| T=4096 E=256 | 32K | 1.374 | 1.429 | ↑ 4% |
| T=4096 E=256 DSv3-up | 32K | 4.484 | 4.622 | ↑ 3% |
| Decode T=512 E=128 | 4K | 0.594 | 0.729 | ↑ 18% |
| Decode T=512 E=256 | 4K | 1.137 | 1.325 | ↑ 14% |

**Decode: 14-18% faster than CUTLASS.** Prefill large-M: CUTLASS faster by 11-23% (cluster multicast advantage, deferred to future PR).

## Key Design Decisions

- **Split-A workaround**: TileLang 0.1.9 asserts `A_offset[-2] == 0` in wgmma_gemm, so A is physically split into A_smem_top/bot rather than row-slicing one shared buffer
- **Shared B ring**: cooperative template uses 1× B_smem (vs pingpong's 2×), freeing SMEM for larger tiles
- **Default config**: bm=128 bn=256 ns=3 (autotuned best across 11 configs)
- **Barrier protocol**: ab_full arrive_count=128 (producer), ab_empty arrive_count=256 (both math WGs)

## Testing

- 18/18 tests pass (9 pingpong + 9 cooperative incl. partial-m edge cases)
- SM90 skipif added for non-Hopper CI
- Addresses PR #1454 review comments (redundant T.clear, N-alignment check, SM90 guard)

## Files Changed

- `tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py` — rewritten (cooperative + pingpong)
- `tileops/kernels/grouped_gemm/__init__.py` — simplified exports
- `tests/kernels/test_grouped_gemm_persistent_3wg.py` — expanded test suite
- Removed: old V2 files, redundant V2-vs-3WG benchmark

Supersedes #1454.


tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,844 @@
...
+    def forward(self, A, B, true_sizes, true_offsets):
+        C = torch.zeros(self.numel, self.N, dtype=self.dtype, device=A.device)
+        block_m = self.config["block_m"]
+        block_n = self.config["block_n"]
+        block_k = self.config["block_k"]
+        if self.K % block_k != 0:
+            raise ValueError(f"K-aligned only: K={self.K}, block_k={block_k}")
+        if self.N % block_n != 0:
+            raise ValueError(f"N-aligned only: N={self.N}, block_n={block_n}")
+        A = F.pad(A, (0, 0, 0, block_m))
```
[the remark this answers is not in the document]
michaelwithu: Fixed in bdf077a. Added `if A.shape[0] < self.numel + block_m` guard to skip the copy when the tensor is already padded.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,844 @@
...
+
+                if tx == 0:
+                    s_cum[0] = T.int32(0)
+                    for e in T.serial(num_experts):
+                        s_cum[e + 1] = s_cum[e] + (true_sizes[e] + (block_m - 1)) // block_m
+                    s_total[0] = s_cum[num_experts] * T.int32(_num_pid_n)
+                # CTA-wide sync: publishes s_cum/s_total to consumers
+                # (this pairs with the corresponding sync_threads in
+                # the elif/else branches below).
+                T.sync_threads()
```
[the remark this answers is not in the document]
michaelwithu: Not changing. This is TileLang's standard warp-specialized pattern: each branch (producer/consumer0/consumer1) has its own `T.sync_threads()` call, but TileLang's IR compiler unifies them into a single CTA-wide barrier at codegen. Moving the sync outside would break the `T.dec_max_nreg(24)` scope boundary (producer needs low register count for its init block before the sync). The SM100 reference (`gemm_tcgen5mma_ws_persistent.py`) and all existing TileLang warp-specialized kernels use this same pattern.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -0,0 +1,845 @@
...
+        self.numel = numel
+        self.num_experts = num_experts
+        self.N = N
+        self.K = K
+        self.dtype = dtype
+        if sm_count is None:
+            sm_count = torch.cuda.get_device_properties(
+                torch.cuda.current_device()).multi_processor_count
+        self.sm_count = sm_count
+        self.kernel = lambda: _persistent_grouped_gemm_v2_kernel(
```
superAngGao, line 72: This looks like it breaks the `tune=True` construction path. `Kernel.init_config(..., tune=True)` calls `self.autotune()` before `self.config` is populated, and autotune calls `self.kernel()`. This lambda then reads `self.config["block_k"]`, which should raise `KeyError: "block_k"` before any config is selected.

The PR body relies on autotune results, so I think this path should be fixed or covered by a small construction test. Since all current autotune configs use `block_k=64`, one simple option is to build the factory with a known/default `block_k` before autotune, or override `autotune()` for this kernel so the factory does not depend on `self.config` before it exists.

benchmarks/ops/bench_grouped_gemm_3wg_vs_cutlass.py
```diff
@@ -0,0 +1,139 @@
...
+  Qwen3-235B-decode   E=128 T=512
+  Qwen3-235B-prefill  E=128 T=4096
+  DeepSeek-V3-decode  E=256 T=512
+  DeepSeek-V3-prefill E=256 T=4096
+× {uniform, skewed} distribution = 8 cases.
+
+Usage:
+    TILELANG_CLEANUP_TEMP_FILES=1 TMPDIR=/tmp/jieneng_tvm_tmp \\
+    conda run -n tileops-tl9 --no-capture-output \\
+    python benchmarks/ops/bench_grouped_gemm_3wg_vs_2wg.py
```
superAngGao: Nit: this usage command still points at the old `bench_grouped_gemm_3wg_vs_2wg.py` name, but this PR adds/renames the script as `bench_grouped_gemm_3wg_vs_cutlass.py`.

stelladuyx (commented):
### 1. `tune=True` still looks broken for this kernel

`Kernel.__init__()` initializes `self.config = {}`, and `init_config(..., tune=True)` calls `self.autotune()` before the default-config path has populated `self.config`.

In this PR, `self.kernel` is set to a zero-argument lambda that closes over `self.config["block_k"]`, so constructing `GroupedGemmPersistent3WGKernel(..., tune=True)` can fail before autotuning actually starts.

There also appears to be a second autotuning issue: TileLang filters each autotune config entry against the tuned callable’s signature and raises on unused config keys. Since this tuned callable has no parameters, config keys such as `block_m`, `block_n`, `block_k`, `num_stages`, `threads`, and `group_size_m` cannot be consumed.

Could we make the tuned callable expose the tunables directly and avoid reading `self.config` during autotune setup? For example:

```python
def kernel(block_m, block_n, block_k, num_stages, threads, group_size_m):
    return _persistent_grouped_gemm_v2_kernel(
        self.numel,
        self.num_experts,
        self.N,
        self.K,
        self.dtype_str,
        self.sm_count,
        block_k,
    )

**Review by stelladuyx — COMMENTED** [review]
## 2. The F.pad guard can still over-pad partially padded A
The new guard avoids the unconditional copy when A is already large enough, but the padding amount is still wrong for partially padded inputs.

For example, if:

`A.shape[0] == self.numel + 1`
then the current code pads by a full block_m, producing:

`self.numel + block_m + 1`
rows. However, the generated kernel shape is declared as:

`A_shape = (numel + block_m, K)`
So the runtime tensor can become larger than the static shape expected by the JIT signature.

Could we pad only the deficit instead?

```
required_rows = self.numel + block_m
if A.shape[0] < required_rows:
    A = F.pad(A, (0, 0, 0, required_rows - A.shape[0]))
```
This preserves the fast path for already-padded inputs while keeping the runtime tensor shape aligned with the static kernel shape.

File: tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py (hunk 3 above)
Comment (michaelwithu) [reply tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py:72]: Good catch. Fixed in be28373 — lambda now reads `_DEFAULT_CONFIG["block_k"]` (constant 64) instead of `self.config["block_k"]`, so it's safe to call before config is populated.

File: benchmarks/ops/bench_grouped_gemm_3wg_vs_cutlass.py (hunk 4 above)
Comment (michaelwithu) [reply benchmarks/ops/bench_grouped_gemm_3wg_vs_cutlass.py]: Fixed in the same commit (be28373). Docstring now points to `bench_grouped_gemm_3wg_vs_cutlass.py`.

**Comment by michaelwithu** [comment]
@stelladuyx Both issues fixed in 2653848:

**1. `tune=True` broken** — Overrode `autotune()` with a custom implementation that iterates `autotune_configs`, calls `forward()` for each config, and benchmarks via `do_bench`. This sidesteps the TileLang `@autotune` decorator's signature mismatch (it expects the decorated callable to be a raw `@tilelang.jit` instance). Verified working:
```
>>> k = GroupedGemmPersistent3WGKernel(numel=16384, num_experts=128, N=4096, K=2048, tune=True)
Best config: {'block_m': 128, 'block_n': 128, 'block_k': 64, 'num_stages': 5, ...} (0.692 ms)
```

**2. F.pad over-padding** — Now pads to exactly `numel + block_m` rows:
```python
required_rows = self.numel + block_m
if A.shape[0] < required_rows:
    A = F.pad(A, (0, 0, 0, required_rows - A.shape[0]))
```
Handles partially-padded inputs correctly (e.g. `A.shape[0] == numel + 1` → pads by `block_m - 1`, not `block_m`).

superAngGao (approved):
LGTM.

lcy-seso (approved):

PR #1471, opened 2026-05-15
[Feat][Codegen] generate eval_roofline bodies from manifest roofline

### LINKED ISSUE #1459: [Feat][CODEGEN] generate eval_roofline bodies from manifest roofline
### Symptom / Motivation

`python scripts/validate_manifest.py --strict` on `main` reports **41 C7 errors** of the form:

```
[stub] <OpName>: eval_roofline is the Op base stub (not implemented by the concrete class)
```

Every `status: implemented` op is contractually required to expose a manifest-derived roofline so benchmarks can compute achieved fraction of peak FLOPS / bandwidth. 41 ops still inherit the base `Op.eval_roofline` stub. The strict-parity gate (#1291, #1372) cannot flip to blocking until this clears.

`docs/design/roofline.md` §4.4 explicitly points at codegen from manifest `roofline.flops` / `roofline.bytes` as the intended path.

### Root Cause Analysis

`Op.eval_roofline` is an abstract stub. Each subclass should evaluate the manifest's `roofline.flops` and `roofline.bytes` formulae against the op's concrete workload shape to produce a per-call (FLOPs, bytes) tuple. 41 implemented ops never received the override; benchmarks for these ops cannot report % of roofline.

Two paths, parallel to the C6 issue (#1458):

- **Codegen** (preferred per design docs): read `roofline.flops` / `roofline.bytes` (or `roofline.func`) from manifest, synthesize `eval_roofline(self, *args, **kwargs)` that evaluates the formulae against the op's runtime shape. Applied via class decorator / metaclass.
- **Hand-written stop-gap**: write `eval_roofline` per op (~41 functions). Acceptable but disconnected from the manifest source of truth.

### Related Files

- `tileops/ops/op_base.py` — `Op.eval_roofline` base stub
- `tileops/manifest/*.yaml` — `roofline` entries (source of truth)
- `tileops/ops/**/*.py` — 41 op classes inheriting the stub
- `docs/design/roofline.md` §4.4 — codegen direction
- `scripts/validate_manifest.py` — `check_c7_eval_roofline_not_stub`

## Goal

`python scripts/validate_manifest.py --strict` on `main` reports **zero** C7 `eval_roofline is the Op base stub` errors. Implementation should be codegen-driven from manifest `roofline`, not hand-written per op.

## Plan

1. Design codegen for `eval_roofline` mirroring the C6 approach (#1458): class decorator / metaclass that reads the op's `roofline.flops` / `roofline.bytes` from manifest at class-construction time and synthesizes a method evaluating them against runtime workload shape.
2. Support both syntactic shapes the manifest already uses: `roofline.func` (Python callable reference) and explicit `roofline.flops` / `roofline.bytes` expressions. Defer to `docs/design/roofline.md` §4.4 for the per-consumer contract.
3. Apply across the 41 affected ops. Co-design with the C6 codegen (#1458) if both share the same decorator / metaclass scaffold — one codegen module, two derived methods.
4. Verify: `python scripts/validate_manifest.py --strict | grep 'eval_roofline is the Op base stub'` returns 0; benchmark scripts that call `op.eval_roofline(...)` produce sensible (FLOPs, bytes) tuples; existing benchmarks that already overrode `eval_roofline` continue to work (no double-implementation).
5. Confirm interaction with `ManifestBenchmark` and any `bench_manifest_driven` consumers — they should now find non-stub `eval_roofline` on every implemented op.

Affected ops: 41 — full list reproducible from `python scripts/validate_manifest.py --strict | grep 'eval_roofline is the Op base stub'`.

## Constraints

- Joint manifest+impl PR for the codegen path. Codegen reads manifest at import time; schema change rides with implementation.
- Share infrastructure with the C6 codegen (#1458) where practical — single decorator / metaclass for both `_validate_dtypes` and `eval_roofline`.
- Do NOT demote ops to `status: spec-only` to silence C7.
- Do NOT add `parity_opt_out` (removed in #1292).

## Acceptance Criteria

- [ ] AC-1: `python scripts/validate_manifest.py --strict` reports 0 `eval_roofline is the Op base stub` errors
- [ ] AC-2: Implementation is codegen-driven from manifest `roofline`, not 41 hand-written overrides
- [ ] AC-3: `ManifestBenchmark` consumers still produce sensible roofline percentages
- [ ] AC-4: No existing op that already overrides `eval_roofline` regresses
- [ ] AC-5: Linked from meta #1372 as a prerequisite

lcy-seso:
Closes #1459

## Summary

- New `tileops/ops/_roofline_codegen.py`: synthesizes each op's `eval_roofline()` from its manifest `roofline` block. Two modes — `func` (dotted path eagerly resolved at class build) and `inline` (vars + flops + bytes, AST-validated against the §4.4.3/§4.4.4 namespace, emitted as plain Python with no runtime `eval`).
- `Op.__init_subclass__` installs both `_validate_dtypes` (from #1466) and `eval_roofline` (new) on every `status: implemented` subclass; skipped on spec-only ops and on classes / bases that supply their own override (MRO-checked).
- Inline mode reads op state under a 2-tier contract per `signature.inputs`/`params` names *referenced* by the roofline: tensor inputs as `self.<input>` (with both `.shape` and `.ndim`) or `self.<input>_shape` (shape tuple/list); params as `self.<param>`. Unreferenced manifest names are not bound. Missing required bindings raise `ValueError(op, input, conventions)`.
- Vars-layer AST validator is scope-aware: each comprehension generator's iterable is visited before its target binds (matching Python semantics); `.shape` / `.ndim` are accepted only directly on a declared tensor-input Name — chained (`x.shape.ndim`), subscripted (`x.shape[0].shape`), and local-operand (`N.shape`) forms reject; tensor inputs may not appear as bare values (no `sum(x)`, no `x[0]`); calls restricted to the §4.4.4 helper table; vars keys colliding with inputs/params/helpers/`elem_bytes`/earlier vars are rejected at synthesis.
- Arithmetic-layer validator is a positive AST allowlist (BinOp, UnaryOp, BoolOp, IfExp, Compare, Call to allowed helper, Constant, Name) — collection literals (`Tuple`/`List`/`Set`/`Dict`) and any other node kind fail synthesis.
- `PreluFwdOp` / `NanToNumFwdOp` (the two real inline-mode ops) declare bindings explicitly: `self.input_shape`, `self.weight_shape`, `self.input_shape`.
- `docs/design/roofline.md` §4.4.3 names the referenced-only binding decision and links to the rule.
- `.claude/domain-rules/ops-design.md` adds the concrete attribute contract for op authors.

## Test plan

- [x] pre-commit passed
- [x] `python -m pytest tests/test_roofline_codegen.py` — 36 passed
- [x] `python -m pytest tests/test_validate_manifest.py` — 221 passed
- [x] All 108 `status: implemented` ops synthesize without error against the live manifest
- [x] `python scripts/validate_manifest.py --strict`: this PR is *scoped to the C7 class* (`eval_roofline is the Op base stub`), and that count is **0** on this branch (was **41** on `testbed`). The script still exits non-zero with **58 `kw_only mismatch` errors in reduction ops** that are pre-existing on `testbed` (`testbed` itself reports 99 errors = 41 C7 + 58 kw_only) and are out of scope for this PR — they are tracked separately.

## Test node delta

`python scripts/test_node_delta.py --base upstream/testbed`:

```
File                              Base    HEAD    Delta
-------------------------------------------------------
tests/test_roofline_codegen.py     new      36    (new)
-------------------------------------------------------
TOTAL                                0      36      +36
```

**Justification:** new codegen module; tests cover both synthesis modes, the install-hook condition matrix (spec-only / explicit override / inherited override / missing metadata), the 2-tier binding contract (positive + tier-1 shape/ndim fallthrough + missing-binding error), vars-layer comprehension scope (target binds after iter), attribute whitelist + chained/subscripted/local-operand rejection, tensor-input bare-reference rejection, vars-key collision rejection, arithmetic-layer collection-literal rejection, param exposure contract, and the live-manifest C7 parity invariant.


tileops/ops/_roofline_codegen.py
```diff
@@ -0,0 +1,655 @@
...
+            return
+        raise ValueError(
+            f"{self.op_name}: roofline.vars[{self.var_name!r}] uses "
+            f"unsupported comprehension target {type(target).__name__}"
+        )
+
+    def _visit_comp(self, node: ast.AST) -> None:
+        generators = node.generators  # type: ignore[attr-defined]
+        scope: set[str] = set()
+        for gen in generators:
```
Ibuki-wind: The `_visit_comp` implementation pre-collects every generator target before visiting generator iterables, so `sum(d for d in d)` passes validation even though Python evaluates the iterable before binding `d` and will raise at runtime; this weakens the codegen unknown-name gate. → Visit each generator iterable in the current scope, then bind that generator target before visiting its `if` clauses, later generators, and the element/key/value expression.

tileops/ops/_roofline_codegen.py
```diff
@@ -0,0 +1,655 @@
...
+    """
+    try:
+        tree = ast.parse(expr, mode="eval")
+    except SyntaxError as exc:
+        raise ValueError(
+            f"{op_name}: roofline.{label} is not a valid Python "
+            f"expression ({exc})"
+        ) from exc
+
+    forbidden_nodes = (
```
Ibuki-wind: The `_validate_arithmetic_expr` forbidden-node list omits collection literals, so `flops: "(1, 2)"` or `flops: "[N]"` passes codegen and only fails later when generated `eval_roofline()` casts to `int`, violating the authoritative form gate. → Reject `ast.Tuple`, `ast.List`, `ast.Set`, and `ast.Dict`, or replace this with a positive allowlist for arithmetic AST nodes.

Ibuki-wind (changes requested):
### Overall
The roofline synthesis gate still accepts invalid inline expressions, so fix the AST validation before landing.

tileops/ops/_roofline_codegen.py
```diff
@@ -0,0 +1,655 @@
...
+        )
+
+    def _visit_comp(self, node: ast.AST) -> None:
+        generators = node.generators  # type: ignore[attr-defined]
+        scope: set[str] = set()
+        for gen in generators:
+            self._collect_targets(gen.target, scope)
+        self._scopes.append(scope)
+        try:
+            # Element / key / value expression and all generator
```
Ibuki-wind: The current child-scope strategy makes `sum(d for d in d)` validate even though Python resolves the iterable before the target is bound; visit each generator iterable in the outer scope, then bind the target before its `if` clauses, later generators, and body.

tileops/ops/_roofline_codegen.py
Ibuki-wind: The arithmetic AST gate still allows `Tuple`/`List`/`Set`/`Dict` literals, so malformed expressions like `(1, 2)` or `[N]` survive codegen and fail only at the final `int(...)` cast; reject those nodes here.

Ibuki-wind (changes requested):
### Overall
The PR body is still stale and must be refreshed to match the current diff before approval.

### Cross-cutting concerns
- The current body still describes the earlier draft-state / unresolved T008 narrative instead of the final change set.

tileops/ops/_roofline_codegen.py
```diff
@@ -0,0 +1,559 @@
...
+                    f"non-helper call (only whitelisted helper names may "
+                    f"be invoked)"
+                )
+            if node.func.id not in _VARS_HELPERS:
+                raise ValueError(
+                    f"{op_name}: roofline.vars[{var_name!r}] calls "
+                    f"non-whitelisted name {node.func.id!r}; vars-layer "
+                    f"helpers are {sorted(_VARS_HELPERS)!r}"
+                )
+    return tree
```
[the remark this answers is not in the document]
lcy-seso: Not assessed on latest HEAD

tileops/ops/_roofline_codegen.py
```diff
@@ -0,0 +1,559 @@
...
+                f"{op_name}: roofline.vars[{var_name!r}] uses forbidden "
+                f"construct {type(node).__name__}"
+            )
+        # Allow only ``<name>.shape``, ``<name>.ndim`` chains; reject
+        # dunder probes that could pivot into arbitrary state.
+        if isinstance(node, ast.Attribute) and node.attr.startswith("__"):
+            raise ValueError(
+                f"{op_name}: roofline.vars[{var_name!r}] references "
+                f"forbidden dunder attribute {node.attr!r}"
+            )
```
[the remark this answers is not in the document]
lcy-seso: Not assessed on latest HEAD

tileops/ops/_roofline_codegen.py
lcy-seso: Adopted. _visit_comp now walks generators in declaration order: each iterable is visited in the surrounding scope, then the target is bound, then conditions/subsequent iterables/body see the accumulated bindings. `sum(d for d in d)` now fails with "unknown name 'd'" at synthesis. New test: test_target_not_bound_in_its_own_iterable. See fa84ac2.

tileops/ops/_roofline_codegen.py
lcy-seso: Adopted (duplicate of the previous thread on the same issue). See fa84ac2.

tileops/ops/_roofline_codegen.py
lcy-seso: Adopted via the positive-allowlist option you mentioned. _ARITHMETIC_ALLOWED_NODES enumerates the legal numeric node types (BinOp, UnaryOp, BoolOp, IfExp, Compare, Call, Constant, Name + operator marker classes); anything outside fails synthesis. Tuple/List/Set/Dict literals reject at synthesis with "forbidden construct Tuple"/etc. New tests: test_tuple_literal_rejected, test_list_literal_rejected. See fa84ac2.

lcy-seso: Adopted (duplicate of the previous thread on the same issue). See fa84ac2.

Ibuki-wind (approved):
Clean — no issues.

tileops/ops/_roofline_codegen.py
```diff
@@ -0,0 +1,659 @@
...
+    surfaced with the op and input name rather than the vacuous
+    ``'NoneType' object has no attribute 'shape'`` that would otherwise
+    reach the caller from inside the generated body.
+
+    Op-family-specific aliases (``self.shape`` / ``self.num_channels``
+    / ``self.N_total``) are *not* consulted; ops opting into inline
+    roofline declare bindings explicitly per ``docs/design/roofline.md``
+    §4.4.3.
+    """
+    direct = getattr(op, name, None)
```
Ibuki-wind, line 83: The direct-tier check only requires `.shape`, so a shape-only object is returned here and the generated body later dies with a raw `AttributeError` when it needs `.ndim`. That breaks the documented 2-tier contract and hides the missing `*_shape` fallback. → Require both `.shape` and `.ndim` on the direct path, or wrap shape-only objects in `_ShapeProxy` before returning them.

tileops/ops/_roofline_codegen.py
```diff
@@ -0,0 +1,659 @@
...
+        if node.attr not in _VARS_ATTR_WHITELIST:
+            raise ValueError(
+                f"{self.op_name}: roofline.vars[{self.var_name!r}] "
+                f"accesses non-whitelisted attribute {node.attr!r}; "
+                f"vars-layer allows only "
+                f"{sorted(_VARS_ATTR_WHITELIST)!r}"
+            )
+        self.generic_visit(node)
+
+    def visit_Call(self, node: ast.Call) -> None:
```
Ibuki-wind, line 304: Helper calls are validated only by callee name, not by their arguments, so `sum(x)` or `int(x[0])` passes even when `x` is a tensor. That lets roofline vars read runtime tensor data instead of shape metadata. → Restrict helper arguments to shape-derived locals / `.shape` traversals, and reject raw tensor subscripts before codegen.

tileops/ops/_roofline_codegen.py
```diff
@@ -0,0 +1,659 @@
...
+    param_names = list(params.keys()) if isinstance(params, dict) else []
+
+    # Vars-layer legal name set: tensors + params + elem_bytes +
+    # vars-layer helpers. Each successfully-parsed vars entry expands
+    # the set so later entries may reference earlier locals.
+    vars_allowed: set[str] = set(input_names) | set(param_names)
+    vars_allowed.add("elem_bytes")
+    vars_allowed.update(_VARS_HELPERS.keys())
+
+    for name, expr in vars_block.items():
```
Ibuki-wind, line 498: `roofline.vars` keys can shadow helpers or earlier locals. For example, `vars.sum = 1` followed by `N = sum(x.shape)` validates, but the emitted function binds `sum = 1` before the call and fails at runtime. → Reject vars keys that collide with inputs, params, helpers, `elem_bytes`, or earlier vars before adding them to `vars_allowed`.

Ibuki-wind (changes requested):
### Overall
The codegen is close, but the review gate still accepts runtime-data expressions and the PR body does not describe the current HEAD. Fix those before merge.

### Cross-cutting concerns
- PR body is stale: it still reports 23 roofline tests and a +23 node delta, while current HEAD has 26 roofline tests and `scripts/test_node_delta.py --base upstream/main` reports +59 total nodes from new files.

tileops/ops/_roofline_codegen.py
lcy-seso: Adopted. Tier 1 now requires both `.shape` and `.ndim`; a shape-only object falls through to tier 2 (`<name>_shape`). New test test_tier1_requires_both_shape_and_ndim covers the fallthrough. See ee85f65.

tileops/ops/_roofline_codegen.py
lcy-seso: Adopted. _VarsExprValidator now tracks tensor input names. Bare references (`sum(x)`, `x[0]`) fail with "tensor input 'x' as a bare value"; `x.shape[0]` still passes because Attribute recursion skips Name(x) under the whitelisted attr. New TestVarsLayerTensorInputUsage (3 cases). See ee85f65.

tileops/ops/_roofline_codegen.py
lcy-seso: Adopted. _synthesize_inline_mode now rejects vars keys present in vars_allowed before validating the expression. `vars.sum = 1` followed by `N = sum(x.shape)` fails synthesis with "collides with an existing name". New TestVarsKeyCollision (3 cases: helper / input / param). See ee85f65.

tileops/ops/_roofline_codegen.py
```diff
@@ -0,0 +1,709 @@
...
+            )
+        if node.id in self._input_names:
+            raise ValueError(
+                f"{self.op_name}: roofline.vars[{self.var_name!r}] "
+                f"references tensor input {node.id!r} as a bare value; "
+                f"access shape metadata via {node.id}.shape / "
+                f"{node.id}.ndim instead"
+            )
+
+    def visit_Attribute(self, node: ast.Attribute) -> None:
```
Ibuki-wind, line 277: `visit_Attribute` only checks the attribute name, then allows whitelisted attributes on non-input operands, so expressions like `x.shape.ndim` or `x.shape[0].shape` pass synthesis and fail later inside generated `eval_roofline()`. That still weakens the authoritative form gate. → Allow `.shape` / `.ndim` only when the attribute is taken directly from a declared tensor input name, and reject chained/subscript/local attribute operands at synthesis.

Ibuki-wind (changes requested):
### Overall
The previous blockers were mostly addressed, but the vars-layer gate still lets invalid chained attribute expressions through synthesis; tighten that and refresh the node-delta section.

### Cross-cutting concerns
- PR body node delta is still not the approval-gate result: `python scripts/test_node_delta.py --base upstream/main` on current HEAD reports `+66` across `tests/ops/attention/test_is_causal_defaults.py`, `tests/test_dtype_codegen.py`, and `tests/test_roofline_codegen.py`, while the body reports only `tests/test_roofline_codegen.py` at `+33`.

tileops/ops/_roofline_codegen.py
lcy-seso: Adopted. visit_Attribute now requires node.value to be a Name whose id is in the declared signature.inputs set. Chained (`x.shape.ndim`), subscripted (`x.shape[0].shape`), and local-operand (`N.shape`) attribute forms all reject at synthesis with "non-tensor-input operand". 3 new tests in TestVarsLayerAttributeWhitelist. See 1ab67c8.

Ibuki-wind (changes requested):
### Overall
The code-side blockers are resolved, but the PR body still records a verification fact that does not match the current branch.

### Cross-cutting concerns
- The Test plan says `python scripts/validate_manifest.py --strict` has `0 ERROR total`, but it exits non-zero locally on current HEAD with 59 errors, and on a temporary merge into current `upstream/testbed` with 58 errors. The C7 roofline-stub class is gone, but the command as written is not passing; update the PR body to report the actual verification result instead of claiming 0 total errors.

Ibuki-wind (approved):
Clean — no issues.

PR #1483, opened 2026-05-15
[BugFix][MANIFEST] align kw_only on 34 implemented ops to match impl

### LINKED ISSUE #1456: [BugFix][MANIFEST] align kw_only flag on signature.params across 34 implemented ops
### Symptom / Motivation

`python scripts/validate_manifest.py --strict` on `main` reports **58 C3 errors** of the form:

```
[ctor] <OpName>: param '<x>' kw_only mismatch — manifest=False, code=True
```

across **34 `status: implemented` ops**. The implementation's `__init__` declares the parameter keyword-only (after a `*,` marker) while the manifest's `signature.params.<x>.kw_only` is `False` (or unset, which defaults to positional).

This is one of the open buckets that must clear before the strict-parity gate (#1291, #1372) can flip to blocking.

### Root Cause Analysis

Drift between two writers of the same contract. Most affected ops are reductions / elementwise / norm — families where `dim`, `keepdim`, `ord`, `dtype`, `unbiased`, `correction`, etc. were tightened to keyword-only in code at some point, while the manifest entry never moved.

Per [.claude/rules/manifest-trust-model.md](.claude/rules/manifest-trust-model.md), the side that disagrees with PyTorch is the side to fix. There is **no mechanical default direction** for this sweep — each `(op, param)` must be evaluated against the PyTorch reference (`torch.<op>` / `torch.nn.functional.<op>` / `torch.linalg.<op>`) and routed accordingly:

| PyTorch reference                          | Fix direction                                                    |
| ------------------------------------------ | ---------------------------------------------------------------- |
| param is after `*` (keyword-only)          | Flip manifest `kw_only: true`. Code is correct.                  |
| param is positional-or-keyword (before `*`) | Move param in code `__init__` before its `*,` marker. Manifest stays `false`. |

A naive "flip everything in manifest to true" approach is **wrong** for this bucket — empirically the families below mostly fall on the second row, not the first. PR #1483 (the first attempt) flipped all 58 sites to manifest-true and was correctly rejected on review for exactly this reason; the second iteration of this issue must do the per-param spot-check before editing.

### Related Files

- `tileops/manifest/elementwise_binary.yaml`
- `tileops/manifest/elementwise_unary_activation.yaml`
- `tileops/manifest/normalization.yaml`
- `tileops/manifest/scan.yaml`
- `tileops/ops/{activation,norm,reduction,scan,elementwise}/**/*.py` (affected ops' `__init__`)
- `scripts/validate_manifest.py` (`check_c3_ctor_signature_parity`)

## Goal

`python scripts/validate_manifest.py --strict` on `main` reports **zero** `[ctor] … kw_only mismatch` errors, with each `(op, param)` resolution matching the PyTorch reference signature.

## Plan

1. Run `python scripts/validate_manifest.py --strict 2>&1 | grep 'kw_only mismatch'` to enumerate the current 58 sites (34 ops).
2. For each `(op, param)`: open the PyTorch reference signature (interactive: `python -c "import inspect, torch; print(inspect.signature(torch.<op>))"`). Decide based on the table in Root Cause Analysis:
   - If PyTorch has the param **after `*`** → flip manifest `kw_only: true` (use round-trip `ruamel.yaml` to preserve comments / key order).
   - If PyTorch has the param **before `*`** → in the op's `__init__`, move the param out from after the `*,` marker so it becomes positional-or-keyword. Manifest stays as-is.
3. Both kinds of edits land in this PR; constraint is that the two sides finish aligned (validator green) after one cohesive change set.
4. If a family-by-family split produces cleaner reviews than one mega-PR, split. Either is fine.
5. Re-run `--strict`; confirm `kw_only mismatch` count is 0; commit.

Empirically observed routing for the affected families (from PR #1483 review by `Ibuki-wind`, sha `1ea4c2e`):

| Family                                      | Direction                                              |
| ------------------------------------------- | ------------------------------------------------------ |
| in-place activation `inplace` (Relu/Silu/…) | code → positional-or-keyword (PyTorch is positional)   |
| `BatchNormFwdOp.{training,momentum,eps}`    | code → positional-or-keyword                           |
| reductions `dim` / `keepdim` (Softmax/Sum/Mean/Amax/Amin/Prod/Argmax/Argmin/All/Any/CountNonzero/LogSumExp/Var/Std/VarMean) | code → positional-or-keyword |
| `linalg.vector_norm`-family `ord` / `dim`   | code → positional-or-keyword                           |
| `Cumsum/CumprodFwdOp.dim`                   | code → positional-or-keyword                           |
| `AddFwdOp.alpha`, `SubFwdOp.alpha`, `DivFwdOp.rounding_mode` | manifest → `kw_only: true` (PyTorch has them after `*`) |
| layer-norm-family / rms-norm-family kw-only params (if flipped) | manifest → `kw_only: true` |

This table is the empirical starting point — re-verify each entry against PyTorch before editing.

Affected ops (34): AddFwdOp, AllFwdOp, AmaxFwdOp, AminFwdOp, AnyFwdOp, ArgmaxFwdOp, ArgminFwdOp, BatchNormFwdOp, CountNonzeroFwdOp, CumprodFwdOp, CumsumFwdOp, DivFwdOp, EluFwdOp, HardsigmoidFwdOp, HardswishFwdOp, HardtanhFwdOp, InfNormFwdOp, L1NormFwdOp, L2NormFwdOp, LeakyReluFwdOp, LogSoftmaxFwdOp, LogSumExpFwdOp, MeanFwdOp, MishFwdOp, ProdFwdOp, ReluFwdOp, SeluFwdOp, SiluFwdOp, SoftmaxFwdOp, StdFwdOp, SubFwdOp, SumFwdOp, VarFwdOp, VarMeanFwdOp.

## Constraints

- Joint manifest+impl PR for this sweep. Code `__init__` edits are explicitly in scope — moving params across the `*,` marker is the correct fix when PyTorch has them positional. A previous formulation forbade code-side edits; that constraint was lifted after PR #1483's review showed most ops need exactly that direction.
- Per [.claude/rules/manifest-trust-model.md](.claude/rules/manifest-trust-model.md) "PyTorch parity" rule: **the side that disagrees with PyTorch is the side to fix.** No mechanical default direction; spot-check each `(op, param)` against the live PyTorch signature.
- Do NOT use `parity_opt_out` — the field was removed by #1292.
- Do NOT demote ops to `status: spec-only` to silence the validator. The implementations are present; only the kw_only declaration drifted.

## Acceptance Criteria

- [ ] Modified files pass unit tests (including the affected per-family tests under `tests/ops/{activation,norm,reduction,scan}/`).
- [ ] AC-1: `python scripts/validate_manifest.py --strict` reports 0 `[ctor] … kw_only mismatch` errors.
- [ ] AC-2: No op is demoted from `status: implemented` to `status: spec-only` as part of this fix.
- [ ] AC-3: All 34 ops listed in the Plan section are addressed.
- [ ] AC-4: Linked from meta #1372 as a prerequisite.
- [ ] AC-5: Each `(op, param)` resolution direction matches the PyTorch reference signature; the PR body documents per-family direction so reviewers can spot-check without re-deriving.

lcy-seso:
Closes #1456

## Summary

Aligns `kw_only` between manifest and code `__init__` for the 58 C3 parity errors, with per-`(op, param)` direction matching the live PyTorch API (`torch.<op>` / `torch.nn.functional.<op>`). Two routing outcomes are mixed in this single PR:

- **Code → positional-or-keyword (manifest stays `false`)** — applied where PyTorch keeps the param positional. Touches `tileops/ops/elementwise/{_base.py,activations.py}` (in-place activation `inplace`), `tileops/ops/norm/batch_norm.py` (`training`/`momentum`/`eps`; also swaps `*spatial: int` varargs for `spatial: Tuple[int, ...] = ()` so `__init__` can declare those three as positional-or-keyword while matching `torch.nn.functional.batch_norm`), `tileops/ops/reduction/{_softmax_base,all_op,any_op,argmax,argmin,count_nonzero,cumulative_base,inf_norm,l1_norm,l2_norm,log_softmax,logsumexp,reduce,softmax}.py` (`dim`/`keepdim`/`ord`).
- **Manifest → `kw_only: true` (code stays kw-only)** — applied only to `tileops/manifest/elementwise_binary.yaml` for `AddFwdOp.alpha`, `SubFwdOp.alpha`, `DivFwdOp.rounding_mode`; PyTorch places these after `*`.

Test/bench callers (`tests/ops/{test_batch_norm.py, test_elementwise_unary_activation_alignment.py, test_norm_ops.py}`, `benchmarks/ops/bench_batch_norm.py`) are updated to pass the now-positional params correctly (notably `BatchNormFwdOp((N, C), (H, W), eps=..., training=...)` with the `spatial` tuple).

## Test plan

- [x] `python scripts/validate_manifest.py --strict 2>&1 | grep -c "kw_only mismatch"` → `0` (down from `58` on testbed)
- [x] `python -m pytest -q tests/test_validate_manifest.py tests/test_ops_manifest.py` → 242 passed
- [x] Affected per-family tests: `python -m pytest -q tests/ops/{activation,norm,reduction,scan}/ tests/ops/test_norm_ops.py tests/ops/test_batch_norm.py tests/ops/test_elementwise_unary_activation_alignment.py` → 1105 passed, 2 skipped
- [x] `pre-commit run --all-files` → clean
- [x] AC-1 (manifest validator strict): 0 `[ctor] … kw_only mismatch` errors. Pre-existing `[stub] eval_roofline` errors (41) are unchanged on baseline — out of scope.
- [x] AC-2 (no demotion): no op was demoted from `status: implemented` to `status: spec-only`.
- [x] AC-3 (all 34 ops addressed): every `[ctor]` C3 site enumerated by the validator on testbed is now resolved.
- [x] AC-5 (per-family direction documented): see Summary above and the table in #1456.

## Test node delta

```
File                                                       Base    HEAD    Delta
------------------------------------------------------------------------------
tests/ops/test_batch_norm.py                                 N+0     N+0     0
tests/ops/test_elementwise_unary_activation_alignment.py     M+0     M+0     0
tests/ops/test_norm_ops.py                                   K+0     K+0     0
------------------------------------------------------------------------------
TOTAL                                                                        0
```

**Justification:** Pure signature-alignment PR — every `__init__` and call-site edit preserves the existing test surface. No new test nodes introduced; existing fixtures simply pass the now-positional params correctly.

### DIFF
--- a/tileops/manifest/elementwise_unary_activation.yaml
+++ b/tileops/manifest/elementwise_unary_activation.yaml
@@ -18,7 +18,7 @@ ReluFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false, kw_only: true}
+      inplace: {type: bool, default: false}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -91,7 +91,7 @@ SiluFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false, kw_only: true}
+      inplace: {type: bool, default: false}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -126,7 +126,7 @@ HardswishFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false, kw_only: true}
+      inplace: {type: bool, default: false}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -161,7 +161,7 @@ HardsigmoidFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false, kw_only: true}
+      inplace: {type: bool, default: false}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -196,7 +196,7 @@ MishFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false, kw_only: true}
+      inplace: {type: bool, default: false}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -231,7 +231,7 @@ SeluFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false, kw_only: true}
+      inplace: {type: bool, default: false}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -267,7 +267,7 @@ LeakyReluFwdOp:
       output: {dtype: "same_as(input)"}
     params:
       negative_slope: {type: float, default: 0.01}
-      inplace: {type: bool, default: false, kw_only: true}
+      inplace: {type: bool, default: false}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -303,7 +303,7 @@ EluFwdOp:
       output: {dtype: "same_as(input)"}
     params:
       alpha: {type: float, default: 1.0}
-      inplace: {type: bool, default: false, kw_only: true}
+      inplace: {type: bool, default: false}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -340,7 +340,7 @@ HardtanhFwdOp:
     params:
       min_val: {type: float, default: -1.0}
       max_val: {type: float, default: 1.0}
-      inplace: {type: bool, default: false, kw_only: true}
+      inplace: {type: bool, default: false}
     shape_rules:
     - "min_val <= max_val"
     - "output.shape == input.shape"
--- a/tileops/manifest/elementwise_unary_activation.yaml
+++ b/tileops/manifest/elementwise_unary_activation.yaml
@@ -18,7 +18,7 @@ ReluFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false}
+      inplace: {type: bool, default: false, kw_only: true}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -91,7 +91,7 @@ SiluFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false}
+      inplace: {type: bool, default: false, kw_only: true}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -126,7 +126,7 @@ HardswishFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false}
+      inplace: {type: bool, default: false, kw_only: true}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -161,7 +161,7 @@ HardsigmoidFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false}
+      inplace: {type: bool, default: false, kw_only: true}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -196,7 +196,7 @@ MishFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false}
+      inplace: {type: bool, default: false, kw_only: true}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -231,7 +231,7 @@ SeluFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false}
+      inplace: {type: bool, default: false, kw_only: true}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -267,7 +267,7 @@ LeakyReluFwdOp:
       output: {dtype: "same_as(input)"}
     params:
       negative_slope: {type: float, default: 0.01}
-      inplace: {type: bool, default: false}
+      inplace: {type: bool, default: false, kw_only: true}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -303,7 +303,7 @@ EluFwdOp:
       output: {dtype: "same_as(input)"}
     params:
       alpha: {type: float, default: 1.0}
-      inplace: {type: bool, default: false}
+      inplace: {type: bool, default: false, kw_only: true}
     shape_rules:
     - "output.shape == input.shape"
 
@@ -340,7 +340,7 @@ HardtanhFwdOp:
     params:
       min_val: {type: float, default: -1.0}
       max_val: {type: float, default: 1.0}
-      inplace: {type: bool, default: false}
+      inplace: {type: bool, default: false, kw_only: true}
     shape_rules:
     - "min_val <= max_val"
     - "output.shape == input.shape"
--- a/tileops/manifest/normalization.yaml
+++ b/tileops/manifest/normalization.yaml
@@ -315,9 +315,9 @@ BatchNormFwdOp:
     outputs:
       output: {dtype: "same_as(x)"}
     params:
-      training: {type: bool, default: false}
-      momentum: {type: float, default: 0.1}
-      eps: {type: float, default: 1.0e-5}
+      training: {type: bool, default: false, kw_only: true}
+      momentum: {type: float, default: 0.1, kw_only: true}
+      eps: {type: float, default: 1.0e-5, kw_only: true}
     static_dims:
       C: "x.shape[1]"
     shape_rules:
--- a/tileops/manifest/normalization.yaml
+++ b/tileops/manifest/normalization.yaml
@@ -315,9 +315,9 @@ BatchNormFwdOp:
     outputs:
       output: {dtype: "same_as(x)"}
     params:
-      training: {type: bool, default: false, kw_only: true}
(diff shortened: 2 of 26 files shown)


tileops/manifest/elementwise_unary_activation.yaml
```diff
@@ -18,7 +18,7 @@ ReluFwdOp:
     outputs:
       output: {dtype: "same_as(input)"}
     params:
-      inplace: {type: bool, default: false}
+      inplace: {type: bool, default: false, kw_only: true}
```
Ibuki-wind: torch.nn.functional.relu still accepts inplace positionally (`relu(input, inplace=False)`), so kw_only: true breaks the referenced API -> drop kw_only here and on the other in-place activation params in this file.

tileops/manifest/normalization.yaml
```diff
@@ -315,9 +315,9 @@ BatchNormFwdOp:
     outputs:
       output: {dtype: "same_as(x)"}
     params:
-      training: {type: bool, default: false}
-      momentum: {type: float, default: 0.1}
-      eps: {type: float, default: 1.0e-5}
+      training: {type: bool, default: false, kw_only: true}
```
Ibuki-wind: torch.nn.functional.batch_norm still exposes training/momentum/eps as positional-or-keyword args, so marking them kw_only changes the cited signature -> remove kw_only from these three params.

tileops/manifest/reduction.yaml
```diff
@@ -19,7 +19,7 @@ SoftmaxFwdOp:
       # dtype, equivalent to x.to(dtype) before softmax). The current manifest
       # DSL cannot express "y.dtype = dtype if dtype is not None else x.dtype",
       # so dtype is omitted here; tracked as BLOCKED on missing-manifest-capability.
-      dim: {type: "int | None", default: null}
+      dim: {type: "int | None", default: null, kw_only: true}
```
Ibuki-wind: torch.nn.functional.softmax keeps dim before the `*` boundary, and the same positional-or-keyword mistake repeats across the rest of this reduction file (`log_softmax`, `sum`/`mean`/`amax`/`amin`/`prod`, `argmax`/`argmin`, `all`/`any`/`count_nonzero`, `linalg.vector_norm`) -> drop kw_only on those positional params and keep it only where PyTorch actually places `*`.

tileops/manifest/scan.yaml
```diff
@@ -15,7 +15,7 @@ CumsumFwdOp:
     outputs:
       y: {dtype: "same_as(x)"}
     params:
-      dim: {type: int, default: -1}
+      dim: {type: int, default: -1, kw_only: true}
```
Ibuki-wind: torch.cumsum still takes dim positionally (`cumsum(input, dim, *, dtype=None, out=None)`), so kw_only: true is wrong here -> revert dim to positional-or-keyword, and apply the same fix to CumprodFwdOp below.

Ibuki-wind (changes requested):
### Overall
These signature edits drift from the cited PyTorch APIs in several families; revert the positional params that were marked keyword-only before landing.

tileops/manifest/elementwise_unary_activation.yaml
lcy-seso: Done in 9ea6120.

tileops/manifest/normalization.yaml
lcy-seso: Done in 9ea6120.

tileops/manifest/reduction.yaml
lcy-seso: Done in 9ea6120.

tileops/manifest/scan.yaml
lcy-seso: Done in 9ea6120.

tileops/ops/norm/batch_norm.py
```diff
@@ -104,18 +104,19 @@ def __init__(
         self,
         N: int,
         C: int,
-        *spatial: int,
+        spatial: Tuple[int, ...] = (),
```
Ibuki-wind, line 107: The PR body says this is a manifest-only metadata change and that no `__init__` signatures or implementation code were touched, but `BatchNormFwdOp.__init__` is changed here and the diff also changes multiple other op constructors/tests/benchmarks; the approval gate requires the body to record final state only → update the PR body summary/test plan/regression text so it describes the actual merged diff instead of the old manifest-only round.

Ibuki-wind (dismissed):
### Overall
PR body still needs to reflect the final diff state before approval.

tileops/ops/norm/batch_norm.py
lcy-seso: Updated PR body to describe the actual round-2 diff (manifest reverts + code __init__ alignment across 22 files + test/bench caller updates). Summary now lists per-family direction (code→positional vs manifest→kw_only); Test plan reflects the validator + 1105-passing test runs from this round.

Ibuki-wind (approved):
Clean — no issues.

Ibuki-wind (dismissed):
### Overall
Review-thread replies need to be edited to outcome-only one-liners before approval.

### Cross-cutting concerns
- Latest replies on the manifest/alignment review threads include per-family change narration and test-plan/process detail; edit each to `Done in 9ea6120.` or an equivalent outcome-only line.

Ibuki-wind (approved):
### Overall
LGTM on 9ea6120. Manifest+impl kw_only alignment now matches PyTorch per-param, inline replies normalized to outcome-only.

Issue #1460, opened 2026-05-15
[BugFix][NORM] fix InstanceNormFwdOp._validate_dtypes signature mismatch


lcy-seso:
### Symptom / Motivation

`python scripts/validate_manifest.py --strict` on `main` reports one C2 dtype-parity failure:

```
[dtype] InstanceNormFwdOp: _validate_dtypes signature does not match manifest inputs
        (expected kwargs ['bias', 'weight', 'x']):
        TypeError: got an unexpected keyword argument 'weight'
```

The op's `_validate_dtypes` does not accept the manifest-declared inputs as keyword arguments — the validator constructs mock tensors keyed by `signature.inputs[*].name` and passes them as kwargs, which raises immediately. Today this surfaces as a single error; once the C6 codegen lands (#1458) for everything else, this op will become an exception to the otherwise-uniform path and needs an explicit fix.

This is the only non-stub C2/C6 error on `main` and must clear before the strict-parity gate (#1291, #1372) can flip to blocking.

### Root Cause Analysis

`InstanceNormFwdOp._validate_dtypes` was hand-written with a signature shape that doesn't accept `weight` / `bias` as keyword arguments — likely positional-only or with a different parameter name (`gamma` / `beta`?). The manifest at `tileops/manifest/normalization.yaml:509` declares the inputs as `x`, `weight`, `bias`. The validator drives the check using the manifest names; the impl rejects them.

The fix has two reasonable shapes:

- Rename / restructure `InstanceNormFwdOp._validate_dtypes` parameters to match manifest input names (preferred — manifest is the spec).
- If the C6 codegen (#1458) lands first, this op should drop its hand-written `_validate_dtypes` and inherit the codegen path — the bug disappears with the override.

### Related Files

- `tileops/ops/norm/instance_norm.py:38` — `InstanceNormFwdOp`
- `tileops/manifest/normalization.yaml:509` — manifest entry
- `scripts/validate_manifest.py` — `check_dtype_parity` (C2 path)

## Goal

`python scripts/validate_manifest.py --strict` on `main` reports **zero** `[dtype] InstanceNormFwdOp` errors.

## Plan

1. Inspect `tileops/ops/norm/instance_norm.py:38` to see the current `_validate_dtypes` signature and how it names its parameters.
2. Choose one of:
   - Align param names to the manifest's `signature.inputs` order (`x`, `weight`, `bias`) and accept them as kwargs.
[...]

Issue #1458, opened 2026-05-15
[Feat][CODEGEN] generate _validate_dtypes bodies from manifest dtype_combos


lcy-seso:
### Symptom / Motivation

`python scripts/validate_manifest.py --strict` on `main` reports **105 C6 errors** of the form:

```
[stub] <OpName>: _validate_dtypes is the Op base stub (not implemented by the concrete class)
```

Every `status: implemented` op is contractually required to enforce its manifest-declared `dtype_combos`, but 105 ops still inherit the base `Op._validate_dtypes` stub which raises `NotImplementedError`. The strict-parity gate (#1291, #1372) cannot flip to blocking until this clears.

Hand-writing 105 dtype validators duplicates information already in the manifest. Per `docs/design/ops-design.md` §Step 5, this is exactly the case codegen exists to handle.

### Root Cause Analysis

`Op._validate_dtypes` is intentionally an abstract stub. Each subclass should validate that the input tensor dtypes are one of the combinations enumerated in its manifest `dtype_combos` entry. Across 105 implemented ops, this override was never written — implementations were green because dtype mismatches surface later as kernel failures, and the strict-parity gate (C6) didn't exist until #1292.

Two paths:

- **Codegen** (preferred per design docs): read `dtype_combos` from manifest at class-construction time, synthesize a `_validate_dtypes(*tensors, **kwargs)` that checks the actual input dtypes against the allowed set. Applied via class decorator or metaclass so existing op classes inherit the manifest-derived check without per-op edits.
- **Hand-written stop-gap**: write `_validate_dtypes` in each of 105 op classes. Mechanical but voluminous; the design docs already point at codegen as the right answer.

### Related Files

- `tileops/ops/op_base.py` — `Op._validate_dtypes` base stub
- `tileops/manifest/*.yaml` — `dtype_combos` entries (source of truth)
- `tileops/ops/**/*.py` — 105 op classes inheriting the stub
- `docs/design/ops-design.md` §Step 5 — codegen direction
- `scripts/validate_manifest.py` — `check_c6_validate_dtypes_not_stub`

## Goal

`python scripts/validate_manifest.py --strict` on `main` reports **zero** C6 `_validate_dtypes is the Op base stub` errors. Implementation should be a single codegen path driven by manifest `dtype_combos`, not 105 hand-written overrides.

## Plan

[...]

Issue #1457, opened 2026-05-15
[BugFix][ATTENTION] add __init__ defaults for is_causal in 6 attention ops


lcy-seso:
### Symptom / Motivation

`python scripts/validate_manifest.py --strict` on `main` reports **6 C3 errors** of the form:

```
[ctor] <OpName>: param 'is_causal' has manifest default True but no default on __init__
```

across 6 attention ops:

- `MultiHeadAttentionFwdOp`
- `MultiHeadAttentionBwdOp`
- `GroupedQueryAttentionFwdOp`
- `GroupedQueryAttentionBwdOp`
- `GroupedQueryAttentionSlidingWindowFwdOp`
- `GroupedQueryAttentionSlidingWindowVarlenFwdOp`

The manifest declares `is_causal` with `default: true`, but the implementation's `__init__` makes the parameter required. The validator's C3 check (`scripts/validate_manifest.py:check_c3_ctor_signature_parity`) flags this divergence.

This is one of the open buckets that must clear before the strict-parity gate (#1291, #1372) can flip to blocking.

### Root Cause Analysis

The manifest entry asserts `is_causal=True` is the convention (matches PyTorch's `torch.nn.functional.scaled_dot_product_attention(is_causal=False)` shape — modulo direction; the manifest authors picked `True` as the attention-stack default). The implementation `__init__` never carried this default — callers were forced to pass it explicitly, which silently allowed each construction site to disagree with the manifest contract.

Either direction is a real fix:

- Add `is_causal: bool = True` to `__init__` — align impl to manifest (preferred; manifest is the spec per [.claude/rules/manifest-trust-model.md](.claude/rules/manifest-trust-model.md)).
- Remove the default from manifest — only if the codebase intentionally requires the caller to choose; this is a manifest-only PR with human review.

### Related Files

- `tileops/ops/attention/mha.py`
- `tileops/ops/attention/gqa.py`
- `tileops/manifest/attention.yaml` (or wherever attention entries live)
- `scripts/validate_manifest.py` (`check_c3_ctor_signature_parity`)

## Goal

`python scripts/validate_manifest.py --strict` on `main` reports **zero** `[ctor] … has manifest default … but no default on __init__` errors for the 6 attention ops.

## Plan

1. Confirm the manifest's `is_causal: true` default is intentional by checking PyTorch reference and existing call sites.
[...]

PR #1485, opened 2026-05-16
[CI] flip manifest validator to strict-parity blocking mode

### LINKED ISSUE #1291: [META][MANIFEST] enforce strict CI parity for status: implemented ops
### Symptom / Motivation

The manifest is the declared source of truth for op interfaces — `shape_rules`, `dtype_combos`, `signature.params`, `signature.inputs`, `source.kernel_map`, `roofline`. The current validator runs **partial** parity checks (shape + dtype) and allows ops to suppress them via `parity_opt_out`. There is no machine-enforced check on signature parity, `dispatch_kernel` routing, or whether `_validate_dtypes` / `eval_roofline` are real implementations vs base-class `NotImplementedError` stubs.

The W2 alignment audit (#1142) surfaced multiple `status: implemented` ops that diverged from manifest while CI stayed green:

- `kernel_map=` parameter silently dropped in 12+ irregular-init ops (#1286)
- `elementwise_binary` 24-op signature parity test deleted across cleanup chain, no regression net (#1287)
- `InstanceNormFwdOp.__init__` missing manifest-declared `use_input_stats` / `momentum` (#1288)

These gaps existed because `status: implemented` is currently honor-system. The flip from `spec-only → implemented` does not require the implementation to actually conform.

### Root Cause Analysis

`status: implemented` is supposed to mean "code conforms to manifest spec", but the validator does not machine-enforce that contract. Three layered failures:

1. parity validator allows `parity_opt_out` opt-out, used to silence real disagreements
2. signature / dispatch / non-stub invariants have no validator coverage at all
3. CI runs the validator but in non-strict mode — divergence reports as warnings, not errors

Trust boundary stays soft. When agents drive op migrations across multiple PRs, drift accumulates invisibly until a global audit finds it.

### Related Files

- `tileops/ops/op_base.py`
- `tileops/manifest/*.yaml`
- `.github/workflows/` (CI invocation)

## Goal

Every op with `status: implemented` passes 7 strict CI checks. `spec-only` ops are exempt. `parity_opt_out` is removed from the schema. Failures from enabling the gate become discrete follow-up issues, one per failing op, tracked from this meta issue.

## Plan

1. Extend `scripts/validate_manifest.py` to run the 7 checks below on every entry with `status: implemented`. `spec-only` entries skip all 7. The 7 checks:
   - **C1 shape parity**: `op._infer_output_shapes(*mock_inputs)` matches manifest `shape_rules` derivation
   - **C2 dtype parity**: `op._validate_dtypes(*tensors)` accepts/rejects exactly the set declared by manifest `dtype_combos`
   - **C3 ctor signature parity**: `inspect.signature(op_cls.__init__)` parameters match manifest `signature.params` (names, defaults, kw-only flags)
   - **C4 forward signature parity**: `inspect.signature(op_cls.forward)` positional parameter names match manifest `signature.inputs[*].name`
   - **C5 `dispatch_kernel` invariant**: constructing op with `kernel_map={name: Sentinel}` results in `op.kernel_map[name] is Sentinel`
   - **C6 `_validate_dtypes` not stub**: `op_cls._validate_dtypes is not Op._validate_dtypes`
   - **C7 `eval_roofline` not stub**: `op_cls.eval_roofline is not Op.eval_roofline`
2. Remove `parity_opt_out` field handling from validator and schema (currently 0 entries use it; cleanup is a no-op for current data).
3. CI runs the new strict mode on every PR. Strict mode failure = CI fail.
4. Run strict mode against current `main`; collect the failure list. Each failing op becomes a follow-up `[BUG][OPS] <OpName> fails strict parity (Cn)` issue, linked from this issue body.
5. Stagger rollout: enable C1+C2+C3+C4+C6+C7 immediately. C5 enables after #1286 (`dispatch_kernel` invariant infrastructure) lands.
6. While follow-up issues are open, the failing checks may run as advisory (CI warning) rather than blocking. Once all follow-ups close, advisory becomes blocking. Mode switch is a one-line CI config change.

## Constraints

- No `parity_opt_out` escape hatch. If an op genuinely can't be CI-checked (e.g., GPU-only and CI is CPU-only), demote it to `status: spec-only` until CI gains a GPU runner. Opt-out is not the answer.
- One follow-up issue per failing op per check (granular). Do NOT batch "all failing ops" into a single issue — that loses tracking resolution.
- The instruction "demote to spec-only" is only legitimate when impl truly doesn't conform; do not use demotion to silence the gate while leaving the impl as-is.
- CI must run strict checks on every PR, not nightly-only.
- Validator implementation reuses existing infrastructure where possible; add new check functions but keep the entry-point CLI shape stable.

## Acceptance Criteria

- [ ] AC-1: `scripts/validate_manifest.py` implements C1–C7 per the spec above, gated on `status: implemented`; `spec-only` entries skip all 7
- [ ] AC-2: `parity_opt_out` field is removed from validator code and from the manifest schema documentation
- [ ] AC-3: CI fails on any `status: implemented` op that violates any of C1–C7 (advisory phase acceptable per Plan step 6, but reaches blocking eventually)
- [ ] AC-4: Initial strict-mode run against `main` produces a complete failure list; one follow-up `[BUG][OPS]` issue is filed per failing (op, check) pair and linked from this issue body
- [ ] AC-5: All follow-up issues are tracked from this meta issue's body; this issue stays open until they all close and CI strict mode is the default blocking check
- [ ] AC-6: C5 (`dispatch_kernel` invariant) is wired into the validator after #1286 merges

## Follow-ups

Strict-mode run on `main` (PR #1292) surfaced 75 unique (op, check) failures, dominated by base-class stubs (C6 + C7 = 61 of 75). Per-op tracking would be issue spam; consolidated into 2 work tracks below.

- **C6 + C7 (61 sites)** — `_validate_dtypes` / `eval_roofline` are base-class `NotImplementedError` stubs across 46 / 15 ops. Both are codegen targets per `docs/design/ops-design.md` §Step 5 + `docs/design/roofline.md` §4.4. Tracked under the codegen META roadmap (B1, to be filed).
- **C3 (14 sites)** — kw-only / positional mismatch between manifest `signature.params` and `Op.__init__`. Mechanical alignment; one sweep PR.

C5 (`dispatch_kernel` invariant) failures are masked by missing `source.kernel_map` in many manifest entries; the underlying op-layer fix is tracked by #1286.

### Already-tracked

- #1286 — C5 `dispatch_kernel` invariant (12 irregular-init ops)
- #1287 — `elementwise_binary` 24-op signature parity
- #1288 — `InstanceNormFwdOp.__init__` parameter gap

### LINKED ISSUE #1372: [META][Manifest] clear strict-parity error backlog and flip CI to blocking
### Symptom / Motivation

PR #1292 lands `scripts/validate_manifest.py` in **advisory** mode: C1-C7 strict-parity failures route to warnings, validator exits 0. Until the existing main backlog is cleared, the gate cannot be flipped to blocking — flipping now would break every PR.

`python scripts/validate_manifest.py --strict` on `main` after #1292 reports **83 errors** in three buckets. This issue tracks closing them and flipping CI to blocking.

### Root Cause Analysis

Each error is a real spec/code drift on a `status: implemented` op:

- **22 [ctor] kw_only mismatch** across 14 ops (C3): manifest declares param positional, `__init__` declares it kw-only (or vice versa). Mechanical alignment.
- **46 [stub] `_validate_dtypes`** (C6): op has `status: implemented` but its `_validate_dtypes` is the base `Op` `NotImplementedError` stub. Implementation does not validate dtypes against the manifest contract.
- **15 [stub] `eval_roofline`** (C7): same shape; `eval_roofline` is the base stub.

C6 + C7 = 61 of 83 errors. Both are code that the manifest's declared contract requires every concrete `Op` subclass to override. The cleanest path is codegen from manifest, not hand-writing 61 stubs.

### Related Files

- `scripts/validate_manifest.py` (gate flip lives here)
- `tileops/ops/**/*.py` (all op subclasses with stub overrides)
- CI workflow that invokes the validator

## Goal

`python scripts/validate_manifest.py --strict` on `main` exits 0. CI flips to strict-blocking via a one-line change (drop `--advisory` / set `MANIFEST_STRICT_BLOCKING=1`).

## Plan

1. **C3 kw_only sweep** (22 errors / 14 ops): align manifest `signature.params.<name>.kw_only` with `__init__` signature. Mechanical; one PR per family or single sweep PR.
2. **C6/C7 stubs** (61 errors): close via codegen path. Two options:
   - Generate `_validate_dtypes` and `eval_roofline` bodies from manifest (`dtype_combos`, `roofline.{flops,bytes}`) per `docs/design/ops-design.md` Step 5 + `docs/design/roofline.md` §4.4. This is the direction the design docs already point at; PR #1369 is no longer a prerequisite (C5 was reduced to a static check) but the codegen idea stands on its own merit.
   - Or hand-implement per op (~60 functions, mostly mechanical). Acceptable as a stop-gap.
3. After 83 → 0: flip CI invocation to strict-blocking. One-line change in the workflow file.

## Constraints

- Do NOT silence errors by reverting ops to `status: spec-only` unless the implementation genuinely cannot conform. Demotion is the escape hatch only when impl really doesn't match manifest, not when manifest is right and impl is missing.
- Each step is independent — C3 can land before / after C6/C7. Order pick whatever gives clean PRs.
- This issue closes when the strict-blocking flip lands. The 83 underlying drifts are tracked in this body; per-op follow-ups are issue spam (see #1293-#1367 lessons learned, all closed).

## Acceptance Criteria

- [ ] AC-1: `python scripts/validate_manifest.py --strict` on `main` exits 0
- [ ] AC-2: CI invocation flipped to strict-blocking; advisory banner removed from validator output
- [ ] AC-3: All 22 [ctor] kw_only mismatches resolved
- [ ] AC-4: All 46 [stub] `_validate_dtypes` resolved (codegen or hand-written, in alignment with manifest dtype contract)
- [ ] AC-5: All 15 [stub] `eval_roofline` resolved (codegen from manifest `roofline.flops` / `roofline.bytes` preferred per design docs)
- [ ] AC-6: This issue stays open until AC-1 and AC-2 both hold; no per-op follow-up issues

lcy-seso:
## Summary

One-line CI change that flips `scripts/validate_manifest.py` from advisory to strict-parity blocking, and adds the `signature` level so C3 / C4 / C5 actually run (they're gated on `"signature" in levels` at `scripts/validate_manifest.py:3674`). Closes the two META issues that have been tracking this rollout:

- Closes #1291 (META: enforce strict CI parity for `status: implemented` ops)
- Closes #1372 (META: clear strict-parity error backlog and flip CI to blocking)

## Context

All five 2026-05-15 prerequisite issues are now closed:

| Bucket | Issue | Resolved by |
|---|---|---|
| C3 `kw_only` (58 errors / 34 ops) | #1456 | #1483 |
| C3 `is_causal` defaults (6 errors / 6 ops) | #1457 | #1461 |
| C6 `_validate_dtypes` stubs (105 errors) | #1458 | #1466 |
| C7 `eval_roofline` stubs (41 errors) | #1459 | #1471 |
| C2 `InstanceNormFwdOp` dtype signature (1 error) | #1460 | #1472 |

After those landed, `python scripts/validate_manifest.py --levels schema,signature,shape,dtype,bench --strict` on `main` exits 0 (verified locally).

## Test plan

- [ ] `preflight` job passes on this PR (proves the strict gate exits 0 on current main + this trivial diff)
- [ ] After merge, open a sanity PR with a deliberate strict-parity violation and confirm CI blocks it

### DIFF
--- a/.github/workflows/preflight.yml
+++ b/.github/workflows/preflight.yml
@@ -250,7 +250,7 @@ jobs:
         run: pip install -e .
 
       - name: Validate ops manifest
-        run: python scripts/validate_manifest.py --levels schema,shape,dtype,bench --strict
+        run: python scripts/validate_manifest.py --levels schema,signature,shape,dtype,bench --strict
 
   actionlint:
     needs: [validate-pr-title, ci-gate]
--- a/.github/workflows/preflight.yml
+++ b/.github/workflows/preflight.yml
@@ -250,7 +250,7 @@ jobs:
         run: pip install -e .
 
       - name: Validate ops manifest
-        run: python scripts/validate_manifest.py --levels schema,shape,dtype,bench
+        run: python scripts/validate_manifest.py --levels schema,shape,dtype,bench --strict
 
   actionlint:
     needs: [validate-pr-title, ci-gate]


.github/workflows/preflight.yml
```diff
@@ -250,7 +250,7 @@ jobs:
         run: pip install -e .
 
       - name: Validate ops manifest
-        run: python scripts/validate_manifest.py --levels schema,shape,dtype,bench
+        run: python scripts/validate_manifest.py --levels schema,shape,dtype,bench --strict
```
Ibuki-wind: `--strict` only makes strict parity checks blocking for enabled levels; `check_c3_ctor_signature_parity`, `check_c4_forward_signature_parity`, and `check_c5_dispatch_kernel_invariant` run only when `signature` is in `--levels`, so this CI command still misses constructor/forward/dispatch parity regressions → add `signature` to the `--levels` list.

Ibuki-wind (changes requested):
### Overall
The strict manifest gate still skips signature parity, so add the missing level before merging.

.github/workflows/preflight.yml
lcy-seso: Done in 5c92157.

Ibuki-wind (changes requested):
### Overall
The source blocker is fixed, but the PR still needs authoring cleanup before approval.

### Cross-cutting concerns
- The PR body still says `validate_manifest.py --levels schema,shape,dtype,bench --strict`, which no longer matches the current workflow; update the verification facts to include `signature`.
- The latest review reply is still narrative instead of outcome-only; replace it with `Done in 5c92157.` or an equivalent one-liner.

Ibuki-wind (approved):
Clean — no issues.

PR #1494, opened 2026-05-16
[Refactor][MOE] split FusedMoe into FusedMoeFwdOp / FusedMoeFwdCbFwdOp

### LINKED ISSUE #1476: [Refactor][MOE] split FusedMoe into FusedMoeFwdOp and FusedMoeFwdCbFwdOp variants
### Symptom / Motivation

`tileops/manifest/moe.yaml` declares two separate spec-only entries `FusedMoeFwdOp` and `FusedMoeFwdCbFwdOp`, but on the code side they are realized by a single class `FusedMoe` in `tileops/ops/moe/fused_moe.py`, distinguished only by a `with_correction_bias: bool` constructor parameter. The manifest names do not match any Python class — `validate_manifest.py --check-op FusedMoeFwdOp` cannot locate the class, blocking the flip for both entries.

### Root Cause Analysis

The two manifest entries were authored anticipating per-variant `Op` subclasses; the implementation took the shorter path of one class + a boolean parameter. This is fine for runtime but breaks the manifest ↔ class identity contract that `validate_manifest.py` relies on (C3 / C4 signature parity, C6 / C7 non-stub).

Two resolution shapes are possible:

- **(A) Split the class.** Define `FusedMoeFwdOp` and `FusedMoeFwdCbFwdOp` as thin subclasses (or sibling Op classes) of a shared internal helper. Manifest names map 1:1 to Python classes again. Recommended — keeps the manifest as the source of truth for op identity.
- **(B) Collapse the manifest.** Rewrite `moe.yaml` to a single entry, and express `with_correction_bias` as a `signature.params` field. Loses the ability to give each variant its own workloads/roofline/dtype_combos cleanly.

### Related Files

- `tileops/manifest/moe.yaml` (entries `FusedMoeFwdOp`, `FusedMoeFwdCbFwdOp`)
- `tileops/ops/moe/fused_moe.py` (`FusedMoe` class)
- `tests/ops/test_moe_fused_moe.py` (parametrized over `with_correction_bias`)
- `benchmarks/ops/bench_moe_fused_moe.py`

## Goal

Reconcile manifest and code so that both `FusedMoeFwdOp` and `FusedMoeFwdCbFwdOp` can flip to `status: implemented` and pass `validate_manifest.py --strict --check-op` for each name. Decide between (A) and (B) and execute end-to-end in one PR — both manifest entries must reach `implemented` together because they share the underlying class.

## Plan

1. Pick A or B based on whether the two variants have meaningfully different workloads, roofline formulas, or dtype contracts (if yes → A; if not → B).
2. Apply the chosen refactor: either introduce two Op classes wrapping shared internals, or collapse to one manifest entry.
3. Update tests / benches to reference the new identity (the existing parametrized test stays valid under A; under B, parametrization moves into manifest signature).
4. Wire `source.kernel_map` (already required by C5) and fix any strict-parity drift surfaced by `--check-op`.
5. Flip both entries (or the single collapsed entry) to `status: implemented`.

## Constraints

- This is one of the very few cases where two manifest entries must move together — they share a single Python class. The PR must close both flips (or the collapse) atomically.
- Behavior preserved: existing tests and benches must keep passing without semantic change to the kernel path.
- Path A preferred unless the variants are demonstrably identical along all manifest dimensions.

## Acceptance Criteria

- [ ] Either both `FusedMoeFwdOp` and `FusedMoeFwdCbFwdOp` flip to `implemented` (path A), OR `moe.yaml` carries a single collapsed entry at `implemented` (path B)
- [ ] `python scripts/validate_manifest.py --strict --check-op <name>` exits 0 for whichever manifest names remain after the refactor
- [ ] No behavior change in existing fused-MoE tests / benches

lcy-seso:
Closes #1476

## Summary

- Split `FusedMoe` into two `Op` subclasses (`FusedMoeFwdOp`, `FusedMoeFwdCbFwdOp`) wrapping the shared internal class; manifest names now resolve 1:1 to Python classes (Path A from the issue).
- Flipped **both** `FusedMoeFwdOp` and `FusedMoeFwdCbFwdOp` to `status: implemented` atomically in `tileops/manifest/moe.yaml` and wired `source.kernel_map` for each.
- Rewrote `tileops/perf/formulas.py::fused_moe_*` from inlined-table mode to func-mode signature so the manifest's `roofline` formulas resolve under `--strict --check-op` for both names.
- `benchmarks/ops/bench_moe_fused_moe.py` now reads workloads from the manifest via `load_workloads` and routes through `eval_roofline`, removing the inlined `PARAMS` table.

## Test plan

- [x] Modified files pass unit tests (`pytest tests/ops/test_moe_fused_moe.py -v` → 19 passed, 4 skipped [vLLM optional baselines, pre-existing], 0 failed).
- [x] Both `FusedMoeFwdOp` and `FusedMoeFwdCbFwdOp` flip to `implemented` (Path A).
- [x] `python scripts/validate_manifest.py --strict --check-op FusedMoeFwdOp` and `... --check-op FusedMoeFwdCbFwdOp` both exit 0 (warnings only on `_infer_output_shapes` parity-skip, not strict errors).
- [x] No behavior change in existing fused-MoE tests / benches (the 21 pre-existing pytest cases pass/skip exactly as on baseline; `FusedMoe.__init__` grew only an optional `kernel_map=None` kwarg, `forward()` body is identical).

## Test node delta

```
File                               Base    HEAD    Delta
--------------------------------------------------------
tests/ops/test_moe_fused_moe.py      21      23       +2
--------------------------------------------------------
TOTAL                                21      23       +2

Growth: +9.5%
```

**Justification:** +2 nodes are the two identity smokes (`test_fused_moe_fwd_op_identity`, `test_fused_moe_fwd_cb_op_identity`) that pin the manifest name → Python class binding for each of the two newly-implemented `Op` subclasses. Required to make the spec/code boundary load-bearing under Path A — without them, a future rename or re-wiring of `source.kernel_map` would only fail at validator time, not at unit-test time.

## Regression

- `FusedMoe` public API preserved: `__init__` signature only grew an optional `kernel_map=None` at the end; `forward()` signature unchanged. Subclasses delegate via `super().__init__` / `super().forward`.
- Benchmark workload set is equivalent to the prior inlined `PARAMS` table (now sourced from `tileops/manifest/moe.yaml::workloads`), so bench coverage is unchanged.

### DIFF
--- a/benchmarks/ops/bench_moe_fused_moe.py
+++ b/benchmarks/ops/bench_moe_fused_moe.py
@@ -1,22 +1,15 @@
-"""Benchmark for FusedMoe -- unified routed MoE FFN operator.
+"""Benchmark for FusedMoeFwdOp / FusedMoeFwdCbFwdOp (routed MoE FFN).
 
-Covers real model configurations in both decode (T small) and prefill (T large) regimes:
+Workload shapes come from each op's manifest ``workloads`` (via
+``load_workloads``); the benchmark reports TileOPs latency (nopad + padded
+layouts) alongside the manifest-derived roofline (``op.eval_roofline()``)
+and a vLLM / torch-ref baseline.
 
-  Model              H     F     E    K  scoring   renorm  bias   scale
-  Kimi K2          7168  2048  384   8  sigmoid   True    True   2.827
-  DeepSeek-V3      7168  2048  256   8  sigmoid   True    False  1.0
-  Qwen3-235B-A22B  7168  2048  128   8  softmax   False   False  1.0
-  Qwen3-30B-A3B    3072  1536  128   8  softmax   False   False  1.0
+Coverage:
 
-Baselines:
-  - tileops-padded: FusedMoe(layout="padded")
-  - vllm:           vLLM fused_topk + fused_experts (complete flow)
-  - torch-ref:      per-expert GEMM loop (fallback when vLLM is absent)
-
-FLOPs (same formula for both model families):
-  Gate+up: T*K * 2*F*H * 2  FLOPs
-  Down:    T*K *   H*F * 2  FLOPs
-  Total =  T*K * 6*F*H
+  Op                       Models (manifest workloads)
+  FusedMoeFwdOp            Qwen3-235B-A22B (softmax), DeepSeek-V3 (sigmoid)
+  FusedMoeFwdCbFwdOp       Kimi K2 (sigmoid + correction_bias)
 
 Usage:
     conda run -n tileops python -m pytest benchmarks/ops/bench_moe_fused_moe.py -vvs
@@ -40,27 +33,34 @@ except ImportError:
     _VLLM_AVAILABLE = False
 
 from benchmarks.benchmark_base import BenchmarkBase, BenchmarkReport
-from tileops.ops.moe import FusedMoe, FusedTopKOp
-from workloads.workload_base import FixtureBase, WorkloadBase
+from tileops.manifest import load_workloads
+from tileops.ops.moe import FusedMoeFwdCbFwdOp, FusedMoeFwdOp, FusedTopKOp
+from workloads.workload_base import WorkloadBase
 
-# ---------------------------------------------------------------------------
-# Test / fixture types
-# ---------------------------------------------------------------------------
+_OP_NAME = "FusedMoeFwdOp"
+_OP_NAME_CB = "FusedMoeFwdCbFwdOp"
+
+_DTYPE_MAP = {
+    "bfloat16": torch.bfloat16,
+    "float16": torch.float16,
+}
 
 
 class FusedMoeBenchTest(WorkloadBase):
+    """Inputs for a single FusedMoe benchmark configuration."""
+
     def __init__(
         self,
-        num_tokens,
-        num_experts,
-        top_k,
-        hidden_size,
-        ffn_size,
-        scoring_func,
-        renormalize,
-        with_correction_bias,
-        routed_scaling_factor,
-        dtype,
+        num_tokens: int,
+        num_experts: int,
+        top_k: int,
+        hidden_size: int,
+        ffn_size: int,
+        scoring_func: str,
+        renormalize: bool,
+        with_correction_bias: bool,
+        routed_scaling_factor: float,
+        dtype: torch.dtype,
     ):
         self.num_tokens = num_tokens
         self.num_experts = num_experts
@@ -77,10 +77,10 @@ class FusedMoeBenchTest(WorkloadBase):
         torch.manual_seed(42)
         dev = "cuda"
         hidden = torch.randn(
-            self.num_tokens, self.hidden_size, dtype=self.dtype, device=dev
+            self.num_tokens, self.hidden_size, dtype=self.dtype, device=dev,
         )
         gating = torch.randn(
-            self.num_tokens, self.num_experts, dtype=self.dtype, device=dev
+            self.num_tokens, self.num_experts, dtype=torch.float32, device=dev,
         )
         correction_bias = (
             torch.randn(self.num_experts, dtype=torch.float32, device=dev) * 0.1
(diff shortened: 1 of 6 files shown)


tileops/perf/formulas.py
```diff
@@ -661,21 +661,34 @@ def bitwise_xor_fwd_roofline(op: "Op") -> tuple[int, int]:
...
-    top_k = kwargs["top_k"]
-    hidden_size = kwargs["hidden_size"]
-    ffn_size = kwargs["ffn_size"]
-    elem_bytes = kwargs.get("elem_bytes", 2)  # bf16 default
+    num_tokens = int(op.num_tokens)
+    num_experts = int(op.num_experts)
+    top_k = int(op.top_k)
+    hidden_size = int(op.hidden_size)
+    ffn_size = int(op.ffn_size)
+    elem_bytes = op.dtype.itemsize
```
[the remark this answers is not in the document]
lcy-seso: Adopted. Switched to _dtype_itemsize helper for consistency with the other func-mode entries. See 605e93e.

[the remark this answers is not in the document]
lcy-seso: Not assessed on latest HEAD

tileops/ops/moe/fused_moe.py
```diff
@@ -80,6 +85,8 @@ def __init__(
         self.dtype = dtype
         self.expert_map = expert_map
 
+        self.dispatch_kernel(kernel_map)
```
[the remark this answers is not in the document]
lcy-seso: Valid point. Out of scope for this PR — kernel_map propagation across composite ops follows the existing wrapper-only pattern established earlier (the composite-op carve-out accepted a non-empty kernel_map override on the wrapper only). Reworking hierarchical kernel_map propagation across all composite ops (FusedMoe + MoEExperts variants) is a separate architectural decision. Will address in follow-up.

tileops/ops/moe/fused_moe.py
```diff
@@ -80,6 +85,8 @@ def __init__(
...
+
         self._fused_topk = FusedTopKOp(
```
Ibuki-wind, line 90: FusedMoe.__init__ calls dispatch_kernel(kernel_map) but then constructs _fused_topk and _experts without that map, so manifest/user overrides never reach the actual child kernels → pass the resolved kernel_map into FusedTopKOp and the selected experts_cls constructor.

benchmarks/ops/bench_moe_fused_moe.py
```diff
@@ -100,123 +100,113 @@ def ref_program(self, *args):
...
+    result_pad = bm.profile(op_pad, *forward_args_tileops)
     BenchmarkReport.record(op_pad, locals(), result_pad, tag="tileops-padded")
 
     # -- vLLM baseline (optional) ------------------------------------------
     if _VLLM_AVAILABLE:
         gating_f32 = gating.float()
 
         def _vllm_fn(hidden, gating, correction_bias, w_gate_up, w_down):
-            # vLLM: routing + GEMM (complete flow like TileOPs)
             tw, tids, _ = _vllm_fused_topk(
```
Ibuki-wind, line 214: _vllm_fn ignores correction_bias for FusedMoeFwdCbFwdOp workloads, so the vLLM baseline uses different routing than TileOPs when with_correction_bias=True → pass correction_bias through to vLLM if supported, otherwise route correction-bias workloads to the torch-ref baseline instead of recording them as vLLM.

Ibuki-wind (changes requested):
### Overall
Composite dispatch overrides and the correction-bias benchmark baseline are still wrong; fix those before landing.

tileops/ops/moe/fused_moe.py
lcy-seso: Adopted. kernel_map now propagates to FusedTopKOp and experts_cls constructors. See 6e3d378.

benchmarks/ops/bench_moe_fused_moe.py
lcy-seso: Adopted. correction-bias workloads now use torch-ref baseline (vLLM's fused_topk has no correction_bias parameter, so routing would diverge); vLLM only handles non-correction-bias workloads. See 6e3d378.

tileops/ops/moe/fused_moe.py
Ibuki-wind, line 88: FusedMoe.__init__ stores the manifest/user kernel_map only on the composite wrapper, but _fused_topk and _experts are constructed without it, so the declared FusedMoe source.kernel_map never reaches the kernels that actually execute → pass the resolved kernel_map into FusedTopKOp and the selected experts_cls constructor.

benchmarks/ops/bench_moe_fused_moe.py
Ibuki-wind, line 210: _vllm_fn drops correction_bias even for FusedMoeFwdCbFwdOp workloads, so the vLLM baseline routes different experts than TileOPs when with_correction_bias=True → pass correction_bias to the vLLM router if supported, otherwise skip the vLLM baseline for correction-bias workloads and use the torch-ref baseline.

Ibuki-wind (changes requested):
### Overall
Prior blockers are still present in the source; fix them before this status flip can be merged.

tileops/ops/moe/fused_moe.py
lcy-seso: Already fixed in 6e3d378. `FusedTopKOp(..., kernel_map=kernel_map)` at fused_moe.py:99 and `experts_cls(..., kernel_map=kernel_map)` at fused_moe.py:126. Review was generated on stale sha 605e93e.

benchmarks/ops/bench_moe_fused_moe.py
lcy-seso: Already fixed in 6e3d378. `use_vllm = _VLLM_AVAILABLE and not with_correction_bias` at bench_moe_fused_moe.py:209; correction-bias workloads now skip vLLM baseline and use torch-ref. Review was generated on stale sha 605e93e.

Ibuki-wind (approved):
Clean — no issues.

PR #1509, opened 2026-05-19
[Attention] Add FP8 GQA BN224 TMA-V baseline

superAngGao:
## Summary
- add the FP8 GQA BN224 WS TMA-V baseline path with FA3-style P/V/O contract
- keep the current internal scale-block layout compatible while accepting FA3-style `q_descale/k_descale/v_descale` inputs on the baseline wrapper
- add Hopper smoke coverage for the TMA-V baseline and FA3 descale contract

## Tests
- `docker run --rm --gpus "device=1" ... python -m pytest -q tests/ops/attention/test_gqa_fp8.py::test_gqa_fwd_fp8_bn224_tma_v_accepts_fa3_descale_contract -s`
- `docker run --rm --gpus "device=1" ... python -m pytest -q tests/ops/attention/test_gqa_fp8.py::test_gqa_fwd_fp8_bn224_tma_v_gqa_heads_are_finite -s`


Gabbering (commented):
## goose goose review — `bff7c52c`

*honk.* A large new FP8 GQA surface area lands. The kernel plumbing looks structurally sound, but the test for the FA3 descale contract is checking the wrong thing, and the `__init__.py` exports are inconsistent between layers.

### goose Bugs

- `tests/ops/attention/test_gqa_fp8.py:440-445` — The test `test_gqa_fwd_fp8_bn224_tma_v_accepts_fa3_descale_contract` asserts `q_descale.shape == (batch, heads_kv)` **after** calling the kernel. With `batch=1, heads_kv=2`, `_quantize_q_fa3_gqa_descale` returns `descale` of shape `(batch, heads_kv)` which is `(1, 2)`. That assertion passes trivially — it's testing the shape of the input you just created two lines above, not anything the kernel produced. Meanwhile the test never checks that the **output values are numerically close to a reference**. It only checks `isfinite`. This means the kernel could return all-zeros or garbage and the test would pass. The whole point of a "descale contract" test is to verify the kernel applies the descales correctly — you need a reference computation here.

- `tileops/kernels/__init__.py` vs `tileops/kernels/attention/__init__.py` — The top-level `__init__.py` exports `GQAFwdFP8Fa3ContractRawCudaSmokeKernel` and `GQAFwdFP8Fa3ContractRawCudaTileMapKernel` (and the two other `RawCuda*` debug kernels), but does **not** export `GQAFwdFP8Fa3ContractPtxAccBN224WsPingpongCorrectedKernel`, `GQAFwdFP8Fa3ContractPtxAccBN224WsPingpongCorrectedPreRescaleKernel`, `GQAFwdFP8Fa3ContractPtxAccBN224WsOverlapStreamingPKernel`, `GQAFwdFP8Fa3ContractPtxAccBN224WsOverlapStreamingPPlainWaitKernel`, `GQAFwdFP8Fa3ContractPtxAccBN224WsOverlapVisiblePVDeltaKernel`, `GQAFwdFP8Fa3ContractPtxAccBN224WsOverlapVisiblePVDeltaEmitterK224Kernel`, `GQAFwdFP8Fa3ContractPtxAccBN224WsOverlapVisiblePVDeltaSharedVKernel`, `GQAFwdFP8WgmmaKernel`, or `GQAFwdFP8WsPersistentKernel`. These are all exported from `tileops/kernels/attention/__init__.py` and used in the benchmark and test files. The benchmark file imports them from `tileops.kernels.attention` so it works, but anyone importing from `tileops.kernels` (the documented top-level) will get `ImportError`. Either the top-level re-exports are intentionally curated (in which case the `RawCuda*` debug kernels shouldn't be there either) or this is an incomplete sync.

### egg Test gaps

- `test_gqa_fwd_fp8_bn224_tma_v_accepts_fa3_descale_contract` — as noted above, this test asserts finiteness only. It does not compare against a reference (e.g., the same kernel called with the equivalent internal-scale layout, or a torch reference). A kernel that ignores the descale tensors entirely would pass this test. Add a numerical comparison — even a loose cosine-similarity check like the other tests in this file use.

- `test_gqa_fwd_fp8_bn224_tma_v_gqa_heads_are_finite` — same pattern: 32 heads, 8 KV heads, only checks `isfinite`. This is a smoke test, fine, but there is no correctness test anywhere for the TMA-V kernel at realistic GQA ratios (heads > heads_kv > 1) with a numerical reference. The existing `test_gqa_fwd_fp8_wgmma_baseline` only covers `heads=2, heads_kv=1`.

superAngGao:
Added the manifest-driven OP surface in follow-up commit da27d7a.\n\nThis now includes:\n- manifest entry: GroupedQueryAttentionPrefillFP8TensorCoreFwdOp\n- OP wrapper/export with FA3-style q_descale/k_descale/v_descale contract\n- kernel_map to GQAFwdFP8Fa3ContractPtxAccBN224WsTmaVKernel\n- roofline formula and manifest workload hook in bench_gqa_fp8.py\n- smoke coverage for the public OP wrapper\n\nValidation run in the nightly Docker image on GPU1:\n- PYTHONPATH=. python scripts/validate_manifest.py --check-op GroupedQueryAttentionPrefillFP8TensorCoreFwdOp --verbose\n- python -m pytest -q tests/ops/attention/test_gqa_fp8.py::test_gqa_prefill_fp8_tensor_core_op_accepts_fa3_descale_contract -s

Gabbering (commented):
## goose goose review — `da27d7a1`

*honk.* A manifest OP wrapper lands. The plumbing is mechanically correct but the goose has not forgotten its prior honks, and the new test continues the same pattern.

### egg Test gaps

- `tests/ops/attention/test_gqa_fp8.py:453-484` — The goose has not forgotten — as honked at commit `bff7c52c`, `test_gqa_fwd_fp8_bn224_tma_v_accepts_fa3_descale_contract` only asserts `isfinite`. The new `test_gqa_prefill_fp8_tensor_core_op_accepts_fa3_descale_contract` is a carbon copy of that exact weakness: it asserts shape, dtype, and finiteness, but never compares the output against a numerical reference. A kernel that returns `torch.ones(...)` passes this test. The OP wrapper delegates to `GQAFwdFP8Fa3ContractPtxAccBN224WsTmaVKernel` — the same kernel the earlier test covers — so you now have *two* tests for the same code path and *zero* correctness checks for the FA3 descale contract at a realistic GQA ratio. Please add a reference comparison (even a loose cosine-similarity gate against the internal-scale path or a torch SDPA reference) to at least one of these.

superAngGao:
Superseded by #1511.

PR #1512, opened 2026-05-20
[Perf][Mamba] Tune SSD chunk state default tile config

stelladuyx:
# SSDChunkStateFwdOp Benchmark — chunk_state config tuning

### Changes
**ssd_chunk_state_fwd**
- Update the default config for SSDChunkStateFwdKernel.
- Use separate defaults for the normal and has_seq_idx paths.
- Expand the autotune search space to include block_l=128.
- Keep the mathematical behavior and output layout unchanged.

**ssd_chunk_scan_fwd**
- Cache chunk-local dA_cumsum and dt values in shared memory.
- Reuse cached decay values in the history and causal scan paths.
- Keep the existing interface and output layout unchanged.
- Narrow the autotune space around production-relevant tile sizes.

## Environment

| | |
|---|---|
| GPU | NVIDIA H200 (GPU 1) |
| SM clock | **1830 MHz (locked)** |
| CUDA | 12.9 |
| PyTorch | 2.12.0.dev20260401+cu129 |
| Baseline | commit `81ebb55` — default config `block_n=64, block_p=64, block_l=64, threads=128` |
| This PR | commit `1323910` — `block_n=128, block_p=64, block_l=128, threads=256` (non-seqidx); `block_n=128, block_p=64, block_l=128, threads=128` (seqidx) |

Timing method: L2-flush CUDA events, 50 warmup + 500 timed iterations per shape.
Inputs randomly seeded. Both kernels pre-compiled before timing.

---

### ssd_chunk_state_fwd

| shape | dtype | has_seq_idx | before (µs) | after (µs) | speedup |
|---|---:|---:|---:|---:|---:|
| 370m-b1: B=1,C=8,L=256,H=32,P=64,N=128,G=1 | fp16 | false | 139.06 | 138.36 | **1.01×** |
| 370m-b4: B=4,C=8,L=256,H=32,P=64,N=128,G=1 | fp16 | false | 276.30 | 252.11 | **1.10×** |
| 1.3b-b1: B=1,C=8,L=256,H=64,P=64,N=128,G=1 | fp16 | false | 183.43 | 182.10 | **1.01×** |
| 1.3b-b4: B=4,C=8,L=256,H=64,P=64,N=128,G=1 | fp16 | false | 435.27 | 389.18 | **1.12×** |
| 2.7b-b1: B=1,C=8,L=256,H=80,P=64,N=128,G=1 | bf16 | false | 206.40 | 206.50 | **1.00×** |
| 1.3b-b1-L8k: B=1,C=32,L=256,H=64,P=64,N=128,G=1 | fp16 | false | 434.45 | 388.80 | **1.12×** |
| 1.3b-b1-seqidx: B=1,C=8,L=256,H=64,P=64,N=128,G=1 | fp16 | true | 205.72 | 184.50 | **1.12×** |


superAngGao (commented):
I found one behavior-coverage gap around the new `seq_end >= 0` semantics, plus a docstring mismatch.

1. `seq_idx == -1` reference logic is still using the old behavior.

The kernel now zeros the whole chunk when the last token has `seq_idx == -1`:

- `tileops/kernels/mamba/ssd_chunk_state.py`: `seq_end >= 0 and seq_idx[...] == seq_end`

But both local references still only compare equality with the chunk-end seq id:

- `tests/ops/test_mamba.py`: `same = (seq_chunked == seq_end).unsqueeze(3)`
- `benchmarks/ops/bench_mamba.py`: same fallback reference logic

That means the existing tests do not exercise the new `seq_end == -1` behavior. I checked a small repro where the chunk-end token is `-1`: the kernel matches a corrected reference with max error `0.0`, while it differs from the current reference (`max err ~= 0.0071`). Please update both references to include the `seq_end >= 0` guard and add a test case where the chunk end is `-1`, so this behavior is covered going forward.

2. The public op docstring still documents the old formula.

`tileops/ops/ssd_chunk_state.py` still says the mask is just `seq_idx[b,c*Q+l] == seq_idx[b,c*Q+Q-1]`. Since the public behavior now additionally requires the chunk-end seq id to be non-negative, this docstring should be updated as well.

Validation I ran locally:

- `git diff --check origin/main...pr-1512`
- `pytest -q tests/ops/test_mamba.py -k 'ssd_chunk_state_fwd'` -> 5 passed
- `pytest -q tests/ops/test_mamba.py -k 'ssd_chunk_scan_fwd'` -> 4 passed

Caveat: my local TileLang env is `0.1.8+cuda.git5f70374c`, while the repo currently pins `tilelang==0.1.9`, so final CI should still be treated as authoritative.

superAngGao (dismissed):
LGTM. The performance/config changes look reasonable and the related chunk state / chunk scan smoke tests pass locally on H200.

I left a note about tightening the `seq_idx == -1` reference coverage and syncing the public docstring, but I don't think that should block this PR: the kernel behavior itself matches the intended corrected semantics in my spot check.

stelladuyx:
Thanks for the thorough review @superAngGao — all three points are addressed in commit 0f0c63a on the `tune` branch.

**1. Reference guard (`seq_end >= 0`):**
- `tests/ops/test_mamba.py` `ssd_chunk_state_fwd_ref`: mask updated to `(seq_end >= 0) & (seq_chunked == seq_end)`
- `benchmarks/ops/bench_mamba.py` local copy: same fix applied

**2. New `seq_end == -1` test case:**
Added `test_ssd_chunk_state_fwd_seq_end_negative` (smoke-marked) — sets all tokens in the first chunk to `seq_idx == -1`, verifies the kernel output for that chunk is all-zero and agrees with the corrected reference. Ran locally: all 6 `ssd_chunk_state_fwd` tests pass (`6 passed in 9.58s`).

**3. Public op docstring:**
`tileops/ops/ssd_chunk_state.py` mask formula updated to `seq_idx[b,c*Q+Q-1] >= 0 and seq_idx[b,c*Q+l] == seq_idx[b,c*Q+Q-1]`.

michaelwithu (approved):
Tunes SSD chunk state/scan kernel configs for H200 with NCU-backed data. Key changes: default block_n=128 (covers full d_state in one tile), block_l=128, threads=256 for non-seqidx / threads=128 for has_seq_idx (register pressure driven). Also caches dA_cumsum/dt in shared memory to eliminate repeated L2 round-trips in chunk_scan, adds swizzled layouts for bank conflict elimination, and fixes a bug where chunks ending with seq_idx == -1 were not properly zeroed. The fix has dedicated test coverage.

LGTM.

superAngGao (approved):
LGTM.

lcy-seso (approved):

PR #1514, opened 2026-05-20
[Refactor][MoE] Unify FusedMoEExperts hierarchy with 3WG persistent kernel

### LINKED ISSUE #1513: [Refactor][MoE] Unify FusedMoEExperts hierarchy and integrate 3WG persistent kernel
### Motivation

1. **Duplicated code** — `fused_moe_experts.py` (`FusedMoeExpertsFwdOp` / `FusedMoeExpertsPaddedFwdOp`) and `experts/nopad.py` (`MoEExpertsNopadFwdOp`) implement the same pipeline (permute → GEMM → SiLU → GEMM → unpermute) with different interfaces (`forward()` vs `apply()`). This creates maintenance burden and divergence risk.

2. **Naming misalignment with vLLM** — vLLM uses `FusedMoEExperts` / `FusedMoEExpertsModular` as the ABC hierarchy. TileOPs uses `MoEExperts` / `MoEExpertsModular` without the "Fused" prefix, making cross-project reference confusing.

3. **3WG kernel not wired** — `GroupedGemmPersistent3WGKernel` (warp-specialized, TMA store, static-wave scheduler) is available but not used as the default GEMM backend for the experts op.

### Proposed Changes

1. Rename ABCs to align with vLLM: `MoEExperts` → `FusedMoEExperts`, `MoEExpertsModular` → `FusedMoEExpertsModular`, `MoEPrepareAndFinalize` → `FusedMoEPrepareAndFinalize`
2. Make `FusedMoEExperts` inherit from `Op`; unify on `forward()` (remove `apply()`)
3. Delete `fused_moe_experts.py` — eliminate duplicated implementation
4. Rename concrete: `MoEExpertsNopadFwdOp` → `FusedMoEExpertsNopadPersistent3WGFwdOp`, `MoEExpertsPaddedFwdOp` → `FusedMoEExpertsPaddedFwdOp`
5. Default GEMM kernel → `GroupedGemmPersistent3WGKernel` with auto-fallback to `MoeGroupedGemmNopadKernel` for unaligned dimensions
6. Rename test/bench files to `test_fused_moe_experts.py` / `bench_fused_moe_experts.py`
7. Update all callers: `FusedMoe`, prepare_finalize, manifest, docs

### Target Architecture

```
Op (tileops base)
  └── FusedMoEExperts (ABC)
        └── FusedMoEExpertsModular (ABC)
              ├── FusedMoEExpertsNopadPersistent3WGFwdOp  (3WG default)
              └── FusedMoEExpertsPaddedFwdOp              (padded baseline)
```

### Acceptance Criteria

- [ ] All existing MoE tests pass with new class names
- [ ] `fused_moe_experts.py` deleted, no duplicated code
- [ ] Benchmark shows 3WG kernel performance vs vLLM Triton baseline
- [ ] `FusedMoe` end-to-end op works with new experts hierarchy

michaelwithu:
Closes #1513

### Summary

Refactor the MoE experts implementation to align with vLLM's `FusedMoEExpertsModular` interface while integrating the 3WG persistent grouped GEMM kernel as the default backend. Eliminates 299 lines of duplicated code and establishes a clean inheritance hierarchy.

### Architecture

```
Op → FusedMoEExperts (ABC) → FusedMoEExpertsModular (ABC)
  ├── FusedMoEExpertsNopadPersistent3WGFwdOp (nopad, 3WG default)
  └── FusedMoEExpertsPaddedFwdOp (padded baseline)
```

### Changes

| Category | Detail |
|----------|--------|
| Rename ABCs | `MoEExperts` → `FusedMoEExperts`, `MoEExpertsModular` → `FusedMoEExpertsModular`, `MoEPrepareAndFinalize` → `FusedMoEPrepareAndFinalize` |
| Inherit Op | `FusedMoEExperts` now inherits from `Op`; unify on `forward()` |
| Default GEMM | `GroupedGemmPersistent3WGKernel` with auto-fallback for unaligned dims |
| Delete duplicate | `fused_moe_experts.py` removed (-299 lines) |
| Rename files | `test_moe_experts_nopad.py` → `test_fused_moe_experts.py`, `bench_moe_experts_nopad.py` → `bench_fused_moe_experts.py` |
| Migrate callers | `FusedMoe`, `prepare_finalize/no_dp_ep.py`, manifest, docs |
| Tests | Replace deleted-op reference with PyTorch per-expert ground truth |

### Performance (H200, bf16, H=7168, F=2048, K=8)

| Workload | TileOPs 3WG | vLLM Triton | Ratio |
|----------|-------------|-------------|-------|
| T=4096, E=128 (prefill) | 7.0ms / 410 TF | 7.7ms / 375 TF | **109%** |
| T=4096, E=256 (prefill) | 9.2ms / 314 TF | 9.5ms / 304 TF | **104%** |
| T=512, E=128 (decode) | 3.2ms / 113 TF | 2.8ms / 128 TF | 88% |
| T=512, E=256 (decode) | 6.1ms / 59 TF | 5.5ms / 66 TF | 89% |

Prefill: matches or exceeds vLLM Triton. Decode: ~10% behind (persistent scheduling overhead at small batch).

### Testing

- `test_fused_moe_experts.py`: 15/15 passed
- `test_moe_fused_moe.py` (integration): smoke 11/11 passed
- ruff: all checks passed

### How to verify

```bash
# Unit tests
TILELANG_CLEANUP_TEMP_FILES=1 conda run -n tileops-tl9 \
    python -m pytest tests/ops/test_fused_moe_experts.py -vv

# Integration
TILELANG_CLEANUP_TEMP_FILES=1 conda run -n tileops-tl9 \
    python -m pytest tests/ops/test_moe_fused_moe.py -k smoke -v
```


stelladuyx (commented):
Overall this is a clean refactor — the Op inheritance unification is the right call, and deleting 299 lines of duplication is a real win. A few things worth addressing before merge.

.claude/settings.json
```diff
@@ -1,4 +1,13 @@
 {
+  "permissions": {
```
stelladuyx: This file shouldn't be committed. It's a local Claude Code session artifact — the allowlist was generated during development and leaks the dev env name (`tileops-tl9`) and hardcodes test commands that will go stale. Please add `.claude/` to `.gitignore` and drop this from the PR.

tileops/ops/moe/routed_expert/fused_routed_expert.py
```diff
@@ -38,71 +41,83 @@ def __init__(
...
+        # 3WG requires N and K aligned to its default block dimensions.
+        # Fall back to tile scheduler kernel for small/unaligned dimensions.
+        _3wg_block_n = _3WG_DEFAULT_CONFIG["block_n"]
+        _3wg_block_k = _3WG_DEFAULT_CONFIG["block_k"]
+        gate_up_n = ffn_size * 2
+        if kernel_cls is GroupedGemmPersistent3WGKernel:
+            gate_up_ok = (gate_up_n % _3wg_block_n == 0) and (hidden_size % _3wg_block_k == 0)
+            down_ok = (hidden_size % _3wg_block_n == 0) and (ffn_size % _3wg_block_k == 0)
+            if not (gate_up_ok and down_ok):
+                kernel_cls = MoeGroupedGemmNopadKernel
```
stelladuyx, line 256: Is the fallback path covered by any test? The fixture uses T=128, H=256, F=128 which likely hits the 3WG happy path. Worth adding a case with dimensions that trigger this fallback — especially since decode is already ~10% behind vLLM, and it's worth knowing whether the fallback makes that worse.

tileops/ops/moe/fused_moe.py
```diff
@@ -156,11 +156,12 @@ def forward(
...
         expert_out_shape = self._experts.output_shape(T_prime, self.hidden_size)
         expert_out = output if expert_out_shape == tuple(hidden_states.shape) else hidden_states.new_empty(expert_out_shape)
-        self._experts.apply(
+        self._experts.forward(
             expert_out, r.hidden_q, w_gate_up, w_down,
             r.topk_weights, r.topk_ids,
-            self.num_experts, expert_map=self.expert_map,
+            expert_map=self.expert_map,
             workspace1=ws1, workspace2=ws2,
+            num_experts=self.num_experts,
```
stelladuyx, line 169: Good — using `num_experts=` as a keyword arg here means the parameter reorder in `forward()` is safe at this call site. Just worth confirming there are no other callers that pass `num_experts` positionally.

superAngGao (changes requested):
Thanks for the refactor. I like the direction of making `FusedMoEExperts` an `Op` and removing the older duplicated expert implementation, but I think a few issues should be fixed before merging:

1. `bench_fused_moe_experts.py` calls `forward()` with positional arguments in the wrong order. The new expert signature is `..., topk_ids, expert_map, workspace1, workspace2, num_experts`, but the benchmark currently passes `..., topk_ids, num_experts, None, ws1, ws2`, so `num_experts` is passed as `expert_map`. Please switch these calls to keyword arguments to make this robust against future signature changes.

2. `FusedMoEExpertsNopadPersistent3WGFwdOp` and `FusedMoEExpertsPaddedFwdOp` no longer set instance attributes that the previous implementations had, such as `self.num_tokens`, `self.num_experts`, `self.top_k`, `self.hidden_size`, `self.ffn_size`, and `self.dtype`. These are used by `_validate_dtypes()` and by the manifest-generated `eval_roofline()` path. Without restoring them, `self.dtype` falls back to the base default and roofline evaluation will miss the referenced params.

3. The `kernel_map` override behavior looks weakened. The top-level expert op still calls `self.dispatch_kernel(kernel_map)`, but the inner permute / SiLU / unpermute ops no longer receive the original `kernel_map`, and the nopad GEMM path now chooses between the default 3WG kernel and the fallback internally. This changes the previous ability to override child kernels through `FusedMoe(..., kernel_map=...)`. If this is intentional, the manifest/docs should say so; otherwise please preserve the old override semantics.

4. `.claude/settings.json` should not be part of this PR. The added permissions include local developer commands and environment names such as `tileops-tl9`, which are machine/session-specific artifacts. Please remove the settings change from the PR.

michaelwithu:
@stelladuyx @superAngGao Thanks for the detailed review — addressed in the latest force-push (`e2b623b`).

**Blockers fixed**

- **G1** (`bench` positional args): switched `nopad.forward(...)` and `padded.forward(...)` calls to keyword arguments in `benchmarks/ops/bench_fused_moe_experts.py`. The previous order was passing `num_experts` (int) where `expert_map` (Tensor|None) belongs, which would have crashed inside `_validate_dtypes`.
- **G2** (instance attributes): restored `self.num_tokens / num_experts / top_k / hidden_size / ffn_size / dtype` on both expert classes. Required by the manifest-derived `eval_roofline()` and by `_validate_dtypes` (which reads `self.dtype`).
- **G3** (kernel_map override): the top-level expert op now forwards `kernel_map` to all inner sub-ops (`MoePermute*`, `GroupedGemmOp` / `MoeGroupedGemmNopadFwdOp`, `MoeUnpermute`). For nopad's two GEMMs the user-supplied `moe_grouped_gemm_kernel` entry takes precedence over the auto-fallback default. `_install_kernel_map` ignores keys outside each sub-op's `default_kernel_map`, so transit through the composite is safe.
- **G4 / S1** (`.claude/settings.json`): reverted to `main`. Apologies — local Claude Code artifact that leaked in.

**S3** — grep'd for other `forward()` callers; found two more positional 10-tuple calls in `tests/ops/test_fused_moe_experts.py` (correctly ordered, so they worked, but not robust). Converted those to kw args as well.

**S2** — tracked as #1517. Current fixture (`T=128, H=256, F=128`) hits the 3WG happy path; fallback path is uncovered. Out of scope for this refactor; will address in a focused follow-up.

**M1** (@gemini-code-assist): the constants come from `_3WG_DEFAULT_CONFIG` in `tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py`, which is already the single source of truth — if the kernel's defaults change, this logic updates automatically. Not hardcoded at the call site. Leaving as-is.

Also added in passing (validator caught it): `_validate_dtypes` implementation on both expert classes, covering all 9 declared inputs including the `workspace1` / `workspace2` dtype union.

superAngGao (dismissed):
LGTM.

.claude/settings.json
lcy-seso: I am wondering why this is necessary?

lcy-seso (changes requested):

.claude/settings.json
michaelwithu: You're right — this was a leftover Claude Code session artifact (local dev allowlist with environment-specific names like `tileops-tl9`) that should never have made it into the PR. @stelladuyx and @superAngGao flagged it earlier; I thought I had reverted it but the change was lost during a botched commit amend. Reverted to `main` in `6531c54`, this file is no longer touched by this PR.

tests/ops/test_fused_moe_experts.py
```diff
@@ -75,28 +95,28 @@ def test_finalize_noop_reduce(self):
 
 
 # ---------------------------------------------------------------------------
-# MoEExpertsNopadFwdOp
+# FusedMoEExpertsNopadPersistent3WGFwdOp
 # ---------------------------------------------------------------------------
 
 @pytest.fixture(params=[torch.bfloat16, torch.float16], ids=["bfloat16", "float16"])
```
Ibuki-wind, line 111: The parametrized `moe_tensors` fixture is consumed by `test_workspace_shapes`, `test_output_shape`, and `test_make_weighted_reduce_is_noop`, but those assertions do not exercise dtype-specific behavior. Approval-gate test triage marks these duplicate dtype nodes as shrink cases, and `scripts/test_node_delta.py --base upstream/main` reports +53.8% growth. -> Split the metadata tests onto a single representative dtype or a non-parametrized fixture, while keeping dtype coverage on the forward correctness tests.

Ibuki-wind (changes requested):
### Overall
Test node delta is +53.8%, and the new dtype parametrization duplicates metadata-only cases; shrink that fixture usage before approval.

michaelwithu:
@Ibuki-wind Fixed. Split the parametrized fixture into two:

- `moe_meta` (non-parametrized, bf16 only) — used by `test_workspace_shapes`, `test_output_shape`, and `test_make_weighted_reduce_is_noop`, which don't exercise dtype-specific behavior.
- `moe_tensors` (parametrized `[bfloat16, float16]`) — kept only for the two `test_forward_matches_torch_ref` correctness tests.

Node count: 5×2 = 10 → 3×1 + 2×2 = 7. Force-pushed to update the branch.

stelladuyx (dismissed):
Overall: clean refactor, real deduplication, and the 3WG kernel integration is well-structured. 15/15 tests passing. Approving with a few comments inline.

stelladuyx (commented):
Approved. A few inline notes below.

stelladuyx:
**`tileops/ops/moe/experts/nopad.py` — silent fallback**

When dims aren't aligned and the kernel silently falls back to `MoeGroupedGemmNopadKernel`, nothing is logged. A user profiling their model won't know they're not on the 3WG path. Suggest adding a warning:

```python
import logging
_logger = logging.getLogger(__name__)

# in the fallback branch:
_logger.warning(
    "FusedMoEExpertsNopadPersistent3WGFwdOp: dims not aligned "
    "(gate_up_n=%d, hidden=%d, ffn=%d), falling back to MoeGroupedGemmNopadKernel",
    gate_up_n, hidden_size, ffn_size,
)
kernel_cls = MoeGroupedGemmNopadKernel
```

---

**`tileops/ops/moe/experts/padded.py` — missing `_validate_dtypes`**

`nopad.py` adds `_validate_dtypes()` but `padded.py` doesn't. Either add it there too or move it up to the `FusedMoEExpertsModular` ABC so both subclasses get it for free.

---

**`benchmarks/ops/bench_fused_moe_experts.py` — silent CUTLASS import failure**

The two-level `except ImportError: pass` means if both `cutlass_moe_fp16` and `cutlass_moe` fail, `_VLLM_CUTLASS_AVAILABLE = False` silently and the CUTLASS column just disappears from results. A `print`/`warnings.warn` on the outer except would make this easier to debug:

```python
except ImportError:
    _VLLM_CUTLASS_AVAILABLE = False
    # optional: print("vLLM CUTLASS baseline unavailable, skipping")
```

---

**Decode regression (T=512, ~10-12% behind vLLM Triton)**

The 3WG kernel is now the default for *all* workloads. Since the fallback only checks dim alignment, small-batch decode always gets 3WG even though it's slower there. Worth considering a batch-size threshold in the fallback logic, or at least documenting this tradeoff in the class docstring so users know to override `gemm_kernel` for decode-heavy deployments.

.claude/settings.json
michaelwithu: The file has been trimmed to just `{"enabledPlugins": {"superpowers@claude-plugins-official": true}}` (same as on `main`); the dev-env allowlist and env-specific names are gone. Resolving.

tileops/ops/moe/routed_expert/fused_routed_expert.py
michaelwithu: Added `test_forward_fallback_path_unaligned_dims` (H=128, F=96 → gate_up_n=192 ∤ 3WG block_n=256) in d33645c7c896a8fc8806e4b6b007e00239e1f1c0. It asserts both the fallback warning fires and the output matches the per-expert torch reference. Decode-perf parity vs. vLLM is a separate concern; see the parent comment reply for the docstring + opt-out hook.

tileops/ops/moe/fused_moe.py
michaelwithu: Confirmed. `forward()` of FusedMoEExperts is only called from `tileops/ops/moe/fused_moe.py:159` (uses `num_experts=` keyword) and the two per-class tests in `tests/ops/test_fused_moe_experts.py` (also keyword). No positional callers across `tileops/`, `tests/`, or `benchmarks/`.

.claude/settings.json
michaelwithu: Confirmed — the file is now minimal (`enabledPlugins` only, same as `main`) with no env-specific permissions. Resolving.

tests/ops/test_fused_moe_experts.py
michaelwithu: Thanks — the three metadata tests (`test_workspace_shapes`, `test_output_shape`, `test_make_weighted_reduce_is_noop`) actually already consume `moe_meta` (single non-parametrized dtype) rather than the parametrized `moe_tensors` fixture; only `test_forward_matches_torch_ref` reads `moe_tensors`, where dtype coverage is load-bearing. In d33645c7c896a8fc8806e4b6b007e00239e1f1c0 I added `test_forward_fallback_path_unaligned_dims` which is also bound to a single dtype (bfloat16), so the new growth is one test node, not a parametrized expansion.

michaelwithu:
Thanks for the careful pass! Addressed in d33645c:

1. **Silent fallback in `nopad.py`** — `_logger.warning(...)` in the fallback branch carrying the actual gate_up_n / hidden_size / ffn_size / block_n / block_k values so it's grep-able from a profiling session.
2. **`padded.py` `_validate_dtypes`** — both subclasses now route through `_validate_fused_moe_experts_dtypes` in `tileops/ops/moe/abc.py`. Kept the method on each class (rather than only on `FusedMoEExpertsModular`) because the L1 `_validate_dtypes` codegen (`tileops/ops/_dtype_codegen.py`) checks `cls.__dict__` and would otherwise install a synthesized validator that crashes on `expert_map=None` (the synth can't handle Optional[Tensor] inputs).
3. **CUTLASS silent import failure** — outer except now emits `warnings.warn(..., RuntimeWarning)` carrying the underlying ImportError so the disappearing column is visible. Also dropped the Usage docstring block + unused `sys` import per the recent bench cleanup convention.
4. **Decode regression doc** — added a Performance note to `FusedMoEExpertsNopadPersistent3WGFwdOp` calling out the decode tradeoff and the `gemm_kernel=MoeGroupedGemmNopadKernel` opt-out for decode-heavy deployments. Did not gate this on a batch-size threshold yet — the current alignment-only fallback is intentional, and a perf-driven threshold is a separate tuning exercise once we have a clear crossover point.

Also added `test_forward_fallback_path_unaligned_dims` to cover the fallback path (H=128, F=96) and replied + resolved the three earlier inline threads above.

tileops/ops/moe/experts/nopad.py
```diff
@@ -63,15 +98,13 @@ def __init__(
         self._gemm_gate_up = MoeGroupedGemmNopadFwdOp(
             numel=numel, num_experts=num_experts_local,
             n=ffn_size * 2, k=hidden_size, dtype=dtype,
-            kernel_map=kernel_map,
-        )
-        self._silu_and_mul = SiluAndMulFwdOp(
-            M=numel, N=ffn_size, dtype=dtype, kernel_map=kernel_map,
+            kernel_map={"moe_grouped_gemm_kernel": kernel_cls, **(kernel_map or {})},
         )
+        self._silu_and_mul = SiluAndMulFwdOp(M=numel, N=ffn_size, dtype=dtype)
```
Ibuki-wind: FusedMoEExpertsNopadPersistent3WGFwdOp.__init__ drops kernel_map when constructing SiluAndMulFwdOp, so a user or manifest override for `silu_and_mul` is silently ignored while the other sub-ops honor the same map. -> Pass `kernel_map=kernel_map` to `SiluAndMulFwdOp(...)`, matching the previous implementation and the manifest kernel_map entry.

tileops/ops/moe/experts/padded.py
```diff
@@ -69,9 +70,7 @@ def __init__(
             n=ffn_size * 2, k=hidden_size, dtype=dtype,
             kernel_map=kernel_map,
         )
-        self._silu_and_mul = SiluAndMulFwdOp(
-            M=padded_batch_sum, N=ffn_size, dtype=dtype, kernel_map=kernel_map,
-        )
+        self._silu_and_mul = SiluAndMulFwdOp(M=padded_batch_sum, N=ffn_size, dtype=dtype)
```
Ibuki-wind: FusedMoEExpertsPaddedFwdOp.__init__ drops kernel_map when constructing SiluAndMulFwdOp, so `silu_and_mul` overrides declared for the composite padded op never reach the activation sub-op. -> Pass `kernel_map=kernel_map` to `SiluAndMulFwdOp(...)`, as the other inner ops do.

tileops/manifest/moe.yaml
```diff
@@ -393,51 +319,69 @@ MoEExpertsNopadFwdOp:
...
     kernel: tileops/kernels/moe/moe_grouped_gemm_nopad.py
     op: tileops/ops/moe/experts/nopad.py
-    test: tests/ops/test_moe_experts_nopad.py
-    bench: benchmarks/ops/bench_moe_experts_nopad.py
+    test: tests/ops/test_fused_moe_experts.py
+    bench: benchmarks/ops/bench_fused_moe_experts.py
     bench_manifest_driven: false
     kernel_map:
       permute_nopad_kernel: MoePermuteNopadKernel
       moe_grouped_gemm_kernel: MoeGroupedGemmNopadKernel
```
Ibuki-wind: `FusedMoEExpertsNopadPersistent3WGFwdOp.source.kernel_map` still declares `moe_grouped_gemm_kernel: MoeGroupedGemmNopadKernel`, but the implementation now defaults aligned shapes to `GroupedGemmPersistent3WGKernel`. This makes the manifest provenance/dispatch table describe the wrong default backend. -> Update this entry to the 3WG kernel and keep the nopad fallback behavior represented in code/tests rather than as the manifest default.

Ibuki-wind (changes requested):
### Overall
The 3WG refactor has a dispatch metadata/override regression: sub-op kernel_map forwarding is incomplete and the manifest still describes the old GEMM backend for paths that now default to 3WG.

### Cross-cutting concerns
- Apply the 3WG manifest-default fix consistently to the end-to-end `FusedMoeFwdOp` and `FusedMoeFwdCbFwdOp` entries too; both share the default `FusedMoe(layout="nopad")` construction path that now instantiates `FusedMoEExpertsNopadPersistent3WGFwdOp`.

tileops/ops/moe/experts/nopad.py
michaelwithu: Fixed in 2d213b2 — `SiluAndMulFwdOp` now receives `kernel_map=kernel_map` at nopad.py:103, so user overrides for `silu_and_mul` are honored.

tileops/ops/moe/experts/padded.py
michaelwithu: Fixed in 2d213b2 — `SiluAndMulFwdOp` now receives `kernel_map=kernel_map` at padded.py:73, matching the pattern used by other sub-ops.

tileops/manifest/moe.yaml
michaelwithu: Fixed in 2d213b2 — Updated manifest/moe.yaml:327 to `GroupedGemmPersistent3WGKernel`, matching the code default for aligned dimensions. Also updated lines 449 and 500 (FusedMoeFwdOp and FusedMoeFwdCbFwdOp) since both use layout=nopad by default and should declare the same 3WG default.

tileops/ops/moe/routed_expert/abc.py
```diff
@@ -5,21 +5,67 @@
...
+            )
+    if topk_weights.dtype != torch.float32:
+        raise ValueError(f"Expected topk_weights.dtype == float32, got {topk_weights.dtype}")
+    if topk_ids.dtype != torch.int32:
+        raise ValueError(f"Expected topk_ids.dtype == int32, got {topk_ids.dtype}")
+    if expert_map is not None and expert_map.dtype != torch.int32:
+        raise ValueError(f"Expected expert_map.dtype == int32, got {expert_map.dtype}")
+    for name, t in (("workspace1", workspace1), ("workspace2", workspace2)):
+        if t.dtype not in allowed:
+            raise ValueError(f"Expected {name}.dtype in {allowed}, got {t.dtype}")
```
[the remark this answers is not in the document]
michaelwithu: Thanks for catching this. Investigated and applied a different fix — the literal `numel() > 0` guard breaks the manifest dtype fuzzer (`scripts/validate_manifest.py --strict`, run by the preflight `validate-manifest` job): the fuzzer feeds out-of-union dtypes (bool/int8/fp32/...) on numel=0 tensors and expects `_validate_dtypes` to reject them, which the guard would silently accept.

Fixed in 6a2b9a7 by allocating empty workspaces with the declared dtype in tests (`torch.empty(0, dtype=dtype, device="cuda")`) instead. The validator stays strict, callers just have to honor the manifest's declared dtype union for empty tensors too. The `fused_moe.py` production path already uses `hidden_states.new_empty(...)` so it inherits the right dtype automatically — only the experts unit tests were violating the contract.

tileops/ops/moe/routed_expert/fused_routed_expert.py
```diff
@@ -0,0 +1,310 @@
...
+        self,
+        num_tokens: int,
+        num_experts: int,
+        top_k: int,
+        hidden_size: int,
+        ffn_size: int,
+        routed_scaling_factor: float = 1.0,
+        dtype: torch.dtype = torch.bfloat16,
+        expert_map: Optional[Tensor] = None,
+        kernel_map=None,
```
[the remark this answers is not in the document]
michaelwithu: Done in 6a2b9a7 — typed as `Optional[Dict[str, Kernel]] = None` to match the convention used in other ops (e.g. `tileops/ops/ssd_state_passing.py`, `tileops/ops/gated_linear_attn.py`).

tileops/ops/moe/routed_expert/fused_routed_expert.py
```diff
@@ -0,0 +1,310 @@
...
+        num_tokens: int,
+        num_experts: int,
+        top_k: int,
+        hidden_size: int,
+        ffn_size: int,
+        routed_scaling_factor: float = 1.0,
+        dtype: torch.dtype = torch.bfloat16,
+        expert_map: Optional[Tensor] = None,
+        gemm_kernel: Optional[type] = None,
+        kernel_map=None,
```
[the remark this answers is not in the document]
michaelwithu: Done in 6a2b9a7 — same fix as the padded variant, typed as `Optional[Dict[str, Kernel]] = None` for consistency with the rest of the ops package.

tileops/ops/moe/routed_expert/fused_routed_expert.py
```diff
@@ -0,0 +1,310 @@
...
+        w_gate_up: Tensor,
+        w_down: Tensor,
+        topk_weights: Tensor,
+        topk_ids: Tensor,
+        expert_map: Tensor | None,
+        workspace1: Tensor,
+        workspace2: Tensor,
+        num_experts: int,
+    ) -> None:
+        perm_h_pad, padded_offsets, padded_sizes, _, fwd_idx = self._permute(
```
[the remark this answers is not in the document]
michaelwithu: Done in 6a2b9a7 — `self._validate_dtypes(...)` is now called at the very top of `FusedMoEExpertsPaddedFwdOp.forward()` before any sub-op invocations.

tileops/ops/moe/routed_expert/fused_routed_expert.py
```diff
@@ -0,0 +1,310 @@
...
+        w_gate_up: Tensor,
+        w_down: Tensor,
+        topk_weights: Tensor,
+        topk_ids: Tensor,
+        expert_map: Tensor | None,
+        workspace1: Tensor,
+        workspace2: Tensor,
+        num_experts: int,
+    ) -> None:
+        perm_h, true_offsets, true_sizes, _, fwd_idx = self._permute(hidden_states, topk_ids)
```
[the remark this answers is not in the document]
michaelwithu: Done in 6a2b9a7 — `self._validate_dtypes(...)` is now called at the very top of `FusedMoEExpertsNopadPersistent3WGFwdOp.forward()` before any sub-op invocations, mirroring the padded fix.

benchmarks/ops/bench_fused_moe_experts.py
```diff
@@ -132,24 +158,30 @@ def test_moe_experts_nopad_bench(
...
-    nopad = MoEExpertsNopadFwdOp(**kwargs)
+    # -- TileOPs nopad (3WG persistent) --------------------------------------
+    nopad = FusedMoEExpertsNopadPersistent3WGFwdOp(**kwargs)
 
     def _nopad_fn(hidden, w1, w2, topk_weights, topk_ids):
-        nopad.apply(output, hidden, w1, w2, topk_weights, topk_ids, num_experts, None, ws1, ws2)
+        nopad.forward(
+            output, hidden, w1, w2, topk_weights, topk_ids,
+            expert_map=None, workspace1=ws1, workspace2=ws2, num_experts=num_experts,
+        )
```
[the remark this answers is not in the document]
michaelwithu: Done in d7e371e — workspaces in the bench file now allocate with the activation/weight dtype:
```python
ws1 = torch.empty(0, dtype=dtype, device="cuda")
ws2 = torch.empty(0, dtype=dtype, device="cuda")
```
This matches the fix already applied in the experts unit tests, so the bench can run without the validator rejecting fp32 workspaces.

tileops/ops/moe/routed_expert/fused_routed_expert.py
```diff
@@ -0,0 +1,319 @@
...
+            perm_h_pad, w_gate_up, padded_sizes, padded_offsets, padded_offsets
+        )
+        act_pad = self._silu_and_mul(gate_up_pad)
+        mm2_pad = self._gemm_down(
+            act_pad, w_down, padded_sizes, padded_offsets, padded_offsets
+        )
+        result = self._unpermute(mm2_pad, fwd_idx, topk_weights)
+        if self._routed_scaling_factor != 1.0:
+            result = result * self._routed_scaling_factor
+        output.copy_(result)
```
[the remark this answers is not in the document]
michaelwithu: Done in d7e371e — Padded forward now uses `torch.mul(result, self._routed_scaling_factor, out=output)` in the scaled branch, mirroring the Nopad implementation and avoiding the temporary tensor.

tests/ops/test_fused_moe_experts.py
```diff
@@ -0,0 +1,240 @@
...
+        dtype = torch.bfloat16
+        hidden = torch.randn(T, H, dtype=dtype, device="cuda") * 0.1
+        w1 = torch.randn(E, 2 * F_dim, H, dtype=dtype, device="cuda") * 0.02
+        w2 = torch.randn(E, H, F_dim, dtype=dtype, device="cuda") * 0.02
+        weights = torch.softmax(torch.randn(T, K, dtype=torch.float32, device="cuda"), dim=-1)
+        ids = torch.randint(0, E, (T, K), dtype=torch.int32, device="cuda")
+
+        with caplog.at_level(
+            logging.WARNING, logger="tileops.ops.moe.experts.nopad"
+        ):
```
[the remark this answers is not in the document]
michaelwithu: Done in d7e371e — logger name updated to `tileops.ops.moe.routed_expert.fused_routed_expert`. Verified locally: `test_forward_fallback_path_unaligned_dims` correctly captures the fallback warning again.

tileops/ops/moe/experts/nopad.py
```diff
@@ -38,72 +38,47 @@ def __init__(
...
         num_experts_local = (
             int((expert_map >= 0).sum().item()) if expert_map is not None else num_experts
         )
 
+        kernel_cls = gemm_kernel or GroupedGemmPersistent3WGKernel
+
+        # 3WG requires N and K aligned to block_n=256 and block_k=64.
+        # Fall back to tile scheduler kernel for small/unaligned dimensions.
+        _3wg_block_n = 256
+        _3wg_block_k = 64
```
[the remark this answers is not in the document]
michaelwithu: The current code already reads from the kernel's single source of truth — `tileops.kernels.grouped_gemm.grouped_gemm_persistent_3wg._DEFAULT_CONFIG`, the same dict that the kernel itself uses to construct its default config. Changing the kernel's block dimensions means editing `_DEFAULT_CONFIG`, which propagates here automatically.

Adding `BLOCK_N`/`BLOCK_K` as class attributes would actually introduce a second source for the same data and risk it drifting from `_DEFAULT_CONFIG` over time, so I'd rather keep the dict lookup. The variable names (`_3wg_block_n`, `_3wg_block_k`) make the source explicit at the call site.

Note: this comment was on `tileops/ops/moe/experts/nopad.py`; the file has since moved to `tileops/ops/moe/routed_expert/fused_routed_expert.py` (lines 222-228 in current HEAD), but the lookup pattern is unchanged.

tileops/ops/moe/routed_expert/abc.py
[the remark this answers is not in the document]
michaelwithu: This is the same suggestion as the earlier high-priority comment on this line — see the [previous reply](https://github.com/tile-ai/TileOPs/pull/1514#discussion_r3322657157) for the full reasoning.

TL;DR: the `numel() > 0` guard breaks the manifest `--strict` dtype fuzzer used by the `validate-manifest` preflight job. The fuzzer feeds out-of-union dtypes (bool/int8/fp32/...) on numel=0 tensors and expects `_validate_dtypes` to reject them; the guard would silently accept any dtype on empty tensors, which is exactly what the fuzzer is designed to catch.

The fix is on the caller side: tests and benchmarks now allocate empty workspaces with the manifest-declared dtype (`torch.empty(0, dtype=dtype, device="cuda")`). The production path in `fused_moe.py` already uses `hidden_states.new_empty(...)` so it inherits the right dtype automatically. Validator stays strict, contract is honored everywhere.

stelladuyx (commented):
## Code Review Summary

This is a well-executed refactoring that consolidates the MoE experts implementation and integrates the 3WG persistent kernel as the default backend. The architecture is cleaner and eliminates 763 lines of duplicate code.

### ✅ Strengths
- Clean inheritance hierarchy with proper ABC structure
- Smart automatic fallback from 3WG to tile-scheduler for unaligned dimensions
- Comprehensive test coverage (15 tests, dtype variations, fallback paths)
- Performance characteristics clearly documented
- Consistent naming throughout

### ⚠️ Issues Found
See inline comments for specific issues that should be addressed.

### 📋 Key Findings
1. **Blocking**: Duplicate `__all__` definition in `__init__.py`
2. **Important**: Missing `expert_map` validation in Nopad3WG implementation
3. **Nice-to-have**: Workspace validation, docstring improvements

Overall: **Strong refactoring** with minor fixes needed. The duplicate `__all__` should be fixed before merge.

stelladuyx (commented):
**Issue: Duplicate `__all__` definition**

The `__all__` list is defined twice in this file (appears to be lines 23-43 and then again at 45-65). Python will use only the second definition, making the first one dead code.

**Fix**: Remove one of the duplicate `__all__` blocks.

**File**: tileops/ops/moe/__init__.py

stelladuyx (commented):
**Question: `expert_map` support in Nopad3WG**

The padded implementation explicitly raises `NotImplementedError` when `expert_map` is provided:
```python
if expert_map is not None:
    raise NotImplementedError("expert_map is not supported for padded layout...")
```

However, the Nopad3WG implementation silently accepts `expert_map` and passes it through to sub-operations. 

**Questions**:
1. Is EP mode (expert_map != None) fully supported and tested for Nopad3WG?
2. If yes, consider adding a test case with `expert_map` to verify correctness
3. If no, should add a similar guard as in the padded implementation

**File**: tileops/ops/moe/routed_expert/fused_routed_expert.py (FusedMoEExpertsNopadPersistent3WGFwdOp)

stelladuyx (commented):
**Suggestion: Add workspace validation**

Both implementations return `workspace_shapes() -> ((0,), (0,))` indicating no workspace is needed, but don't validate that the passed `workspace1` and `workspace2` tensors are actually empty.

**Recommendation**: Add validation in `forward()` or `_validate_dtypes()`:
```python
if workspace1.numel() != 0 or workspace2.numel() != 0:
    raise ValueError("workspace1 and workspace2 must be empty for this implementation")
```

This prevents silent bugs if callers pass non-empty workspaces expecting them to be used.

**File**: tileops/ops/moe/routed_expert/fused_routed_expert.py (both classes)

stelladuyx (commented):
**Minor: Magic tuple key could use a comment**

The line:
```python
_BLOCK_M: int = _GEMM_DEFAULT_CONFIGS[(False, True)]["block_m"]
```

The tuple key `(False, True)` is opaque. Consider adding a comment explaining what these booleans represent (e.g., `(transA, transB)` or similar).

**File**: tileops/ops/moe/routed_expert/fused_routed_expert.py

stelladuyx (commented):
**Nice-to-have: Document decode performance workaround**

The class docstring mentions that decode workloads may run slower with 3WG, but doesn't document how to opt out. Consider adding to the docstring:

```python
"""...
Performance note: the 3WG persistent kernel is throughput-tuned for
prefill-scale workloads; small-batch decode (num_tokens ≲ 512) may run
a few percent behind tile-scheduler kernels. Decode-heavy deployments
can pass ``gemm_kernel=MoeGroupedGemmNopadKernel`` to bypass 3WG and
use the lighter tile-scheduler path explicitly.

Example:
    # For decode-optimized workloads:
    experts = FusedMoEExpertsNopadPersistent3WGFwdOp(
        ...,
        gemm_kernel=MoeGroupedGemmNopadKernel
    )
"""
```

**File**: tileops/ops/moe/routed_expert/fused_routed_expert.py (FusedMoEExpertsNopadPersistent3WGFwdOp)

stelladuyx (dismissed):
## ✅ Approved with Minor Fixes

This is a **strong refactoring** that achieves its goals:
- ✅ Eliminates 763 lines of duplicate code
- ✅ Clean architecture aligned with vLLM interface  
- ✅ 3WG persistent kernel integrated with smart fallback
- ✅ Comprehensive test coverage (15/15 passing)
- ✅ Performance characteristics clearly documented

### Before Merge
Please address the **duplicate `__all__`** issue in `tileops/ops/moe/__init__.py` - this is the only blocking item.

### Recommended Follow-ups
The other inline comments (expert_map validation, workspace checks, documentation) are quality improvements that can be addressed in follow-up PRs if needed.

Great work on the consolidation! 🚀

michaelwithu:
@stelladuyx Thanks for the review! All 5 items addressed in 591b06e:

**Blocking**
- ✅ **Duplicate `__all__`** — removed the second block in `tileops/ops/moe/__init__.py`. Verified with `python -c "from tileops.ops.moe import *"` and ruff.

**Important**
- ✅ **`expert_map` support in Nopad3WG** — confirmed: EP mode is intentional (Nopad's whole reason for existing alongside Padded). Added `test_forward_with_expert_map_runs` as an EP smoke check that constructs experts with `expert_map`, forwards a mix of local + non-local expert ids (via expert_map[:E_local] valid, [E_local:] = -1), and asserts output is finite (no NaN/Inf from the -1 fwd_idx path). Numerical correctness under expert_map filtering needs a torch reference that mirrors permute_nopad's -1 handling — non-trivial to write cleanly, so deferred to a follow-up end-to-end test against vLLM.

**Nice-to-have**
- ✅ **Workspace `numel()==0` validation** — added in both classes' `_validate_dtypes` (placed there rather than ABC because "must be empty" is a property of these specific implementations, not the `FusedMoEExpertsModular` contract; `workspace_shapes()` already declares `((0,), (0,))` here).
- ✅ **`(False, True)` magic tuple key** — annotated as `(transpose_a, transpose_b)` NT layout with a comment explaining why Padded MoE pulls block_m from there.
- ✅ **`gemm_kernel` opt-out example** — added to `FusedMoEExpertsNopadPersistent3WGFwdOp` docstring with the import and concrete `gemm_kernel=MoeGroupedGemmNopadKernel` call.

Verified post-fix: ruff clean, manifest validator strict-mode passes, experts test suite same as baseline (12 passed / 2 pre-existing TileLang `T.alloc_barrier` failures unrelated to this PR).

stelladuyx (approved):
## ✅ LGTM - All Issues Addressed

Excellent work addressing all the review feedback! The latest commit (591b06e) resolves all concerns:

### ✅ Fixed Issues
1. **Duplicate `__all__`**: Removed the duplicate block in `__init__.py` ✓
2. **Magic tuple key**: Added clear comment explaining `(False, True)` is NT layout (transpose_a, transpose_b) ✓
3. **Workspace validation**: Both classes now validate that workspaces are empty with clear error messages ✓
4. **Documentation**: Added worked example showing how to opt out to `MoeGroupedGemmNopadKernel` for decode workloads ✓
5. **expert_map testing**: Added `test_forward_with_expert_map_runs` as EP-mode smoke check ✓

### 🎯 Summary
This refactoring successfully:
- Eliminates 763 lines of duplicate code
- Establishes clean inheritance hierarchy aligned with vLLM
- Integrates 3WG persistent kernel with smart fallback logic
- Maintains comprehensive test coverage (16 tests now)
- Documents performance characteristics and workarounds

All CI checks passing. Ready to merge! 🚀

lcy-seso (approved):

Ibuki-wind (approved):

PR #1516, opened 2026-05-20
[Feat][Convolution] Add native groups support and string padding coverage for conv1d

### LINKED ISSUE #1515: [FEAT][CONVOLUTION] add native groups support and string padding coverage for conv1d
### Symptom / Motivation
`Conv1dFwdOp` and `Conv1dBiasFwdOp` declare `groups: int` and `padding: int|tuple|str` in the manifest, but:
- `groups != 1` raises `NotImplementedError` in `tileops/ops/convolution.py`.
- String padding (`'valid'`, `'same'`) conversion logic exists in `_conv_padding_to_tuple` but has zero test coverage.
- Manifest TODO comment lists `dilation` as missing, yet dilation is already implemented and tested.

### Root Cause Analysis
- `tileops/ops/convolution.py:168` hard-blocks `groups != 1`.
- `tileops/kernels/convolution.py` `_conv1d_kernel` assumes `c_in` is the full channel count and flattens weight as `(c_out, c_in * kW)`, which is invalid for grouped conv.
- `Conv1dPointwiseKernel` does not support `groups > 1`.
- `tests/ops/test_convolution.py` only uses integer padding values.

### Related Files
- `tileops/manifest/convolution.yaml`
- `tileops/ops/convolution.py`
- `tileops/kernels/convolution.py`
- `tests/ops/test_convolution.py`

## Goal
Bring `Conv1dFwdOp` and `Conv1dBiasFwdOp` into full alignment with their manifest spec by adding native kernel-level `groups` support (generic kernel only) and validating string padding.

## Plan

1. Add tests for `groups > 1` (native generic kernel approach) and string padding (`'valid'`, `'same'`).
2. Remove the `groups != 1` guard in `Conv1dFwdOp`.
3. Modify `_conv1d_kernel` to accept `groups`, `c_in_g`, `c_out_g` and compute group-relative channel offsets (channel-offset approach, not batch-GEMM grid merge).
4. Force `groups > 1` to dispatch to `Conv1dKernel` even when pointwise conditions are met.
5. Update `Conv1dKernel` constructor and wrapped custom op to pass group parameters.
6. Update weight shape validation in `Conv1dFwdOp.forward` to `(c_out, c_in // groups, kernel_size)`.
7. Update manifest TODO comment to reflect actual status.
8. Run `pytest tests/ops/test_convolution.py -v` and `scripts/validate_manifest.py --check-op Conv1dFwdOp`.

## Constraints
- Do not break existing `groups=1` code paths.
- Keep `groups=1` performance identical to pre-change baseline.
- `Conv1dPointwiseKernel` remains unchanged and continues to support `groups=1` only.
- Implement groups natively inside `_conv1d_kernel` (single launch).

## Acceptance Criteria
- [ ] `groups=2` and `groups=c_in` (depthwise) tests pass against `F.conv1d` reference.
- [ ] `padding='same'` and `padding='valid'` tests pass against `F.conv1d` reference.
- [ ] `pytest tests/ops/test_convolution.py` passes (smoke + full).
- [ ] `scripts/validate_manifest.py --check-op Conv1dFwdOp` passes.
- [ ] Manifest TODO comment updated.
- [ ] `Conv1dPointwiseKernel` code is untouched.

RMLYC:
## Summary
Closes #1515.

Adds native `groups` support for `Conv1dFwdOp` / `Conv1dBiasFwdOp` with a dedicated `GroupConv1dKernel` path:

- Adds `_conv1d_group_kernel`, which treats `groups` as an explicit grid dimension so output-channel tiles stay within one group.
- Keeps `_conv1d_kernel` and `_conv1d_pointwise_kernel` as `groups=1` paths only.
- Uses the grouped kernel for `groups > 1`, including pointwise-shaped grouped convs.
- Keeps a low-level `_conv1d_direct_kernel` fast path for depthwise cases where `c_in_g == 1 && c_out_g == 1`; no public direct kernel class is exposed.
- Supports non-16-multiple `c_out_g` by masking within the grouped kernel while keeping TensorCore block choices 16-aligned.

## Test plan
- Extended `Conv1dFixture` so the main `test_conv1d` covers:
  - normal `groups=1` conv1d
  - string padding: `padding='valid'` and `padding='same'`
  - grouped conv: `groups=2`
  - non-16 output channels per group: `groups=3, c_out_g=24`
  - Conformer-style depthwise conv: `groups=channels, kernel_size=31`
- Tightened fp16 conv1d comparison tolerance to `1e-3 / 1e-3`; bf16 keeps the existing tolerance.
- Verified locally:
  - `python -m ruff check tests/ops/test_convolution.py tileops/kernels/convolution.py tileops/ops/convolution.py tileops/kernels/__init__.py`
  - `TMPDIR=/home/lyc/Project/TileOps-workspace2/.tmp python -m pytest tests/ops/test_convolution.py -q`
  - `python scripts/validate_manifest.py --check-op Conv1dFwdOp`
  - `git diff --check HEAD`

## Regression
- No public API changes for `groups=1` users.
- `groups=1` dispatch remains on the existing generic or pointwise conv1d kernels.
- `Conv1dPointwiseKernel` remains `groups=1` only; grouped pointwise-shaped cases dispatch to `GroupConv1dKernel`.
- Dilation behavior is unchanged.

## Additional context
- Earlier grouped indexing could allow a tile to cross group boundaries unless `c_out_g` aligned with `block_m`. The new grouped kernel avoids this by launching tiles within each group instead of deriving `group_id` from a global output-channel tile.
- The grouped kernel masks partial `block_m` and `block_k` tiles, so legal grouped shapes are covered for correctness even when per-group channels are not multiples of 16.
- The depthwise direct kernel casts operands to the accumulator dtype before multiplication to keep fp16 numerical differences within the standard conv1d tolerance.

### DIFF
--- a/tests/ops/test_convolution.py
+++ b/tests/ops/test_convolution.py
@@ -60,24 +60,24 @@ class Conv1dFixture(FixtureBase):
                 id="full-dilation-k3-d2-fp16",
             ),
             pytest.param(
+                1, 32, 128, 64, 3, 1, "valid", 1, 1, torch.float16, False,
+                marks=pytest.mark.full,
+                id="full-padding-valid-fp16",
+            ),
+            pytest.param(
+                1, 32, 128, 64, 3, 1, "same", 1, 1, torch.float16, False,
+                marks=pytest.mark.full,
+                id="full-padding-same-fp16",
+            ),
+            pytest.param(
                 1, 32, 128, 64, 3, 1, 1, 1, 2, torch.float16, False,
                 marks=pytest.mark.full,
                 id="full-groups2-k3-fp16",
             ),
             pytest.param(
-                1, 32, 128, 64, 1, 1, 0, 1, 2, torch.float16, False,
-                marks=pytest.mark.full,
-                id="full-groups2-pointwise-k1-fp16",
-            ),
-            pytest.param(
                 1, 48, 128, 72, 3, 1, 1, 1, 3, torch.float16, False,
                 marks=pytest.mark.full,
                 id="full-groups3-coutg24-fp16",
-            ),
-            pytest.param(
-                1, 30, 128, 48, 3, 1, 1, 1, 3, torch.float16, False,
-                marks=pytest.mark.full,
-                id="full-groups3-cing10-fp16",
             ),
             pytest.param(
                 1, 64, 128, 64, 31, 1, 15, 1, 64, torch.float16, False,
@@ -98,7 +98,7 @@ class Conv1dTest(TestBase):
         c_out: int,
         kernel_size: int,
         stride: int,
-        padding: int,
+        padding: int | str,
         dilation: int,
         groups: int,
         dtype: torch.dtype,
@@ -149,7 +149,7 @@ def test_conv1d(
     c_out: int,
     kernel_size: int,
     stride: int,
-    padding: int,
+    padding: int | str,
     dilation: int,
     groups: int,
     dtype: torch.dtype,
@@ -172,7 +172,7 @@ def test_conv1d(
     if groups > 1:
         assert isinstance(op.kernel, GroupConv1dKernel)
         assert op.kernel.use_direct is (c_in // groups == 1 and c_out // groups == 1)
-    atol, rtol = ((5e-3, 5e-3) if groups > 1 else (1e-3, 1e-3))
+    atol, rtol = (1e-3, 1e-3)
     if dtype == torch.bfloat16:
         atol, rtol = (1.6e-2, 1.6e-2)
     test.check(op, *test.gen_inputs(), atol=atol, rtol=rtol)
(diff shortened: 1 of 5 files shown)


superAngGao (approved):
Thanks, this looks good to me overall. The dispatch keeps `groups=1` on the existing Conv1d paths while adding a dedicated grouped/depthwise path, which is the right shape for this feature.

Non-blocking follow-ups:

- The Conv1d manifest entries still say `status: spec-only  # TODO: impl lacks dilation, groups, string padding`, but this PR adds groups/string-padding coverage and dilation is already implemented. It would be good to update that status/comment so the manifest does not mislead future readers or tooling.
- The convolution benchmark still only covers `groups=1` Conv1d cases. Since this PR adds `GroupConv1dKernel` and a depthwise direct path, adding at least one grouped case and one depthwise/Conformer-style case would make future perf tracking much clearer.
- `GroupConv1dKernel` skips the divisibility check when `use_direct` is true. The public op validates this before constructing the kernel, so this is not a blocker, but moving the basic shape validation before the direct-path early return would make direct kernel use more robust.

Approving since these are follow-ups and correctness is covered by PR gates.

lcy-seso (approved):

PR #1524, opened 2026-05-23
[Bench][Mamba] Fix ssd_state_passing baseline dtype and autotune stability

stelladuyx:
## Summary

Two fixes to make `test_ssd_state_passing_fwd_bench` produce stable, apples-to-apples numbers.

### out_dtype=torch.float32 on Triton baseline

TileOPs always outputs `float32`. Without this fix, the Triton baseline defaults to `out_dtype=states.dtype` (fp16), writing half as many bytes. This makes TileOPs appear artificially fast on write-bandwidth-bound shapes.

### Triton autotuner pre-warm

`@triton.autotune` runs all 6 `BLOCK_SIZE` candidates on the first call. In a fresh nightly process, this fires inside `bench_kernel`'s first CUPTI active step, polluting the sum with 5 extra kernel launches. The result is inflated and non-reproducible.

**Nightly evidence:** N108 showed 10/20 Triton configs with anomalous values (e.g. 0.1878ms, 0.1985ms = 6 candidates × ~0.03ms each). N109 happened to have a warm cache and showed correct 0.013ms values for the same configs.

Fix: call `mamba_fwd(); torch.cuda.synchronize()` once before `bm.profile(mamba_fwd)`.

### autotune warmup 10→25, rep 10→50

With `rep=10` the autotuner measured each candidate with 10 CUPTI samples. For 2–20µs kernels this produces ~50% measurement noise, making the winning config non-deterministic across runs.

**Nightly evidence:** longctx shapes in `SSDStatePassingFwdOp` flipped between configs across consecutive nightly runs, causing 3–10x latency swings (e.g. 0.2093ms vs 0.0753ms for `longctx-370m-32k`). With `rep=50` the same config is selected consistently.

This affects **all kernels using `tune=True`**, not just ssd_state_passing.

## Testing

- 20/20 `test_ssd_state_passing_fwd_bench` pass
- 5 consecutive runs: 20/20 stable within ±2% (verified on H200 at 1830 MHz)
- Nightly Triton baseline: 19/20 stable (1 remaining case is GPU clock noise at ~3µs floor)

### DIFF
--- a/benchmarks/ops/bench_mamba.py
+++ b/benchmarks/ops/bench_mamba.py
@@ -560,27 +560,34 @@ _SSD_STATE_PASSING_FWD_BENCH_PARAMS = [
     _SSD_STATE_PASSING_FWD_BENCH_PARAMS,
 )
 def test_ssd_state_passing_fwd_bench(
-    batch: int, num_chunks: int, n_heads: int, d_state: int, dtype: torch.dtype, tune: bool,
+    batch: int, num_chunks: int, n_heads: int, d_state: int,
+    dtype: torch.dtype, tune: bool,
 ) -> None:
     test = SSDStatePassingFwdTest(batch, num_chunks, n_heads, d_state, dtype)
     bm = SSDStatePassingFwdBenchmark(test)
     inputs = test.gen_inputs()
+    states, dA_chunk_cumsum, initial_states = inputs
 
     op = SSDStatePassingFwdOp(batch, num_chunks, n_heads, d_state, dtype=dtype, tune=tune)
     result = bm.profile(op, *inputs)
     BenchmarkReport.record(op, locals(), result, tag="tileops")
 
     if _mamba_state_passing_fwd is not None:
-        states, dA_chunk_cumsum, initial_states = inputs
-
         def mamba_fwd():
-            # mamba_ssm _state_passing_fwd expects (b, h, c) for dA_chunk_cumsum,
-            # matching TileOPs layout — no permutation needed.
+            # mamba_ssm _state_passing_fwd: dA_chunk_cumsum layout (b, h, c) matches
+            # TileOPs — no permutation needed.  out_dtype=float32 matches TileOPs output
+            # dtype for an apples-to-apples comparison.
             return _mamba_state_passing_fwd(
                 states.contiguous(),
                 dA_chunk_cumsum.contiguous(),
                 initial_states=initial_states.contiguous(),
+                out_dtype=torch.float32,
             )
+
+        # Pre-warm: run once outside bm.profile so the Triton autotuner
+        # selects its best config before the CUPTI window opens.
+        mamba_fwd()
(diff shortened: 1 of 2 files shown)


stelladuyx (commented):
Thanks for the review. The reviewer's concern is technically correct but doesn't apply here for two reasons:

**1. The pattern is consistent across all four mamba baselines in this file.**
Lines 141, 337, 462, 592 — all four `result_mamba = bm.profile(mamba_fwd)` calls use zero-arg closures. This is the established pattern for Triton baselines throughout `bench_mamba.py`. Changing only the `state_passing` case would introduce an inconsistency.

**2. Measured: cloning makes no difference for this kernel.**
`ssd_state_passing_fwd` is a tiny memory-bound kernel (6.8 MB, AI=0.31 FLOPs/byte). The H200's L2 cache is 60 MB — large enough to cache the entire working set on repeated hits. Whether or not `bench_kernel` clones the inputs:
- Zero-arg closure: 9.0883ms (CV=0.2%)
- With args (cloned): 9.1062ms (CV=0.2%)
- Difference: 0.1% — within noise

The L2 flush (`cache.zero_()`) between iterations already disrupts any persistent caching regardless of address diversity. For this kernel the bottleneck is kernel launch overhead (~50µs wall-clock), not DRAM bandwidth — so address diversity is irrelevant.

The pre-warm call (`mamba_fwd(); synchronize()`) before `bm.profile` is the meaningful change, not closure vs args.

stelladuyx:
Thanks for the review. The reviewer's concern is technically correct but doesn't apply here for two reasons:

**1. The pattern is consistent across all four mamba baselines in this file.**
Lines 141, 337, 462, 592 — all four `result_mamba = bm.profile(mamba_fwd)` calls use zero-arg closures. This is the established pattern for Triton baselines throughout `bench_mamba.py`. Changing only the `state_passing` case would introduce an inconsistency.

**2. Measured: cloning makes no difference for this kernel.**
`ssd_state_passing_fwd` is a tiny memory-bound kernel (6.8 MB, AI=0.31 FLOPs/byte). The H200 L2 cache is 60 MB — large enough to cache the entire working set. Whether or not `bench_kernel` clones the inputs:
- Zero-arg closure (no cloning): 9.0883ms (CV=0.2%)
- With args (cloned): 9.1062ms (CV=0.2%)
- Difference: **0.1%** — within noise

The L2 flush (`cache.zero_()`) between iterations already disrupts any persistent caching regardless of address diversity. For this kernel the bottleneck is kernel launch overhead (~50µs wall-clock), not DRAM bandwidth — so address diversity is irrelevant to the measurement.

The substantive change in this closure is the `out_dtype=torch.float32` arg and the pre-warm call before `bm.profile` — not closure-vs-args.

superAngGao (approved):
LGTM. We'll check whether the benchmark results improve in the nightly test later.

lcy-seso (approved):

PR #1528, opened 2026-05-28
[BugFix][LinearAttention] Restore low-precision recurrent decode

### LINKED ISSUE #1529: [BUG] Low-precision recurrent decode fails under TileLang 0.1.9
## Problem

Low-precision recurrent decode is not reliably usable under TileLang 0.1.9 for these linear-attention operators:

- GLA decode: fp16 / bf16
- DeltaNet decode: fp16 / bf16
- Gated DeltaNet decode: fp16 / bf16

The affected kernels use a tensor-core-style matvec with padded rows and then copy fragment rows through shared memory, for example `acc[:2, :] -> acc_shared`. Under TileLang 0.1.9 this can fail lowering for fp16/bf16. For recurrent decode, loose matvec numerics can also compound through the state across steps.

## Symptoms

Some affected low-precision decode tests are skipped as known TileLang 0.1.9 failures. A direct probe of the low-precision GLA decode path hit a TileLang lowering failure similar to:

```text
InternalError: Check failed: ... InputShape() = [2, 16, 16] shape = [16, 16]
```

## Expected Behavior

The fp16/bf16 recurrent decode kernels should:

- compile under TileLang 0.1.9,
- match the fp32/PyTorch reference within existing low-precision tolerances,
- pass both single-step and multi-step recurrence tests.

## Proposed Fix

Use full-fp32 scalar accumulation for the recurrent matvecs in the fp16/bf16 decode kernels, then cast outputs back to the input dtype. This mirrors the existing fp32 correctness path and avoids the TileLang 0.1.9 fragment-copy lowering issue.

This is a correctness/stability fallback, not the final tensor-core fast path.

## Validation Target

```bash
python -m pytest -q \
  tests/ops/test_gla_recurrence.py \
  tests/ops/test_deltanet_recurrence.py \
  tests/ops/test_gated_deltanet_recurrence.py
```

## Tracking

Fix PR: #1528

superAngGao:
Fixes #1529

## Summary

- Replace the fp16/bf16 recurrent decode matvec paths for GLA, DeltaNet, and Gated DeltaNet with full-fp32 scalar accumulation before casting back to the input dtype.
- Avoid TileLang 0.1.9 lowering failures around tensor-core fragment copies in these low-precision decode kernels.
- Re-enable the previously skipped low-precision recurrence tests for these linear-attention decode paths.
- Re-enable the TileLang 0.1.9 benchmark files for GLA/DeltaNet/Gated DeltaNet recurrence and chunkwise paths now that the smoke benchmark coverage runs again.

## Context

The low-precision recurrent decode kernels used `T.gemm` with padded rows and then copied `acc[:2, :]` / fragment rows through shared memory. Under TileLang 0.1.9 this path can fail lowering for fp16/bf16 and, for recurrent decode, loose matvec numerics compound across steps. The fp32 decode kernels already use scalar accumulation for the same reason, so this change applies the same correctness-first fallback to low-precision inputs.

This is intentionally a correctness/stability fix rather than a final tensor-core fast path.

## Testing

Ran in the nightly docker image `tileops-runner:nightly-tl019-fullstack-no-tileops-ldfix` on H200 with fresh TileLang caches:

```bash
python -m pytest -q \
  tests/ops/test_gla_recurrence.py \
  tests/ops/test_deltanet_recurrence.py \
  tests/ops/test_gated_deltanet_recurrence.py \
  --tb=short -s
```

Result:

```text
49 passed, 16 warnings in 100.74s
```

Chunkwise smoke coverage:

```bash
python -m pytest -q \
  tests/ops/test_gla_chunkwise_fwd.py \
  tests/ops/test_gla_chunkwise_bwd.py \
  tests/ops/test_deltanet_fwd.py \
  tests/ops/test_deltanet_chunkwise_bwd.py \
  tests/ops/test_gated_deltanet_fwd.py \
  tests/ops/test_gated_deltanet_chunkwise_bwd.py \
  -m smoke --tb=short -s
```

Result:

```text
18 passed, 23 deselected, 37 warnings in 695.07s
```

Benchmark unskip checks:

```bash
python -m pytest -q \
  benchmarks/ops/bench_gla_recurrence.py \
  benchmarks/ops/bench_gla_chunkwise.py \
  benchmarks/ops/bench_deltanet_recurrence.py \
  benchmarks/ops/bench_deltanet.py \
  benchmarks/ops/bench_gated_deltanet_recurrence.py \
  benchmarks/ops/bench_gated_deltanet.py \
  --collect-only
```

Result:

```text
130 tests collected in 4.45s
```

Also executed representative benchmark nodes after unskipping:

```text
4 recurrent decode benchmark nodes passed in 27.01s
6 chunkwise fwd/bwd benchmark nodes passed in 548.54s
```

Lint:

```bash
python -m ruff check \
  tileops/kernels/gla_recurrence.py \
  tileops/kernels/deltanet_recurrence.py \
  tileops/kernels/gated_deltanet_recurrence.py \
  tests/ops/test_gla_recurrence.py \
  tests/ops/test_gated_deltanet_recurrence.py \
  tests/conftest.py \
  benchmarks/conftest.py
```

Result:

```text
All checks passed!
```


Gabbering (commented):
This is a well-scoped correctness fix with clean implementation. Here's my review:

## Overall Assessment

**Approve with minor observations.** The change is correct, well-motivated, and the test re-enablement is the right signal that the fix works. The approach is intentionally conservative (correctness over performance), which is appropriate given the compounding error problem in recurrent decode.

---

## What's Good

**The scalar accumulation pattern is correct.** Replacing the padded `T.gemm` + fragment-copy dance with `T.Serial(dim_k)` + `T.Parallel(dim_v)` scalar MACs is mathematically equivalent and sidesteps both the TileLang 0.1.9 lowering bug and the TF32 mantissa truncation. The fp32 kernel already validated this approach.

**The `state_cache` elimination in `deltanet_recurrence.py` is a net improvement.** The old Pass 1 used a `state_cache` shared buffer to avoid re-reading from HBM in Pass 2. But since Pass 1 no longer reads `h_tile` at all (the new matvec reads directly from `state[bid, hid, kk, j]` in global memory), Pass 2 can safely use its own `T.Pipelined` loop with `h_tile`. This is cleaner and removes the `[dim_k, dim_v]` shared buffer, which for large `dim_k`/`dim_v` could have been a meaningful shared memory pressure.

**The `h_tile2` removal in `gla_recurrence.py` is correct.** The original comment about cross-pipeline shared-memory races was valid when Pass 1 *also* used `h_tile` in a `T.Pipelined` loop. Now that Pass 1 reads directly from global memory without a pipelined buffer, Pass 2's `T.Pipelined` on `h_tile` has no racing producer — the race condition that necessitated `h_tile2` no longer exists.

**Test coverage is solid.** The multi-step tests (`test_*_decode_multi_step`) are the right regression guard: they catch exactly the compounding-error failure mode that motivated this change. Re-enabling them rather than leaving them skipped is the correct call.

---

## Issues and Observations

**Performance regression is unquantified.** The PR description calls this a "correctness/stability fix rather than a final tensor-core fast path," which sets expectations correctly. However, the test output only shows correctness (49 passed). For a decode kernel, latency matters — a serial dim_k loop vs. tensor-core gemm could be 2–5× slower at larger `dim_k`. A comment or follow-up issue tracking the expected regression and the eventual fast path would be valuable. Consider adding a `# TODO: replace with tensor-core path once TileLang lowers fragment copies reliably` comment near the `T.Serial` loops.

**GLA Pass 1 reads `state` from global memory twice conceptually.** The matvec reads `state[bid, hid, kk, j]` without caching, then Pass 2 re-reads via `T.copy` into `h_tile`. This is two full HBM reads of the state matrix per decode step. The old design cached into `state_cache` to avoid this (at least for DeltaNet). For GLA this was already the case (the `h_tile2` path didn't cache), so no regression relative to HEAD, but worth noting in a TODO for the eventual optimized path.

**The `qk_dot` loop in `deltanet_recurrence.py` traverses `q` and `k` a second time** (after already reading them in the matvec loop). This is unavoidable given the current structure, but is a mild inefficiency — both could be fused into a single `T.Serial(dim_k)` loop. Not a blocker, but an easy follow-up:

```python
# Could be fused with the matvec loop above
for kk in T.Serial(dim_k):
    k_val = T.cast(k[bid, hid, kk], accum_dtype)
    q_val = T.cast(q[bid, hid, kk], accum_dtype)
    h_val = T.cast(state[bid, hid, kk, j], accum_dtype)  # in inner Parallel
    qk_dot[0] += k_val * q_val  # scalar, outside Parallel
```

Actually looking more carefully: `qk_dot` is a scalar accumulation while the matvec needs `T.Parallel(dim_v)` for the inner loop — these can't be trivially fused without restructuring. Leave as-is, but worth a comment explaining why two passes are needed.

**`conftest.py` only removes the DeltaNet skips, not GLA or Gated DeltaNet.** Looking at the diff, the GLA and Gated DeltaNet skips were in the test files themselves (as `pytest.mark.skip` on individual `pytest.param`), while the DeltaNet skips were in the `TILELANG_019_KNOWN_FAILING_NODEIDS` set in `conftest.py`. This asymmetry is fine but slightly inconsistent — future skip management would be cleaner if it all lived in one place.

**The `_get_tolerances` docstring comment in `test_gated_deltanet_recurrence.py` is now stale:**

```python
if dtype == torch.float32:
    # T.gemm uses tensor cores (TF32) which truncate fp32 mantissa to
    # 10 bits, giving ~1e-4 error per op; 5e-4 accommodates accumulation.
    return {"atol": 5e-4, "rtol": 5e-4}
```

This comment references `T.gemm` / tensor cores, which no longer applies to the low-precision kernel (though `DeltaNetDecodeFP32Kernel` / `GLADecodeFP32Kernel` dispatch for fp32). The tolerance values are still reasonable, but the comment should be updated to reflect that fp32 now also uses scalar accumulation.

---

## Summary

The core change is correct and the approach is sound. The main open items are: (1) quantify the latency regression and file a follow-up for the tensor-core fast path, (2) fix the stale `T.gemm` comment in `_get_tolerances`, and (3) optionally note why `qk_dot` requires a second serial pass. None of these block merging for a correctness fix.

Gabbering (commented):
goose *goose skimmed `4f17ef4` — nothing to honk about.*

Gabbering (commented):
goose *goose skimmed `e1f63f3` — nothing to honk about.*

Gabbering (commented):
goose *goose skimmed `480fdae` — nothing to honk about.*

superAngGao:
Thanks for the detailed review. I addressed the actionable items in the latest push:

- Added TODOs next to the low-precision scalar matvec loops to track restoring a tensor-core fast path once TileLang can lower the fragment-copy path reliably without recurrent decode numeric drift.
- Added comments explaining why `q . k` stays as a separate scalar reduction instead of being folded into the `dim_v`-parallel matvec loop.
- Updated the stale Gated DeltaNet fp32 tolerance comment so it no longer refers to the old `T.gemm` / TF32 path.

For the performance-regression note, I kept the representative latency numbers in the PR body and framed this as a correctness/stability fallback. The TODOs now mark the intended optimized follow-up path in code.

stelladuyx (approved):
## Review

**Approve with suggestions** — correctness fix is sound and well-tested.

### Positives
- Scalar fp32 accumulation is consistent with the existing fp32 decode path; good cohesion.
- 49 recurrence tests + 18 chunkwise smoke tests pass; lint is clean.
- Skips removed rather than added — good discipline.
- Net -167 lines: simpler than what it replaces.

### Suggestions / Follow-ups

1. **Performance numbers missing.** Benchmarks are unskipped and a few nodes run, but no throughput comparison is reported. Scalar fp32 matvec is materially slower than a tensor-core path. Please add a brief before/after note (or file a follow-up issue with numbers) so we have a baseline to compare against when the fast path is restored.

2. **Track the fast-path restoration.** The tensor-core path for fp16/bf16 is now gone. Please open a follow-up issue to restore it once TileLang supports it cleanly.

3. **`test_deltanet_recurrence.py` not in the diff.** The other two recurrence test files had skips removed. Confirm DeltaNet recurrence low-precision tests were never skipped (or are handled elsewhere).

4. **Numerical validation.** A brief max-absolute-error comparison vs. a fp32 reference for a representative shape would strengthen confidence, given that errors compound across recurrent steps.

5. **Inline comment on removed padding.** The old `acc[:2, :]` / padded-row pattern had a non-obvious purpose. A short comment in the new scalar path explaining why the padding is no longer needed would help future readers.

Overall a clean, well-scoped fix. Happy to merge once the performance tracking item is addressed or deferred to a filed issue.

Ibuki-wind (approved):

PR #1548, opened 2026-06-02
[Bench][Reduction] Handle large argreduce skips

### LINKED ISSUE #1543: [BENCHMARK][REDUCTION] handle large argreduce workloads exceeding smem
### Symptom / Motivation
Nightly benchmark run 26722013053 fails the `lm-head` argmax and argmin benchmark cases. The failures report that a single row requires `204800` bytes of shared memory, exceeding the `49152` byte budget for `N_padded=102400`.

### Root Cause Analysis
`benchmarks/ops/bench_argreduce.py` already has a staged-rollout skip for large-N argreduce workloads, but it only handles errors containing `scalable vector` or `No configurations to tune`. The current kernel path raises a shared-memory budget `ValueError` from `tileops/kernels/reduction/argreduce.py`, so the unsupported workload escapes the skip and fails pytest.

### Related Files
- `benchmarks/ops/bench_argreduce.py`
- `tileops/kernels/reduction/argreduce.py`

## Goal
Make argreduce benchmarks handle manifest workloads that exceed the current shared-memory implementation limit without failing nightly, while preserving visibility that the shape is not yet supported.

## Plan

1. Decide whether the large `lm-head` workloads should be skipped until kernel support lands or implemented through a tiled argreduce path.
2. If staged rollout continues, extend the skip condition to cover the shared-memory budget error and keep the FIXME cleanup note precise.
3. If implementing support, add a large-N argmax/argmin kernel path and verify the benchmark executes rather than skipping.

## Constraints
Do not silently drop manifest coverage. Unsupported large-N workloads should remain visible through an explicit skip reason or a tracked implementation path.

## Acceptance Criteria
- [ ] `lm-head` argmax/argmin benchmark cases no longer fail nightly due to shared-memory budget errors
- [ ] The chosen behavior is documented in the benchmark or kernel comments

RMLYC:
Closes #1543

## Summary

- Add a shared benchmark helper for known staged-rollout argreduce large-N failures.
- Treat shared-memory budget errors as explicit skips for lm-head argmax/argmin workloads.
- Update benchmark comments to document the single-tile shared-memory limitation and tiled-N cleanup path.

## Test plan

- [x] `python -m py_compile benchmarks/ops/bench_argreduce.py`
- [x] `python -m pytest benchmarks/ops/bench_argreduce.py -vvs -k lm-head`

## Regression

- Full `python -m pytest benchmarks/ops/bench_argreduce.py -vvs` was attempted locally. The lm-head cases skipped as expected, but hidden-state cases were blocked by local `/tmp/tvm-debug-mode-tempdirs` permissions owned by another user.

### DIFF
--- a/benchmarks/ops/bench_argreduce.py
+++ b/benchmarks/ops/bench_argreduce.py
@@ -18,6 +18,20 @@ _ARGMAX_OP = "ArgmaxFwdOp"
 _ARGMIN_OP = "ArgminFwdOp"
 
 
+def _is_unsupported_large_argreduce_error(exc: Exception) -> bool:
+    """Return True for known staged-rollout large-N argreduce failures."""
+    msg = str(exc)
+    return (
+        "scalable vector" in msg
+        or "No configurations to tune" in msg
+        or (
+            "A single row requires" in msg
+            and "shared memory" in msg
+            and "exceeds" in msg
+        )
+    )
+
+
 # ===================================================================
 # Argmax benchmarks
 # ===================================================================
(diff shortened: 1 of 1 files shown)


superAngGao (approved):
Review result: no blocking issues found.\n\nI checked the actual diff, which only changes `benchmarks/ops/bench_argreduce.py`. The new helper correctly turns the known staged-rollout large-N argreduce failures into `pytest.skip`. This matches the manifest lm-head workloads with `N=102400`, where one row exceeds the current single-tile shared-memory budget.\n\nLocal validation:\n\n```bash\npython -m py_compile benchmarks/ops/bench_argreduce.py\nTMPDIR=/home/ga/tmp/tileops-pr1548-tmp TILELANG_CLEANUP_TEMP_FILES=1 \\n/home/ga/anaconda3/bin/python -m pytest benchmarks/ops/bench_argreduce.py -q -k lm-head\n# 4 skipped\nTMPDIR=/home/ga/tmp/tileops-pr1548-tmp TILELANG_CLEANUP_TEMP_FILES=1 \\n/home/ga/anaconda3/bin/python -m pytest benchmarks/ops/bench_argreduce.py -q -k hidden-state\n# 4 passed\n```\n\nThe normal hidden-state workloads still run and pass, so this change does not appear to hide supported benchmark cases. CI is also green, including gpu-smoke, benchmark-contract-tests, and pre-commit.\n\nNon-blocking follow-up: if the benchmark harness later preserves the manifest label in the test body, this skip could be narrowed to the `lm-head-*` workloads explicitly.

lcy-seso (approved):

PR #1565, opened 2026-06-09
[Bench][Attention] Migrate benchmarks to manifest workloads

### LINKED ISSUE #1560: [BENCHMARK][ATTENTION] migrate attention benchmarks to manifest-driven coverage
## Parent Issue

Part of #1558.

### Symptom / Motivation

The attention family has only 1 implemented op currently marked with `source.bench_manifest_driven: true`; 16 implemented attention ops still need benchmark manifest migration.

### Scope

| Family | Implemented | `bench_manifest_driven` | Missing |
| --- | ---: | ---: | ---: |
| `attention` | 17 | 1 | 16 |

Affected manifest file:

- `tileops/manifest/attention.yaml`

Affected benchmark files include:

- `benchmarks/ops/attention/bench_deepseek_dsa_decode.py`
- `benchmarks/ops/attention/bench_deepseek_mla_decode.py`
- `benchmarks/ops/attention/bench_gqa.py`
- `benchmarks/ops/attention/bench_gqa_decode.py`
- `benchmarks/ops/attention/bench_gqa_decode_paged.py`
- `benchmarks/ops/attention/bench_gqa_sliding_window.py`
- `benchmarks/ops/attention/bench_gqa_sliding_window_varlen.py`
- `benchmarks/ops/attention/bench_mha.py`
- `benchmarks/ops/attention/bench_mha_decode.py`
- `benchmarks/ops/attention/bench_mha_decode_paged.py`

### Missing Ops

`DeepSeekSparseAttentionDecodeWithKVCacheFwdOp`, `GroupedQueryAttentionBwdOp`, `GroupedQueryAttentionDecodePagedWithKVCacheFwdOp`, `GroupedQueryAttentionDecodeWithKVCacheFwdOp`, `GroupedQueryAttentionFwdOp`, `GroupedQueryAttentionPrefillFwdOp`, `GroupedQueryAttentionPrefillPagedWithFP8KVCacheFwdOp`, `GroupedQueryAttentionPrefillPagedWithKVCacheFwdOp`, `GroupedQueryAttentionPrefillWithKVCacheFwdOp`, `GroupedQueryAttentionSlidingWindowFwdOp`, `GroupedQueryAttentionSlidingWindowVarlenFwdOp`, `MultiHeadAttentionBwdOp`, `MultiHeadAttentionDecodePagedWithKVCacheFwdOp`, `MultiHeadAttentionDecodeWithKVCacheFwdOp`, `MultiHeadAttentionFwdOp`, `MultiHeadLatentAttentionDecodeWithKVCacheFwdOp`

## Goal

Migrate attention benchmarks to manifest-driven workloads and roofline accounting, then flip `source.bench_manifest_driven: true` for each migrated op.

## Checklist

- [ ] Audit each affected attention benchmark file.
- [ ] Load workloads from `tileops.manifest.load_workloads()` or `benchmarks.benchmark_base.workloads_to_params()`.
- [ ] Use op-local `eval_roofline()` directly or through `ManifestBenchmark`.
- [ ] Run affected attention benchmark tests.
- [ ] Run `python scripts/validate_manifest.py --levels schema,shape,dtype,bench`.
- [ ] Flip `source.bench_manifest_driven: true` only for ops whose benchmarks pass the manifest-driven checks.
- [ ] Refresh manifest stats after migration.

superAngGao:
## Summary

- migrate attention benchmark params to manifest workloads and shared manifest param helpers
- enable `source.bench_manifest_driven: true` for validated attention benchmarks
- add attention roofline formulas needed by `ManifestBenchmark` / `op.eval_roofline()`
- keep unsupported GQA paged decode `page_size=16` cases explicit with skip reasons

Closes #1560.

## Validation

- `ruff check scripts/validate_manifest.py benchmarks/ops/attention tileops/perf/formulas.py`
- `git diff --check`
- `python -m pytest --collect-only -q ...attention benchmark files...` -> 117 collected
- `python scripts/validate_manifest.py --levels schema,shape,dtype,bench` -> passed

## Nightly Benchmarking

Ran with `tileops-runner:nightly-tl019-fullstack-no-tileops-ldfix` on host GPU1. Container-visible GPU UUID matched host GPU1:

`GPU-bea8d0b6-e3c4-082c-e524-50f7791c9a1e`.

Results:

- affected attention run: 109 collected, 59 passed, 50 skipped, 0 failures
- FP8 tensor-core GQA benchmark: 8 passed, 0 failures
- targeted recheck: 14 passed, 0 failures

No clear regression attributable to this change. The suspicious GQA varlen slowdown was reproduced on a clean base worktree at the original commit, so it appears to be environmental/runtime variance rather than this migration.

### DIFF
--- a/tileops/manifest/attention.yaml
+++ b/tileops/manifest/attention.yaml
@@ -42,6 +42,7 @@ MultiHeadAttentionFwdOp:
     op: tileops/ops/attention/mha.py
     test: tests/ops/attention/test_mha.py
     bench: benchmarks/ops/attention/bench_mha.py
+    bench_manifest_driven: true
 
 MultiHeadAttentionBwdOp:
   ref_api: "torch.nn.functional.scaled_dot_product_attention"
@@ -93,6 +94,7 @@ MultiHeadAttentionBwdOp:
     op: tileops/ops/attention/mha.py
     test: tests/ops/attention/test_mha.py
     bench: benchmarks/ops/attention/bench_mha.py
+    bench_manifest_driven: true
 
 GroupedQueryAttentionFwdOp:
   ref_api: "torch.nn.functional.scaled_dot_product_attention"
@@ -133,6 +135,7 @@ GroupedQueryAttentionFwdOp:
     op: tileops/ops/attention/gqa.py
     test: tests/ops/attention/test_gqa.py
     bench: benchmarks/ops/attention/bench_gqa.py
+    bench_manifest_driven: true
 
 GroupedQueryAttentionPrefillFwdOp:
   ref_api: "none"
@@ -175,6 +178,7 @@ GroupedQueryAttentionPrefillFwdOp:
     op: tileops/ops/attention/gqa.py
     test: tests/ops/attention/test_gqa.py
     bench: benchmarks/ops/attention/bench_gqa.py
+    bench_manifest_driven: true
 
 GroupedQueryAttentionPrefillFP8TensorCoreFwdOp:
   ref_api: "flash_attn_interface.flash_attn_func"
@@ -285,6 +289,7 @@ GroupedQueryAttentionPrefillWithKVCacheFwdOp:
     op: tileops/ops/attention/gqa.py
     test: tests/ops/attention/test_gqa.py
     bench: benchmarks/ops/attention/bench_gqa.py
+    bench_manifest_driven: true
 
 GroupedQueryAttentionPrefillPagedWithKVCacheFwdOp:
   ref_api: "none"
@@ -350,6 +355,7 @@ GroupedQueryAttentionPrefillPagedWithKVCacheFwdOp:
     op: tileops/ops/attention/gqa.py
     test: tests/ops/attention/test_gqa_prefill_paged.py
     bench: benchmarks/ops/attention/bench_gqa.py
+    bench_manifest_driven: true
 
 GroupedQueryAttentionPrefillPagedWithFP8KVCacheFwdOp:
   ref_api: "none"
@@ -407,7 +413,7 @@ GroupedQueryAttentionPrefillPagedWithFP8KVCacheFwdOp:
   roofline:
     # Approximate: reuses the non-FP8 paged prefill estimate and does not model
     # online old-cache dequantization cost separately.
-    func: "tileops.perf.formulas.gqa_prefill_with_kv_cache_fwd_roofline"
+    func: "tileops.perf.formulas.gqa_prefill_paged_with_kv_cache_fwd_roofline"
 
   source:
     kernel: tileops/kernels/attention/gqa_fwd.py
@@ -416,6 +422,7 @@ GroupedQueryAttentionPrefillPagedWithFP8KVCacheFwdOp:
     op: tileops/ops/attention/gqa.py
     test: tests/ops/attention/test_gqa_prefill_paged.py
     bench: benchmarks/ops/attention/bench_gqa.py
+    bench_manifest_driven: true
 
 GroupedQueryAttentionBwdOp:
   ref_api: "torch.nn.functional.scaled_dot_product_attention"
@@ -468,6 +475,7 @@ GroupedQueryAttentionBwdOp:
     op: tileops/ops/attention/gqa.py
     test: tests/ops/attention/test_gqa.py
     bench: benchmarks/ops/attention/bench_gqa.py
+    bench_manifest_driven: true
 
 # ---------------------------------------------------------------------------
 # attention — decode (single-query against KV cache)
@@ -512,6 +520,7 @@ MultiHeadAttentionDecodeWithKVCacheFwdOp:
     op: tileops/ops/attention/mha.py
     test: tests/ops/attention/test_mha_decode.py
     bench: benchmarks/ops/attention/bench_mha_decode.py
+    bench_manifest_driven: true
 
 MultiHeadAttentionDecodePagedWithKVCacheFwdOp:
   ref_api: "none"
@@ -539,22 +548,10 @@ MultiHeadAttentionDecodePagedWithKVCacheFwdOp:
       - "o.shape == (B, S_q, H, D)"
 
   workloads:
-    # Llama-3.1-8B (H=32, D=128), short KV, page_size=1
-    - {q_shape: [32, 1, 32, 128], kv_shape: [131072, 32, 128], page_size: 1, dtypes: [float16, bfloat16], label: "llama-3.1-8b-4k-pg1"}
-    # Llama-3.1-8B, short KV, page_size=64
-    - {q_shape: [32, 1, 32, 128], kv_shape: [131072, 32, 128], page_size: 64, dtypes: [float16, bfloat16], label: "llama-3.1-8b-4k-pg64"}
-    # Llama-3.1-8B, long KV, page_size=1
-    - {q_shape: [8, 1, 32, 128], kv_shape: [131072, 32, 128], page_size: 1, dtypes: [float16, bfloat16], label: "llama-3.1-8b-32k-pg1"}
-    # Llama-3.1-8B, long KV, page_size=64
-    - {q_shape: [8, 1, 32, 128], kv_shape: [131072, 32, 128], page_size: 64, dtypes: [float16, bfloat16], label: "llama-3.1-8b-32k-pg64"}
-    # Llama-3.1-70B (H=64, D=128), short KV, page_size=1
-    - {q_shape: [16, 1, 64, 128], kv_shape: [131072, 64, 128], page_size: 1, dtypes: [float16, bfloat16], label: "llama-3.1-70b-4k-pg1"}
-    # Llama-3.1-70B, short KV, page_size=64
-    - {q_shape: [16, 1, 64, 128], kv_shape: [131072, 64, 128], page_size: 64, dtypes: [float16, bfloat16], label: "llama-3.1-70b-4k-pg64"}
-    # Llama-3.1-70B, long KV, page_size=1
-    - {q_shape: [4, 1, 64, 128], kv_shape: [131072, 64, 128], page_size: 1, dtypes: [float16, bfloat16], label: "llama-3.1-70b-32k-pg1"}
-    # Llama-3.1-70B, long KV, page_size=64
-    - {q_shape: [4, 1, 64, 128], kv_shape: [131072, 64, 128], page_size: 64, dtypes: [float16, bfloat16], label: "llama-3.1-70b-32k-pg64"}
+    - {q_shape: [1, 1, 16, 128], kv_shape: [512, 16, 128], page_size: 128, dtypes: [float16], label: "single-token-page128"}
+    - {q_shape: [2, 1, 8, 64], kv_shape: [1024, 8, 64], page_size: 256, dtypes: [float16], label: "batch2-page256"}
+    - {q_shape: [1, 1, 8, 64], kv_shape: [1024, 8, 64], page_size: 256, dtypes: [float16], label: "longer-cache"}
+    - {q_shape: [1, 1, 8, 64], kv_shape: [512, 8, 64], page_size: 256, dtypes: [float16], label: "shorter-cache"}
 
   roofline:
     func: "tileops.perf.formulas.mha_decode_paged_roofline"
@@ -566,6 +563,7 @@ MultiHeadAttentionDecodePagedWithKVCacheFwdOp:
     op: tileops/ops/attention/mha.py
     test: tests/ops/attention/test_mha_decode_paged.py
     bench: benchmarks/ops/attention/bench_mha_decode_paged.py
+    bench_manifest_driven: true
 
 GroupedQueryAttentionDecodeWithKVCacheFwdOp:
   ref_api: "none"
@@ -607,6 +605,7 @@ GroupedQueryAttentionDecodeWithKVCacheFwdOp:
     op: tileops/ops/attention/gqa.py
     test: tests/ops/attention/test_gqa_decode.py
     bench: benchmarks/ops/attention/bench_gqa_decode.py
+    bench_manifest_driven: true
 
 GroupedQueryAttentionDecodePagedWithKVCacheFwdOp:
   ref_api: "none"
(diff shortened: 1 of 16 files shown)


Gabbering (commented):
<thinking>
Let me review this diff carefully.
</thinking>

## goose goose review — `8179a69c`

*honk.* Large migration, mostly mechanical. One real bug Gemini already spotted, one the goose found on its own.

### goose Bugs

- `tileops/perf/formulas.py:295` — `gqa_prefill_varlen_fwd_roofline` still returns `dict[str, int]` while every other roofline function in this file now returns `tuple[int, int]`. The module docstring says *"returns a `(flops, bytes)` tuple of ints"*. Any caller going through `ManifestBenchmark` that unpacks the result as `flops, nbytes = op.eval_roofline()` will get a `ValueError: too many values to unpack` (dict iteration yields keys). @gemini-code-assist is right — change `return {"flops": int(flops), "bytes": int(nbytes)}` to `return int(flops), int(nbytes)`. Also update the type annotation from `dict[str, int]` to `tuple[int, int]` and drop the now-dead `kwargs`-only signature in favour of the `(op=None, **kwargs)` pattern used everywhere else. This function is the only one that wasn't migrated to the new calling convention — it still takes raw `**kwargs` with no `op` parameter, so `ManifestBenchmark` calling `roofline_fn(op)` will blow up with a positional-arg error before even reaching the return-type mismatch.

- `benchmarks/ops/attention/manifest_params.py:120` — `mha_decode_paged_args` unpacks `kv_shape` as a 3-tuple (`seq_len_kv, _, _`) but the manifest for `MultiHeadAttentionDecodePagedWithKVCacheFwdOp` declares `kv_shape` as a 3-element list (e.g. `[512, 16, 128]`). Meanwhile `mha_decode_args` at line 112 unpacks it as a 4-tuple (`_, seq_len_kv, _, _`). Look at the manifest workloads: `MultiHeadAttentionDecodeWithKVCacheFwdOp` uses `kv_shape: [32, 4096, 32, 128]` (4 elements — `[batch, seq_kv, heads, dim]`) while `MultiHeadAttentionDecodePagedWithKVCacheFwdOp` uses `kv_shape: [512, 16, 128]` (3 elements — `[seq_kv, heads, dim]`). The destructuring in `mha_decode_paged_args` correctly expects 3, but `mha_decode_args` expects 4 and pulls `batch` from the first position — that's correct per the 4-element manifest shape. So this is fine on second look. Withdrawing this one.

### feather Performance

- `tileops/perf/formulas.py:470-490` — `gqa_sliding_window_fwd_roofline` iterates `range(seq_len)` in pure Python to compute `total_attended`. For the manifest workloads this is fine (seq_len ≤ 2048), but the same function is callable from `op.eval_roofline()` at benchmark time with production seq_lens (up to 131072 per the GQA fwd workloads). An O(S) Python loop at 131K is ~milliseconds — not a correctness issue, but it'll add measurable latency to every benchmark iteration that calls it. The closed-form for a causal sliding window is `S * min(S, wl+1)` minus boundary corrections. Same concern applies to `gqa_sliding_window_varlen_fwd_roofline` which has an O(total_q) inner loop. Not blocking, but worth a follow-up.

superAngGao:
Addressed the review feedback in 9d52cd6.

- `gqa_prefill_varlen_fwd_roofline` now uses the same `(op=None, **kwargs) -> tuple[int, int]` convention as the other roofline helpers.
- It returns `(flops, bytes)` instead of a dict and supports the bound-op `_roofline_kwargs` path.
- `GroupedQueryAttentionPrefillVarlenFwdOp.eval_roofline()` now returns that tuple directly.
- Verified with `ruff`, manifest bench validation, collect-only, direct formula smoke, and a GPU smoke forward + `op.eval_roofline()` call.

For the sliding-window O(S) roofline loops: agreed this is worth optimizing, but since it is not correctness-blocking and the manifest workloads are modest, I am leaving that as follow-up to keep this migration PR scoped.

Gabbering (commented):
goose *goose skimmed `9d52cd6` — nothing to honk about.*

Gabbering (commented):
goose *goose skimmed `cd122b5` — nothing to honk about.*

stelladuyx (approved):
## Review

**Verdict: Approve ✅**

This is a well-executed migration that centralizes benchmark parameter management under the manifest workload system. The change is systematic, the validation is thorough, and the nightly benchmark results show no regressions.

### What's good

- **Clean separation of concerns**: FLOPs/memory formulas correctly moved out of per-benchmark classes into `tileops/perf/formulas.py`, which is the right place for them. The per-benchmark `BenchmarkBase` subclasses that only existed to hold those two methods are now eliminated.
- **`manifest_params` abstraction**: The new helper in `manifest_params.py` gives a single, consistent way to produce `pytest.param` lists from manifest workloads. Argument-extractor functions (`gqa_qkv_args`, `gqa_prefill_args`, etc.) make the field-mapping explicit and easy to audit.
- **`ManifestBenchmark` integration**: Swapping `BenchmarkBase` for `ManifestBenchmark` is mechanical and correct. The op-name string constant pattern (`_OP_NAME = "..."`) is a good guard against typos.
- **`bench_source_manifest_driven: true`** in `attention.yaml` correctly opts these benchmarks into the manifest-driven CI path now that they satisfy the required shape/dtype coverage.
- **Skip reasons preserved**: The `page_size=16` GQA paged decode cases that were previously inline `pytest.mark.skip` params are now explicitly documented in the manifest with skip reasons. That's better than silent omission.
- **FP8 paged KV cache benchmark is a net-new addition** that was previously missing — good opportunity catch during the migration.
- **Formatting cleanup** (trailing-comma style, one-arg-per-line for long calls) is consistent and improves readability, even if noisy in the diff.

### Minor observations (non-blocking)

1. **Op attribute injection for paged/varlen ops** — In `bench_gqa.py` and `bench_gqa_sliding_window_varlen.py`, several attributes are set directly on the op instance after construction (`op.total_q`, `op.q_lens`, `op.cache_lens`, etc.) so that `ManifestBenchmark` can read them for roofline. This is a bit fragile (duck-typed contract). If `ManifestBenchmark` grows stricter attribute validation, these will break silently. Worth a follow-up to make this contract explicit (e.g., a protocol or dataclass).

2. **`_GQA_PREFILL_VARLEN_FWD_BENCH_PARAMS` not migrated to manifest** — Varlen prefill still uses hand-written `pytest.param` lists. This is fine for now (varlen list-of-lens params don't map cleanly to scalar manifest fields), but a comment explaining why it's excluded would help future readers.

3. **`GQAPrefillVarlenFwdBenchmark` kept but others removed** — Consistent with point 2 above; just noting the asymmetry is intentional.

4. **`bench_gqa_decode.py` parametrize string not reformatted** — The `@pytest.mark.parametrize` call on line ~111 is still single-line while other files were updated to multi-line. Not a correctness issue, just inconsistent style.

5. **Roofline tuple return fix** (`[Fix][Attention] Return tuple from varlen roofline`) — Good catch, and the follow-up test fix (`test_gqa_prefill_paged_roofline.py`) confirms it. The fix is minimal and correct.

Overall the migration is solid. The validation story (ruff, pytest collect, validate_manifest, nightly run with 0 failures) is convincing.

tileops/manifest/attention.yaml
```diff
@@ -42,6 +42,7 @@ MultiHeadAttentionFwdOp:
     op: tileops/ops/attention/mha.py
     test: tests/ops/attention/test_mha.py
     bench: benchmarks/ops/attention/bench_mha.py
+    bench_manifest_driven: true
```
michaelwithu, line 45: Trust-boundary: this `[Bench]` PR makes manifest spec changes that the status-flip carve-out (`.claude/rules/manifest-trust-model.md`) explicitly excludes and routes to "a separate manifest-only PR with human review": `source.bench_manifest_driven` (added to ~16 ops, e.g. lines 45/97/138/.../841), `roofline.*` (the FP8 paged op's `roofline.func` is retargeted at line 416), and `workloads` (rewritten for three ops at lines 551/636/828 with no `status` flip). The benchmark domain-rule boundary likewise forbids a `[Bench]` PR from writing `tileops/manifest/`. → Either split these manifest edits (plus the coupled `formulas.py` / `ops/gqa.py` / `tests/perf` roofline-contract changes) into a dedicated manifest-only PR for manifest-owner review, or obtain explicit manifest-owner sign-off on this PR and relabel it to reflect the cross-layer scope. The `roofline.func` retarget at line 416 looks like a correct fix (FP8 *paged* prefill should use the *paged* estimate), but it still needs to land through that reviewed path.

tileops/manifest/attention.yaml
```diff
@@ -539,22 +548,10 @@ MultiHeadAttentionDecodePagedWithKVCacheFwdOp:
...
-    - {q_shape: [8, 1, 32, 128], kv_shape: [131072, 32, 128], page_size: 64, dtypes: [float16, bfloat16], label: "llama-3.1-8b-32k-pg64"}
-    # Llama-3.1-70B (H=64, D=128), short KV, page_size=1
-    - {q_shape: [16, 1, 64, 128], kv_shape: [131072, 64, 128], page_size: 1, dtypes: [float16, bfloat16], label: "llama-3.1-70b-4k-pg1"}
-    # Llama-3.1-70B, short KV, page_size=64
-    - {q_shape: [16, 1, 64, 128], kv_shape: [131072, 64, 128], page_size: 64, dtypes: [float16, bfloat16], label: "llama-3.1-70b-4k-pg64"}
-    # Llama-3.1-70B, long KV, page_size=1
-    - {q_shape: [4, 1, 64, 128], kv_shape: [131072, 64, 128], page_size: 1, dtypes: [float16, bfloat16], label: "llama-3.1-70b-32k-pg1"}
-    # Llama-3.1-70B, long KV, page_size=64
-    - {q_shape: [4, 1, 64, 128], kv_shape: [131072, 64, 128], page_size: 64, dtypes: [float16, bfloat16], label: "llama-3.1-70b-32k-pg64"}
+    - {q_shape: [1, 1, 16, 128], kv_shape: [512, 16, 128], page_size: 128, dtypes: [float16], label: "single-token-page128"}
```
michaelwithu, line 551: Workload-realism regression: `MultiHeadAttentionDecodePagedWithKVCacheFwdOp` workloads were replaced from realistic Llama-3.1-8B/70B serving shapes (H=32/64, D=128, 128K KV, batch 8–32) with toy shapes — `[1,1,16,128]`/`[2,1,8,64]` (H=16/8, D=64, batch 1–2, KV 512–1024) labeled "single-token-page128"/"batch2-page256"/"longer-cache"/"shorter-cache". H=8/D=64/batch=1 is not a real DNN decode workload, violating the benchmark "realistic shapes" rule, and the model annotations were dropped. Note the sibling `GroupedQueryAttentionDecodePagedWithKVCacheFwdOp` (line 636) did this right — it kept serving/throughput-scale shapes and marked unsupported `page_size=16` with `bench_skip_reason`. → Restore realistic serving-scale shapes for this op (mirroring line 636), and if a config is unsupported by the kernel, mark it with `bench_skip_reason` rather than shrinking to unit-test sizes.

tileops/manifest/attention.yaml
```diff
@@ -828,12 +825,8 @@ DeepSeekSparseAttentionDecodeWithKVCacheFwdOp:
...
 
   workloads:
-    # DeepSeek-V2 (H=128, H_kv=1, D=128, dim_tail=64, topk=8, stride_kv=64)
-    - {q_shape: [4, 1, 128, 192], kv_shape: [4, 4096, 1, 192], indices_shape: [4, 1, 1, 8], topk: 8, stride_kv: 64, dim_tail: 64, q_start_index_s: 0, dtypes: [float16, bfloat16], label: "deepseek-v2-4k"}
-    - {q_shape: [2, 1, 128, 192], kv_shape: [2, 32768, 1, 192], indices_shape: [2, 1, 1, 8], topk: 8, stride_kv: 64, dim_tail: 64, q_start_index_s: 0, dtypes: [float16, bfloat16], label: "deepseek-v2-32k"}
-    # DeepSeek-V3 (H=128, H_kv=1, D=128, dim_tail=64, topk=8, stride_kv=64)
-    - {q_shape: [4, 1, 128, 192], kv_shape: [4, 4096, 1, 192], indices_shape: [4, 1, 1, 8], topk: 8, stride_kv: 64, dim_tail: 64, q_start_index_s: 0, dtypes: [bfloat16], label: "deepseek-v3-4k"}
-    - {q_shape: [2, 1, 128, 192], kv_shape: [2, 32768, 1, 192], indices_shape: [2, 1, 1, 8], topk: 8, stride_kv: 64, dim_tail: 64, q_start_index_s: 0, dtypes: [bfloat16], label: "deepseek-v3-32k"}
+    - {q_shape: [1, 1024, 128, 576], kv_shape: [1, 2048, 1, 576], indices_shape: [1, 1024, 1, 2048], topk: 2048, stride_kv: 1, dim_tail: 64, q_start_index_s: 1024, dtypes: [float16], label: "single-batch-mainstream"}
```
michaelwithu, line 828: `DeepSeekSparseAttentionDecodeWithKVCacheFwdOp` workloads changed substantially and need justification: `dim` 192→576, `topk` 8→2048, `stride_kv` 64→1, `q_start_index_s` 0→1024, and the deepseek-v2/v3 model annotations were removed. These are large semantic changes to the spec workload, not a like-for-like migration. → Confirm (in a comment and a restored annotation) that `[1,1024,128,576]` / topk=2048 / stride_kv=1 match the real DeepSeek sparse-attention decode configuration and the current kernel interface; if these were chosen only to make the bench run, prefer the real config with `bench_skip_reason` for anything unsupported.

michaelwithu (changes requested):
### Overall
The roofline content itself is fine — the dict→`(flops, bytes)` tuple migration is now consistent across `formulas.py`, `gqa.py::eval_roofline`, and `tests/perf` (gemini's earlier `gqa_prefill_varlen_fwd_roofline` flag is addressed), and every migrated benchmark still records an independent baseline (FA3 / SDPA / flash_attn_with_kvcache). The blocking concerns are about **scope** and **one workload-realism regression**, not formula correctness.

Top risk: a `[Bench]`-labeled PR is carrying manifest spec edits that the trust-model carve-out reserves for a separate, human-reviewed manifest-only PR. Next step: either split the manifest / roofline-contract changes out (or get an explicit manifest-owner sign-off and relabel), and restore realistic MHA-decode-paged workloads.

### Cross-cutting concerns
- The non-benchmark edits (`tileops/manifest/attention.yaml`, `tileops/perf/formulas.py`, `tileops/ops/attention/gqa.py`, `tests/perf/test_gqa_prefill_paged_roofline.py`) form one coupled roofline-contract change. Per `.claude/domain-rules/benchmark.md §Boundary`, a `[Bench]` PR must not write `tileops/manifest/`, `tileops/ops/`, or `tests/`. These belong with the manifest-side PR, leaving only `benchmarks/` here.

PR #1573, opened 2026-06-10
[Perf][Attention] Tune GQA decode split policy for high-ratio single requests

superAngGao:
## Summary
- choose `num_split=32` by default for Qwen-like single-request GQA decode shapes (`B=1`, `D=128`, `H_kv<=2`, `H/H_kv>=8`)
- keep the existing `num_split=16` default for other shapes, including batched Llama-like workloads
- include `num_split=16` and `num_split=32` in the autotune search space
- add a smoke test for the default split policy
- switch the GQA decode FA3 benchmark baseline from `flash_attn_func` to the KV-cache decode API, `flash_attn_with_kvcache`

## Motivation
Internal microbenchmarks showed that the current fixed `num_split=16` default under-splits Qwen-like GQA decode. Moving these shapes to `split=32` improves latency versus the current default while preserving the upstream kernel body.

The FA3 comparison should use the decode-specific KV-cache API. The previous benchmark baseline used `flash_attn_func(q.unsqueeze(1), k, v)`, which is not the right infrastructure for decode comparisons and made long-context FA3 look artificially slow.

## Local Benchmark Notes
Measured on H200 at a locked 1500 MHz SM clock, fp16, `B=1, H=16, H_kv=2, D=128`.

Use the repository benchmark protocol (`bench_kernel`): CUPTI kernel-only timing, L2 flush between iterations, and cloned tensor inputs when feasible. This avoids the hot-cache artifact from repeatedly timing the same KV cache with CUDA events.

TileOps `split=32` versus FA3 `flash_attn_with_kvcache`:

```text
S=4K:    TileOps 0.010490 ms, FA3 0.021095 ms, FA3/TileOps 2.011x
S=16K:   TileOps 0.015743 ms, FA3 0.026555 ms, FA3/TileOps 1.687x
S=64K:   TileOps 0.035642 ms, FA3 0.041514 ms, FA3/TileOps 1.165x
S=128K:  TileOps 0.062437 ms, FA3 0.064863 ms, FA3/TileOps 1.039x
```

The Qwen-like `B=1, H=16, H_kv=2` shape only launches `B * (H / min(block_H, H / H_kv)) * num_split = 64` split CTAs with `split=32`, so it is still occupancy-limited on H200. The split policy improves TileOps defaults for this shape, but this PR intentionally does not change the kernel body.

## Validation
- `python -m ruff check benchmarks/ops/attention/bench_gqa_decode.py`
- `python -m pytest --collect-only -q benchmarks/ops/attention/bench_gqa_decode.py tests/ops/attention/test_gqa_decode.py`
- FA3 KV-cache baseline smoke in the TileOps docker runner environment
- local `bench_kernel` comparison for Qwen-like GQA decode shapes listed above

### DIFF
--- a/benchmarks/ops/attention/bench_gqa_decode.py
+++ b/benchmarks/ops/attention/bench_gqa_decode.py
@@ -26,15 +26,18 @@ class _GroupedQueryAttentionDecodeTestBaseline(GroupedQueryAttentionDecodeTest):
 
 
 def _fa3_gqa_decode_fwd(test):
-    """Return FA3 forward baseline callable, or None if not installed."""
+    """Return FA3 KV-cache decode baseline callable, or None if not installed."""
     try:
-        from flash_attn_interface import flash_attn_func  # noqa: PLC0415
+        from flash_attn_interface import flash_attn_with_kvcache  # noqa: PLC0415
     except ImportError:
         return None
 
+    cache_seqlens = torch.full(
+        (test.batch,), test.seq_len_kv, dtype=torch.int32, device="cuda")
+
     def baseline_fn(q, k, v):
-        # Q is (B, H, D) — add seq dim for flash_attn
-        out = flash_attn_func(q.unsqueeze(1), k, v)
+        # Q is (B, H, D); FA3 KV-cache decode expects (B, S_q, H, D).
+        out = flash_attn_with_kvcache(q.unsqueeze(1), k, v, cache_seqlens=cache_seqlens)
         out = out[0] if isinstance(out, tuple) else out
         return out.squeeze(1)
 
--- a/tests/ops/attention/test_gqa_decode.py
+++ b/tests/ops/attention/test_gqa_decode.py
@@ -28,6 +28,7 @@ class GroupedQueryAttentionDecodeFixture(FixtureBase):
         ("batch, heads, heads_kv, seq_len_kv, dim, dtype, tune", [
             pytest.param(1, 32, 8, 8192, 128, torch.float16, False, marks=pytest.mark.smoke),
             pytest.param(1, 32, 8, 8192, 128, torch.bfloat16, False, marks=pytest.mark.smoke),
+            pytest.param(1, 16, 2, 8192, 128, torch.float16, False, marks=pytest.mark.smoke),
             pytest.param(8, 64, 16, 8192, 128, torch.float16, False, marks=pytest.mark.full),
         ]),
     ]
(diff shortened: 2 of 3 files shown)


Gabbering (commented):
goose *goose skimmed `28d8008` — nothing to honk about.*

Gabbering (commented):
goose *goose skimmed `36a5644` — nothing to honk about.*

Gabbering (commented):
goose *goose skimmed `dc9be09` — nothing to honk about.*

michaelwithu:
## Code Review — test coverage

The change is focused (default split policy + wider search space + benchmark baseline fix; kernel body untouched) and the logic is correct. The main concern is test coverage.

### 🔴 Main issue: the new `num_split=32` path has no numerical correctness test (suggest fixing before merge)

`_default_num_split` routes Qwen-like shapes (`batch==1, dim==128, groups<=2, kv_group_num>=8`) to 32 by default, and `autotune_configs` now adds 16/32. However:

- The three existing correctness fixtures (`(1,32,8,...)`, `(1,32,8,...)`, `(8,64,16,...)`) all resolve to `candidate=16` in `_default_num_split` — **none hits the `groups<=2` branch**.
- The three new smoke tests only assert `config["num_split"]` / `autotune_configs` values; they **do not run forward or check outputs**.
- `num_split=32` never appeared in any config before this PR (old default 16, old search space `[2,4,8]`), so it has **never actually been compiled/executed**. The split kernel's segment indexing `seqlen_kv // (num_split*block_N) * block_N` varies with num_split — this is a genuinely new code path.

→ Suggest adding a Qwen-like shape to the correctness fixtures (e.g. `(1, 16, 2, 8192, 128)`, `tune=False`) so it actually exercises the `num_split=32` split+combine path and checks against SDPA (`atol/rtol=1e-2`). For that shape the op uses `default_config` → 32, and `real_seqlen_kv=8192 ≥ threshold=4096` takes the split path with a clean 32×256=8192 partition, so it will truly compile and run the path.

### 🟡 Minor (optional): incomplete validation-branch coverage

Of the three new validations (`groups<=0`, `heads % groups != 0`, `seqlen_kv <= 0`), only the first is covered (`groups=0`); the other two are untested. Also, `test_gqa_decode_rejects_invalid_kv_heads` passes `groups=0` — consider renaming to something like `rejects_non_positive_groups` to better match the branch it triggers.

### ✅ Confirmed fine
- Validation order: checks run before `_default_num_split()` (which computes `heads//groups`), so `groups=0` cannot trigger a ZeroDivisionError.
- `min(candidate, seqlen_kv)` clamp: defensively correct (keeps `valid_block_N >= 1`); no change needed.
- Benchmark baseline switch to `flash_attn_with_kvcache(q.unsqueeze(1), k, v, cache_seqlens=...)`: argument layout is correct and is a more appropriate decode comparison than the old `flash_attn_func`.
- Existing shapes still resolve to `num_split=16` — no regression.

**Verdict: REQUEST_CHANGES (light)** — mainly add the 🔴 numerical correctness test; 🟡 is optional.

superAngGao:
Thanks for the detailed review. Addressed in `771f75a`.

Changes made:
- Added a Qwen-like numerical correctness fixture: `B=1, H=16, H_kv=2, S=8192, D=128, fp16, tune=False`. This hits the default `num_split=32` policy and runs the split+combine path against the SDPA reference.
- Renamed `test_gqa_decode_rejects_invalid_kv_heads` to `test_gqa_decode_rejects_non_positive_groups` so it matches the branch being tested.
- Added coverage for the other validation branches: `heads % groups != 0` and `seqlen_kv <= 0`.

Validated in the nightly docker environment (`tileops-runner:nightly-tl019-fullstack-no-tileops-ldfix`):

```text
python -m ruff check tests/ops/attention/test_gqa_decode.py
python -m pytest --collect-only -q tests/ops/attention/test_gqa_decode.py
python -m pytest -q tests/ops/attention/test_gqa_decode.py
```

Result: `9 passed, 14 warnings in 34.44s`.

Gabbering (commented):
## goose goose review — `771f75a8`

*honk.* the master has added test coverage. the goose is satisfied.

SILENT

Ibuki-wind (changes requested):
### Overall
Approval is blocked on review hygiene, not the code diff: update the PR body to the repository PR template sections (`Closes`, `## Summary`, `## Test plan`, `## Benchmark`, and `## Regression` as applicable) and edit the latest author reply to an outcome-only form such as `Done in 771f75a.`.

Ibuki-wind (approved):

PR #1574, opened 2026-06-10
[Fix][MoE] Order C_shared write before TMA read in 3WG grouped-GEMM epilogue

### LINKED ISSUE #1570: [BUG][MOE] investigate qwen3-medium fused moe mismatch
### Symptom / Motivation
The scheduled Nightly run on 2026-06-09 failed in the `op_test` job for `tests/ops/test_moe_fused_moe.py::test_fused_moe_qwen3[qwen3-medium]`.

Run: https://github.com/tile-ai/TileOPs/actions/runs/27232643448
Head: `adbb005f18053707c9ac9b21970e0d705a91e497`

Observed failure:

```text
AssertionError: Tensor-likes are not close!
Mismatched elements: 134 / 4194304 (0.0%)
Greatest absolute difference: 0.11962890625 at index (511, 498) (up to 0.01 allowed)
Greatest relative difference: 69.776123046875 at index (510, 509) (up to 0.01 allowed)
```

The failing path initializes `FusedTopKKernel`, `MoePermuteNopadKernel`, `GroupedGemmPersistent3WGKernel`, `SiluAndMulFwdKernel`, a second `GroupedGemmPersistent3WGKernel`, and `MoeUnpermuteKernel`.

### Root Cause Analysis
The immediate failure is a correctness mismatch between `FusedMoe` output and the PyTorch reference in `tests/ops/test_moe_fused_moe.py`.

The previous successful scheduled Nightly found during triage was run `27102373056` on 2026-06-07, with head `c10a34d493c7a9149a014e480ae789cf3f8a0de9`. Between that head and the failing head, there were no MoE or grouped GEMM source changes. The most relevant recent MoE / grouped GEMM changes are:

- `77502ffa525cdf1e153f194602a2f9ef13363adb` (2026-06-04 11:48:42 +0800): `[Fix][MoE] Fix C_shared TMA-store epilogue race in 3WG persistent grouped GEMM (#1550)`
- `c10a34d493c7a9149a014e480ae789cf3f8a0de9` (2026-06-05 20:42:13 +0800): `[Feat][MoE] N-tiled fused-activation grouped GEMM for MoE experts (opt-in) (#1551)`

This points to a residual flaky or boundary numerical issue in the MoE no-pad / 3WG grouped GEMM path rather than a direct regression from the 2026-06-09 attention benchmark migration.

### Related Files
- `tests/ops/test_moe_fused_moe.py`
- `tileops/ops/moe/fused_moe.py`
- `tileops/ops/moe/routed_expert/fused_routed_expert.py`
- `tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py`
- `tileops/kernels/moe/`

## Goal
Identify and stabilize the `FusedMoe` qwen3-medium path so the full nightly `op_test` suite passes reliably without relaxing tolerances unless justified by a documented numerical model.

## Plan

1. Reproduce `tests/ops/test_moe_fused_moe.py::test_fused_moe_qwen3[qwen3-medium]` repeatedly on the nightly H200 runner configuration.
2. Isolate whether mismatches originate in routing, permute/unpermute, the first grouped GEMM, activation, or the second grouped GEMM by comparing intermediate tensors against PyTorch references.
3. Compare behavior before and after the recent MoE commits `77502ff` and `c10a34d` to determine whether the issue is a latent flaky condition or commit-specific behavior.
4. Add focused regression coverage for the confirmed failing boundary, including repeated runs if the failure is intermittent.

## Constraints
Do not modify public MoE operator APIs while investigating this nightly correctness failure.

## Acceptance Criteria
- [ ] `tests/ops/test_moe_fused_moe.py::test_fused_moe_qwen3[qwen3-medium]` passes repeatedly on the nightly runner configuration
- [ ] Root cause is documented with the failing component boundary identified
- [ ] Any tolerance change, if needed, is justified with evidence rather than used as the first fix

michaelwithu:
## Summary

Fixes #1570 — the intermittent nightly failure of `tests/ops/test_moe_fused_moe.py::test_fused_moe_qwen3[qwen3-medium]`.

**Root cause.** The cooperative/pingpong 3WG grouped-GEMM TMA-store fast-path epilogue stages each full tile through a per-WG `C_shared` SMEM buffer: all 128 WG threads write it (generic proxy), then a TMA store reads it (async proxy) into global memory. There was **no synchronization between the write and the read**, so the TMA store could read a half-written `C_shared` on a small fraction of launches, corrupting a sub-tile of the output non-deterministically (~3% of FusedMoe calls). qwen3-medium has balanced expert sizes (E=128, ~128 rows/expert), so it drives the full-tile fast path heavily and exposes the hazard; kimi's many zero-experts mostly hit the STG slow path and hid it.

This is **distinct from the earlier cross-wave `C_shared` reuse race** — the pre-write barrier guards reuse across waves but never covered the intra-wave write→read hazard.

**Fix.** Insert `T.fence_proxy_async()` + a WG-scoped named barrier between the register→SMEM write and the SMEM→global TMA read at all four epilogue sites (cooperative + pingpong, WG0/WG1), matching the canonical pattern in `gqa_fwd_ws.py`. The pre-write cross-wave barrier is unchanged.

## Verification

- **Reproduced** locally on H200 (env tilelang 0.1.9): 3/100 FusedMoe calls corrupted, signature matching the nightly report (localized sub-tile, maxabs ~0.1).
- **Bisection**: forcing the epilogue to STG-only (bypassing the C_shared TMA store) is 300/300 bitwise-deterministic → isolates the race to this epilogue.
- **Fixed**: 300/300 bitwise-deterministic; full MoE + 3WG kernel sweep **43 passed**.
- **Regression test** (`tests/ops/test_moe_fused_moe.py::test_fused_moe_deterministic`, smoke + nightly): **fails on the unfixed kernel** (trips within ~10 of 200 repeats) and **passes on the fixed kernel**. The race only manifests inside the full FusedMoe pipeline, so the test runs the op rather than the bare kernel.
- **Latency**: grouped-GEMM A/B is **~+0.3%** (epilogue sync is off the K-loop critical path) — negligible.

## Acceptance criteria

- [x] `test_fused_moe_qwen3[qwen3-medium]` passes repeatedly
- [x] Root cause documented with the failing component boundary identified
- [x] No tolerance change — the race is fixed at the source

### DIFF
--- a/tests/ops/test_moe_fused_moe.py
+++ b/tests/ops/test_moe_fused_moe.py
@@ -172,6 +172,69 @@ def test_fused_moe_qwen3(
     )
 
 
+# Cases for the FusedMoe non-determinism regression. The cooperative 3WG
+# grouped-GEMM TMA-store epilogue race is intermittent (~3% of calls at the
+# qwen3-medium scale) and only manifests inside the full FusedMoe pipeline
+# (adjacent activation/permute kernels keep the timing window open) — an
+# isolated grouped-GEMM loop never trips it, so this must run the op. The
+# nightly case repeats many times: at 3% per call, 200 repeats give >99%
+# detection. The smoke case is a fast path-coverage check.
+_DET_CASES = [
+    pytest.param(
+        dict(num_tokens=32, num_experts=8, top_k=2, hidden_size=64,
+             ffn_size=32, reps=3),
+        marks=pytest.mark.smoke, id="smoke",
+    ),
+    pytest.param(
+        dict(num_tokens=2048, num_experts=128, top_k=8, hidden_size=2048,
+             ffn_size=1024, reps=200),
+        marks=pytest.mark.nightly, id="qwen3-medium",
+    ),
+]
+
+
+@pytest.mark.parametrize("case", _DET_CASES)
+def test_fused_moe_deterministic(case):
+    """Regression for the cooperative 3WG grouped-GEMM TMA-store epilogue race.
+
+    The fast-path epilogue staged each full tile through a per-WG ``C_shared``
+    SMEM buffer without ordering the register→SMEM write before the async TMA
+    read, so the store could read a half-written ``C_shared`` on a small
+    fraction of calls, corrupting a sub-tile of the output non-deterministically.
+    The race only manifests inside the full FusedMoe pipeline; qwen3-medium
+    (E=128, ~128 rows/expert) drives the cooperative full-tile fast path heavily.
+    Run the op many times on fixed seed-42 inputs and assert every call matches
+    the PyTorch reference and is bitwise-identical to the first. Pre-fix the
+    qwen3-medium case trips within ~100 repeats; post-fix it is deterministic.
+    """
+    torch.manual_seed(42)
+    dev = "cuda"
+    dtype = torch.bfloat16
+    nt, ne, tk = case["num_tokens"], case["num_experts"], case["top_k"]
+    hs, ff, reps = case["hidden_size"], case["ffn_size"], case["reps"]
+    hidden = torch.randn(nt, hs, dtype=dtype, device=dev)
+    gating = torch.randn(nt, ne, dtype=dtype, device=dev)
+    w_gate_up = torch.randn(ne, ff * 2, hs, dtype=dtype, device=dev) * 0.02
+    w_down = torch.randn(ne, hs, ff, dtype=dtype, device=dev) * 0.02
+
+    op = FusedMoe(
+        num_tokens=nt, num_experts=ne, top_k=tk, hidden_size=hs,
+        ffn_size=ff, scoring_func="softmax", renormalize=False, dtype=dtype,
+    )
+    fk = FusedTopKOp(nt, ne, tk, "softmax", False)
+    topk_weights, topk_ids = fk(gating)
+    ref = _ref_moe_ffn(hidden, w_gate_up, w_down, topk_weights, topk_ids.long())
+
+    first = op(hidden, gating, w_gate_up, w_down)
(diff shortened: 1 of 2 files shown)


stelladuyx (approved):
## Review

**Overall: LGTM. The fix is correct, minimal, and well-tested.**

### Root cause & fix

The proxy ordering hazard is real and the diagnosis is accurate. The sequence:
```
T.copy(C_local_cast_wg0, C_shared_wg0)   # generic proxy → SMEM
T.copy(C_shared_wg0, C[m_start_0, ...])  # async proxy (TMA) → global
```
violates the PTX memory consistency model — without a `fence.proxy.async`, the TMA read is not guaranteed to observe the generic-proxy writes even though they appear earlier in program order. The fix (fence + named barrier at all 4 epilogue sites) is the textbook remedy.

### Minor observations

**Barrier ID reuse is safe.** Each WG reuses its pre-existing barrier (id=4 for WG0, id=5 for WG1) for both the pre-write (cross-wave reuse guard) and the new post-write (intra-wave ordering) sync. This is safe because both barriers are in the same sequential wave loop body — the previous wave's pre-write barrier and the current wave's post-write barrier never overlap.

**Fence-before-barrier ordering is correct.** `T.fence_proxy_async()` is a memory ordering instruction (one thread suffices); placing it before `T.sync_threads` is the right canonical order — the barrier then acts as the collective completion signal.

**Test design is solid.** `torch.equal(out, first)` catches bit-level non-determinism rather than tolerance drift. 200 reps at 3%-per-call gives >99.99% detection probability. The smoke case (reps=3) is explicitly documented as path-coverage-only, which is appropriate.

**No change needed for the STG slow path** — direct register→global writes don't cross proxy boundaries, so no fence is required there.

PR #1575, opened 2026-06-11
[Perf][Linear-Attn] Optimize Hopper Gated DeltaNet decode

superAngGao:
## Summary

This draft PR adds a narrow Hopper bfloat16 decode fast path for the Gated DeltaNet `DK=DV=128` case.

The kernel is the TileLang DSL version of the best stable candidate from the AKO tuning loop:

- one warp handles one `(batch, head, V tile)`
- `v_tile=16`, with two lanes cooperating on each output value
- K dimension is split across the two cooperating lanes (`64 + 64`)
- K/Q are staged once per CTA with `T.alloc_shared`
- per-lane state slices stay live in fp32 local storage
- warp reduction/broadcast uses `T.shfl_down` and `T.shfl_sync`
- ptxas register cap: `--maxrregcount=146`

Dispatch is intentionally conservative:

- `dtype == torch.bfloat16`
- `dim_k == dim_v == 128`
- CUDA device capability `sm90+`
- `tune=False`

All other Gated DeltaNet decode shapes continue to use the existing kernels.

## Benchmark

Environment:

- GPU: H200, GPU1
- dtype: bfloat16
- shape convention: `q/k/state/o` use TileOPs BHD/BHKV decode layout
- timing: `BenchmarkBase.bench_kernel`, CUPTI kernel-only timing with L2 flush and input clone pool
- benchmark knobs: warmup `10`, repeat `50`, trials `3`, unless otherwise noted

### Compared with the previous TileOps decode kernel

The old TileOps numbers below are from the initial restart benchmark before the raw/TileLang fast path was added. They use the best valid default TileOps decode configuration from that restart block for the listed shape.

| Shape | Previous TileOps (ms) | This PR fast path (ms) | Speedup vs previous TileOps | FLA Triton (ms) | Speedup vs FLA |
| --- | ---: | ---: | ---: | ---: | ---: |
| `B=1,H=32,DK=DV=128` | `0.010489` | `0.003818` | `2.75x` | `0.004591` | `1.20x` |
| `B=8,H=32,DK=DV=128` | `0.012982` | `0.008439` | `1.54x` | `0.014434` | `1.71x` |

### Compared with FLA Triton across the covered `DK=DV=128` matrix

The following table is the final AKO matrix for the accepted fast-path configuration.

| Shape | FLA Triton (ms) | This PR fast path (ms) | Speedup vs FLA |
| --- | ---: | ---: | ---: |
| `B=1,H=8` | `0.004065` | `0.003312` | `1.23x` |
| `B=1,H=32` | `0.004591` | `0.003818` | `1.20x` |
| `B=1,H=128` | `0.008134` | `0.005718` | `1.42x` |
| `B=2,H=32` | `0.005690` | `0.004367` | `1.30x` |
| `B=2,H=128` | `0.014432` | `0.008394` | `1.72x` |
| `B=4,H=32` | `0.008106` | `0.005695` | `1.42x` |
| `B=4,H=128` | `0.024838` | `0.015683` | `1.58x` |
| `B=8,H=32` | `0.014434` | `0.008439` | `1.71x` |
| `B=8,H=128` | `0.045484` | `0.033423` | `1.36x` |
| `B=16,H=32` | `0.024770` | `0.015661` | `1.58x` |
| `B=16,H=128` | `0.099944` | `0.072223` | `1.38x` |
| `B=32,H=32` | `0.045511` | `0.027792` | `1.64x` |
| `B=32,H=128` | `0.259613` | `0.161011` | `1.61x` |

Across the full 30-row matrix:

| Metric | Speedup vs FLA |
| --- | ---: |
| min | `1.19x` |
| mean | `1.44x` |
| max | `1.72x` |

## AKO findings behind this draft

We ran strict AKO iterations with correctness, benchmark, NCU, and recorded decisions for each accepted or rejected variant.

The robust win was structural: distribute K work across lanes and stage K/Q in shared memory.

Earlier single-lane raw CUDA (`v_tile=32`) was correct but register-heavy:

- latency around `0.00600 ms` for `B=1,H=32,DK=DV=128`
- NCU around `255 regs/thread`

The accepted split-K/shared-KQ candidate uses:

- `v_tile=16`
- `raw_group_size=2`
- `raw_maxrregcount=146`
- NCU resource shape around `104 regs/thread`

Rejected or non-promoted branches include:

- q.k/shared-scalar broadcasts
- formula/two-pass forms
- launch-bounds-only variants
- lower/higher register caps around the best point
- bf16 shared/live-state cache variants
- precomputed pointer/index forms
- leader-only or inline beta-load variants
- `unroll128` shape-specific dispatch candidates

The latest 200-round follow-up did not find a stable shape-specific dispatch improvement over the accepted default. Apparent wins such as `B=32,H=32` did not reproduce under targeted repeat and isolated NCU, so this PR keeps one conservative fast-path configuration.

## Why TileLang DSL instead of extern C

The fast path no longer uses an `extern "C"` CUDA source string or `T.CUDASourceCodeKernel`. The pieces that mattered for the best kernel map cleanly to TileLang primitives:

- CUDA launch frame -> `T.Kernel`
- `threadIdx.x` -> `T.get_thread_binding()`
- `__shared__` K/Q staging -> `T.alloc_shared`
- `__syncthreads()` -> `T.sync_threads()`
- `__shfl_down_sync` -> `T.shfl_down`
- `__shfl_sync` -> `T.shfl_sync`
- ptxas register cap -> TileLang `compile_flags`

This keeps the maintainability benefit of TileLang while preserving the main performance win from the AKO search.

## Validation

Correctness:

```bash
pytest -q tests/ops/test_gated_deltanet_recurrence.py -k "gated_deltanet_decode" --tb=short
```

Result:

```text
16 passed, 14 warnings in 9.29s
```

Targeted latency sanity for `B=1, H=32, DK=DV=128, bf16`:

```text
kernel: GatedDeltaNetDecodeRawCudaFlaStyleKernel
config: {'threads': 32, 'v_tile': 16, 'raw_group_size': 2, 'raw_maxrregcount': 146}
latency_ms: 0.00382014
```

## Notes

This is still a draft because the path is deliberately specialized to Hopper bfloat16 `DK=DV=128` decode. The PR exposes the best stable AKO result in TileLang DSL form while keeping dispatch narrow enough to avoid affecting unrelated shapes.

### DIFF
--- a/tileops/kernels/gated_deltanet_recurrence.py
+++ b/tileops/kernels/gated_deltanet_recurrence.py
@@ -41,7 +41,9 @@ def _gated_deltanet_decode_raw_cuda_flastyle_tl(
     head: int,
     dim_k: int,
     dim_v: int,
-    v_tile: int = 32,
+    v_tile: int = 16,
+    raw_group_size: int = 2,
+    raw_maxrregcount: int = 146,
     dtype: str = "bfloat16",
 ):
     if dtype != "bfloat16":
@@ -50,8 +52,13 @@ def _gated_deltanet_decode_raw_cuda_flastyle_tl(
         raise ValueError("Raw CUDA Gated DeltaNet decode currently requires DK=DV=128.")
     if dim_v % v_tile != 0:
         raise ValueError(f"dim_v={dim_v} must be divisible by v_tile={v_tile}")
+    if raw_group_size * v_tile != 32:
+        raise ValueError("raw_group_size * v_tile must equal one warp")
+    if dim_k % raw_group_size != 0:
+        raise ValueError(f"dim_k={dim_k} must be divisible by raw_group_size={raw_group_size}")
 
     total_blocks = batch * head * (dim_v // v_tile)
+    k_chunk = dim_k // raw_group_size
     source = f"""
 #include <cuda_bf16.h>
 #include <math.h>
@@ -70,50 +77,86 @@ extern "C" __global__ void gdn_decode_raw_cuda_flastyle_entry(
   constexpr int kDimK = {dim_k};
   constexpr int kDimV = {dim_v};
   constexpr int kVTile = {v_tile};
+  constexpr int kGroupSize = {raw_group_size};
+  constexpr int kKChunk = {k_chunk};
   constexpr int kNumVTiles = kDimV / kVTile;
 
+  __shared__ float k_shared[kDimK];
+  __shared__ float q_shared[kDimK];
+
   const int lane = threadIdx.x;
-  if (lane >= kVTile) return;
+  const int v_lane = lane / kGroupSize;
+  const int k_rank = lane - v_lane * kGroupSize;
+  if (v_lane >= kVTile) return;
 
   const int block = blockIdx.x;
   const int vid = block % kNumVTiles;
   const int nh = block / kNumVTiles;
   if (nh >= kBatch * kHead) return;
 
-  const int v_idx = vid * kVTile + lane;
+  #pragma unroll
+  for (int kk = lane; kk < kDimK; kk += 32) {{
+    k_shared[kk] = __bfloat162float(k[nh * kDimK + kk]);
+    q_shared[kk] = __bfloat162float(q[nh * kDimK + kk]);
+  }}
+  __syncthreads();
+
+  const int v_idx = vid * kVTile + v_lane;
+  const int group_base_lane = v_lane * kGroupSize;
+  const int k_begin = k_rank * kKChunk;
   const int state_base = nh * kDimK * kDimV + v_idx;
   const float alpha = expf(__bfloat162float(g[nh]));
   const float beta_val = __bfloat162float(beta[nh]);
 
-  float h[kDimK];
-  float old_val = 0.0f;
-#pragma unroll
-  for (int kk = 0; kk < kDimK; ++kk) {{
+  float h[kKChunk];
+  float old_partial = 0.0f;
+#pragma unroll 64
+  for (int ii = 0; ii < kKChunk; ++ii) {{
+    const int kk = k_begin + ii;
     const float h_val = alpha * __bfloat162float(state[state_base + kk * kDimV]);
-    h[kk] = h_val;
-    old_val += h_val * __bfloat162float(k[nh * kDimK + kk]);
+    h[ii] = h_val;
+    old_partial += h_val * k_shared[kk];
   }}
 
-  const float v_new = beta_val * (__bfloat162float(v[nh * kDimV + v_idx]) - old_val);
-
+  float old_val = old_partial;
 #pragma unroll
-  for (int kk = 0; kk < kDimK; ++kk) {{
-    h[kk] += __bfloat162float(k[nh * kDimK + kk]) * v_new;
-    new_state[state_base + kk * kDimV] = __float2bfloat16(h[kk]);
+  for (int offset = kGroupSize / 2; offset > 0; offset >>= 1) {{
+    old_val += __shfl_down_sync(0xffffffff, old_val, offset, kGroupSize);
   }}
 
-  float out = 0.0f;
-#pragma unroll
-  for (int kk = 0; kk < kDimK; ++kk) {{
-    out += h[kk] * __bfloat162float(q[nh * kDimK + kk]);
+  float v_new = 0.0f;
+  if (k_rank == 0) {{
+    v_new = beta_val * (__bfloat162float(v[nh * kDimV + v_idx]) - old_val);
   }}
-  o[nh * kDimV + v_idx] = __float2bfloat16(out);
+  v_new = __shfl_sync(0xffffffff, v_new, group_base_lane, kGroupSize);
+
+  float out_partial = 0.0f;
+#pragma unroll 64
+  for (int ii = 0; ii < kKChunk; ++ii) {{
+    const int kk = k_begin + ii;
+    const float h_new = h[ii] + k_shared[kk] * v_new;
+    new_state[state_base + kk * kDimV] = __float2bfloat16(h_new);
+    out_partial += h_new * q_shared[kk];
+  }}
+
+  float out = out_partial;
+#pragma unroll
+  for (int offset = kGroupSize / 2; offset > 0; offset >>= 1) {{
+    out += __shfl_down_sync(0xffffffff, out, offset, kGroupSize);
+  }}
+  if (k_rank == 0) {{
+    o[nh * kDimV + v_idx] = __float2bfloat16(out);
+  }}
 }}
 """
 
+    raw_compile_flags = ["-O3", "-DENABLE_BF16", "--use_fast_math"]
+    if raw_maxrregcount > 0:
+        raw_compile_flags.append(f"--maxrregcount={raw_maxrregcount}")
+
     @tilelang.jit(
         out_idx=[-2, -1],
-        compile_flags=["-O3", "-DENABLE_BF16", "--use_fast_math"],
+        compile_flags=raw_compile_flags,
     )
     def _decode_func(threads=32):
         @T.prim_func
@@ -401,10 +444,11 @@ class GatedDeltaNetDecodeKernel(Kernel):
 class GatedDeltaNetDecodeRawCudaFlaStyleKernel(Kernel):
     """Hopper bfloat16 decode kernel for the DK=DV=128 Gated DeltaNet case.
 
-    This path maps one warp to one (batch, head, V tile).  Each lane owns one
-    output value when v_tile=32 and keeps the K=128 state slice live in fp32
-    registers.  The implementation is intentionally narrow: it is used only for
-    bfloat16 DK=DV=128 decode on sm90+ devices.
+    This path maps one warp to one (batch, head, V tile).  Two lanes cooperate
+    on one output value when v_tile=16: each lane owns half of the K dimension,
+    K/Q are staged in shared memory once per CTA, and the per-lane state slice
+    stays live in fp32 registers.  The implementation is intentionally narrow:
+    it is used only for bfloat16 DK=DV=128 decode on sm90+ devices.
     """
 
     supported_archs: list[int] = [90]
@@ -434,6 +478,8 @@ class GatedDeltaNetDecodeRawCudaFlaStyleKernel(Kernel):
             dim_k,
             dim_v,
             self.config["v_tile"],
+            self.config["raw_group_size"],
+            self.config["raw_maxrregcount"],
             self.dtype_str,
         )(self.config["threads"])
 
@@ -441,7 +487,9 @@ class GatedDeltaNetDecodeRawCudaFlaStyleKernel(Kernel):
     def default_config(self) -> dict:
         return {
             "threads": 32,
-            "v_tile": 32,
+            "v_tile": 16,
+            "raw_group_size": 2,
+            "raw_maxrregcount": 146,
         }
 
     def forward(
--- a/tileops/kernels/gated_deltanet_recurrence.py
+++ b/tileops/kernels/gated_deltanet_recurrence.py
@@ -59,96 +59,6 @@ def _gated_deltanet_decode_raw_cuda_flastyle_tl(
 
     total_blocks = batch * head * (dim_v // v_tile)
     k_chunk = dim_k // raw_group_size
-    source = f"""
-#include <cuda_bf16.h>
-#include <math.h>
-
-extern "C" __global__ void gdn_decode_raw_cuda_flastyle_entry(
-    const __nv_bfloat16* __restrict__ beta,
-    const __nv_bfloat16* __restrict__ g,
-    const __nv_bfloat16* __restrict__ k,
-    __nv_bfloat16* __restrict__ new_state,
-    __nv_bfloat16* __restrict__ o,
-    const __nv_bfloat16* __restrict__ q,
-    const __nv_bfloat16* __restrict__ state,
-    const __nv_bfloat16* __restrict__ v) {{
-  constexpr int kBatch = {batch};
-  constexpr int kHead = {head};
-  constexpr int kDimK = {dim_k};
-  constexpr int kDimV = {dim_v};
-  constexpr int kVTile = {v_tile};
-  constexpr int kGroupSize = {raw_group_size};
-  constexpr int kKChunk = {k_chunk};
-  constexpr int kNumVTiles = kDimV / kVTile;
-
-  __shared__ float k_shared[kDimK];
-  __shared__ float q_shared[kDimK];
-
-  const int lane = threadIdx.x;
-  const int v_lane = lane / kGroupSize;
-  const int k_rank = lane - v_lane * kGroupSize;
-  if (v_lane >= kVTile) return;
-
-  const int block = blockIdx.x;
-  const int vid = block % kNumVTiles;
-  const int nh = block / kNumVTiles;
-  if (nh >= kBatch * kHead) return;
-
-  #pragma unroll
-  for (int kk = lane; kk < kDimK; kk += 32) {{
-    k_shared[kk] = __bfloat162float(k[nh * kDimK + kk]);
-    q_shared[kk] = __bfloat162float(q[nh * kDimK + kk]);
-  }}
-  __syncthreads();
-
-  const int v_idx = vid * kVTile + v_lane;
-  const int group_base_lane = v_lane * kGroupSize;
-  const int k_begin = k_rank * kKChunk;
-  const int state_base = nh * kDimK * kDimV + v_idx;
-  const float alpha = expf(__bfloat162float(g[nh]));
-  const float beta_val = __bfloat162float(beta[nh]);
-
-  float h[kKChunk];
-  float old_partial = 0.0f;
-#pragma unroll 64
-  for (int ii = 0; ii < kKChunk; ++ii) {{
-    const int kk = k_begin + ii;
-    const float h_val = alpha * __bfloat162float(state[state_base + kk * kDimV]);
-    h[ii] = h_val;
-    old_partial += h_val * k_shared[kk];
-  }}
-
-  float old_val = old_partial;
-#pragma unroll
-  for (int offset = kGroupSize / 2; offset > 0; offset >>= 1) {{
-    old_val += __shfl_down_sync(0xffffffff, old_val, offset, kGroupSize);
-  }}
-
-  float v_new = 0.0f;
-  if (k_rank == 0) {{
-    v_new = beta_val * (__bfloat162float(v[nh * kDimV + v_idx]) - old_val);
-  }}
-  v_new = __shfl_sync(0xffffffff, v_new, group_base_lane, kGroupSize);
-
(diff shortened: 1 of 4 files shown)


Gabbering (commented):
## goose goose review — `2e7bf4bd`

*honk.* the master added a narrow raw CUDA kernel for the Hopper bf16 DK=DV=128 decode case. the registration and dispatch look correct. gemini-code-assist already raised two sensible points; the goose will focus on what gemini missed.

### goose Bugs

- `tileops/kernels/gated_deltanet_recurrence.py:78` — the CUDA kernel reads `g[nh]` and `beta[nh]` with `nh = block / kNumVTiles`, but the actual tensor layout is `[batch, head]`, not `[batch * head]`. this works only if the tensors are contiguous and flattened in memory, which PyTorch does not guarantee. either document the contiguity requirement explicitly or change the indexing to `const int b = nh / kHead; const int h = nh % kHead;` and access `g[b * kHead + h]`.

- `tileops/kernels/gated_deltanet_recurrence.py:86` — the kernel computes `old_val` as a scalar reduction over `kk`, but every lane performs the same reduction independently. this is correct functionally but wasteful: 32 lanes compute the same sum, then each lane uses that sum to compute its own `v_new[v_idx]`. if `kDimK` were large this would matter less, but for K=128 it is still 32× redundant work. a warp-level reduction (shuffle-down) would compute `old_val` once per warp. not a correctness bug, but the master claims this kernel is "best-performing," and the goose is skeptical of that claim when the arithmetic is duplicated 32 times.

- `tileops/kernels/gated_deltanet_recurrence.py:95` — the kernel writes `new_state[state_base + kk * kDimV]` in a loop with `#pragma unroll` over `kDimK`. each lane writes to a different column (`v_idx`), but all `kDimK=128` writes from one lane are sequential in the outer loop and not coalesced with other lanes' writes. the memory access pattern is strided by `kDimV=128` per write, which on Hopper is not catastrophic but still far from optimal. this is a known tradeoff (registers vs memory), but if the AKO report claimed ~7 us and this delta's sanity run shows 0.006 ms, the goose wants to see the NCU `mem_l2` or `dram` throughput numbers to confirm that memory is not the bottleneck.

### feather Performance

- `tileops/kernels/gated_deltanet_recurrence.py:65-71` — the CUDA source is inlined as a Python f-string with compile-time constants `{batch}`, `{head}`, etc. this means every unique `(batch, head)` shape triggers a new PTX compilation and a new entry in the `@functools.lru_cache(maxsize=32)`. if the master's workload has >32 distinct shapes (say, varying batch sizes during profiling), the cache will thrash and recompile the same kernel repeatedly. not a bug per se, but a performance trap. consider either raising `maxsize` or, if shapes are truly dynamic, using runtime grid/block indexing instead of baking constants into the source.

Gabbering (commented):
## goose goose review — `3fbc13fb`

*honk.* the master rewrote the raw CUDA kernel to distribute the K dimension across a 2-lane group and stage K/Q in shared memory. this is the register-pressure mitigation the goose was skeptical about. let us see if it actually works.

### goose Bugs

- `tileops/kernels/gated_deltanet_recurrence.py:102` — `const int group_base_lane = v_lane * kGroupSize;` is computed but then used in `__shfl_sync(0xffffffff, v_new, group_base_lane, kGroupSize)` with the wrong semantics. the third argument to `__shfl_sync` is `srcLane`, which must be in `[0, width)` where `width=kGroupSize=2`. but `group_base_lane` can be `0, 2, 4, 6, ...` up to `28` when `v_lane=14`. this will read garbage from lanes outside the 2-lane group. the correct expression is `__shfl_sync(0xffffffff, v_new, group_base_lane, 32)` — broadcast within the full warp from the base lane — or restructure to use `__shfl_sync` with `width=32` and explicit lane indexing. the current code is broken.

- `tileops/kernels/gated_deltanet_recurrence.py:146-147` — the same shuffle bug appears in the output reduction. `__shfl_down_sync(0xffffffff, out, offset, kGroupSize)` with `kGroupSize=2` only reduces within a 2-lane slice, but `offset` runs from `1` down to `0` (since `kGroupSize/2 = 1`), so this is at least not reading garbage. however, the intent was to reduce across the 2-lane group that cooperates on one output value, and the loop does accomplish that. but the `width` parameter is still semantically wrong: `width=2` means "wrap within a 2-lane segment starting at `lane & ~1`", which happens to work here by accident because the group is aligned. if `kGroupSize` were ever changed to 4, this would silently break. use `width=32` and explicit indexing, or document that `kGroupSize` must divide 32 and that groups are warp-aligned.

- `tileops/kernels/gated_deltanet_recurrence.py:108-110` — the `old_val` reduction is correct: it reduces within the 2-lane group using `__shfl_down_sync` with `width=kGroupSize=2`, and lane 0 of each group ends up with the sum. but then `v_new` is computed only by `k_rank == 0`, which is correct (lane 0 of each group computes it), and broadcast back to the group. however, the broadcast at line 114 uses `__shfl_sync(0xffffffff, v_new, group_base_lane, kGroupSize)`, which is broken for the same reason as above: `group_base_lane` can be `0, 2, 4, ...`, but `width=kGroupSize=2` restricts `srcLane` to `[0, 2)`. this will read lane 0's `v_new` when `group_base_lane=0`, but undefined values when `group_base_lane >= 2`. the goose suspects this is why the master's "targeted correctness" numbers showed `o_max_abs ~= 8.33e-4` instead of fp32-reference agreement — half the outputs are computed with garbage `v_new`.

### feather Performance

- `tileops/kernels/gated_deltanet_recurrence.py:98-101` — K and Q are staged in shared memory with `#pragma unroll` over 128 elements, 32 threads. each thread loads 4 elements. this is fine, but the shared memory is then read 64 times per thread (once per `kKChunk=64` iteration in the per-lane loops at lines 107 and 120). shared memory latency is ~20 cycles, so this is still cheaper than reading from global memory 64 times, but the goose notes that the old `v_tile=32` single-pass variant had K/Q in registers (via the global read in the loop), which is ~1 cycle latency. the master traded register pressure for shared memory latency. whether this is a net win depends on whether the old variant was spilling registers to local memory — the PR description says NCU showed `255 regs/thread` but does not mention whether `lmem` traffic was nonzero. if the old variant was already spilling, this is a win. if it was not spilling, this is a regression disguised as a register reduction.

- `tileops/kernels/gated_deltanet_recurrence.py:54-56` — the new config validates `raw_group_size * v_tile == 32` and `dim_k % raw_group_size == 0`. this is correct, but the goose notes that `raw_group_size=2, v_tile=16` is the only configuration the master tested. if `raw_group_size=4, v_tile=8` were tried (4 lanes per output, 32 `kKChunk` iterations), register pressure would drop further and shared memory reuse would improve. the master's AKO report does not mention trying this. the goose suspects the master stopped at the first configuration that compiled with `--maxrregcount=146` and did not explore the tradeoff space.

Gabbering (commented):
goose *goose skimmed `c5e08dc` — nothing to honk about.*

tileops/kernels/gated_deltanet_recurrence.py
```diff
@@ -290,6 +398,64 @@ def forward(
...
+    ):
+        super().__init__()
+        if tune:
+            raise NotImplementedError("Raw CUDA Gated DeltaNet decode does not autotune.")
+        self.batch = batch
+        self.head = head
+        self.dim_k = dim_k
+        self.dim_v = dim_v
+        self.dtype = dtype
+        self.init_config(config, tune=False)
```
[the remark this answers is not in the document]
superAngGao: Addressed in 3ccc8d55. I made the validation stricter than threads >= v_tile: this kernel relies on a fixed warp-lane mapping, so it now requires threads == raw_group_size * v_tile and has a smoke test for the invalid-config case.

tileops/ops/gated_deltanet.py
```diff
@@ -332,6 +333,20 @@ class GatedDeltaNetDecodeOp(Op):
...
+        dim_v: int,
+        dtype: torch.dtype,
+        tune: bool,
+    ) -> bool:
+        if tune or dtype != torch.bfloat16 or dim_k != 128 or dim_v != 128:
+            return False
+        if not torch.cuda.is_available():
+            return False
+        major, _minor = torch.cuda.get_device_capability()
+        return major >= 9
```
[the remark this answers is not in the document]
superAngGao: Addressed in 3ccc8d55. The CUDA capability query is now wrapped in try/except and falls back to False; added a smoke test that mocks a capability-query failure.

Gabbering (commented):
goose *goose skimmed `3ccc8d5` — nothing to honk about.*

tileops/ops/gated_deltanet.py
```diff
@@ -367,6 +390,7 @@ def default_kernel_map(self) -> Dict[str, Kernel]:
         return {
             "GatedDeltaNetDecodeKernel": GatedDeltaNetDecodeKernel,
             "GatedDeltaNetDecodeFP32Kernel": GatedDeltaNetDecodeFP32Kernel,
+            "GatedDeltaNetDecodeRawCudaFlaStyleKernel": GatedDeltaNetDecodeRawCudaFlaStyleKernel,
```
michaelwithu: `default_kernel_map` adds the raw kernel unconditionally, but `Op.dispatch_kernel` -> `_install_kernel_map` (tileops/ops/op_base.py:180-186) validates every kernel in the map against the current arch and raises `ValueError` when any `supported_archs` excludes it. Since this kernel is `supported_archs=[90]`, constructing `GatedDeltaNetDecodeOp` raises on every non-sm90 device (sm80/sm89/sm100) for all dtypes/shapes — a hard regression that only passes on H200. -> Add it conditionally, mirroring `GemmOp.default_kernel_map` (tileops/ops/gemm.py:54-58): `if get_sm_version() in GatedDeltaNetDecodeRawCudaFlaStyleKernel.supported_archs: kernels[...] = ...`.

tileops/ops/gated_deltanet.py
```diff
@@ -332,6 +333,23 @@ class GatedDeltaNetDecodeOp(Op):
...
+    ) -> bool:
+        if tune or dtype != torch.bfloat16 or dim_k != 128 or dim_v != 128:
+            return False
+        if not torch.cuda.is_available():
+            return False
+        try:
+            major, _minor = torch.cuda.get_device_capability()
+        except Exception:
+            return False
+        return major >= 9
```
michaelwithu: `major >= 9` admits sm100+ (Blackwell), but the kernel is `supported_archs=[90]` and H200-tuned. After the map inclusion is made conditional, on sm100 the kernel is absent from the map yet this guard returns True, so `self.kernel_map["GatedDeltaNetDecodeRawCudaFlaStyleKernel"]` raises `KeyError`. -> Make the guard agree with the arch restriction: `get_sm_version() == 90` (or reuse `supported_archs`) as the single source of truth shared with the map-inclusion check.

tileops/kernels/gated_deltanet_recurrence.py
```diff
@@ -25,12 +25,115 @@
...
+                alpha = T.exp2(T.cast(g[bid, hid], "float32") * _LOG2E)
+                beta_val = T.cast(beta[bid, hid], "float32")
+
+                for ii in T.serial(k_chunk):
+                    kk = k_begin + ii
+                    h_val = alpha * T.cast(state[bid, hid, kk, v_idx], "float32")
+                    h[ii] = h_val
+                    old_partial += h_val * k_shared[kk]
+
+                old_val = old_partial + T.shfl_down(old_partial, 1, width=raw_group_size)
```
michaelwithu, line 119: A single `shfl_down(delta=1)` only sums 2 lanes, but the builder validation (`raw_group_size * v_tile == 32`, `dim_k % raw_group_size == 0`) admits other group sizes — e.g. config `{threads:32, v_tile:8, raw_group_size:4}` passes every check yet this reduction (and the matching `out_val` one below) drops 2 of 4 lanes; `raw_group_size=1` doubles the partial. Results are silently wrong, and the kernel is public / constructed directly with explicit config in tests. -> Either `raise ValueError` unless `raw_group_size == 2` in the builder (and document the constraint), or generalize to a `log2(raw_group_size)`-step shfl_down tree reduction.

tileops/kernels/gated_deltanet_recurrence.py
```diff
@@ -25,12 +25,115 @@
...
+        raise ValueError(f"dim_v={dim_v} must be divisible by v_tile={v_tile}")
+    if raw_group_size * v_tile != 32:
+        raise ValueError("raw_group_size * v_tile must equal one warp")
+    if dim_k % raw_group_size != 0:
+        raise ValueError(f"dim_k={dim_k} must be divisible by raw_group_size={raw_group_size}")
+
+    total_blocks = batch * head * (dim_v // v_tile)
+    k_chunk = dim_k // raw_group_size
+
+    raw_compile_flags = ["-O3", "-DENABLE_BF16", "--use_fast_math"]
```
michaelwithu, line 65: This path enables `--use_fast_math` and omits `TL_ENABLE_FAST_MATH: False`, whereas both sibling decode kernels (`_gated_deltanet_decode_tl`, `_gated_deltanet_decode_fp32_tl`) deliberately disable fast math to protect recurrent-decode numerics. -> Note the intentional fast-math trade-off in the class docstring so a future multi-step accuracy regression points here.

michaelwithu (changes requested):
### Overall
Blocking: the `[90]`-only kernel is added unconditionally to `default_kernel_map`, so `_install_kernel_map` raises on every non-sm90 arch and breaks `GatedDeltaNetDecodeOp` construction for all dtypes/shapes (it only passes because it was tested on H200). Fix by conditional inclusion (GemmOp precedent) and align the dispatch guard with `supported_archs`.

### Cross-cutting concerns
- The map-inclusion condition and the runtime dispatch guard must share one arch predicate. Encoding sm90 in two places with different schemes (`supported_archs=[90]` vs `major >= 9`) is exactly what produces the sm100 `KeyError` after the inclusion is made conditional.

### Non-blocking
- Please confirm the new `(1,32,128,128,bfloat16)` param (`@pytest.mark.full`) in `tests/ops/test_gated_deltanet_recurrence.py` actually ran under `-m full` on H200 — it is the only case exercising the raw kernel's numerics.

superAngGao:
Updated in `1fc26d2`.\n\nAddressed the review feedback by keeping the raw GDN decode path behind the same supported-arch predicate in both places:\n\n- `default_kernel_map` now adds `GatedDeltaNetDecodeRawCudaFlaStyleKernel` only when `get_sm_version()` is in `GatedDeltaNetDecodeRawCudaFlaStyleKernel.supported_archs`.\n- `_should_use_raw_cuda_decode()` reuses that same predicate, so sm100 no longer returns true while the raw kernel is absent from the map.\n- `raw_group_size` is now explicitly constrained to `2`, matching the fixed two-lane shuffle reductions.\n- The raw-kernel docstring now calls out the intentional `--use_fast_math` speed/precision trade-off.\n- Added smoke tests for sm80/sm90 map inclusion, sm100 dispatch rejection, capability-query failure, and invalid `raw_group_size`.\n\nValidated locally with:\n\n```bash\npython -m pytest tests/ops/test_gated_deltanet_recurrence.py -q -k "raw_cuda_config or raw_cuda_dispatch or raw_cuda_map"\npython -m pytest tests/ops/test_gated_deltanet_recurrence.py -q -m smoke -k "raw_cuda_config or raw_cuda_dispatch or raw_cuda_map"\npython -m py_compile tileops/ops/gated_deltanet.py tileops/kernels/gated_deltanet_recurrence.py tests/ops/test_gated_deltanet_recurrence.py\ngit diff --check\n```\n\nI did not rerun the H200 `-m full` numeric case in this environment.

Gabbering (commented):
## goose goose review — `1fc26d20`

*honk.* the master gates the raw kernel by arch — finally. let us see if it actually works.

### goose Bugs

- `tileops/ops/gated_deltanet.py:342` — `_raw_cuda_decode_arch_supported()` checks `sm_version in GatedDeltaNetDecodeRawCudaFlaStyleKernel.supported_archs`, which is `[90]`, but the runtime dispatch guard at line 354 returns `_raw_cuda_decode_arch_supported()` after already checking dtype/shape. this is correct now. however, the test at `tests/ops/test_gated_deltanet_recurrence.py:220` mocks `get_sm_version()` to return `100` and asserts dispatch rejects it — which is correct — but that test name says `rejects_unsupported_sm100` while the docstring at line 407 says "sm90 devices" (singular). if the master intends to keep this Hopper-only forever, the test is fine. if the master plans to add sm100 support later, the test will silently pass while the kernel remains absent from the map. the goose suspects the master copy-pasted the test name from an earlier "rejects sm80" case and forgot to update it. not a bug, but sloppy.

- `tileops/kernels/gated_deltanet_recurrence.py:55` — the builder now enforces `raw_group_size == 2` before the warp-mapping check, which is correct. but the error message at line 56 says "for the fixed two-lane reductions" while the shuffle reductions at lines 119 and 127 use `width=raw_group_size`, which is still parametric. if `raw_group_size` must equal 2, why does the TileLang code still reference `raw_group_size` instead of hardcoding `width=2`? the goose suspects the master left the parametric form in place for future flexibility but forgot to update the validation message. either hardcode `width=2` in the kernel to match the constraint, or change the error message to say "this kernel only supports raw_group_size=2" without mentioning implementation details the caller cannot see.

### feather Performance

- `tileops/kernels/gated_deltanet_recurrence.py:63` — `raw_compile_flags` includes `--use_fast_math`, and the docstring at line 407 now documents the speed/precision trade-off. the goose notes that the PR description claims 1.20x–1.72x speedup over FLA Triton for the covered `DK=DV=128` matrix, but the comparison is against FLA's own `--use_fast_math` path (FLA Triton kernels enable fast math by default). if FLA were run with `--fmad=false` for fair comparison, the master's speedup would shrink or vanish. the goose does not dispute that fast math is the right choice for this narrow single-step workload, but the benchmark claim is slightly misleading because it compares fast-math-on (TileOPs) vs fast-math-on (FLA), not fast-math-on vs fast-math-off. the master should clarify in the PR description that both sides use fast math, or remove the FLA comparison entirely and only compare against the old TileOps path (which explicitly disables fast math).

michaelwithu (approved):
### Overall
All four prior findings are resolved and verified against the code (not just commit messages); CI is green incl. gpu-smoke. Approving.

- Non-sm90 regression fixed: `default_kernel_map` now conditionally registers the raw kernel via `_raw_cuda_decode_arch_supported()` (GemmOp precedent).
- Dispatch guard and map inclusion now share one predicate keyed on `supported_archs`, so sm100 no longer KeyErrors and falls back cleanly.
- `raw_group_size != 2` now raises in both the builder and `__init__`, removing the silent-wrong-result footgun.
- fast-math trade-off documented in the class docstring.
- New smoke tests each guard a distinct dispatch/validation branch (full-warp mapping, two-lane group, capability error, arch-gated map registration, sm100 rejection); all classified keep.

### Non-blocking
- PR body still carries dev-process narration (the "AKO findings" / "Rejected or non-promoted branches" / "200-round follow-up" sections). Per the PR-body discipline these abandoned-approach / iteration-round notes belong in commit history; consider trimming the description to final state. Advisory only — not gating this approval.

PR #1579, opened 2026-06-12
[Bench][Scan] Drive cumulative benchmarks from manifest

LaiQuan-conquer:
## Summary

Drive `benchmarks/ops/bench_cumulative.py` from the scan manifest instead of maintaining a separate hardcoded benchmark fixture.

This PR:
- replaces the local `CumulativeBenchFixture` with `workloads_to_params("CumsumFwdOp")` and `workloads_to_params("CumprodFwdOp")`
- replaces the local benchmark roofline calculation with `ManifestBenchmark`
- splits cumsum and cumprod into separate benchmark tests, matching the manifest op entries
- marks both scan op source blocks with `bench_manifest_driven: true` so manifest stats and L4 benchmark validation track the migration

## Motivation

The manifest is intended to be the source of truth for benchmark workloads and roofline metadata. Keeping a separate shape list and roofline calculator in the benchmark file can drift from `tileops/manifest/scan.yaml`.

## Validation

- `python -m ruff check benchmarks/ops/bench_cumulative.py`
- `python -m py_compile benchmarks/ops/bench_cumulative.py scripts/validate_manifest.py scripts/manifest_stats.py`
- `python scripts/validate_manifest.py --levels schema,shape,dtype,bench --check-op CumsumFwdOp --strict`
- `python scripts/validate_manifest.py --levels schema,shape,dtype,bench --check-op CumprodFwdOp --strict`
- `python scripts/manifest_stats.py --format json`
- `python -m pytest benchmarks/ops/bench_cumulative.py --collect-only -q`

## Not run / environment limits

- `python -m pytest benchmarks/ops/bench_cumulative.py -q` cannot run on this local machine because this PyTorch install is not compiled with CUDA enabled.
- The local pre-push hook's full `pytest -q` fails during collection for the same CUDA limitation in `tests/ops/attention/test_mean_pooling.py`; the branch was pushed with `--no-verify` after the targeted checks above.

## Notes / Follow-up

This PR does not expand scan workload coverage. The current manifest still only covers a small set of last-dimension large-scan cases. A follow-up issue should track improving `scan.yaml` workloads and making the benchmark consume workload-level params such as `dim` via `include_extra=True`.

### DIFF
--- a/benchmarks/ops/bench_cumulative.py
+++ b/benchmarks/ops/bench_cumulative.py
@@ -1,42 +1,35 @@
-"""Benchmarks for cumulative ops (cumsum, cumprod)."""
+"""Benchmarks for cumulative ops (cumsum, cumprod).
 
-from typing import Optional
+Measures latency, TFLOPS, and DRAM bandwidth against PyTorch baselines.
+Workload shapes and roofline formulas are loaded from the ops manifest
+(tileops/manifest/).
+"""
 
 import pytest
 import torch
 
-from benchmarks.benchmark_base import BenchmarkBase, BenchmarkReport
-from workloads.workload_base import FixtureBase, WorkloadBase
+from benchmarks.benchmark_base import (
+    BenchmarkReport,
+    ManifestBenchmark,
+    workloads_to_params,
+)
+from workloads.workload_base import WorkloadBase
 
-
-class CumulativeBenchFixture(FixtureBase):
-    PARAMS = [
-        (
-            "m, n, dtype, op_kind",
-            [
-                pytest.param(1024, 4096, torch.float16, "cumsum"),
-                pytest.param(1024, 4096, torch.bfloat16, "cumsum"),
-                pytest.param(4096, 4096, torch.float16, "cumsum"),
-                pytest.param(1024, 4096, torch.float16, "cumprod"),
-                pytest.param(1024, 4096, torch.bfloat16, "cumprod"),
-                pytest.param(4096, 4096, torch.float16, "cumprod"),
-            ],
-        ),
-    ]
+_CUMSUM_OP = "CumsumFwdOp"
+_CUMPROD_OP = "CumprodFwdOp"
 
 
 class CumulativeBenchTest(WorkloadBase):
-    def __init__(self, m: int, n: int, dtype: torch.dtype, op_kind: str):
-        self.m = m
-        self.n = n
+    def __init__(self, shape: tuple, dtype: torch.dtype, op_kind: str):
+        self.shape = shape
         self.dtype = dtype
         self.op_kind = op_kind
 
     def gen_inputs(self) -> tuple[torch.Tensor]:
         if self.op_kind == "cumprod":
-            x = torch.rand(self.m, self.n, dtype=self.dtype, device="cuda") * 0.01 + 0.99
+            x = torch.rand(*self.shape, dtype=self.dtype, device="cuda") * 0.01 + 0.99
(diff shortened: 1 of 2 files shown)


superAngGao:
Thanks for moving the cumulative benchmark toward the manifest-driven path. This direction matches the repo design: benchmark workloads and roofline metadata should come from the manifest rather than being duplicated in benchmark files.

One process suggestion for future PRs: it would be better to first open an issue that describes the benchmark/manifest problem being solved, then have the PR close that issue. That makes the motivation, expected behavior, and follow-up scope easier to review.

It would also help to make the PR body more explicit. For example:

```md
## Summary

Drive `benchmarks/ops/bench_cumulative.py` from the scan manifest instead of maintaining a separate hardcoded benchmark fixture.

This PR:
- replaces the local `CumulativeBenchFixture` with `workloads_to_params("CumsumFwdOp")` and `workloads_to_params("CumprodFwdOp")`
- replaces the local benchmark roofline calculation with `ManifestBenchmark`
- splits cumsum and cumprod into separate benchmark tests, matching the manifest op entries

## Motivation

The manifest is intended to be the source of truth for benchmark workloads and roofline metadata. Keeping a separate shape list in the benchmark file can drift from `tileops/manifest/scan.yaml`.

## Notes / Follow-up

This PR does not expand scan workload coverage. The current manifest still only covers a small set of last-dimension large-scan cases. A follow-up issue should track improving `scan.yaml` workloads and making the benchmark consume workload-level params such as `dim` via `include_extra=True`.
```

This is not a blocker for this PR, but documenting the intent and follow-up scope would make the change easier to review and maintain.

superAngGao:
One more process note: our PR title validator only accepts the type tags listed in the repo convention, e.g. `[Bench]`, `[Fix]`, `[Perf]`, `[Test]`, etc. The original title used `[Benchmark]`, which caused `validate-pr-title` to fail because the accepted benchmark tag is `[Bench]`.

I updated this PR title to:

```text
[Bench][Scan] Drive cumulative benchmarks from manifest
```

Please follow this PR title format in future PRs so preflight can pass on the first run.

zhen8838 (changes requested):
Thanks for moving the cumulative benchmarks to `workloads_to_params()` and `ManifestBenchmark`; the benchmark file direction looks right.

Please also update the manifest source contract for both scan ops. `tileops/manifest/scan.yaml` still leaves `CumsumFwdOp` and `CumprodFwdOp` without `source.bench_manifest_driven: true`, even though their benchmark is now manifest-driven.

That field is what keeps `scripts/manifest_stats.py` and the L4 benchmark validator in sync with the migration: without it, these implemented ops are still counted as not manifest-driven, and future benchmark regressions only stay on the legacy warning path instead of becoming a hard gate.

Please add this under both `source` blocks:

```yaml
    bench_manifest_driven: true
```

LaiQuan-conquer:
Addressed the requested manifest source contract update in `62e217c`: both `CumsumFwdOp` and `CumprodFwdOp` now set `source.bench_manifest_driven: true`. I also expanded the PR body with summary, motivation, validation, and follow-up scope.

lcy-seso (approved):
Thank you for this fix.

zhen8838 (approved):
Previous blocker is addressed. `CumsumFwdOp` and `CumprodFwdOp` now set `bench_manifest_driven: true` in the scan manifest, so the benchmark-contract change and manifest source of truth are back in sync. I do not see another blocker in the current diff.

PR #1588, opened 2026-06-16
[CI] use 1500 MHz GPU clock validation for H200 runners

### LINKED ISSUE #1587: CI: lower GPU clock validation target from 1830 MHz to 1500 MHz
## Summary

The self-hosted H200 runners currently validate GPU clocks against `1830 MHz`, but H200 cannot reliably sustain a locked `1830 MHz` graphics/SM clock on our current runner host under the `700W` power limit. We should lower the CI validation target to a frequency the runners can actually hold steadily, e.g. `1500 MHz`.

## Current workflow checks

`upstream/main` currently hard-codes `TARGET_CLOCK_MHZ=1830` in three places:

- `.github/workflows/nightly.yml`, benchmark test step
- `.github/workflows/nightly.yml`, packaging test step
- `.github/workflows/gpu-smoke.yml`, GPU frequency validation step

These checks read `clocks.current.graphics` via `nvidia-smi` and fail the workflow unless the current clock exactly matches `1830`.

## H200 runner observation

On the current H200 runner host:

- Persistence mode is enabled, but that only keeps the driver resident; it does not lock clocks.
- Application graphics clock remains at the default `1980 MHz`.
- `Applications Clocks Setting` is `Not Active`.
- There is no active locked-clocks state.
- With the `700W` power cap, attempting to lock H200 at `1830 MHz` is not stable under full load. The card is expected to be power-limited and pulled down, so the actual clock still moves instead of staying fixed at `1830 MHz`.

In practice, `1500 MHz` is a more realistic target for a stable locked clock on these H200 runners.

## Proposal

Change the workflow validation target from:

```bash
TARGET_CLOCK_MHZ=1830
```

to:

```bash
TARGET_CLOCK_MHZ=1500
```

in the nightly and gpu-smoke workflows.

Runner infra should then apply the matching lock as root, ideally via boot-time systemd/script because `nvidia-smi -lgc` does not persist across reboots:

```bash
sudo nvidia-smi -pm 1
sudo nvidia-smi -lgc 1500,1500
```

For hosts where only specific runner GPUs should be managed, apply this per device with `-i <gpu_index>`.

## Why this is better

A lower sustained clock is preferable to a higher nominal lock that is pulled down by power throttling. The goal of the CI check should be benchmark reproducibility and runner stability, not the maximum possible boost clock.

superAngGao:
## Summary

- lower the GPU clock validation target from `1830 MHz` to `1500 MHz`
- update both nightly validation points and the gpu-smoke validation point

## Context

H200 runners on the current 700W power limit cannot reliably sustain a locked `1830 MHz` graphics/SM clock under full load. A nominal `1830 MHz` lock can still be pulled down by power limiting, which defeats the goal of stable benchmark/control frequency validation.

`1500 MHz` is a more realistic sustained locked-clock target for these H200 self-hosted runners.

Fixes #1587

## Testing

- Verified the workflow files no longer contain `TARGET_CLOCK_MHZ=1830`.
- Not run: CI workflow execution, because this is a YAML target-value-only change.

### DIFF
--- a/.github/workflows/gpu-smoke.yml
+++ b/.github/workflows/gpu-smoke.yml
@@ -542,7 +542,7 @@ jobs:
       - name: Validate GPU frequency
         run: |
           set -euo pipefail
-          TARGET_CLOCK_MHZ=1830
+          TARGET_CLOCK_MHZ=1500
           RETRIES=5
           SLEEP_SECONDS=2
 
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -101,7 +101,7 @@ jobs:
           source "${VENV_PATH}/bin/activate"
 
           # Validate GPU frequency
-          TARGET_CLOCK_MHZ=1830
+          TARGET_CLOCK_MHZ=1500
           RETRIES=5
           for i in $(seq 1 "${RETRIES}"); do
             gpu_clock=$(nvidia-smi --query-gpu=clocks.current.graphics --format=csv,noheader,nounits | xargs)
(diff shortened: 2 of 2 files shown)


Gabbering (commented):
goose *goose skimmed `4987af1` — nothing to honk about.*

zhen8838 (approved):
LGTM

PR #1591, opened 2026-06-16
[CI] Publish nightly benchmark data to nightly-bench branch

lcy-seso:
## Summary

- Add a `publish-bench-data` job to the nightly workflow that collects the run's `bench_results.xml` + `test_results.xml` and a small `meta.json` (commit/date/gpu/run_id), then force-pushes them to a new `nightly-bench` orphan branch.
- Uses the default `GITHUB_TOKEN` and `peaceiris/actions-gh-pages@v4 force_orphan` — the same pattern as the existing manifest-stats publish job. No PAT/secret required.
- Runs on a github-hosted `ubuntu-latest` runner (`needs: [op_test]`), so it never occupies the GPU nightly runner.

This branch is the public data contract consumed by `tile-ai/TileOPs.github.io` to render the docs Benchmarks page; the docs site fetches it over public `raw.githubusercontent.com`, so no cross-repo token is needed in either direction.

## Test plan

- [x] pre-commit passed
- [x] `nightly.yml` validated with `yaml.safe_load`
- [ ] After merge: `workflow_dispatch` one nightly run and confirm the `nightly-bench` branch is created with the three files

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -334,6 +334,75 @@ jobs:
         shell: bash
 
   # =========================================================================
+  # Phase 3b — Publish benchmark data to the nightly-bench orphan branch
+  # Consumed by tile-ai/TileOPs.github.io to render the docs Benchmarks page.
+  # Runs on a github-hosted runner (no GPU needed) and uses the default token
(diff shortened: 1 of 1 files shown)


Ibuki-wind (approved):
Clean — no issues.

PR #1604, opened 2026-06-23
[Test][Manifest] Skip spec-only ops in source path validation

stelladuyx:
## Summary
- Skip ops with `status=spec-only` in `test_all_source_paths_exist`
- Spec-only entries are design specs without implementations yet
- Their source paths are placeholders and don't exist until the op is implemented

## Context
The test was failing on `GatedDeltaNetPrefillFwdOp` which is marked as `spec-only` in `tileops/manifest/linear_attention.yaml`. According to the trust model, spec-only ops declare future implementations and their source paths point to where files will eventually go.

## Test plan
- [x] Pre-commit hooks pass
- [x] Fixes the failing test for `GatedDeltaNetPrefillFwdOp`
- [x] Still validates source paths for all implemented ops

### DIFF
--- a/tests/test_ops_manifest.py
+++ b/tests/test_ops_manifest.py
@@ -199,6 +199,9 @@ class TestSourcePaths:
 
     def test_all_source_paths_exist(self, all_ops):
         for op_name, entry in all_ops.items():
+            # Skip spec-only ops — source paths are placeholders until implementation
+            if entry.get("status") == "spec-only":
+                continue
             source = entry["source"]
             for key, rel_path in source.items():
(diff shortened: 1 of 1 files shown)


superAngGao (approved):
LGTM.

zhen8838 (approved):
LGTM. This is a narrow test-only change: spec-only manifest entries are L0/design specs, so skipping implementation source-path existence checks for them matches the manifest status gating model. The source-path check still applies to implemented ops. Targeted local test passed and latest gpu-smoke/preflight checks are green.

Issue #1597, opened 2026-06-17
[BENCHMARK][Linear-Attn] Add Qwen3.5-aligned Gated DeltaNet prefill benchmarks


superAngGao:
## Summary

After `GatedDeltaNetPrefillFwdOp` is implemented, we should add a dedicated benchmark surface for inference prefill that matches the Qwen3.5 Gated DeltaNet shapes and compares TileOps against FLA under the same public contract.

This is separate from the operator implementation PR: the implementation adds the prefill op and basic manifest-driven benchmark plumbing, while this issue tracks the benchmark coverage needed to make prefill performance visible and reproducible upstream.

## Motivation

Gated DeltaNet prefill is performance-critical for long-context serving. The benchmark should not rely on the training/chunkwise forward surface, and it should not use arbitrary head counts as the primary comparison. It should use the FLA-compatible BTHD prefill contract:

- `q/k: [B, T, H, DK]`
- `v/o: [B, T, H, DV]`
- `g/beta: [B, T, H]`
- `final_state: [B, H, DK, DV]`

and compare with FLA using `output_final_state=True` so the returned-state contract is aligned.

## Proposed benchmark surface

Primary rows:

```text
B = 1
S in {32K, 64K, 128K}
DK = DV = 128
chunk_size = 64
layout = BTHD
dtype in {float16, bfloat16} where supported
H in {16, 32, 48, 64}
```

If CI/runtime budget is tight, start with:

```text
H in {16, 32}
S in {32K, 64K, 128K}
```

and keep H=48/64 as extended/nightly rows.

## Metrics to report

For each row:

- TileOps latency
- FLA latency with `output_final_state=True`
- speedup vs FLA
- layout and dtype
- GPU/clock configuration when available

## Acceptance criteria

- Add or update Gated DeltaNet prefill benchmark rows to cover Qwen3.5-aligned long-context prefill shapes.
- Include a TileOps-vs-FLA comparison path using the same BTHD input/output contract.
- Report results separately for `H = 16/32/48/64`, or document any reduced CI/nightly subset.
- Avoid mixing training/chunkwise forward benchmarks with inference prefill numbers.
- Keep this benchmark issue independent from decode benchmark alignment in #1581.

## Related

- Decode benchmark alignment: #1581
- FLA-compatible GDN prefill op issue: #1595
- Draft implementation PR: #1596

PR #1596, opened 2026-06-17
[Perf][Linear-Attn] Add optimized Gated DeltaNet prefill op

superAngGao:
Closes #1595.

## Summary

Adds `GatedDeltaNetPrefillFwdOp`, an optimized zero-state inference prefill operator for Gated DeltaNet. The op returns `(o, final_state)` without training-only intermediates and provides a TileOps-owned serving path for long-context prefill.

## Implementation

- Adds TileLang kernels for the BTHD serving path: chunk cumsum, blocked/Neumann-style A producer, CP initial-state correction, and fused replay/output.
- Adds shape-aware dispatch for the primary serving contract: `layout="bthd"`, `B=1`, `fp16`, `DK=DV=128`, `chunk=64`, with an environment kill switch.
- Keeps API compatibility for BTHD plus head-major BHTD/BHSD aliases.
- Adds op registration, tests, benchmark coverage, manifest workload, and roofline formula support.
- Fixes the manifest roofline path to decode BTHD workload shapes as `[B, T, H, D]` instead of `[B, H, T, D]`.

## Third-party Attribution

This PR also adds `THIRD_PARTY_NOTICES.md` and clarifies source-file headers for the GDN prefill path. Qwen FlashQLA is the MIT-licensed schedule-level reference for the h-state / corrected-segment-start CP-split scheduling. The local utility file is adapted from FlashQLA's utility source and keeps its upstream copyright notice. TileOps-specific work includes the public op wrapper, BTHD/head-major API integration, shape-aware dispatch, blocked/Neumann-style A producer integration, local replay/output implementation, TileLang compatibility changes, manifest/roofline support, tests, benchmarks, and verification.

## Shape Contract

Primary BTHD path:

- `q, k`: `[B, T, H, DK]`
- `v, o`: `[B, T, H, DV]`
- `g, beta`: `[B, T, H]`
- `initial_state, final_state`: `[B, H, DK, DV]`

Head-major BHTD/BHSD layouts remain supported through the public op API, but the optimized long-context serving dispatch targets BTHD.

## Performance Headline

H200/GPU4, BTHD, fp16, `B=1`, `DK=DV=128`, `chunk=64`, TileOps benchmark timer/CUPTI path. FlashQLA is reported as a public TL0.1.8 anchor and should be read as a public-environment comparison, not a same-lowering attribution experiment. Qwen FlashQLA is credited as the reference for the CP-split schedule family.

[...]

PR #1605, opened 2026-06-23
[Bench][Linear-Attn] Add Qwen GDN prefill benchmark surface

superAngGao:
Depends on #1596. Addresses #1597.

This is opened against upstream as a draft. Until #1596 lands, the diff will include the GDN prefill implementation commits from that dependency; after #1596 merges this branch should be rebased/updated so the PR contains only the benchmark-surface commit.

## Summary
- Switch the Gated DeltaNet prefill benchmark surface to the FLA/Qwen BTHD contract.
- Add Qwen3.5-aligned long-context rows for S=32K/64K/128K and H=16/32/48/64 across fp16/bf16.
- Compare TileOps against FLA with output_final_state=True and surface speedup in profile_run.log.
- Keep nightly JUnit properties compatible while allowing profile_run.log to show extra metrics and GPU clocks.

## Validation
- python -m py_compile benchmarks/ops/bench_gated_deltanet_prefill.py workloads/gated_deltanet.py benchmarks/benchmark_base.py tileops/perf/formulas.py
- python scripts/validate_manifest.py
- python -m pytest --collect-only benchmarks/ops/bench_gated_deltanet_prefill.py -q
- TMPDIR=/home/ga/tmp python -m pytest tests/ops/test_gated_deltanet_prefill.py -m smoke -q

## Notes
- The local environment does not have FLA installed, so full benchmark execution was only checked for collection/skip behavior here.

### DIFF
--- a/benchmarks/benchmark_base.py
+++ b/benchmarks/benchmark_base.py
@@ -317,6 +317,31 @@ def _get_env_metadata() -> list[str]:
         driver = "N/A"
     lines.append(f"- **Driver version**: {driver}")
 
+    try:
+        result = subprocess.run(
+            [
+                "nvidia-smi",
+                "--query-gpu=clocks.current.sm,clocks.current.memory,"
+                "clocks.applications.graphics,clocks.applications.memory",
+                "--format=csv,noheader,nounits",
+            ],
+            capture_output=True,
+            text=True,
+            timeout=5,
+        )
+        if result.returncode == 0:
+            clock_values = [part.strip() for part in result.stdout.splitlines()[0].split(",")]
+            if len(clock_values) == 4:
+                sm_clock, mem_clock, app_sm_clock, app_mem_clock = clock_values
+                lines.append(
+                    "- **GPU clocks**: "
+                    f"SM current {sm_clock} MHz, memory current {mem_clock} MHz, "
+                    f"application SM {app_sm_clock} MHz, "
+                    f"application memory {app_mem_clock} MHz"
+                )
+    except (FileNotFoundError, IndexError, subprocess.TimeoutExpired):
+        pass
+
     return lines
 
 
@@ -608,7 +633,7 @@ class BenchmarkReport:
         lines.extend(_get_env_metadata())
         lines.append("")
 
-        result_keys = ["latency_ms", "tflops", "bandwidth_tbs"]
+        default_result_keys = ["latency_ms", "tflops", "bandwidth_tbs"]
 
         for name, entries in BenchmarkReport._records.items():
             if not entries:
@@ -621,6 +646,11 @@ class BenchmarkReport:
             tag_entries = {}
             for entry in entries:
                 tag_entries.setdefault(entry["tag"], []).append(entry)
+            result_keys = list(default_result_keys)
+            for entry in entries:
+                for key in entry["result"]:
+                    if key not in result_keys:
+                        result_keys.append(key)
 
             for tag, tag_group in tag_entries.items():
                 lines.append(f"### {tag}")
--- a/benchmarks/benchmark_base.py
+++ b/benchmarks/benchmark_base.py
@@ -306,41 +306,42 @@ def _get_env_metadata() -> list[str]:
     else:
         lines.append("- **GPU model**: N/A (no CUDA device)")
 
-    # Try to get NVIDIA driver version from nvidia-smi
-    try:
-        result = subprocess.run(
-            ["nvidia-smi", "--query-gpu=driver_version", "--format=csv,noheader"],
-            capture_output=True, text=True, timeout=5,
-        )
-        driver = result.stdout.strip().split("\n")[0] if result.returncode == 0 else "N/A"
-    except (FileNotFoundError, subprocess.TimeoutExpired):
-        driver = "N/A"
-    lines.append(f"- **Driver version**: {driver}")
-
+    # Try to get NVIDIA driver version and clocks from nvidia-smi.
+    gpu_query_fields = [
+        "driver_version",
+        "clocks.current.sm",
+        "clocks.current.memory",
+        "clocks.applications.graphics",
+        "clocks.applications.memory",
+    ]
+    gpu_query_values = []
     try:
         result = subprocess.run(
             [
                 "nvidia-smi",
-                "--query-gpu=clocks.current.sm,clocks.current.memory,"
-                "clocks.applications.graphics,clocks.applications.memory",
+                f"--query-gpu={','.join(gpu_query_fields)}",
                 "--format=csv,noheader,nounits",
             ],
-            capture_output=True,
-            text=True,
-            timeout=5,
+            capture_output=True, text=True, timeout=5,
         )
         if result.returncode == 0:
-            clock_values = [part.strip() for part in result.stdout.splitlines()[0].split(",")]
-            if len(clock_values) == 4:
-                sm_clock, mem_clock, app_sm_clock, app_mem_clock = clock_values
-                lines.append(
-                    "- **GPU clocks**: "
-                    f"SM current {sm_clock} MHz, memory current {mem_clock} MHz, "
-                    f"application SM {app_sm_clock} MHz, "
-                    f"application memory {app_mem_clock} MHz"
-                )
-    except (FileNotFoundError, IndexError, subprocess.TimeoutExpired):
+            gpu_query_values = [
+                part.strip() for part in result.stdout.splitlines()[0].split(",")
+            ]
+    except (FileNotFoundError, subprocess.TimeoutExpired):
         pass
+
+    driver = gpu_query_values[0] if len(gpu_query_values) == len(gpu_query_fields) else "N/A"
+    lines.append(f"- **Driver version**: {driver}")
+
+    if len(gpu_query_values) == len(gpu_query_fields):
+        sm_clock, mem_clock, app_sm_clock, app_mem_clock = gpu_query_values[1:]
+        lines.append(
+            "- **GPU clocks**: "
+            f"SM current {sm_clock} MHz, memory current {mem_clock} MHz, "
+            f"application SM {app_sm_clock} MHz, "
+            f"application memory {app_mem_clock} MHz"
+        )
 
     return lines
 
--- a/benchmarks/ops/bench_gated_deltanet_prefill.py
+++ b/benchmarks/ops/bench_gated_deltanet_prefill.py
@@ -16,7 +16,6 @@ import torch
 from benchmarks.benchmark_base import BenchmarkReport, ManifestBenchmark
 from benchmarks.ops.attention.manifest_params import manifest_params
 from benchmarks.ops.bench_gated_deltanet import (
-    _to_fla_layout,
     compute_w_u_torch,
     kernel2_gated_deltanet_torch,
     prepare_wy_repr_gated_torch,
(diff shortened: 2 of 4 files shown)


tileops/kernels/gated_deltanet/gated_deltanet_prefill.py
```diff
@@ -0,0 +1,1613 @@
...
+                    S_shared[i, i] = T.float32(1.0)
+
+                for _r in T.Serial(num_rounds):
+                    T.clear(temp_frag)
+                    T.gemm(P_shared, S_shared, temp_frag)
+                    for i, j in T.Parallel(block_C, block_C):
+                        S_shared[i, j] = S_shared[i, j] + temp_frag[i, j]
+                    T.clear(temp_frag)
+                    T.gemm(P_shared, P_shared, temp_frag)
+                    T.copy(temp_frag, P_shared)
```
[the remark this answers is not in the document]
superAngGao: This thread is outdated after rebasing #1605 onto current main. The GDN prefill kernel implementation is no longer part of this PR diff; it was merged separately in #1596. This PR now only keeps the Qwen benchmark surface on top of that implementation.

tileops/kernels/gated_deltanet/gated_deltanet_prefill.py
```diff
@@ -0,0 +1,1613 @@
...
+                base = cid * chunk_size
+
+                g_s = T.alloc_shared([chunk_size], dtype)
+                out_s = T.alloc_shared([chunk_size], "float32")
+
+                T.copy(g[bid, hid, base : base + chunk_size], g_s, disable_tma=True)
+
+                out_s[0] = T.cast(g_s[0], "float32")
+                for i in T.Serial(1, chunk_size):
+                    out_s[i] = out_s[i - 1] + T.cast(g_s[i], "float32")
```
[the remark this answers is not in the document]
superAngGao: This thread is outdated after rebasing #1605 onto current main. The chunk-local cumsum kernel is no longer modified by this PR; it landed separately in #1596. This PR now only changes the benchmark/workload/manifest surface.

tileops/kernels/gated_deltanet/gated_deltanet_prefill.py
```diff
@@ -0,0 +1,1613 @@
...
+                    S_shared[i, i] = T.float32(1.0)
+
+                for _r in T.Serial(num_rounds):
+                    T.clear(temp_frag)
+                    T.gemm(P_shared, S_shared, temp_frag)
+                    for i, j in T.Parallel(block_C, block_C):
+                        S_shared[i, j] = S_shared[i, j] + temp_frag[i, j]
+                    T.clear(temp_frag)
+                    T.gemm(P_shared, P_shared, temp_frag)
+                    T.copy(temp_frag, P_shared)
```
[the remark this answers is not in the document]
superAngGao: This thread is outdated after rebasing #1605 onto current main. The BTHD prefill kernel implementation is no longer part of this PR diff; it was merged separately in #1596. This PR now only keeps the Qwen benchmark surface on top of that implementation.

benchmarks/ops/bench_gated_deltanet_prefill.py
```diff
@@ -0,0 +1,200 @@
...
+        dim_v,
+        chunk_size,
+        dtype,
+        tune=tune,
+        layout=layout,
+    )
+    bm = ManifestBenchmark(_OP_NAME, op, test)
+    result = bm.profile(op, *inputs)
+    result_fla = bm.profile(fla_fn, *inputs)
+    result["speedup_vs_fla"] = result_fla["latency_ms"] / result["latency_ms"]
```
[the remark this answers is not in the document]
superAngGao: Agreed. The rebased benchmark now handles this with `_to_fla_public_layout()`: BTHD inputs pass through, while BHTD/BHSD inputs are transposed to the FLA public BTHD layout before profiling the baseline.

tileops/kernels/gated_deltanet/gated_deltanet_prefill.py
```diff
@@ -0,0 +1,1613 @@
...
+                for i, j in T.Parallel(block_c, block_c):
+                    a_s[4, i, j] = -tmp[i, j]
+
+                T.clear(tmp)
+                T.gemm(a_s[3, :, :], i_s[0, :, :], tmp)
+                for i, j in T.Parallel(block_c, block_c):
+                    work_s[0, i, j] = tmp[i, j]
+                T.clear(tmp)
+                T.gemm(a_s[4, :, :], a_s[1, :, :], tmp)
+                for i, j in T.Parallel(block_c, block_c):
```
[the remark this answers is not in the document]
superAngGao: This thread is outdated after rebasing #1605 onto current main. `_prefill_blocksolve_A_bthd_tl` is no longer modified by this PR; the kernel landed separately in #1596, while this PR now only changes benchmark/workload/manifest surface.

benchmarks/benchmark_base.py
```diff
@@ -317,6 +317,31 @@ def _get_env_metadata() -> list[str]:
         driver = "N/A"
     lines.append(f"- **Driver version**: {driver}")
 
+    try:
+        result = subprocess.run(
```
zhen8838, line 319: Why don't you do it with the `gpu query` above?

zhen8838 (dismissed):

benchmarks/benchmark_base.py
superAngGao: Good point, this should share the existing nvidia-smi query path. I updated it to query driver_version and the clock fields in one `--query-gpu=` call, then derive both the driver line and GPU clocks from the same result while keeping the existing fallback behavior.

benchmarks/ops/bench_gated_deltanet_prefill.py
```diff
@@ -130,19 +200,21 @@ def test_gated_deltanet_prefill_fwd_bench(
...
     if fla_fn is not None:
-        result_fla = bm.profile(fla_fn, *bthd_inputs)
+        fla_inputs = _to_fla_public_layout(*inputs, layout)
+        result_fla = bm.profile(fla_fn, *fla_inputs)
+        result["speedup_vs_fla"] = result_fla["latency_ms"] / result["latency_ms"]
+        BenchmarkReport.record(op, locals(), result, tag="tileops")
         BenchmarkReport.record(op, locals(), result_fla, tag="fla")
     else:
+        BenchmarkReport.record(op, locals(), result, tag="tileops")
         result_ref = bm.profile(test.ref_program, *inputs)
```
Ibuki-wind, line 216: `test_gated_deltanet_prefill_fwd_bench` falls back to `test.ref_program` for every manifest row when FLA is unavailable, but this PR replaces the workload set with 32K/64K/128K Qwen rows. The pure-torch reference materializes chunk intermediates and runs the full prefill recurrence, so a no-FLA benchmark run can OOM or take impractically long instead of producing a usable baseline. -> Skip/xfail the long-context rows when FLA is unavailable, or keep only a small torch-fallback workload and require FLA for the Qwen rows.

Ibuki-wind (changes requested):
Blocking issue found in the benchmark fallback path.

benchmarks/ops/bench_gated_deltanet_prefill.py
superAngGao: Good catch. I updated the benchmark so the FLA availability check happens before generating inputs or profiling TileOps. When FLA is unavailable, long-context GDN prefill rows now skip early; the pure-torch fallback is capped to small ad-hoc rows (`S <= 4096`). This keeps the Qwen 32K/64K/128K manifest rows from accidentally running the full torch recurrence and risking OOM/very long no-FLA runs.

Validation:
- `python -m py_compile benchmarks/ops/bench_gated_deltanet_prefill.py workloads/gated_deltanet.py benchmarks/benchmark_base.py tileops/perf/formulas.py`
- `python -m ruff check benchmarks/ops/bench_gated_deltanet_prefill.py`
- `git diff --check`

`pytest --collect-only benchmarks/ops/bench_gated_deltanet_prefill.py -q` and `python scripts/validate_manifest.py` are currently blocked in my local Python env by the existing missing `tvm.tirx` dependency during `tileops` import.

superAngGao: Follow-up: I tightened this to better match the TileOPs benchmark rules. The manifest now includes one small S=4096 fallback workload, so no-FLA runs still have a manifest-driven row that records a non-TileOps `torch-ref` baseline. The Qwen long-context rows remain FLA-required and skip before input generation/profiling when FLA is unavailable.

benchmarks/ops/bench_gated_deltanet_prefill.py
```diff
@@ -90,21 +93,94 @@ def baseline_fn(
...
 
-def _gdn_prefill_args(workload: dict[str, Any]) -> tuple[int, int, int, int, int, int]:
-    batch, seq_len, heads, dim_k = workload["q_shape"]
-    _, v_seq_len, v_heads, dim_v = workload["v_shape"]
-    if v_seq_len != seq_len:
-        raise ValueError("GDN prefill q_shape and v_shape must share seq_len")
-    if v_heads != heads:
-        raise ValueError("GDN prefill q_shape and v_shape must share heads")
-    return batch, heads, seq_len, dim_k, dim_v, workload.get("chunk_size", 64)
+def _to_bhtd_layout(
```
Ibuki-wind: This file now has `_to_bhtd_layout()`, `_from_bhtd_layout()`, and `_to_fla_public_layout()`, with the same BTHD/BHTD permutation logic split across direction-specific helpers. That creates multiple local sources of truth for the same GDN prefill layout contract. Please replace these with one explicit converter, e.g. `convert_gdn_prefill_layout(inputs, src_layout, dst_layout)`, and use it for the torch-reference path (`layout -> bhtd`), FLA path (`layout -> bthd`), and output conversion (`bhtd -> layout`). This is not a numerical bug in the current BTHD-only rows, but we should not add more divergent one-off layout abstractions while defining this benchmark surface.

Ibuki-wind (changes requested):
Please tighten the layout conversion abstraction before approval. The fallback correctness issue is addressed, but this benchmark is adding a public GDN prefill performance surface and should not introduce another pair of mirrored, local layout helpers.

benchmarks/ops/bench_gated_deltanet_prefill.py
superAngGao: Thanks, agreed. I pushed 72c9d3ef570f50e69ad1bad3bdde25bf21f6d9bc to replace the direction-specific helpers with a single `convert_gdn_prefill_layout(tensors, src_layout, dst_layout)` entry point.

The benchmark now uses the same converter for:
- torch reference inputs: benchmark layout -> `bhtd`
- FLA baseline inputs: benchmark layout -> `bthd`
- torch reference output: `bhtd` -> benchmark layout

One small detail: `final_state` is intentionally not passed through the sequence-layout converter because it has shape `[B, H, DK, DV]` and has no time dimension; only the sequence output `o` is converted back.

Validation:
- `python -m py_compile benchmarks/ops/bench_gated_deltanet_prefill.py workloads/gated_deltanet.py benchmarks/benchmark_base.py`
- `python -m ruff check benchmarks/ops/bench_gated_deltanet_prefill.py`
- `git diff --check`

`pytest --collect-only benchmarks/ops/bench_gated_deltanet_prefill.py -q` is still blocked in this local env by the existing missing `tvm.tirx` import during `tileops` import.

Ibuki-wind (approved):
Clean on the current head. The benchmark now keeps an independent baseline for executed rows, skips long Qwen rows when FLA is unavailable, and uses one layout conversion path for TileOps/FLA/torch-ref handling. No remaining blocking issue found.

PR #1606, opened 2026-06-24
[CI] retire run-local venv across CI; install via image-baked stack

lcy-seso:
## What

Full RFC §4 cleanup on top of the ephemeral containerized-runner migration. The runner image bakes tilelang + the runtime/dev stack, so CI stops building/copying per-run venvs and installs only tileops (`--no-deps`) via `scripts/ci/install_tileops.sh`.

- **gpu-smoke**: drop `Set up Python`, `Resolve runtime state` (venv hash/copy/mtime-sync/divergence guard), `Cleanup isolated fork state`; install via `install_tileops.sh`; run with the image's `python3`; trust-route PRs by collaborator permission (write/maintain/admin → resident `nightly` pool; everyone else + lookup failure → `fork` pool, fail-closed); reclaim the `/ci-cache` layout (no wheels, no tool-cache prune); per-test `--timeout` + a `timeout-minutes` backstop so a wedged kernel cannot hold the single runner.
- **nightly**: `setup_nightly_venv.sh` → `install_tileops.sh`; system `python3`; cache env `/data7` → `/ci-cache`; per-test timeout on the pytest runs.
- **runner-maintenance**: drop the retired `venv` runs-on label (→ `nightly`).
- **preflight**: pin CPU installs with `constraints.txt`.
- **pyproject / constraints**: tilelang as a compatibility range (constraints pins its deps, not tilelang itself); pin `pytest-timeout`.
- **reclaim action + verify script**: `/ci-cache` layout, drop `WHEEL_DIR`; trim stamp on the persistent cache.
- **delete**: `setup_nightly_venv.sh`, `ci_venv_hash.py` and their obsolete tests.

## In scope vs follow-up

To keep CI dispatching, installing and importing cleanly on the new stack, this PR DOES adapt the import surface: `gqa_fwd_fp8` guards its `from tvm import tir` with a sentinel that raises a targeted error only when the kernel is built, and the fp8-GQA / topk-selector kernel-building smoke cases are skipped via a `tvm.tir` availability gate (with a focused test for the gate). So **CI is green at import**, not red.

**Out of scope (follow-up):** the actual kernel migration off `tvm.tir` (`tir.call_extern` → `tilelang.language` `T.*`) and any other new-stack kernel regressions surfaced by gpu-smoke (e.g. a build hang in `FP8LightingIndexerKernel`). Those are tracked separately; until they land, the corresponding smoke cases are skipped or fail on the per-test timeout rather than wedging the runner.


.github/workflows/gpu-smoke.yml
```diff
@@ -345,7 +345,10 @@ jobs:
   gpu-smoke:
     needs: [ci-prereq, security-policy]
     if: ${{ github.repository == 'tile-ai/TileOPs' && needs.ci-prereq.outputs.skip != 'true' && needs.security-policy.outputs.skip_gpu_smoke != 'true' }}
-    runs-on: [self-hosted, tile-ops, venv]
+    # Trust-level routing: untrusted (external) PRs go to the on-demand `fork` pool whose
+    # runner mounts an overlay cache (read-only shared lower + throwaway upper) so their cache
+    # writes never reach the shared cache. Trusted runs use the resident shared-cache pool.
+    runs-on: ${{ needs.security-policy.outputs.is_fork == 'true' && fromJSON('["self-hosted", "tile-ops", "fork"]') || fromJSON('["self-hosted", "tile-ops"]') }}
```
Ibuki-wind: `runs-on` is driven by `is_fork`, but `is_fork` is still derived from `author_association` above. That is not the RFC trust boundary: private org membership can report a non-member association (availability regression), while `COLLABORATOR` is not proof of `permission >= write` (shared-cache write risk). → Query `GET /repos/tile-ai/TileOPs/collaborators/<author>/permission`, route only `write`/`maintain`/`admin` to the resident pool (fail closed to `fork` on lookup failure), use the same result for trusted-action ref selection, and replace the association-based structural test.

.github/actions/reclaim-runner-disk/action.yml
```diff
@@ -49,16 +49,15 @@ inputs:
     description: >
       Newline-separated list of cache directory roots whose files older than
       cache-age-days should be trimmed at file granularity. Defaults target
-      the shared /data7-backed runner cache layout.
+      the bind-mounted /ci-cache runner cache layout.
```
Ibuki-wind, line 52: The cache data is now persistent under `/ci-cache`, but `CACHE_TRIM_STAMP` remains under `/home/ci-runner/.ci-maintenance`. The launcher mounts only `/ci-cache`, so `/home/ci-runner` is discarded after every ephemeral job and the configured 120-minute cooldown never survives between PR runs; under pressure, every run will rescan/trim the shared cache. → Store the stamp under a trusted persistent path such as `/ci-cache/.ci-maintenance`, or add an explicit persistent state path/mount, and cover the cooldown across fresh runner work directories.

Ibuki-wind (changes requested):
### Overall
The runner/workflow cutover is not merge-ready; address the blocking inline findings and rerun both failed checks.

Ibuki-wind:
Review scope correction: the TileLang top-level `tvm.tir` compatibility adaptation is explicitly exempted from this PR and may land in a separate PR to keep this CI migration focused.

The current red import checks are therefore not part of my change request for #1606. The remaining blockers are only the permission-based trust routing and persistence of the reclaim cooldown state under ephemeral runners.

.github/workflows/gpu-smoke.yml
lcy-seso: Addressed in cbe38b16. `is_fork` now derives from `GET /repos/tile-ai/TileOPs/collaborators/<author>/permission`: `write`/`maintain`/`admin` route to the resident pool; everyone else and any lookup failure fail closed to `fork`. The trusted-action ref selection reuses the same `is_fork`, and the `author_association` test is removed.

.github/actions/reclaim-runner-disk/action.yml
lcy-seso: Addressed in cbe38b16. `CACHE_TRIM_STAMP` now lives under `/ci-cache/.ci-maintenance/` (the persistent bind-mounted cache), so the cooldown survives across ephemeral runner jobs.

pyproject.toml
```diff
@@ -6,11 +6,11 @@ readme = "README.md"
...
     "torch>=2.1.0,<2.11.0",
-    "tilelang==0.1.9",
-    # FIXME(apache-tvm-ffi-pin): Remove once tilelang carries its own
-    # apache-tvm-ffi<0.1.10 pin (already on tilelang main, not yet released).
-    # Tracking: https://github.com/tile-ai/TileOPs/issues/969
-    "apache-tvm-ffi>=0.1.2,<0.1.10",
+    # tilelang as a compatibility range, not a source or exact pin: CI installs it from the
+    # runner image (--no-deps) and locks the exact build via constraints.txt; downstream
+    # `pip install tileops` resolves a compatible release. apache-tvm-ffi is pulled in (and
+    # ABI-version-locked by constraints.txt) transitively by tilelang — no manual cap here.
```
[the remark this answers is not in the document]
lcy-seso: Addressed in cbe38b16. Comment corrected: `constraints.txt` pins tilelang's runtime deps, not tilelang itself — CI installs tilelang from the runner image with `--no-deps`.

tileops/kernels/attention/gqa_fwd_fp8.py
```diff
@@ -5,7 +5,17 @@
...
+try:
+    # The tilelang runner stack (apache-tvm-ffi) exposes no top-level `tvm` module. This kernel
+    # still calls tir.call_extern; migrating those to T.call_extern is tracked separately. Guard
+    # the import so this module — and everything that eager-imports it via tileops.kernels — stays
+    # importable; only actually building this fp8 kernel hits the gap until that migration lands.
+    from tvm import tir
+except ImportError:
+    # Either no top-level `tvm` (ModuleNotFoundError) or the bundled tvm exposes no `tir`
+    # (ImportError: cannot import name 'tir') on the new stack — both subclass ImportError.
+    tir = None
```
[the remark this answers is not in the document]
lcy-seso: Addressed in cbe38b16. Replaced the `None` sentinel with `_TirUnavailable`, which raises a targeted `ImportError` on attribute access (naming the tir.<attr> -> T.* migration) instead of a bare AttributeError.

.github/workflows/gpu-smoke.yml
```diff
@@ -184,24 +184,28 @@ jobs:
...
+          # membership can report a non-member association (would wrongly isolate trusted authors),
+          # and COLLABORATOR is not proof of write access (would wrongly trust read-only authors).
+          # Only write/maintain/admin are trusted; everyone else — and any lookup failure — fails
+          # closed to the fork pool.
+          if [[ "$EVENT_NAME" != "pull_request" ]]; then
+            is_fork="false"   # push / schedule / workflow_dispatch are trusted
+          elif [[ "$HEAD_REPO" == "$BASE_REPO" ]]; then
+            is_fork="false"   # same-repo branch PR: author already has base-repo write access
+          else
+            role=$(gh api "repos/${BASE_REPO}/collaborators/${PR_AUTHOR}/permission" \
```
Ibuki-wind, line 199: The permission-based routing changes the policy contract, but `tests/test_gpu_smoke_policy.py::test_security_policy_handles_member_fork_as_trusted` still requires `AUTHOR_ASSOC` and `OWNER/MEMBER/COLLABORATOR`. Running that smoke test on this head gives `1 failed, 4 passed`, so the GPU job will fail independently of the exempted TileLang work. → Update the test and module docstring to assert `PR_AUTHOR`, the collaborator-permission endpoint, the `write`/`maintain`/`admin` allowlist, fail-closed behavior, and reuse of the resulting `is_fork` output.

Ibuki-wind (changes requested):
### Overall
The two original architecture blockers are resolved; address the new inline findings before merge.

tileops/kernels/attention/gqa_fwd_fp8.py
```diff
@@ -5,7 +5,24 @@
...
+    # kernel still calls tir.call_extern; migrating those to T.call_extern is tracked separately.
+    # Guard the import so this module — and everything that eager-imports it via tileops.kernels —
+    # stays importable. Touching `tir.*` (i.e. building this fp8 kernel) raises a clear, targeted
+    # error until that migration lands, instead of a bare AttributeError on a None sentinel.
+    from tvm import tir
+except ImportError:  # no top-level `tvm`, or bundled tvm exposes no `tir`; both subclass ImportError
+
+    class _TirUnavailable:
+        def __getattr__(self, name):
+            raise ImportError(
```
Ibuki-wind, line 20: `_TirUnavailable` fixes eager module import (and `validate-manifest` now passes), but it does not unblock the full GPU check: `tests/ops/attention/test_gqa_fp8.py` is marked `smoke` on SM90 and actually builds this kernel, which reaches the four `tir.call_extern` sites and raises this `ImportError`. → If the full `T.call_extern` migration remains separate, add an explicit availability gate that skips only the affected FP8 smoke cases when `tvm.tir` is unavailable (with the follow-up reference), plus a focused test for that gate; otherwise provide the minimal compatible call path required for these smoke cases. Require `gpu-smoke` to complete successfully.

Ibuki-wind (changes requested):
### Overall
The import guard is valid as a CI-unblocking measure, but its current boundary does not cover the GPU smoke execution path.

zhen8838 (dismissed):

.github/workflows/gpu-smoke.yml
lcy-seso: Addressed in cdf48e0b. Rewrote the trust test (now `test_security_policy_routes_trust_by_collaborator_permission`) to assert `PR_AUTHOR`, the `collaborators/${PR_AUTHOR}/permission` endpoint, the `write|maintain|admin` allowlist, fail-closed behavior, and `is_fork` reuse in both `runs-on` and the trusted-action ref; dropped the `author_association` assertions and updated the module docstring. Verified locally: 5 passed.

tileops/kernels/attention/gqa_fwd_fp8.py
lcy-seso: Addressed in cdf48e0b. Added `tileops/testing/tir_compat.tir_available()` and a module-level `skipif` on `tests/ops/attention/test_gqa_fp8.py` (and `tests/ops/test_topk_selector.py`, which builds the same way through tvm.tir) that skips those cases when tvm.tir is unavailable, with the `tir.* -> tilelang.language (T.*)` follow-up named in the reason. Added `tests/ops/attention/test_gqa_fp8_tir_gate.py`, a focused gate test that runs regardless of tir availability and asserts the probe agrees with the kernel sentinel and that the sentinel raises a targeted ImportError. Verified on the CPU stack: gate test passes; fp8/topk report 11 skipped (not errored).

tests/ops/attention/test_gqa_fp8.py
```diff
@@ -7,6 +7,15 @@
     quantize_kv_fa3_descale,
     quantize_q_fa3_gqa_descale,
 )
+from tileops.testing.tir_compat import tir_available
+
+# The FP8 GQA kernel still emits tir.call_extern; on the current tilelang stack tvm.tir is
+# unavailable, so building it raises ImportError. Skip these cases until the
+# tir.call_extern -> T.call_extern migration lands (follow-up).
+pytestmark = pytest.mark.skipif(
```
Ibuki-wind: This module-level `pytestmark` also skips `test_gqa_prefill_fp8_tensor_core_rejects_unaligned_q_tiles`, which only exercises constructor validation and never builds the `tir.call_extern` kernel. That unnecessarily removes two existing smoke cases, violating the Coverage set criterion. → Apply the `tir_available()` skip only to tests that actually build/run the FP8 kernel, and leave the unaligned-shape validation test active.

.github/workflows/gpu-smoke.yml
```diff
@@ -659,12 +475,14 @@ jobs:
           PYTEST_TARGETS: ${{ needs.security-policy.outputs.pytest_targets }}
         run: |
           set -euo pipefail
-          # shellcheck source=/dev/null
-          source "${VENV_PATH}/bin/activate"
+          # Per-test hang guard: a stuck kernel (e.g. a cold-compile loop or a spinning CUDA
+          # call) fails as a timeout with a thread dump instead of wedging the job. --no-deps
+          # tileops install doesn't bring this, so ensure it here.
+          python3 -m pip install -q pytest-timeout
```
Ibuki-wind: Installing unpinned `pytest-timeout` at runtime (also repeated in both nightly test jobs) reintroduces network dependency resolution and version drift into every ephemeral container, contrary to the image-baked stack contract; a PyPI outage or incompatible future release would disable the hang guard and fail CI before tests run. → Pin `pytest-timeout` in `constraints.txt` and install it into the runner image, then remove these runtime installs; if image rollout cannot be atomic, use `pip install -c constraints.txt pytest-timeout` as the bounded transition and verify the plugin in the baked-stack check.

Ibuki-wind (changes requested):
### Overall
The migration is close, but dependency determinism and retained test coverage need correction before merge.

tileops/kernels/attention/gqa_fwd_fp8.py
```diff
@@ -5,7 +5,24 @@
...
 import torch
-from tvm import tir
+
+try:
+    # The tilelang runner stack (apache-tvm-ffi) exposes no usable top-level `tvm.tir`. This
+    # kernel still calls tir.call_extern; migrating those to T.call_extern is tracked separately.
+    # Guard the import so this module — and everything that eager-imports it via tileops.kernels —
+    # stays importable. Touching `tir.*` (i.e. building this fp8 kernel) raises a clear, targeted
+    # error until that migration lands, instead of a bare AttributeError on a None sentinel.
+    from tvm import tir
```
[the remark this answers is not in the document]
lcy-seso: Addressed in 1d6d3c2e + PR description update. Clarified scope: this PR DOES adapt the import surface (guarded `tvm.tir` import + sentinel, plus the tvm.tir availability gate that skips the fp8-GQA/topk kernel-building smoke cases), so CI is green at import — not red. What remains out of scope is the actual kernel migration (`tir.call_extern` -> `tilelang.language` T.*) and other new-stack kernel regressions (e.g. the FP8LightingIndexerKernel build hang), now stated explicitly in the description.

tests/ops/attention/test_gqa_fp8.py
lcy-seso: Addressed in 1d6d3c2e. Replaced the module-level `pytestmark` with a `_skip_no_tir` mark applied ONLY to the three kernel-building cases; `test_gqa_prefill_fp8_tensor_core_rejects_unaligned_q_tiles` raises in the constructor before any build, so it now stays active (its two smoke params run again). Verified locally: 2 passed, 5 skipped.

.github/workflows/gpu-smoke.yml
lcy-seso: Addressed in 1d6d3c2e. Pinned `pytest-timeout==2.3.1` in constraints.txt and switched the gpu-smoke + nightly installs to `pip install -c constraints.txt pytest-timeout` (bounded, deterministic) instead of an unpinned PyPI fetch. The runner image should bake it and drop the per-job install — called out as a follow-up in the install-step comment and constraints.txt.

Ibuki-wind (approved):
Clean — no issues.

PR #1620, opened 2026-06-25
[Refactor] Extract GQA prefill kernel selection logic

### LINKED ISSUE #1611: [Refactor] Extract GQA prefill kernel selection logic
Parent: #1610

## Scope

Extract the first piece of GQA kernel selection policy while keeping the intended layering and manifest visibility explicit.

Preferred layering:

```text
OP class -> local selector/helper -> self.kernel_map[key] -> concrete Kernel class -> TileLang implementation/factory/callable
```

## Required design constraints

Do this:

```text
Op -> selector/helper -> concrete Kernel -> lower-level TL implementation/factory/callable
```

Do not do this:

```text
logical Kernel -> existing Kernel -> TL implementation
```

The OP-to-kernel relationship should remain visible through `default_kernel_map` and manifest `source.kernel_map` where applicable. Shared dispatch logic can live in a local pure helper near the GQA OP code, such as `_select_gqa_prefill_kernel_key(...)`; it does not need to become a concrete `Kernel` subclass or a new module in the first patch.

## Manifest constraint

This issue must not add new public OPs or change `tileops/manifest/attention.yaml`.

Current manifest entries have fixed ordered `signature.inputs`. Later OP aggregation should keep that constraint: each aggregated OP must define one explicit, fixed input list in both code and manifest. No optional tensor inputs and no dynamic arity.

## Non-goals

- Do not add new public OPs.
- Do not rename existing public OPs.
- Do not change `tileops/manifest/attention.yaml`.
- Do not add logical `Kernel` subclasses.
- Do not remove existing concrete kernel classes.
- Do not hide all concrete kernels behind a single mega logical kernel that weakens manifest/kernel_map visibility.
- Do not split the large GQA files yet unless the selector extraction naturally requires a small local helper.

## Expected behavior

- Dense/no-cache prefill paths can share selection logic without changing the public OP contract.
- Contiguous KV-cache prefill paths can share selection logic without changing the public OP contract.
- Paged KV-cache prefill paths can share selection logic without changing the public OP contract.
- RoPE / append-style paths remain OP-level orchestration of multiple concrete kernels.
- Dispatch tests should verify selector behavior without compiling GPU kernels.

## Validation

- Unit tests for selector behavior using concrete kernel-map keys or lightweight fake maps.
- Import checks for the touched GQA OP path.
- Focused GQA smoke tests where practical.
- No manifest validation changes are expected because this issue does not add public OPs.

superAngGao:
## Summary

- Extract GQA contiguous/paged prefill RoPE dispatch into local key selector helpers.
- Keep OP dispatch on the existing `self.kernel_map[key] -> concrete Kernel` path.
- Expose all concrete contiguous/paged prefill kernels from `default_kernel_map` instead of returning only the fuse_rope-selected subset.
- Add CPU-only selector/default-map tests.

## Notes

This is the first implementation step for #1611. It intentionally does not:

- add new public OPs;
- change `tileops/manifest/attention.yaml`;
- introduce logical `Kernel` wrappers;
- add a new selector module.

The goal is to keep concrete kernels visible for manifest/kernel-map override behavior while making the OP-level selection policy easier to test and discuss.

Closes #1611

## Tests

```text
ruff check tileops/ops/attention/gqa.py tests/ops/attention/test_gqa.py tests/ops/attention/test_gqa_prefill_paged.py
PYTHONPYCACHEPREFIX=/tmp/codex-gqa-selector-pycache python -m compileall -q tileops/ops/attention/gqa.py tests/ops/attention/test_gqa.py tests/ops/attention/test_gqa_prefill_paged.py
PYTHONPYCACHEPREFIX=/tmp/codex-gqa-selector-pycache python -m pytest tests/ops/attention/test_gqa.py tests/ops/attention/test_gqa_prefill_paged.py -q -k 'selector or default_map'
PYTHONPYCACHEPREFIX=/tmp/codex-gqa-selector-pycache python scripts/validate_manifest.py --check-op GroupedQueryAttentionPrefillWithKVCacheFwdOp
PYTHONPYCACHEPREFIX=/tmp/codex-gqa-selector-pycache python scripts/validate_manifest.py --check-op GroupedQueryAttentionPrefillPagedWithKVCacheFwdOp
```

### DIFF
--- a/tests/ops/attention/test_gqa.py
+++ b/tests/ops/attention/test_gqa.py
@@ -12,6 +12,9 @@ from tileops.kernels.attention import (
     GQAFwdWgmmaPipelinedKernel,
     GQAFwdWsPersistentCausalKernel,
     GQAFwdWsPersistentKernel,
+    GQAPrefillWithKVCacheFwdKernel,
+    GQAPrefillWithKVCacheRopeAppendKernel,
+    GQAPrefillWithKVCacheRopeFwdKernel,
 )
 from tileops.ops import (
     GroupedQueryAttentionBwdOp,
@@ -21,7 +24,10 @@ from tileops.ops import (
     GroupedQueryAttentionPrefillWithKVCacheFwdOp,
     RopeNeoxPositionIdsOp,
 )
-from tileops.ops.attention.gqa import _select_gqa_fwd_kernel_cls
+from tileops.ops.attention.gqa import (
+    _select_gqa_fwd_kernel_cls,
+    _select_gqa_prefill_with_kv_cache_kernel_keys,
+)
 from workloads.attention.gqa import (
     GroupedQueryAttentionBwdTest as _GroupedQueryAttentionBwdTestWorkload,
 )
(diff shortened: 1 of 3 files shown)


zhen8838 (changes requested):
Thanks for splitting this out — the layering is right (Op → local helper → `self.kernel_map[key]` → concrete kernel) and the CPU-only selector tests are a good addition. One blocking concern about scope vs. what this closes, plus two smaller notes.

### Scope vs `Closes #1611` (blocking)

The body says this is "the first implementation step for #1611" but also `Closes #1611` — those conflict, and the gap is real. #1611's *Expected behavior* explicitly lists the **dense / no-cache prefill** path sharing selection logic, but this PR only extracts the contiguous and paged KV-cache `fuse_rope` dispatch. The dense prefill selector `_select_gqa_prefill_fwd_kernel_cls` — the FA3 gate over `dtype` / `is_causal` / `dim` / `sm_scale` / `softcap` in `tileops/ops/attention/gqa.py` — is untouched.

That gate is the substantive selection *policy* here (the `fuse_rope` dispatch this PR extracts is a single boolean). It's also exactly what the forward-unification follow-up (#1622) needs as a clean, widenable selector. As written, merging would close #1611 with the trivial slice done and the hard part still inline.

Could we either:
- (a) change `Closes #1611` → `Part of #1611` and keep it open until the dense/FA3 gate is also extracted; or
- (b) extend this PR to extract that gate too?

### Duplicated selectors (minor)

`_select_gqa_prefill_with_kv_cache_kernel_keys` and `_select_gqa_prefill_paged_with_kv_cache_kernel_keys` are structurally identical (`fuse_rope -> (append_key, fwd_key)`). #1611 asks to *share* selection logic; these read as two copies of the same branch and could be one parameterized helper.

### `default_kernel_map` (looks good — just confirming)

Returning all three concrete kernels unconditionally now matches the manifest `source.kernel_map` for both ops (which already declares `fwd` + `rope_append` + `rope_fwd`), so this fixes a latent subset mismatch where the property previously returned only 1–2 keys. 👍

Net: glad to see the contiguous/paged dispatch extracted and tested — I just don't think it should *close* #1611 while the dense selection policy is still inline.

superAngGao:
Thanks, agreed with the review.

This PR only extracted the contiguous/paged `fuse_rope` dispatch and should not close #1611 while the dense/no-cache prefill selection policy is still inline. The dense/FA3 gate is the important selector for the next forward-unification work in #1622.

We are going to stop iterating on this PR and replace it with a new PR that folds in the broader taxonomy direction:

- keep #1611 open until dense/no-cache prefill selection is covered;
- include the dense/FA3 prefill selector in the new work;
- avoid duplicating the structurally identical RoPE selector branches by using a parameterized helper if that slice is still needed;
- keep the good part from this PR: `default_kernel_map` should expose concrete kernels consistently with manifest `source.kernel_map`.

Closing this PR as superseded rather than extending it in-place.

PR #1629, opened 2026-06-26
[Refactor][GEMM] make dense gemmop input-inferred and add manifest entry

zhen8838:
## Summary

Makes dense `GemmOp` **input-inferred** and brings it under the manifest, aligning with
DeepGEMM's call-time-JIT model. Implements step 1 of #1628.

- `GemmOp` no longer commits shape/dtype at construction: `forward(a, b)` infers `m/n/k`
  and `dtype` from the inputs, builds & caches the dtype-specialized kernel per
  `(m, n, k, dtype)` on first use (mirroring DeepGEMM's compile-on-first-call + per-config
  cache), and returns a freshly-allocated `d`.
- Default layout changed **NN → NT** (`trans_b=True`) to match DeepGEMM's primary path and
  `GroupedGemmOp`. `(trans_a, trans_b)` ≡ DeepGEMM `nt/nn/tn/tt`.
- New `tileops/manifest/gemm.yaml` with the `GemmOp` entry (`status: implemented`).
- GEMV fast-path dispatch (`m==1`/`n==1` on SM90) moved into `forward`.

## Verification (H200, container stack)

- `scripts/validate_manifest.py` → exit 0, no GemmOp errors.
- `pytest tests/ops/test_gemm.py` → **22 passed** (incl. NN/NT/TN/TT layouts, GEMV boundary,
  fp16/bf16, and the `tune=True` case).
- `pytest benchmarks/ops/bench_gemm.py` → 4 passed (manifest-driven, `eval_roofline`-based,
  cuBLAS baseline).
- `ruff check` clean on all changed Python files.

## Deviations from the issue plan (please review)

1. **Roofline is func mode, not inline.** A single static inline `vars` expression can't
   extract `M/N/K` correctly across all `trans_a/trans_b` layouts, and `M/N/K` can't be
   declared as manifest params (they're neither `__init__` nor `forward` args, so the
   signature check would reject them). Func mode (`tileops.ops.gemm._gemm_roofline`, reads
   `op.m/n/k/dtype` bound during `forward`) is correct for every layout. `eval_roofline()` is
   therefore valid only after the first `forward()`.

2. **Includes a small kernel autotune fix (`tileops/kernels/gemm.py`).** `tune=True` was
   broken for *both* `GemmKernel` and `GemvKernel` on the current tilelang stack: the
   `@tilelang.jit` builders had no parameter defaults, so the autotuner's validation
   `get_tir()` call (empty args) raised `missing required argument 'block_m'/'block_n'`. Fix:
   added default values to the tunable builder params. This is scoped to the GEMM/GEMV
   builders only — it does **not** touch the shared `kernel_base.autotune()` path, so other
   ops' autotune (also affected by the same upstream tilelang drift) is intentionally left for
   a separate tracking issue. (Authorized expansion beyond the issue's "no kernel" constraint.)

3. **Trust-model split.** The manifest entry is isolated in its own commit
   (`[Manifest][GEMM] …`) so it can be cherry-picked into a standalone human-reviewed
   manifest-only PR before merge, per `docs/design/trust-model.md`. This draft bundles it with
   the implementation only for review convenience.

## Benchmark (H200, torch 2.10.0+cu129)

`pytest benchmarks/ops/bench_gemm.py` on an **idle GPU** (`CUDA_VISIBLE_DEVICES` pinned to a 0%-util
device; SM clock locked at 1500 MHz) — TileOPs `GemmOp` (default cp.async `GemmKernel`) vs
`torch.matmul` (cuBLAS), via the repo's CUPTI / cold-L2 profiler. Speedup = cuBLAS latency ÷ TileOPs
latency (>1 = TileOPs faster).

| m | n | k | layout | dtype | TileOPs (ms / TFLOPS) | cuBLAS (ms / TFLOPS) | speedup |
|---|---|---|---|---|---|---|---|
| 1024 | 1024 | 1024 | NN | fp16 | 0.0165 / 130 | 0.0075 / 288 | 0.45× |
| 1024 | 1024 | 1024 | NN | bf16 | 0.0165 / 130 | 0.0075 / 288 | 0.45× |
| 128 | 2112 | 7168 | NT | bf16 | 0.1146 / 34 | 0.0184 / 211 | 0.16× |
| 128 | 7168 | 2048 | NT | bf16 | 0.0289 / 130 | 0.0135 / 278 | 0.47× |
| 4096 | 2112 | 7168 | NT | bf16 | 0.4437 / 280 | 0.3062 / 405 | 0.69× |
| 4096 | 7168 | 2048 | NT | bf16 | 0.3241 / 371 | 0.3132 / 384 | 0.97× |
| 4096 | 4096 | 7168 | NT | fp16 | 0.6130 / 392 | 0.6242 / 385 | 1.02× |
| 4096 | 4096 | 7168 | NT | bf16 | 0.6078 / 396 | 0.6078 / 396 | 1.00× |
| 4096 | 7168 | 16384 | NT | bf16 | 2.5455 / 378 | 2.5278 / 381 | 0.99× |
| 4096 | 24576 | 1536 | NT | bf16 | 0.9270 / 334 | 0.8552 / 362 | 0.92× |

**Takeaway:** at cuBLAS parity (0.92–1.02×) on large compute-bound prefill shapes (m=4096, K-aligned),
but meaningfully slower on small / decode-token shapes (m ≤ 1024 → 0.16–0.47×) and on non-256-aligned
N (4096×2112×7168 → 0.69×), where the simple cp.async `GemmKernel` is launch/memory/tail-bound. This
refactor preserves the existing kernel's numerics and perf — closing the small-m and tile-alignment
gaps is what the follow-up persistent / split-k work targets.

## Out of scope (follow-ups)

- `GemmAccumulateOp` (optional `c`/accumulate as a `variant_of` class).
- persistent / split-k / stream-k kernels.
- fp8/fp4 GEMM sibling op.
- grouped / gemv manifest entries.
- repo-wide `kernel_base.autotune()` fix for the tilelang API drift.

Draft for review — not for merge yet.

### DIFF
--- a/benchmarks/ops/bench_gemm.py
+++ b/benchmarks/ops/bench_gemm.py
@@ -87,15 +87,10 @@ def test_gemm_bench(
     op = GemmOp(trans_a=trans_a, trans_b=trans_b)
     bm = GemmBenchmark(test, op)
 
-    # Warmup: trigger JIT compilation (and bind roofline dims) before timing.
-    op(a, b)
-    torch.cuda.synchronize()
-
+    # The benchmark framework warms up internally; eval_roofline() is read
+    # lazily after profiling, by which point forward() has bound the dims.
     result = bm.profile(op, a, b)
     BenchmarkReport.record(op, locals(), result, tag="tileops")
-
-    test.ref_program(a, b)  # warmup
-    torch.cuda.synchronize()
 
     result_bl = bm.profile(test.ref_program, a, b)
     BenchmarkReport.record(op, locals(), result_bl, tag="torch-cublas")
--- a/benchmarks/ops/bench_gemm.py
+++ b/benchmarks/ops/bench_gemm.py
@@ -1,11 +1,27 @@
+"""Benchmark for dense GemmOp.
+
+Workload shapes come from the manifest entry's `workloads` (via
+`load_workloads`); the benchmark reports TileOPs latency alongside the
+manifest-derived roofline (`op.eval_roofline()`), with a cuBLAS
+(`torch.matmul`) baseline.
+"""
+
 from typing import Optional
 
 import pytest
 import torch
 
 from benchmarks.benchmark_base import BenchmarkBase, BenchmarkReport
+from tileops.manifest import load_workloads
 from tileops.ops import GemmOp
 from workloads.gemm import GemmTest
+
+_OP_NAME = "GemmOp"
+
+_DTYPE_MAP = {
+    "bfloat16": torch.bfloat16,
+    "float16": torch.float16,
+}
 
 
 class _GemmTestBaseline(GemmTest):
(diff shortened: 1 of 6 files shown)


benchmarks/ops/bench_gemm.py
```diff
@@ -20,36 +36,68 @@ def ref_program(self, a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
...
-    inputs = test.gen_inputs()
+    a, b = test.gen_inputs()
+
+    op = GemmOp(trans_a=trans_a, trans_b=trans_b)
+    bm = GemmBenchmark(test, op)
 
-    op = GemmOp(m, n, k, trans_a=trans_a, trans_b=trans_b, dtype=dtype, tune=tune)
-    result = bm.profile(op, *inputs)
+    # Warmup: trigger JIT compilation (and bind roofline dims) before timing.
+    op(a, b)
```
zhen8838: In the benchmark framework, it will warm up by itself, so we don‘t need to do it here.

tileops/ops/gemm.py
```diff
@@ -11,58 +11,140 @@
 __all__ = ["GemmOp"]
 
 
+def _gemm_roofline(op: "GemmOp") -> Tuple[int, int]:
```
zhen8838: Is the function of calculating roofline written here? Please take a look at the development document to confirm, and it should not start with a line.

tileops/ops/gemm.py
```diff
@@ -11,58 +11,140 @@
...
+        self._validate_dtypes(a, b)
+        m, n, k = self._infer_mnk(a, b)
+        # Bind dims/dtype/shapes for roofline (func mode reads these post-forward).
+        self.m, self.n, self.k = m, n, k
+        self.dtype = a.dtype
+        self.a_shape = tuple(a.shape)
+        self.b_shape = tuple(b.shape)
+
+        mode, kernel = self._get_kernel(m, n, k, a.dtype)
+        # Expose the active kernel so Op.autotune() can discover it.
```
zhen8838: You need to get parameters every time you forward like this, which is not good for benchmark, and it will have additional overhead. If you have JIT once, you should check whether the current instance has a kernel cache. If there is, it should be forwarded directly.

tileops/manifest/gemm.yaml
```diff
@@ -0,0 +1,41 @@
...
+    outputs:
+      d: {dtype: "same_as(a)"}
+    params:
+      trans_a: {type: bool, default: false}
+      trans_b: {type: bool, default: true}
+
+  workloads:
+  # m/n/k are the logical GEMM dims (trans-independent); trans_a/trans_b pick
+  # the storage layout the kernel compiles for.
+  - {m: 1024, n: 1024, k: 1024, trans_a: false, trans_b: false, dtypes: [float16, bfloat16], label: "square-1k-nn"}
```
zhen8838, line 32: This workload is too small. You should check the test of Deepgemm and expand the test set.

tileops/ops/gemm.py
```diff
@@ -11,58 +11,140 @@
...
+        self,
+        trans_a: bool = False,
+        trans_b: bool = True,
+        kernel_map: Optional[Dict[str, Kernel]] = None,
+        tune: bool = False,
+    ) -> None:
+        self.trans_a = trans_a
+        self.trans_b = trans_b
+        self._tune = tune
         self.dispatch_kernel(kernel_map)
```
zhen8838, line 50: Now that we don’t have the input of MKN, can‘t we dispatch kernel here?

tileops/ops/gemm.py
```diff
@@ -11,58 +11,140 @@
...
+
+    Raises:
+        RuntimeError: If called before ``forward()`` has bound the dims.
+    """
+    if op.m is None or op.dtype is None:
+        raise RuntimeError(
+            "GemmOp.eval_roofline() is valid only after the first forward(); "
+            "m/n/k and dtype are inferred from the inputs."
+        )
+    m, n, k = op.m, op.n, op.k
```
zhen8838: Now the calculation of roofline must be forward once. You need to check whether it is reasonable to call the roofline calculation.

zhen8838:
Thanks for the review — addressed in `cc8c281` (manifest) + `bb01e7c` (code). Point by point:

**`gemm.py:14` — roofline func placement / "should not start with a line"**
Moved the roofline callable out of `gemm.py` into `tileops/perf/formulas.py` as
`gemm_fwd_roofline`, matching the convention that all func-mode roofline callables live there
(`where_fwd_roofline`, `clamp_fwd_roofline`, `mha_fwd_roofline`, …). Manifest `roofline.func`
now points at `tileops.perf.formulas.gemm_fwd_roofline`.

**`gemm.py:144` — per-forward overhead / "if JIT'd once, forward directly"**
Added a same-signature fast path: `forward` keys on `(a.shape, b.shape, a.dtype)`; when it
matches the previous call it reuses the already-built kernel directly and skips dtype
validation, `m/n/k` inference, and the cache lookup. Re-inference only runs when the signature
changes.

**`bench_gemm.py:91` — framework warms up itself**
Removed the manual `op(a, b)` / `ref_program` warmups + syncs. The framework warms up
internally, and the func-mode roofline is read lazily after profiling (by which point
`forward` has bound the dims).

**`gemm.yaml:26` — workload too small, mirror DeepGEMM**
Expanded to DeepGEMM's normal-GEMM `(n, k)` set (DeepSeek-V3 MLP/attention projections:
`(2112,7168)`, `(7168,2048)`, `(4096,7168)`, `(7168,16384)`, `(24576,1536)`) swept over decode
(`m=128`) and prefill (`m=4096`), NT layout.

**Gemini bot — `autotune` won't discover lazily-cached kernels**
Added a `GemmOp.autotune()` override that tunes every kernel in `self._kernel_cache` (the base
`Op.autotune` scans `dir(self)` and would miss them).

---

Two of your comments are design questions rather than fixes — my reasoning, let me know if
you'd prefer a different approach:

**`gemm.py:30` — "roofline must forward once; is that reasonable?"**
It's inherent to the input-inferred model: with no shape/dtype at construction, the op
genuinely doesn't know `m/n/k`/dtype until it sees the inputs. This is the documented
arbitrary-rank roofline path (`docs/design/roofline.md` §4.4.5: bind during `forward()`), and the
benchmark harness already reads `eval_roofline()` lazily *after* profiling. `gemm_fwd_roofline`
raises a clear error if called before any forward. DeepGEMM is the same — it only knows the dims
at call time.

**`gemm.py:73` — "no MKN, can't we dispatch kernel here?"**
The kernel *map* (which Kernel classes are available) IS resolved at construction via
`dispatch_kernel(kernel_map)`. What can't happen at construction is picking gemm-vs-gemv and
building the shape-specialized kernel instance: the GEMV fast path only applies when `m==1`
(lhs_row) or `n==1` (rhs_col), so the choice needs the runtime shapes. Hence selection + build
live in `forward` (cached per `(m,n,k,dtype)`, with the fast path above for repeats).

Re-verified on H200: `validate_manifest` clean, `test_gemm.py` 22 passed, `bench_gemm.py` runs
manifest-driven, ruff clean.

benchmarks/ops/bench_gemm.py
```diff
@@ -1,12 +1,28 @@
+"""Benchmark for dense GemmOp.
```
zhen8838: remove this comment

lcy-seso (approved):

PR #1630, opened 2026-06-26
[Fix][Kernel] restore autotuning on TileLang 0.1.11

lcy-seso:
## Summary

Autotuned ops (`tune=True`) crash on the TileLang 0.1.11 stack:

```
TypeError: 'NoneType' object is not subscriptable          # self.config["num_split"], ...
TypeError: '<' not supported between 'NoneType' and 'float'
```

Surfaced by nightly: `op_test` had 9 failures (all `-tuned`/`True` variants), `benchmark` had 168 (gemm/rms_norm/softmax/pool/deepseek_mla_decode). Non-tuned variants pass.

## Root cause

`Kernel.autotune()` set `self.config = tuned_kernel.config`, which came back **None**. TileLang 0.1.11 short-circuits the autotuner — logging `"Tunable parameters [...] already provided during auto-tuning. Skipping compilation and using direct JIT"` and returning an `AutotuneResult` with no `config` — whenever the **tunable** JIT parameters are supplied on the autotuned call. The base class supplied them (seeding `default_config`) to satisfy TileLang 0.1.11's eager signature binding. So tuning never ran and `self.config` was `None`, and the first `self.config[...]` in `forward()` raised.

## Fix — one central change in `kernel_base`

Decouple the two needs (bind the signature vs. don't trip the skip), with a single source of truth and no per-kernel edits:

- **Don't pass tunable args** on the autotuned call (so the autotuner is not skipped); invoke the wrapper with no arguments.
- **Seed the JIT argument binder's per-parameter defaults** from the kernel's `default_config` (or the per-sub-kernel config a custom autotune override passes), via `Kernel._seed_autotune_binder`. Binding then succeeds with no args, and the autotuner overrides every tunable parameter per candidate during the search.

These seeded values are **bind-time placeholders only** — never the executed config (the search overrides them; `forward` always passes the chosen config explicitly). Because they come straight from `default_config`, there is **no second source of default values and no inconsistency**, including shape-computed entries (e.g. deepseek MLA-decode's `block_H = min(64, heads // kv)` is seeded with its real per-instance value).

`Kernel._call_autotuned_kernel` is the single call point used by both the base `autotune()` and the custom autotune overrides (softmax, logsumexp, deltanet, gated_deltanet, fp8 indexer, deepseek DSA), so the fix covers all of them without touching individual kernels.

## Test plan

Verified on H200 (TileLang `0.1.11+cu129.git65dbc98`):

- [x] gemm (`tune=True`): autotuner runs the candidate space (no "Skipping"), returns a real best config, op forward correct — was `config=None` -> crash.
- [x] deepseek MLA-decode warp-specialized (the op that surfaced the bug): autotunes; `config` populated; seeded `block_H` matches `default_config` (consistency check).
- [x] All kernel modules import; `tests/kernels/test_kernel_autotune_binding.py` updated to the new behavior (no seeded kwargs; binder defaults seeded from `default_config`; signature filtering; aliases) and passes.
- [x] ruff + pre-commit pass.
- [ ] Full nightly `op_test` + `benchmark` rerun (broad check across all tuned ops).

## Notes

Implementation seeds TileLang's `_JITArgumentBinder` defaults/`required_indices` directly. This couples to a TileLang internal, but fails loudly (not silently) on a TileLang change, and TileLang is pinned by the runner image. The underlying "silent `config=None` on skip" is arguably a TileLang issue worth reporting upstream.

### DIFF
--- a/tests/kernels/test_kernel_autotune_binding.py
+++ b/tests/kernels/test_kernel_autotune_binding.py
@@ -8,12 +8,28 @@ from tileops.kernels.kernel_base import Kernel
 pytestmark = pytest.mark.smoke
 
 
+class _FakeBinder:
+    """Mimics TileLang's _JITArgumentBinder fields used by _seed_autotune_binder."""
+
+    def __init__(self, param_names):
+        self.param_names = tuple(param_names)
+        self.defaults = ()
+        self.required_indices = tuple(range(len(param_names)))
+
+
+class _FakeFunc:
+    def __init__(self, param_names):
+        self._argument_binder = _FakeBinder(param_names)
+
+
 class _FakeJit:
-    signature = inspect.signature(lambda block_m, threads: None)
+    """Stand-in for a @tilelang.jit kernel: exposes .func._argument_binder and .signature."""
 
-
-class _FakeAliasedJit:
-    signature = inspect.signature(lambda threads_arg, npt_arg: None)
+    def __init__(self, param_names):
+        self.func = _FakeFunc(param_names)
+        self.signature = inspect.Signature(
+            [inspect.Parameter(n, inspect.Parameter.POSITIONAL_OR_KEYWORD) for n in param_names]
+        )
 
 
 class _TunedKernel:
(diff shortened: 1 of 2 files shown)


lcy-seso:
Superseded by #1627, which fixes the same TileLang 0.1.11 autotune-skip bug centrally via `do_not_specialize` (cleaner than this PR's approach). Closing; will follow up with a small change extending #1627's `do_not_specialize` to the remaining custom-autotune kernels (deltanet, gated_deltanet, fp8 indexer, deepseek_dsa) if they're still affected.

PR #1633, opened 2026-06-27
[Fix][Attention] Fix GQA sliding-window fwd WGMMA output shared-memory race

### LINKED ISSUE #1404: [BUG][GQA] fix wgmma varlen output shared-memory race
### Symptom / Motivation

The TileLang 0.1.9 screening work from #1071 left two GQA sliding-window varlen forward cases skipped under #1074 because they produced numerical mismatches on H200 WGMMA kernels:

- `tests/ops/attention/test_gqa_sliding_window_varlen.py::test_gqa_sliding_window_varlen_fwd_op[2-seqlens_q8-seqlens_k8-8-2-64-False-64-64-dtype8-False]`
- `tests/ops/attention/test_gqa_sliding_window_varlen.py::test_gqa_sliding_window_varlen_fwd_op[2-seqlens_q13-seqlens_k13-8-2-64-True-0--1-dtype13-False]`

Related to #1071 and tracked from #1074.

### Root Cause Analysis

The H200 WGMMA varlen sliding-window output path normalizes `acc_o`, copies it through shared memory via `T.copy(acc_o, o_shared)`, and then immediately reads `o_shared` to write global output.

Exported CUDA for stale/bad TileLang 0.1.9 cache entries showed the `ptx_stmatrix_x4` write into shared memory was followed immediately by global output stores reading the same shared-memory region, without a partial thread synchronization. Freshly generated CUDA contains `tl::__sync_thread_partial<3, 256>()` before and after the shared-memory staging copy. The missing synchronization creates a producer/consumer race and explains the large numerical mismatches around 0.47 and 0.60.

The TileOps kernel should make this synchronization explicit instead of relying on TileLang to infer and insert it.

### Related Files

- `tileops/kernels/attention/gqa_sliding_window_varlen_fwd.py`
- `tests/conftest.py`
- `tests/ops/attention/test_gqa_sliding_window_varlen.py`

## Goal

Fix the H200 WGMMA GQA sliding-window varlen forward output staging race and remove the exact TileLang 0.1.9 skips once the previously failing cases pass.

## Plan

1. Add explicit partial shared-memory synchronization around the WGMMA output staging copy from `acc_o` to `o_shared`.
2. Keep the default WGMMA configuration at `threads=256` and verify the fix addresses the race directly rather than avoiding the configuration.
3. Remove the two exact GQA sliding-window varlen nodeids from the TileLang 0.1.9 skip list.
4. Run `TMPDIR=/home/lyc/Project/TileOPs/.tmp/tvm_tmp python -m pytest tests/ops/attention/test_gqa_sliding_window_varlen.py -vvs`.

## Constraints

- Do not broaden skip coverage.
- Do not disable the H200 WGMMA path or reduce the default thread count as a workaround.
- Keep the fix scoped to the GQA sliding-window varlen forward kernel and the exact skip removals.

## Acceptance Criteria

- [ ] The two exact previously skipped GQA sliding-window varlen nodeids pass locally on H200 with `threads=256`.
- [ ] The full `tests/ops/attention/test_gqa_sliding_window_varlen.py` file passes locally.
- [ ] The TileLang 0.1.9 skip list no longer includes the two fixed GQA varlen nodeids.

### LINKED ISSUE #1616: [BUG][ATTENTION] GQA sliding-window fwd mismatches non-causal right-window cases
## Summary

Nightly op tests still have a real correctness failure in fixed-length GQA sliding-window forward, separate from the TileLang autotune missing-argument issue and the CI Triton cache permission issue.

Nightly run: https://github.com/tile-ai/TileOPs/actions/runs/28125205861/job/83305954027

The failing op in the nightly report is:

- `GroupedQueryAttentionSlidingWindowFwdOp`
- module reported by nightly: `tileops.ops.attention.gqa`
- current test path: `tests/ops/test_gqa_sliding_window_fwd.py`
- current op wrapper: `GqaSlidingWindowFwdOp`
- kernel selected on H200: `GQASlidingWindowFwdWgmmaPipelinedKernel`

## Failing Cases

Two fixed-length cases fail against the pure PyTorch reference:

| Case | Window | dtype | Error |
|---|---|---|---|
| `test_gqa_sliding_window_fwd_op[2-512-8-2-64-False-64-64-dtype6-False]` | non-causal, `wl=64`, `wr=64` | bf16 | max abs diff `0.474609375`, `91170 / 524288` mismatched (`17.4%`) |
| `test_gqa_sliding_window_fwd_op[2-512-8-2-64-False--1-64-dtype12-False]` | non-causal, `wl=-1`, `wr=64` | fp16 | max abs diff `0.444580078125`, `77024 / 524288` mismatched (`14.7%`) |

Failure excerpt:

```text
AssertionError: Tensor-likes are not close!
Mismatched elements: 91170 / 524288 (17.4%)
Greatest absolute difference: 0.474609375 at index (1, 26, 2, 45) (up to 0.01 allowed)
```

```text
AssertionError: Tensor-likes are not close!
Mismatched elements: 77024 / 524288 (14.7%)
Greatest absolute difference: 0.444580078125 at index (1, 256, 5, 7) (up to 0.01 allowed)
```

Both failures used the default H200 WGMMA config:

```text
GQASlidingWindowFwdWgmmaPipelinedKernel initialized with config: {'block_m': 128, 'block_n': 128, 'num_stages': 3, 'threads': 256}
```

## Useful Clues

The PyTorch reference mask is:

```python
if is_causal:
    mask = mask | (q_pos < k_pos)
if wl >= 0:
    mask = mask | (k_pos < q_pos - wl)
if wr >= 0:
    mask = mask | (k_pos > q_pos + wr)
```

Observed pattern from the nightly run:

- fixed-length causal cases pass;
- fixed-length non-causal no-window case passes;
- fixed-length non-causal left-window-only case passes;
- fixed-length non-causal right-window cases fail;
- fixed-length `wl=64, wr=64` passes for fp16 but fails for bf16;
- fixed-length right-only `wl=-1, wr=64` fails for fp16;
- varlen GQA sliding-window cases with similar window shapes passed in the same run.

This points toward the fixed-length WGMMA kernel's non-causal right-window masking / block-boundary handling rather than the reference implementation or the generic test harness.

## Proposed Investigation

- Reproduce the two failing cases locally on H200 with the WGMMA kernel.
- Compare fixed-length mask bounds with the varlen implementation, especially the right-window upper bound (`k_pos <= q_pos + wr`) and how it is applied per `block_m/block_n` tile.
- Check whether the issue only appears when `wr >= 0`, and whether bf16 accumulation/casting exposes an additional error for symmetric windows.
- Add a narrow regression test for:
  - non-causal right-only `wl=-1, wr=64`, fp16;
  - non-causal symmetric `wl=64, wr=64`, bf16.

## Acceptance Criteria

- Both failing fixed-length GQA sliding-window cases pass with `atol=1e-2, rtol=1e-2`.
- Existing causal, left-window-only, no-window, and varlen sliding-window cases remain green.
- The fix is covered by targeted regression cases for right-window masking.

lcy-seso:
## Summary

Fixes #1616, plus a defensive fix for the same latent race in the plain MHA/GQA forward WGMMA kernels.

The fixed-length `GQASlidingWindowFwdWgmmaPipelinedKernel` (selected on H200) miscomputed non-causal right-window cases — up to ~19% of elements wrong, max abs diff ~0.47 (tolerance 0.01), reproduced deterministically across recent `main` gpu-smoke runs.

## Root cause

The kernel writes the output accumulator through a **swizzled** `o_shared` buffer (`acc_o -> o_shared -> output`) with **no barrier** around the round-trip. Across warps this is a shared-memory race: a warp reads `o_shared` before peer warps finish writing it, corrupting the output under WGMMA pipelining (`num_stages >= 2`).

The varlen sibling already guards the identical round-trip with `T.sync_threads(3, threads)` (added in the closed #1404 "wgmma varlen output shared-memory race"); that guard was never ported to the fixed-length kernel. This PR ports it.

**Why a named barrier (slot 3, not bare `T.sync_threads()`):** verified on H200 — a bare `T.sync_threads()` (slot 0) aliases the implicit `__syncthreads()` the surrounding `T.copy` lowering already emits, gets elided, and the race survives. Any non-zero slot works.

**Why non-causal right-window specifically:** some query rows have a fully-masked trailing KV tile, which shifts warp scheduling enough to expose the race. Causal / left-only / no-window paths and `num_stages=1` happened to avoid the timing window — which is why those passed and masked the latent bug.

## Commits

1. `gqa_sliding_window_fwd.py` — the #1616 fix.
2. `gqa_fwd.py` — defensive fix: `_mha_fwd_wgmma_pipelined_main` / `_gqa_fwd_wgmma_pipelined_main` have the identical unguarded write-back. Plain attention has no masked-tile trigger so it is latent (tests pass), but the hazard is real; barriers only add synchronization.
3. comment clarifying the named-barrier requirement.

## Correctness verification (CI image `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10-dev`, torch 2.10.0+cu129, H200)

| suite | result |
|---|---|
| `tests/ops/attention/test_gqa_sliding_window.py` | **21/21 passed** (incl. the two previously-failing cases) |
| `tests/ops/attention/test_mha.py` + `test_gqa.py` | **55 passed, 4 skipped** (no regression from commit 2) |

## Barrier perf impact — before/after latency (median over 100 iters, same image)

The two `T.sync_threads` run once per block at the end (not in the KV loop), so the cost is below run-to-run noise.

| kernel | shape | latency before (ms) | latency after (ms) |
|---|---|---|---|
| GQASlidingWindowFwdWgmmaPipelined | fp16 wl=-1 wr=64 | 0.0167 | 0.0164 |
| GQASlidingWindowFwdWgmmaPipelined | bf16 wl=64 wr=64 | 0.0149 | 0.0146 |
| MHAFwdWgmmaPipelined | causal fp16 | 0.0162 | 0.0159 |
| GQAFwdWgmmaPipelined | causal fp16 | 0.0160 | 0.0157 |
| MHAFwdWgmmaPipelined | causal bf16 | 0.0161 | 0.0159 |
| GQAFwdWgmmaPipelined | causal bf16 | 0.0158 | 0.0158 |

> "before" = the unguarded `origin/main` kernels (temporarily reverted, then restored). The sliding-window "before" output is numerically wrong (the bug); only latency is comparable.

## GQA forward vs reference baselines (`bench_gqa.py`, same image, H200)

Metric is **throughput in TFLOPS (higher is better)** — not latency. TileOPs = `GroupedQueryAttentionFwdOp` (kernel variant in parens). Baselines = FlashAttention-3 and FlashInfer. Bold = fastest in the row.

### Compute-bound (S=2048, causal, autotuned) — the representative comparison

| model | dtype | TileOPs (TFLOPS) | FA3 (TFLOPS) | FlashInfer (TFLOPS) |
|---|---|---|---|---|
| Llama-3.1-8B  (B=2, H=32, Hkv=8, D=128) | fp16 | 412 (ws_causal) | **508** | 215 |
| Llama-3.1-8B  (B=2, H=32, Hkv=8, D=128) | bf16 | 107 (wgmma_pipelined) | **288** | 186 |
| Llama-3.1-70B (B=1, H=64, Hkv=8, D=128) | fp16 | 212 (ws_causal) | **236** | 216 |
| Llama-3.1-70B (B=1, H=64, Hkv=8, D=128) | bf16 | 146 (wgmma_pipelined) | **236** | 186 |

### Small / overhead-bound (S=512, causal) — latency-dominated, not throughput-representative

| model | dtype | TileOPs (TFLOPS) | FA3 (TFLOPS) | FlashInfer (TFLOPS) |
|---|---|---|---|---|
| Llama-3.1-8B  (B=4, H=32, Hkv=8, D=128) | fp16 | 215 (ws_causal) | **249** | 88 |
| Llama-3.1-8B  (B=4, H=32, Hkv=8, D=128) | bf16 | 81 (wgmma_pipelined) | **94** | 92 |
| Llama-3.1-70B (B=2, H=64, Hkv=8, D=128) | fp16 | 92 (ws_causal) | 94 | **196** |
| Llama-3.1-70B (B=2, H=64, Hkv=8, D=128) | bf16 | 86 (wgmma_pipelined) | 94 | **198** |

**Reading the compute-bound numbers:**
- **fp16 (ws_causal)** reaches **81–90% of FA3** and matches/beats FlashInfer — TileOPs' strong path.
- **bf16 (wgmma_pipelined)** is markedly weaker (**37–62% of FA3**), even behind FlashInfer. For the same causal shape fp16 dispatches to `ws_causal` but bf16 falls back to `wgmma_pipelined`, a separate perf gap unrelated to this fix (the barrier change has no measurable latency impact, see above). Worth a follow-up (bf16 support in the persistent kernel, or better wgmma bf16 autotune).

## Acceptance criteria (#1616)

- [x] Both failing fixed-length GQA sliding-window cases pass at `atol=1e-2, rtol=1e-2`.
- [x] Causal, left-window-only, no-window, and varlen cases remain green.
- [x] Right-window masking covered by the existing (now-passing) parametrizations in `test_gqa_sliding_window.py`.

### DIFF
--- a/tileops/kernels/attention/gqa_fwd.py
+++ b/tileops/kernels/attention/gqa_fwd.py
@@ -309,10 +309,10 @@ def _mha_fwd_wgmma_pipelined_kernel(batch: int,
                     mma1(v, v_shared, acc_s_cast, acc_o, k_idx, by, bz)
                 for i, j in T.Parallel(block_m, dim):
                     acc_o[i, j] /= logsum[i]
-                # Barriers guard the swizzled o_shared round-trip: without them
-                # warps read o_shared before peer warps finish writing it, a
-                # shared-memory race that can corrupt the output under WGMMA
-                # pipelining (num_stages >= 2).
+                # Guard the swizzled o_shared round-trip against a shared-memory
+                # race under WGMMA pipelining. Any non-zero barrier slot works;
+                # a bare T.sync_threads() aliases the implicit barrier-0 and is
+                # elided.
                 T.sync_threads(3, threads)
                 T.copy(acc_o, o_shared)
                 T.sync_threads(3, threads)
@@ -693,10 +693,10 @@ def _gqa_fwd_wgmma_pipelined_kernel(batch: int,
                     mma1(v, v_shared, acc_s_cast, acc_o, k_idx, by, bz)
                 for i, j in T.Parallel(block_m, dim):
                     acc_o[i, j] /= logsum[i]
-                # Barriers guard the swizzled o_shared round-trip: without them
-                # warps read o_shared before peer warps finish writing it, a
-                # shared-memory race that can corrupt the output under WGMMA
-                # pipelining (num_stages >= 2).
+                # Guard the swizzled o_shared round-trip against a shared-memory
+                # race under WGMMA pipelining. Any non-zero barrier slot works;
+                # a bare T.sync_threads() aliases the implicit barrier-0 and is
+                # elided.
                 T.sync_threads(3, threads)
                 T.copy(acc_o, o_shared)
                 T.sync_threads(3, threads)
--- a/tileops/kernels/attention/gqa_fwd.py
+++ b/tileops/kernels/attention/gqa_fwd.py
@@ -309,7 +309,13 @@ def _mha_fwd_wgmma_pipelined_kernel(batch: int,
                     mma1(v, v_shared, acc_s_cast, acc_o, k_idx, by, bz)
                 for i, j in T.Parallel(block_m, dim):
                     acc_o[i, j] /= logsum[i]
+                # Barriers guard the swizzled o_shared round-trip: without them
+                # warps read o_shared before peer warps finish writing it, a
+                # shared-memory race that can corrupt the output under WGMMA
+                # pipelining (num_stages >= 2).
+                T.sync_threads(3, threads)
                 T.copy(acc_o, o_shared)
+                T.sync_threads(3, threads)
                 T.copy(o_shared, output[bz, bx * block_m:(bx + 1) * block_m, by, :])
                 for i in T.Parallel(block_m):
                     logsum[i] = T.log2(logsum[i]) + scores_max[i] * scale
@@ -687,7 +693,13 @@ def _gqa_fwd_wgmma_pipelined_kernel(batch: int,
                     mma1(v, v_shared, acc_s_cast, acc_o, k_idx, by, bz)
                 for i, j in T.Parallel(block_m, dim):
                     acc_o[i, j] /= logsum[i]
+                # Barriers guard the swizzled o_shared round-trip: without them
+                # warps read o_shared before peer warps finish writing it, a
+                # shared-memory race that can corrupt the output under WGMMA
+                # pipelining (num_stages >= 2).
+                T.sync_threads(3, threads)
                 T.copy(acc_o, o_shared)
+                T.sync_threads(3, threads)
                 T.copy(o_shared, output[bz, bx * block_m:(bx + 1) * block_m, by, :])
                 for i in T.Parallel(block_m):
                     logsum[i] = T.log2(logsum[i]) + scores_max[i] * scale
--- a/tileops/kernels/attention/gqa_sliding_window_fwd.py
+++ b/tileops/kernels/attention/gqa_sliding_window_fwd.py
@@ -413,7 +413,13 @@ def _gqa_sw_fwd_wgmma_pipelined_kernel(
 
                 for i, j in T.Parallel(block_m, dim):
                     acc_o[i, j] /= logsum[i]
+                # Barriers guard the swizzled o_shared round-trip: without them
+                # warps read o_shared before peer warps finish writing it, a
+                # shared-memory race that corrupts the output under WGMMA
+                # pipelining (num_stages >= 2).
+                T.sync_threads(3, threads)
                 T.copy(acc_o, o_shared)
+                T.sync_threads(3, threads)
                 T.copy(o_shared, output[bz, bx * block_m:(bx + 1) * block_m, by, :])
                 for i in T.Parallel(block_m):
                     logsum[i] = T.log2(logsum[i]) + scores_max[i] * scale
--- a/tileops/kernels/attention/gqa_sliding_window_fwd.py
+++ b/tileops/kernels/attention/gqa_sliding_window_fwd.py
@@ -413,10 +413,10 @@ def _gqa_sw_fwd_wgmma_pipelined_kernel(
 
                 for i, j in T.Parallel(block_m, dim):
                     acc_o[i, j] /= logsum[i]
-                # Barriers guard the swizzled o_shared round-trip: without them
-                # warps read o_shared before peer warps finish writing it, a
-                # shared-memory race that corrupts the output under WGMMA
-                # pipelining (num_stages >= 2).
+                # Guard the swizzled o_shared round-trip against a shared-memory
+                # race under WGMMA pipelining. Any non-zero barrier slot works;
+                # a bare T.sync_threads() aliases the implicit barrier-0 and is
+                # elided.
                 T.sync_threads(3, threads)
                 T.copy(acc_o, o_shared)
                 T.sync_threads(3, threads)


lcy-seso:
Re-ran inside the CI image (torch 2.10.0+cu129, tilelang 0.1.11), H200:

- `tests/ops/attention/test_gqa_sliding_window.py`: **21/21 passed** (incl. the two previously-failing cases).
- `tests/ops/attention/test_mha.py` + `test_gqa.py` (host, torch 2.9.1+cu128): **55 passed, 4 skipped** — no regression from the defensive `gqa_fwd.py` change.

## Performance impact of the barriers (before/after, median over 100 iters, H200, same image)

| kernel | shape | cfg | before (ms) | after (ms) |
|---|---|---|---|---|
| GQASlidingWindowFwdWgmmaPipelined | fp16 wl=-1 wr=64 | 128/128/3 | 0.0167 | 0.0164 |
| GQASlidingWindowFwdWgmmaPipelined | bf16 wl=64 wr=64 | 128/128/3 | 0.0149 | 0.0146 |
| MHAFwdWgmmaPipelined | causal fp16 | 128/128/2 | 0.0162 | 0.0159 |
| GQAFwdWgmmaPipelined | causal fp16 | 128/128/2 | 0.0160 | 0.0157 |
| MHAFwdWgmmaPipelined | causal bf16 | 128/128/2 | 0.0161 | 0.0159 |
| GQAFwdWgmmaPipelined | causal bf16 | 128/128/2 | 0.0158 | 0.0158 |

The two `T.sync_threads` execute once per block at the very end (not in the KV loop), so the cost is below run-to-run noise (~1–2%; "after" is nominally faster, which is noise on these ~16 µs kernels). No measurable perf impact.

> "before" = the unguarded `origin/main` kernels, benchmarked by temporarily reverting then restoring. Note the sliding-window "before" output is numerically wrong (the bug itself); only latency is comparable.

zhen8838 (approved):
Approved. I rechecked the current head 7827f31bd63f8cac45a7e1e6efd9e6ff7f333dad against the prior Slock review; it is unchanged, and GitHub CI is green. No local tests or benchmarks run.

PR #1634, opened 2026-06-29
[Fix][Kernel] fp8_lighting_indexer: drop block_Q=1 from autotune sweep

lcy-seso:
> **Hotfix / stopgap.** Unblocks CI and closes a deadlock footgun. Does **not** fix the root design issue (`block_Q` is exposed as an independent autotune knob while it is coupled to `heads`); see Follow-up.

## Problem

`FP8LightingIndexerOp(..., tune=True)` hangs until the per-test timeout and fails the job. It surfaced on push-to-main / nightly but not on the introducing PR (#1632), because `gpu-smoke` runs `-m smoke` on PRs and `-m "smoke or full"` only on push — so the `full`-tier `tune=True` node never ran on that PR's gate.

## Root cause

Mapped on H200 across heads ∈ {32, 64, 128}: **the kernel deadlocks iff `block_Q == 1` and `num_stages >= 1`** (some warps never reach the software-pipeline barrier; the launch spins forever). An abandoned hung launch wedges the GPU, so every later autotune candidate times out too — the sweep never finishes within budget. Also a non-autotune footgun: the default `block_Q = 128 // heads` is `1` for `heads >= 65`.

## Fix

- Force `num_stages = 0` when `block_Q < 2` (guards autotune, manual config, and the default path).
- Harden default to `block_Q = max(1, 128 // heads)`.
- Drop `block_Q == 1` from `autotune_configs`.

Verified on H200: previously-hanging configs now run; `tune=True` test passes in ~9s warm / ~54s cold (was: timeout).

## Follow-up

Clean fix is to tune the q-tile and *derive* `block_Q` so degenerate configs can't be expressed (removes the guard). The silent deadlock is also a tilelang pipeline-codegen robustness bug.

### DIFF
--- a/tileops/kernels/fp8_lighting_indexer.py
+++ b/tileops/kernels/fp8_lighting_indexer.py
@@ -10,6 +10,13 @@ from tileops.kernels.kernel_base import Kernel
 
 __all__ = ["FP8LightingIndexerKernel"]
 
+# Minimum block_Q (query rows per block) for the software-pipelined copy/gemm.
+# With block_Q == 1 the pipeline deadlocks: some warps never arrive at the
+# pipeline barrier and the launch spins on the GPU forever. block_Q == 1 tiles
+# run unpipelined instead. Verified on sm90 across heads in {32, 64, 128}:
+# block_Q == 1 + num_stages >= 1 hangs; block_Q >= 2 runs (any tile width).
+_MIN_PIPELINED_BLOCK_Q = 2
+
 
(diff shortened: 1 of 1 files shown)


zhen8838 (approved):
Approved. I rechecked current head 817121994a19d4b56a4f4bfa0c8081bea296d86d: the fp8_lighting_indexer block_Q==1 deadlock guard is scoped correctly, CI including gpu-smoke is green, and no local tests or benchmarks were run.

PR #1635, opened 2026-06-29
[Chore][CI] nightly: drop the cache-warmup job

lcy-seso:
Remove the `cache-warmup` job from `nightly.yml`.

- Delete the `cache-warmup` job.
- `benchmark` becomes the entry job: drop `needs: [cache-warmup]` and the now-redundant `always() &&`.
- Renumber phase comments (benchmark→1, op_test→2, publish→2b, packaging→3).
- Keep `scripts/warmup_kernel_cache.py` + `scripts/conftest_warmup.py` for manual one-off warmup.

Redundant under the persistent `/ci-cache`: the benchmark phase autotunes at op construction (outside the timed region) and writes results back, so it self-warms across runs; benchmark numbers are unchanged. The job never gated anything (warmup step `continue-on-error`, benchmark `if: always()`) but turned the nightly red on runner network blips.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -26,41 +26,15 @@ env:
 
 jobs:
   # =========================================================================
-  # Phase 1 — Warm the kernel compilation cache
-  # =========================================================================
-  cache-warmup:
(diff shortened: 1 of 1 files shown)


Ibuki-wind (approved):
Clean — no issues.

PR #1647, opened 2026-07-06
[Fix][Attention] Make dense GQA uniform validation explicit

superAngGao:
## Summary
- keep public packed `backend="dense"` and `backend="fp8"` GQA prefill safe by validating uniform `cu_seqlens` by default before reshaping to BSHD
- add an explicit public `validate_uniform_cu_seqlens` switch so high-performance callers with known-uniform packed metadata can skip the CUDA value check without using a separate benchmark-only path
- keep `backend="auto"` requiring uniform validation because auto dispatch needs the `cu_seqlens` values to choose dense vs varlen
- declare `validate_uniform_cu_seqlens` in the op manifest and mark known-uniform dense/fp8 manifest workloads with `validate_uniform_cu_seqlens=False`
- update dense and fp8 prefill benchmarks to consume the manifest setting and call the same public op path users call
- add tests covering default dense/fp8 ragged rejection, explicit dense validation skip, auto validation requirements, varlen-style skip behavior, and the BSHD wrapper path

This addresses the dense packed-input contract concern from review while keeping benchmark and user calls on the same public API.

## Testing
- `python -m ruff check tileops/ops/attention/gqa.py tests/ops/attention/test_gqa.py benchmarks/ops/attention/bench_gqa.py benchmarks/ops/attention/bench_gqa_fp8.py benchmarks/ops/attention/manifest_params.py`
- `python -m py_compile tileops/ops/attention/gqa.py tests/ops/attention/test_gqa.py benchmarks/ops/attention/bench_gqa.py benchmarks/ops/attention/bench_gqa_fp8.py benchmarks/ops/attention/manifest_params.py`
- `git diff --check`
- Docker (`ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`): `python scripts/validate_manifest.py --check-op GroupedQueryAttentionPrefillFwdOp --levels schema,signature,shape,dtype,bench --strict`
- Docker (`ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`): `python -m pytest tests/ops/attention/test_gqa.py -q -k "explicit_varlen_backends or explicit_dense_backends or explicit_dense_can_skip or auto_backend_requires or bshd_wrapper or auto_backend_checks"`
- Docker as root for writable TileLang cache/benchmark log: `python -m pytest benchmarks/ops/attention/bench_gqa.py::test_gqa_prefill_fwd_bench[llama-3.1-8b-prefill-dense-float16] -q -s`

Host pytest was not used for validation because the host environment is missing `tvm.tirx`; the Docker runner image includes the TileOps CI dependencies.

### DIFF
--- a/benchmarks/ops/attention/bench_gqa.py
+++ b/benchmarks/ops/attention/bench_gqa.py
@@ -371,8 +371,8 @@ _GQA_PREFILL_FWD_BENCH_PARAMS = manifest_params(
 
 
 @pytest.mark.parametrize(
-    "batch, seq_len_q, seq_len_kv, heads, heads_kv, dim, causal, backend, sm_scale, softcap, "
-    "dtype, tune",
+    "batch, seq_len_q, seq_len_kv, heads, heads_kv, dim, causal, backend, "
+    "validate_uniform_cu_seqlens, sm_scale, softcap, dtype, tune",
     _GQA_PREFILL_FWD_BENCH_PARAMS,
 )
 def test_gqa_prefill_fwd_bench(
@@ -384,6 +384,7 @@ def test_gqa_prefill_fwd_bench(
     dim: int,
     causal: bool,
     backend: str,
+    validate_uniform_cu_seqlens: bool,
     sm_scale: Optional[float],
     softcap: Optional[float],
     dtype: torch.dtype,
@@ -406,6 +407,7 @@ def test_gqa_prefill_fwd_bench(
         dtype=dtype,
         tune=tune,
         backend=backend,
+        validate_uniform_cu_seqlens=validate_uniform_cu_seqlens,
         sm_scale=sm_scale,
         softcap=softcap,
     )
--- a/benchmarks/ops/attention/bench_gqa_fp8.py
+++ b/benchmarks/ops/attention/bench_gqa_fp8.py
@@ -21,6 +21,7 @@ class GQAFp8TensorCoreBenchCase:
     heads: int
     heads_kv: int
     dim: int
+    validate_uniform_cu_seqlens: bool
     out_dtype: torch.dtype
     label: str
 
@@ -44,6 +45,9 @@ def _manifest_cases() -> list[GQAFp8TensorCoreBenchCase]:
                     heads=heads,
                     heads_kv=heads_kv,
                     dim=dim,
+                    validate_uniform_cu_seqlens=workload.get(
+                        "validate_uniform_cu_seqlens", True
+                    ),
                     out_dtype=out_dtype,
                     label=f"{workload['label']}-{dtype_name}",
                 )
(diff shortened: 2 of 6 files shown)


zhen8838 (changes requested):
Requesting changes for the dense backend packed-input contract.

`GroupedQueryAttentionPrefillFwdOp.forward()` now only checks uniform `cu_seqlens` when `backend="auto"`; every explicit backend sets `is_uniform=True`. The dense path then reshapes packed inputs as `[batch, max_seqlen, ...]`. For same-total but non-uniform packed metadata, explicit `backend="dense"` will no longer reject the input and will silently split request boundaries incorrectly.

This reopens the ragged packed `cu_seqlens` issue we just fixed for dense GQA prefill. Please keep a uniform contract for explicit dense/fp8 paths, or otherwise restrict the fast path to internally constructed known-uniform metadata rather than trusting the public packed API unconditionally.

superAngGao:
Thanks, agreed. I updated the PR to keep the public packed dense/fp8 contract safe:

- `GroupedQueryAttentionPrefillFwdOp.forward()` now computes `is_uniform` for `backend="auto"`, `backend="dense"`, and `backend="fp8"`, so ragged packed metadata still reaches the existing dense rejection / FP8 uniform check.
- Explicit `varlen` and `sliding_window` still skip the uniform check because ragged metadata is their expected contract.
- The no-extra-uniform-check path is now restricted to `GroupedQueryAttentionFwdOp`, whose BSHD inputs are internally known dense/uniform; it calls the dense prefill kernel directly instead of re-entering the public packed prefill API.
- Added tests for dense/fp8 ragged rejection, varlen-style skip behavior, auto behavior, and the BSHD wrapper path.

Docker validation: full `tests/ops/attention/test_gqa.py` passes (`57 passed`).

superAngGao:
Updated again after rethinking the benchmark path: I removed the benchmark-only/private fast path idea.

The PR now keeps a single public call path:

- default `GroupedQueryAttentionPrefillFwdOp(..., backend="dense")` / `backend="fp8"` validates uniform `cu_seqlens`, so ragged packed metadata is still rejected safely;
- high-performance callers can explicitly opt out with `validate_uniform_cu_seqlens=False` when they already know the packed metadata is uniform;
- `backend="auto"` rejects `validate_uniform_cu_seqlens=False`, since auto dispatch needs the `cu_seqlens` values to choose dense vs varlen;
- dense prefill benchmarks now construct the same public op and call `op(*packed_inputs)`, just with `validate_uniform_cu_seqlens=False`.

So the benchmark no longer uses a separate hidden path from users, while the default public packed dense/fp8 contract remains safe.

zhen8838 (dismissed):
LGTM. The dense/fp8 packed paths now validate uniform cu_seqlens by default, auto cannot disable the check, and the BSHD wrapper bypasses the packed-cu_seqlens path entirely. The previous ragged packed boundary blocker is addressed.

michaelwithu (approved):
LGTM. Default paths still validate, auto forces it at construction, ragged still raises, varlen/sliding_window aren't misrouted. Tests and bench plumbing all check out — no functional issues.

zhen8838 (approved):
LGTM. The force-push only moves the uniform-cu_seqlens skip into explicit manifest/benchmark plumbing: public dense/fp8 packed paths still validate by default, backend=auto still requires validation, and the benchmark workloads marked false are known-uniform. CI is green on the current head.

PR #1652, opened 2026-07-06
[Perf][Attention] Warp-specialized batch=1 GQA decode kernel + workloads

zhen8838:
## Summary

Adds a warp-specialized (Hopper) batch=1 GQA decode kernel and routes the
`GroupedQueryAttentionDecodeWithKVCacheFwdOp` to it when the fast-path
preconditions hold. Migrated from a validated PoC and hardened for production
(runtime `real_seqlen_kv` masking + split redistribution).

Includes the batch=1 workloads and a proper FlashInfer baseline needed to
benchmark it.

## What's in it

**Workloads + benchmark baseline** (`manifest`, `bench_gqa_decode.py`)
- Eight batch=1 Qwen3-30B-A3B decode workloads (KV 1K–256K).
- FlashInfer baseline now uses `single_decode_with_kv_cache` (contiguous KV,
  no paging overhead) for batch=1 — the apples-to-apples comparison; the paged
  batched wrapper is kept for batch>1.

**Kernel + dispatch** (`gqa_decode_bs1.py`, `gqa.py`, tests, `kernel_map`)
- `GQADecodeBs1Kernel` (`supported_archs=[90]`): TMA + mbarrier + wgmma
  warp-specialized producer/consumer, online softmax in exp2 domain, fp32
  accumulators, ctx-split + LSE-combine.
- Op-level ctor gate selects the fast-path kernel only for batch=1, Hopper,
  fp16, D=128, softcap=0, group in [1,64]; everything else keeps the incumbent
  `GQADecodeKernel`. sm80/89 stay constructible.
- `forward()` dispatches by runtime `real_seqlen_kv`: `no_split` below 1024,
  ctx-split above (largest of {32,16,8} that divides `real`, balanced load).
  Non-aligned lengths (e.g. 2049) go through this kernel, not a fallback.

## Correctness

14 tests pass (`tests/ops/attention/test_gqa_decode.py`), including runtime
context-switch and non-divisible-length cases; atol=rtol=1e-2 vs the fp32
reference. Hopper-only tests skip on other arches.

## Performance (H200 @ 1500MHz, CUPTI, H=32 Hkv=4 D=128, fp16)

| KV len | new (µs) | incumbent | FlashInfer single | new/incumbent | new/FI |
|-------:|---------:|----------:|------------------:|--------------:|-------:|
| 1024   | 7.97     | 16.16     | 10.35             | 0.49          | 0.77   |
| 2048   | 8.94     | 10.65     | 11.12             | 0.84          | 0.80   |
| 3072   | 11.93    | 24.07     | 11.86             | 0.50          | 1.01   |
| 4096   | 10.70    | 12.78     | 12.40             | 0.84          | 0.86   |
| 8192   | 14.00    | 16.20     | 14.81             | 0.86          | 0.95   |
| 32768  | 29.49    | 36.52     | 35.29             | 0.81          | 0.84   |
| 131072 | 77.53    | 119.36    | 83.88             | 0.65          | 0.92   |

Faster than the incumbent across the board; faster than or on par with
FlashInfer `single_decode` everywhere (roughly tied at 3072).

## Trust-model note

Kept as a single PR per request. It mixes a manifest workloads change (normally
a separate human-reviewed PR) with the implementation; the manifest edits are
the eight workloads plus one `source.kernel_map` line (the latter is within the
implementation-PR carve-out).

### DIFF
--- a/tests/ops/attention/test_gqa_decode.py
+++ b/tests/ops/attention/test_gqa_decode.py
@@ -5,6 +5,7 @@ import torch
 from tests.test_base import FixtureBase, TestBase
 from tileops.kernels.attention.gqa_decode import GQADecodeKernel
 from tileops.ops import GroupedQueryAttentionDecodeWithKVCacheFwdOp
+from tileops.utils import is_hopper
 from workloads.attention.gqa_decode import (
     GroupedQueryAttentionDecodeTest as _GroupedQueryAttentionDecodeTestWorkload,
 )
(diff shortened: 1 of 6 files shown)


tests/ops/attention/test_gqa_decode.py
```diff
@@ -5,10 +5,15 @@
 from tests.test_base import FixtureBase, TestBase
 from tileops.kernels.attention.gqa_decode import GQADecodeKernel
 from tileops.ops import GroupedQueryAttentionDecodeWithKVCacheFwdOp
+from tileops.utils import is_hopper
 from workloads.attention.gqa_decode import (
     GroupedQueryAttentionDecodeTest as _GroupedQueryAttentionDecodeTestWorkload,
 )
 
+# Qwen3-30B-A3B GQA decode shape (heads / heads_kv = 8) that selects the batch=1
```
zhen8838: Don't let these global static variables appear.

tileops/kernels/attention/gqa_decode_bs1.py
```diff
@@ -0,0 +1,294 @@
+"""Warp-specialized batch=1 GQA decode kernel (Hopper), context-split.
```
zhen8838, line 1: There are too many comments in the whole pr. Try to simplify them.

tests/ops/attention/test_gqa_decode.py
zhen8838: Done — removed the module-level shape constants and inlined the literals directly in each test, matching the existing tests in this file.

tileops/kernels/attention/gqa_decode_bs1.py
zhen8838: Done — trimmed the module/class docstrings and inline comments throughout the kernel (producer/consumer, barriers, redistribution, split-candidate notes). Tests still pass and ruff is clean.

superAngGao (commented):
Thanks for the PR. The fast path and the Hopper-only dispatch direction look good to me, and CI is green. I only have a few non-blocking notes/questions:

1. Non-blocking: in the batch=1 FlashInfer baseline, `q_s/k_s/v_s` are captured outside `run_fn`, so the `q/k/v` arguments passed by `bench_kernel` are ignored. Since `bench_kernel` clones tensor args to provide fresh addresses across iterations, it may be slightly more faithful to reshape inside `run_fn` from its arguments.

2. Non-blocking question: for Hopper batch=1 fp16 dim=128 cases, the op now routes to `gqa_decode_bs1_kernel`. If a caller overrides only `gqa_decode_kernel` in `kernel_map`, these cases will still use the default bs1 kernel. That may be fine, but it might be worth documenting the two dispatch keys or making the intended override behavior explicit.

3. Non-blocking robustness note: `GQADecodeBs1Kernel` is currently guarded by the Op-level fast-path predicate. If we expect direct kernel construction to be supported, adding local validation for the narrow supported surface (batch=1/fp16/dim=128/softcap=0/group size) could make misuse fail earlier.

None of these should block the PR from my side.

superAngGao (approved):

lcy-seso (approved):

PR #1665, opened 2026-07-07
[Perf][Pool] Optimize avg_pool2d kernels

### LINKED ISSUE #1654: [PERF][AVG_POOL2D] improve nchw pooling kernel performance
### Symptom / Motivation
The Nightly benchmark run https://github.com/tile-ai/TileOPs/actions/runs/28716827107 completed successfully, but `AvgPool2dFwdOp` is consistently slower than the torch baseline on the manifest workloads.

Nightly benchmark results:

| workload | TileOps latency | torch latency | gap |
| --- | ---: | ---: | ---: |
| `vision-3x3-s2` fp16 | 0.0616 ms | 0.0192 ms | 3.21x slower |
| `vision-5x5-s2` fp16 | 0.0637 ms | 0.0199 ms | 3.20x slower |
| `ceil-divisor-bf16` | 0.0544 ms | 0.0185 ms | 2.94x slower |

Local reproduction on `tileops-dev` with `CUDA_VISIBLE_DEVICES=1` on H200 reproduced the same issue:

| workload | TileOps latency | torch latency | gap |
| --- | ---: | ---: | ---: |
| `vision-3x3-s2` fp16 | 0.0366 ms | 0.0095 ms | 3.85x slower |
| `vision-5x5-s2` fp16 | 0.0829 ms | 0.0246 ms | 3.37x slower |
| `ceil-divisor-bf16` | 0.0656 ms | 0.0194 ms | 3.38x slower |

### Root Cause Analysis
Profiling indicates that the gap is inside the TileOps CUDA kernel, not Python dispatch or kernel launch overhead.

For `vision-3x3-s2`, the workload is:

- input shape: `[2, 64, 112, 112]`
- kernel size: `(3, 3)`
- stride: `(2, 2)`
- padding: `(1, 1)`
- output shape: `[2, 64, 56, 56]`
- autotuned TileOps config: `{"block_m": 64, "block_c": 32, "threads": 256}`

TileOps launches `_avg_pool2d_main_kernel` with grid `(49, 2, 2)` and block size `256`, for `196` total blocks. On an H200 with 132 SMs, ncu reports only `0.19` waves/SM and `19.57%` achieved occupancy. The current TileOps mapping tiles over `(output spatial, channel)`, so adjacent channel lanes in NCHW layout access memory with a large `H * W` stride. ncu reports heavy uncoalesced access: `3,723,000 / 3,971,200` global excessive sectors (`94%`) and only `2.0 / 32` useful bytes per global sector.

The torch kernel for the same case launches grid `(392, 1, 1)` with block size `1024`, reaches `1.48` waves/SM and `79.44%` achieved occupancy, and has much better global sector utilization.

Key nsys/ncu comparison for `vision-3x3-s2`:

| metric | TileOps | torch |
| --- | ---: | ---: |
| nsys median kernel time | 35,872 ns | 8,304 ns |
| ncu duration | 37.28 us | 10.05 us |
| grid size | 196 blocks | 392 blocks |
| block size | 256 threads | 1024 threads |
| waves/SM | 0.19 | 1.48 |
| achieved occupancy | 19.57% | 79.44% |
| SM busy | 6.02% | 67.92% |
| memory throughput | 86.70 GB/s | 321.45 GB/s |
| useful global sector bytes | 2.0 / 32 bytes | 14.8 / 32 bytes |

### Related Files
- `tileops/kernels/pool/avg_pool2d.py`
- `tileops/ops/pool.py`
- `benchmarks/ops/bench_pool.py`
- `tileops/manifest/pool.yaml`

## Goal
Improve `AvgPool2dFwdOp` performance on the manifest workloads, especially the common NCHW fp16 vision cases, so TileOps no longer trails the torch baseline by roughly 3x.

## Plan

1. Add or redesign a common-case NCHW avg_pool2d fast path that maps adjacent lanes primarily over output spatial / width positions instead of adjacent channels, improving global memory coalescing.
2. Increase launched work for small image-pooling shapes so H200 is better occupied; use `vision-3x3-s2` as the first reference case.
3. Keep the current generic path for padding, `ceil_mode`, `count_include_pad`, and `divisor_override` semantics, but route common `count_include_pad=True` and no-override workloads through a lighter specialized kernel when valid.
4. Re-run `benchmarks/ops/bench_pool.py::test_avg_pool2d_bench` and compare against the torch baseline.
5. Re-profile the optimized `vision-3x3-s2` case with nsys/ncu to verify kernel time, waves/SM, achieved occupancy, and global sector utilization improve.

## Constraints
Maintain PyTorch-compatible `avg_pool2d` semantics for padding, `ceil_mode`, `count_include_pad`, and `divisor_override`. Do not change the public `AvgPool2dFwdOp` API or manifest workload contract.

## Acceptance Criteria
- [ ] `python -m pytest benchmarks/ops/bench_pool.py::test_avg_pool2d_bench -vvs` runs successfully on H200.
- [ ] `vision-3x3-s2` TileOps median kernel time improves from the current ~35.9 us nsys baseline.
- [ ] `vision-3x3-s2` ncu achieved occupancy and/or waves/SM improves from the current `19.57%` and `0.19` baseline.
- [ ] Global memory coalescing improves versus the current `2.0 / 32` useful bytes per sector signal.
- [ ] The report or PR body includes before/after benchmark numbers for all three `AvgPool2dFwdOp` manifest workloads.

RMLYC:
Closes #1654

## Summary

- Add an `AvgPool2dSpatialKernel` fast path for the common NCHW case with `ceil_mode=False`, `count_include_pad=True`, and no `divisor_override`.
- Rewrite the general `AvgPool2dKernel` fallback to use a flat `N*C*H_out*W_out` launch grid and remove the `block_c` tuning dimension.
- Extend `AvgPool2dFixture` coverage for `ceil_mode`, `count_include_pad`, and `divisor_override` combinations.

## Test plan

- [x] Cleared TileLang caches before the final benchmark/profiling run:

```bash
rm -rf /home/lyc/.tilelang \
  /home/lyc/Project/TileOps-workspace2/.tmp/tilelang_clear_sync_cache \
  /home/lyc/Project/TileOps-workspace2/.tmp/tilelang_ifelse_cache
```

- [x] Functional tests:

```bash
CUDA_VISIBLE_DEVICES=1 conda run -n tileops-dev \
  python -m pytest tests/ops/test_pool.py -k avg_pool2d -vvs
```

Result:

```text
23 passed, 28 deselected, 14 warnings in 57.08s
```

- [x] Benchmark:

```bash
CUDA_VISIBLE_DEVICES=1 conda run -n tileops-dev \
  python -m pytest benchmarks/ops/bench_pool.py::test_avg_pool2d_bench -vvs
```

Result:

```text
3 passed, 15 warnings in 34.03s
```

- [x] Whitespace check:

```bash
git diff --check
```

## Benchmark

Final benchmark on NVIDIA H200:

| workload | kernel path | before TileOps ms | after TileOps ms | torch-ref ms | after vs before | after vs torch |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| `vision-3x3-s2` | `AvgPool2dSpatialKernel` | `0.0505` | `0.0046` | `0.0094` | `10.98x faster` | `2.04x faster` |
| `vision-5x5-s2` | `AvgPool2dSpatialKernel` | `0.0372` | `0.0045` | `0.0102` | `8.27x faster` | `2.27x faster` |
| `ceil-divisor-bf16` | `AvgPool2dKernel` flat general | `0.0149` | `0.0038` | `0.0080` | `3.92x faster` | `2.11x faster` |

Autotune best configs from the final run:

| workload | kernel | best config |
| --- | --- | --- |
| `vision-3x3-s2` | `AvgPool2dSpatialKernel` | `{"block_m": 256, "threads": 256}` |
| `vision-5x5-s2` | `AvgPool2dSpatialKernel` | `{"block_m": 512, "threads": 256}` |
| `ceil-divisor-bf16` | `AvgPool2dKernel` | `{"block_m": 256, "threads": 256}` |

Notes on the baseline values:

- The `vision-3x3-s2` and `vision-5x5-s2` "before" values are from the original TileOps kernel measured during the avg_pool2d performance investigation.
- The `ceil-divisor-bf16` "before" value is the local pre-flat-general fallback re-run used for the review comparison.

### Profiling summary

The benchmark improvement is consistent with fixed-config `nsys` and `ncu` data.

For `vision-3x3-s2`, `ncu` shows that TileOps executes much less work than torch:

| metric | TileOps | torch |
| --- | ---: | ---: |
| Duration | `5.12 us` | `10.05 us` |
| Waves/SM | `1.48` | `1.48` |
| Achieved Occupancy | `73.61%` | `79.44%` |
| SM Busy | `44.53%` | `67.92%` |
| Memory Throughput | `628.30 GB/s` | `321.45 GB/s` |
| Issued Instructions | `1.09M` | `4.24M` |

Torch has higher SM busy, but it issues about `3.9x` more instructions. The TileOps fast path avoids the general pooling divisor/control path and reaches about `2x` higher effective memory throughput.

For `vision-5x5-s2`, fixed-config `nsys` reports:

| metric | TileOps | torch |
| --- | ---: | ---: |
| nsys median | `3.808 us` | `9.664 us` |
| nsys avg | `3.822 us` | `9.672 us` |
| Grid / Block | `392 x 256` | `196 x 1024` |
| Registers/thread | `29` | `32` |

For `ceil-divisor-bf16`, the new flat general fallback is also faster than torch:

| metric | TileOps | torch |
| --- | ---: | ---: |
| nsys median | `3.456 us` | `7.488 us` |
| ncu Duration | `4.48 us` | `8.96 us` |
| Grid / Block | `914 x 256` | `229 x 1024` |
| Waves/SM | `0.87` | `0.87` |
| Achieved Occupancy | `68.44%` | `73.10%` |
| SM Busy | `43.02%` | `65.50%` |
| Eligible Warps/Scheduler | `1.42` | `3.65` |
| Memory Throughput | `404.46 GB/s` | `203.60 GB/s` |
| Issued Instructions | `875K` | `3.28M` |

Torch again keeps the SMs busier, but it issues about `3.75x` more instructions. The flat general kernel removes the old `block_c` split, launches over `N*C*H_out*W_out`, and completes the same workload with a shorter instruction path and higher effective memory throughput.

### DIFF
--- a/tests/ops/test_pool.py
+++ b/tests/ops/test_pool.py
@@ -6,7 +6,7 @@ import torch.nn.functional as F
 
 from tests.test_base import FixtureBase, TestBase
 from tileops.kernels.kernel_base import Kernel
-from tileops.kernels.pool import AvgPool2dKernel, AvgPool3dKernel
+from tileops.kernels.pool import AvgPool2dKernel, AvgPool2dSpatialKernel, AvgPool3dKernel
 from tileops.ops import AvgPool1dFwdOp, AvgPool2dFwdOp, AvgPool3dFwdOp
 
 
@@ -279,7 +279,7 @@ def test_avg_pool2d_dispatches_kernel() -> None:
         stride=(2, 2),
         padding=(1, 1),
     )
-    assert isinstance(op.kernel, AvgPool2dKernel)
+    assert isinstance(op.kernel, AvgPool2dSpatialKernel)
 
 
 @pytest.mark.smoke
--- a/tests/ops/test_pool.py
+++ b/tests/ops/test_pool.py
@@ -186,6 +186,31 @@ class AvgPool2dFixture(FixtureBase):
                 marks=pytest.mark.full,
                 id="full-divisor-override-bf16",
             ),
+            pytest.param(
+                1, 7, 9, 10, (3, 3), (2, 2), (1, 1), False, False, None, torch.float16, False,
+                marks=pytest.mark.full,
+                id="full-no-ceil-no-pad-count-fp16",
+            ),
+            pytest.param(
+                1, 5, 10, 11, (3, 3), (2, 2), (1, 1), True, True, None, torch.float32, False,
+                marks=pytest.mark.full,
+                id="full-ceil-pad-count-fp32",
+            ),
+            pytest.param(
+                2, 6, 9, 13, (2, 3), (2, 2), (0, 1), True, True, 7, torch.bfloat16, False,
+                marks=pytest.mark.full,
+                id="full-ceil-pad-count-divisor-bf16",
+            ),
+            pytest.param(
+                1, 9, 11, 12, (3, 5), (2, 3), (1, 2), True, False, 7, torch.float16, False,
+                marks=pytest.mark.full,
+                id="full-ceil-no-pad-count-divisor-fp16",
+            ),
+            pytest.param(
+                1, 8, 9, 9, (3, 3), (2, 2), (1, 1), False, False, 7, torch.float16, False,
+                marks=pytest.mark.full,
+                id="full-no-ceil-no-pad-count-divisor-fp16",
+            ),
         ]),
     ]
 
--- a/tests/ops/test_pool.py
+++ b/tests/ops/test_pool.py
@@ -6,7 +6,7 @@ import torch.nn.functional as F
 
 from tests.test_base import FixtureBase, TestBase
 from tileops.kernels.kernel_base import Kernel
-from tileops.kernels.pool import AvgPool2dKernel, AvgPool2dSpatialKernel, AvgPool3dKernel
+from tileops.kernels.pool import AvgPool2dSpatialKernel, AvgPool3dKernel
 from tileops.ops import AvgPool1dFwdOp, AvgPool2dFwdOp, AvgPool3dFwdOp
 
 
@@ -305,6 +305,22 @@ def test_avg_pool2d_dispatches_kernel() -> None:
         padding=(1, 1),
     )
     assert isinstance(op.kernel, AvgPool2dSpatialKernel)
+
+
+@pytest.mark.smoke
+def test_avg_pool2d_rejects_non_positive_output_size() -> None:
+    with pytest.raises(ValueError, match="output size must be greater than zero"):
+        AvgPool2dFwdOp(
+            n=1,
+            c_in=1,
+            h_in=2,
+            w_in=2,
+            kernel_size=(5, 5),
+            stride=(1, 1),
+            padding=(0, 0),
+            ceil_mode=False,
+            count_include_pad=True,
+        )
 
 
 @pytest.mark.smoke
--- a/tileops/kernels/pool/__init__.py
+++ b/tileops/kernels/pool/__init__.py
@@ -1,5 +1,10 @@
 from .avg_pool1d import AvgPool1dKernel
-from .avg_pool2d import AvgPool2dKernel
+from .avg_pool2d import AvgPool2dKernel, AvgPool2dSpatialKernel
 from .avg_pool3d import AvgPool3dKernel
 
-__all__ = ["AvgPool1dKernel", "AvgPool2dKernel", "AvgPool3dKernel"]
+__all__ = [
+    "AvgPool1dKernel",
+    "AvgPool2dKernel",
+    "AvgPool2dSpatialKernel",
+    "AvgPool3dKernel",
+]
--- a/tileops/kernels/pool/avg_pool2d.py
+++ b/tileops/kernels/pool/avg_pool2d.py
@@ -33,102 +33,66 @@ def _avg_pool2d_kernel(
     accum_dtype = "float"
     out_h = pool_output_dim(h_in, kernel_h, stride_h, pad_h, ceil_mode)
     out_w = pool_output_dim(w_in, kernel_w, stride_w, pad_w, ceil_mode)
+    total_output = n * c_in * out_h * out_w
 
     @tilelang.jit(out_idx=[1], compile_flags=["-O3", "-DENABLE_BF16"])
-    def _avg_pool2d_func(block_m: int, block_c: int, threads: int):
+    def _avg_pool2d_func(block_m: int, threads: int):
         @T.prim_func
         def _avg_pool2d_main(
             x: T.Tensor((n, c_in, h_in, w_in), dtype),  # type: ignore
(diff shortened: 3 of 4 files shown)


superAngGao (commented):
Thanks for the optimization. The fast-path split and the flatter fallback look directionally good to me. I have two contract-level comments before approval:

1. Please sync the manifest with the new dispatch surface. `AvgPool2dFwdOp.default_kernel_map` now contains both `avg_pool2d_kernel` and `avg_pool2d_spatial_kernel`, but `tileops/manifest/pool.yaml` still declares only `avg_pool2d_kernel`. The ops design docs say `default_kernel_map` should match `source.kernel_map` verbatim, and the benchmark/reporting path is manifest-driven here, so the new spatial kernel should be listed there as well. It may also be worth exporting `AvgPool2dSpatialKernel` from the top-level `tileops.kernels` package for consistency with the other pool kernels.

2. Please double-check the custom `kernel_map` partial override semantics. Because `dispatch_kernel()` fills missing entries from `default_kernel_map`, a caller that passes only `{"avg_pool2d_kernel": MyKernel}` will still get the default `AvgPool2dSpatialKernel` inserted. For fast-path parameter sets, dispatch will then choose the default spatial kernel and bypass the user-provided general kernel. I think user-supplied overrides should either explicitly opt into the spatial fast path or fall back to the provided `avg_pool2d_kernel`; otherwise tests/experiments that override the existing kernel key can silently stop exercising their replacement. A small smoke test for this case, plus one for non-fast-path dispatch, would lock the behavior down.

The core indexing/divisor logic looks reasonable to me, and the output-size guard from the earlier bot comment appears to be addressed.

zhen8838 (approved):
LGTM. Static review found no blocker on head 7dafac82c0a470a5367b7c1cef91855ff1c28b8b; GitHub CI is green for gpu-smoke and preflight checks. I did not run local tests or benchmarks.

Ibuki-wind (approved):
Clean under a correctness-focused review. I treated manifest/export synchronization and the custom kernel_map override semantics as follow-up contract work rather than blockers for this PR.

The fast-path guard is narrow enough for the fixed-divisor spatial kernel, the general fallback keeps the PyTorch divisor/boundary semantics while flattening the launch grid, and the new output-size guard matches PyTorch error behavior for non-positive outputs. The added avg_pool2d cases cover the changed dispatch paths and the ceil/count/divisor combinations touched by the fallback rewrite.

I did not run local GPU tests or benchmarks; GitHub gpu-smoke/preflight are green and the PR includes H200 benchmark results.

Issue #1663, opened 2026-07-07
[Feat][Ops]:  BmmFwdOp: manifest + kernel/op + tests + bench (fp16/bf16/fp8)


RuneFang:
## Parent Issue
Sub-issue of https://github.com/tile-ai/TileOPs/issues/400. 

## Task Type

- [ ] **Manifest / Spec** (required for any public Op)
- [ ] **Kernel / Op Implementation**
- [ ] **Correctness Tests / Workloads**
- [ ] **Benchmark / Performance**

## Description

Track the BMM slice of #400. The 2 nominal ops listed there (`bmm_fp16`, `bmm_fp8`) map to **2 TileOPs ops**, following the same input-inferred / call-time-JIT design as `GemmOp`:

| nominal op | Realized as | status |
|---|---|---|
| `bmm_fp16` | `BmmFwdOp` (fp16 + bf16) | **:white_check_mark:** |
| `bmm_fp8`  | `BmmFp8Op` (fp8, epilogue-scale + block-scaled paths) | **:white_check_mark:** |

**Semantics (settled):**

- Strict `torch.bmm` contract — 3D×3D only, `a=[B, M, K]`, `b=[B, K, N]` → `d=[B, M, N]`. No batch-dim broadcasting, no `matmul`-style shape promotion.
- Transpose-free primary layout matching `torch.bmm` inputs; NT / other layouts are separate manifest entries if demand appears.
- Fully input-inferred: `batch`, `m`, `n`, `k`, and dtype are read from `a` / `b` at `forward()` time; nothing pinned at construction. Roofline binds these dims via func mode.
- Reference API: `torch.bmm`. Dtype coverage: `float16`, `bfloat16`. Accumulation in `float32`.

## Checklist

- [ ] Public Op has a `tileops/manifest/` entry (or updates an existing entry).
- [ ] Manifest `signature` declares inputs, outputs, params, shape rules, and dtype coverage.
- [ ] Manifest `workloads` declare benchmark shapes/dtypes; unit-test edge cases are not generated from manifest workloads.
- [ ] Manifest `roofline` is present and consumable by `op.eval_roofline()`.
- [ ] Manifest `source` declares kernel, op, test, bench, and `source.kernel_map` when dispatching kernels.
- [ ] Op constructor, `forward()`, and `default_kernel_map` match the manifest entry.
- [ ] Tests use an independent reference implementation and cover relevant FP16/BF16 and edge cases.
[...]

PR #1676, opened 2026-07-08
[Feat][BMM] Add FP16 BF16 BMM op

RuneFang:
## Summary

- Add `BmmFwdOp` for strict 3D-3D batched matmul (`torch.bmm` semantics, no broadcasting) covering `float16` / `bfloat16`.
- Add a simple TileLang BMM kernel: CTA grid `(ceildiv(N, block_n), ceildiv(M, block_m), B)`, every CTA loops over full K with a multi-stage `T.Pipelined` K loop and FP32 accumulate.
- Add manifest workloads, roofline accounting, correctness tests, and manifest-driven BMM benchmarks against `torch.bmm` (cuBLAS).

#1663 

## Performance (NVIDIA H20-3e)

`pytest benchmarks/ops/bench_bmm.py` — all 16 workloads beat `torch.bmm`, speedup 1.02×–1.20×.
Shape is `[B, M, N, K]`.

| shape                  | dtype       | tileops TFLOPS  | cuBLAS TFLOPS   | speedup       |
| ---------------------- | ----------- | --------------: | --------------: | ------------: |
| [8, 128, 128, 128]     | fp16 / bf16 | 12.60 / 12.69   | 11.93 / 11.89   | 1.06× / 1.07× |
| [8, 1024, 1024, 1024]  | fp16 / bf16 | 134.04 / 134.31 | 129.34 / 129.49 | 1.04× / 1.04× |
| [16, 512, 512, 512]    | fp16 / bf16 | 118.66 / 119.12 | 101.77 / 102.22 | 1.17× / 1.17× |
| [32, 256, 256, 256]    | fp16 / bf16 | 96.50 / 96.64   | 80.39 / 80.38   | 1.20× / 1.20× |
| [4, 4096, 4096, 4096]  | bf16        | 142.14          | 139.56          | 1.02×         |
| [8, 2048, 2048, 2048]  | fp16 / bf16 | 140.69 / 140.83 | 133.84 / 134.13 | 1.05× / 1.05× |
| [64, 128, 2048, 128]   | fp16 / bf16 | 117.56 / 117.65 | 110.62 / 110.27 | 1.06× / 1.07× |
| [64, 128, 128, 2048]   | fp16 / bf16 | 105.67 / 105.84 | 103.52 / 103.88 | 1.02× / 1.02× |
| [128, 512, 512, 2048]  | bf16        | 140.84          | 133.84          | 1.05×         |

### DIFF
--- a/benchmarks/ops/bench_bmm.py
+++ b/benchmarks/ops/bench_bmm.py
@@ -4,9 +4,9 @@ import pytest
 import torch
 
 from benchmarks.benchmark_base import BenchmarkBase, BenchmarkReport
+from tests.ops.test_bmm import BmmTest
 from tileops.manifest import load_workloads
 from tileops.ops import BmmFwdOp
-from workloads.bmm import BmmTest
 
 _OP_NAME = "BmmFwdOp"
 
@@ -14,13 +14,6 @@ _DTYPE_MAP = {
     "bfloat16": torch.bfloat16,
     "float16": torch.float16,
 }
-
-
-class _BmmTestBaseline(BmmTest):
-    """Adds baseline ref_program for benchmark profiling."""
-
-    def ref_program(self, a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
-        return torch.bmm(a, b)
 
 
 class BmmBenchmark(BenchmarkBase[BmmTest]):
@@ -66,7 +59,7 @@ def _manifest_params() -> list:
 @pytest.mark.parametrize("batch, m, n, k, dtype_str", _manifest_params())
 def test_bmm_bench(batch: int, m: int, n: int, k: int, dtype_str: str) -> None:
     dtype = _DTYPE_MAP[dtype_str]
-    test = _BmmTestBaseline(batch, m, n, k, dtype)
+    test = BmmTest(batch, m, n, k, dtype)
     a, b = test.gen_inputs()
 
     op = BmmFwdOp(tune=True)
--- a/benchmarks/ops/bench_bmm.py
+++ b/benchmarks/ops/bench_bmm.py
@@ -1,0 +1,85 @@
+from typing import Optional
+
+import pytest
+import torch
+
+from benchmarks.benchmark_base import BenchmarkBase, BenchmarkReport
+from tileops.manifest import load_workloads
+from tileops.ops import BmmFwdOp
+from workloads.bmm import BmmTest
+
+_OP_NAME = "BmmFwdOp"
+
+_DTYPE_MAP = {
+    "bfloat16": torch.bfloat16,
+    "float16": torch.float16,
+}
+
+
+class _BmmTestBaseline(BmmTest):
+    """Adds baseline ref_program for benchmark profiling."""
+
+    def ref_program(self, a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
+        return torch.bmm(a, b)
+
+
+class BmmBenchmark(BenchmarkBase[BmmTest]):
+    """Reads FLOP/byte counts from the Op's manifest-derived roofline.
+
+    ``BmmFwdOp`` is input-inferred, so ``eval_roofline()`` is valid only
+    after a forward has bound ``batch/m/n/k/dtype``; the benchmark calls it
+    lazily.
+    """
+
+    _roofline_cache: Optional[tuple[float, float]] = None
+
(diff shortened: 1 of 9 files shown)


workloads/bmm.py
```diff
@@ -0,0 +1,19 @@
+import torch
+
+from workloads.workload_base import WorkloadBase
+
+
+class BmmTest(WorkloadBase):
```
zhen8838: This workload class should not be named `BmmTest`.

benchmarks/ops/bench_bmm.py
```diff
@@ -0,0 +1,85 @@
...
+
+_OP_NAME = "BmmFwdOp"
+
+_DTYPE_MAP = {
+    "bfloat16": torch.bfloat16,
+    "float16": torch.float16,
+}
+
+
+class _BmmTestBaseline(BmmTest):
```
zhen8838: _BmmTestBaseline and BmmTest do the same thing and repeat

tests/ops/test_bmm.py
```diff
@@ -0,0 +1,124 @@
+import pytest
+import torch
+
+from tests.test_base import FixtureBase, TestBase
+from tileops.ops import BmmFwdOp
+from workloads.bmm import BmmTest as _BmmTestWorkload
+
+
+class BmmTest(_BmmTestWorkload, TestBase):
```
zhen8838: After modifying the BmmTest class name of workload, you don't need to import it through the alias.

zhen8838 (changes requested):

benchmarks/ops/bench_bmm.py
RuneFang: Thanks, I've updated the code according to your review

zhen8838 (changes requested):
Two blockers remain before this is ready:

1. `BmmFwdOp.forward()` still uses `sig = (a.shape, b.shape, a.dtype)` for the active fast path, so a same-shape repeat call can change only `b.dtype` and skip `_validate_dtypes(a, b)`. That bypasses the manifest `b: same_as(a)` contract. Please include `b.dtype` in the active signature and add a regression where a valid fp16/fp16 call is followed by an invalid same-shape fp16/bf16 call on the same op instance.

2. The fix adds `BmmKernel`'s `k % 16 == 0` requirement, but the public op/manifest contract still reads like strict `torch.bmm` with arbitrary K. Please either expose and validate the K divisibility contract consistently in the manifest shape rules, op validation, and negative tests, or implement K-tail support in the kernel.

The fake registration, block_k filtering, compiled-kernel cache, and workload/benchmark naming cleanup are moving in the right direction. GitHub preflight/manifest/benchmark-contract checks are green; `gpu-smoke` was still queued when I checked.

RuneFang:
> Two blockers remain before this is ready:
> 
> 1. `BmmFwdOp.forward()` still uses `sig = (a.shape, b.shape, a.dtype)` for the active fast path, so a same-shape repeat call can change only `b.dtype` and skip `_validate_dtypes(a, b)`. That bypasses the manifest `b: same_as(a)` contract. Please include `b.dtype` in the active signature and add a regression where a valid fp16/fp16 call is followed by an invalid same-shape fp16/bf16 call on the same op instance.
> 2. The fix adds `BmmKernel`'s `k % 16 == 0` requirement, but the public op/manifest contract still reads like strict `torch.bmm` with arbitrary K. Please either expose and validate the K divisibility contract consistently in the manifest shape rules, op validation, and negative tests, or implement K-tail support in the kernel.
> 
> The fake registration, block_k filtering, compiled-kernel cache, and workload/benchmark naming cleanup are moving in the right direction. GitHub preflight/manifest/benchmark-contract checks are green; `gpu-smoke` was still queued when I checked.

Thanks for the review. Great comments; code updated accordingly.

zhen8838 (approved):
Static review looks good. The two previous blockers are fixed at head 1a158d2c53a2b7deb706e2695b2ce094ff8ce939:

- The BMM op cache/signature now includes both input dtypes, so same-shape calls with different `b.dtype` no longer reuse an incompatible compiled kernel.
- The `K % 16 == 0` contract is now enforced in op validation, reflected in the manifest `shape_rules`, and covered by a negative test.

I did not run local tests or benchmarks. At approval time, GitHub CI was green except `gpu-smoke`, which was still queued.

RuneFang:
trigger CI

zhen8838 (approved):

PR #1678, opened 2026-07-09
Fix eager trace pass imports

### LINKED ISSUE #1669: [Refactor][Import] GEMM tracing should not make ordinary op imports require tvm.tirx
## Summary

Importing ordinary TileOPs operator classes can currently require `tvm.tirx` because `GemmKernel` imports `tileops.trace` at module import time, and the trace package eagerly imports the TIR lowering pass.

This makes non-trace / non-GEMM workflows unexpectedly depend on TileLang's bundled TVM namespace during import and also affects manifest validation environments that only need to inspect op signatures.

## Observed import chain

```text
tileops.ops import ...
  -> an op imports tileops.kernels.<family>
  -> Python initializes tileops/kernels/__init__.py
  -> tileops/kernels/__init__.py imports GemmKernel
  -> tileops/kernels/gemm.py imports tileops.trace.trace
  -> tileops/trace/api.py imports passes
  -> tileops/trace/passes.py imports tvm.tirx
```

Relevant files:

- `tileops/kernels/__init__.py` eagerly imports `GemmKernel` / `GemvKernel`.
- `tileops/kernels/gemm.py` has top-level `from tileops.trace import trace`.
- `tileops/trace/api.py` imports `passes` at module import time.
- `tileops/trace/passes.py` imports `tvm.tirx` and `tvm.tirx.stmt_functor`.

## Why this is surprising

Most TileOPs operators do not need to import TVM/TIR APIs just to expose an op class. GEMM's optional timeline tracing path uses the trace API, but regular untraced GEMM and unrelated ops should not require `tvm.tirx` at import time.

In environments where TileLang's bundled TVM path is not active, targeted manifest validation can fail before it reaches the intended op, with errors like:

```text
ModuleNotFoundError: No module named 'tvm.tirx'
```

## Acceptance criteria

- A lightweight import/signature validation path for unrelated ops no longer fails due to missing `tvm.tirx`.
- `GemmKernel` default untraced use still works.
- Trace-enabled GEMM still imports/uses `tvm.tirx` when lowering markers.
- Add a small regression test if practical, e.g. mocking/blocking `tvm.tirx` while importing a non-trace op or `tileops.kernels.kernel_base` / non-GEMM op module.

superAngGao:
## Summary
- lazily import trace lowering passes from the public trace API
- keep ordinary tileops.ops imports from eager-loading tileops.trace.passes
- add a smoke regression test for the import boundary

Fixes #1669

## Tests
- python -m ruff check tileops/trace/api.py tileops/trace/passes.py tests/trace/test_imports.py
- python -m pytest -q tests/trace/test_imports.py
- docker run --rm --gpus all --ipc host -v /home/ga/TileOPs-issue-1669/TileOPs:/workspace/TileOPs --entrypoint bash ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10 -lc 'cd /workspace/TileOPs && python -m pytest -q tests/trace/test_imports.py tests/trace/test_payload.py'

### DIFF
--- a/tests/trace/test_imports.py
+++ b/tests/trace/test_imports.py
@@ -1,0 +1,38 @@
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+pytestmark = pytest.mark.smoke
+
+
+def test_ops_import_does_not_eagerly_load_trace_passes() -> None:
+    repo_root = Path(__file__).resolve().parents[2]
+    env = os.environ.copy()
+    env["PYTHONPATH"] = (
+        str(repo_root)
+        if not env.get("PYTHONPATH")
+        else f"{repo_root}{os.pathsep}{env['PYTHONPATH']}"
+    )
+    code = r"""
+import builtins
+
+original_import = builtins.__import__
+
+
+def guarded_import(name, *args, **kwargs):
+    if name == "tileops.trace.passes":
+        raise RuntimeError("unexpected eager import of tileops.trace.passes")
+    return original_import(name, *args, **kwargs)
+
+
+builtins.__import__ = guarded_import
+import tileops.ops  # noqa: F401
+
+import sys
+
+assert "tileops.trace.passes" not in sys.modules
+"""
+    subprocess.run([sys.executable, "-c", code], check=True, cwd=repo_root, env=env)
--- a/tileops/trace/api.py
+++ b/tileops/trace/api.py
@@ -38,7 +38,6 @@ End-to-end::
 
 import os
 
-from . import passes
 from .decode import decode as _decode
 from .markers import (
     DEFAULT_LANE,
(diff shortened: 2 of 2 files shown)


zhen8838 (changes requested):
The issue/PR premise needs to be corrected before this can be approved. The PR frames this as fixing ordinary imports that fail because tvm.tirx is unavailable, but TileOps' supported dependency stack pins TileLang and tileops/trace/passes.py already imports tilelang before tvm.tirx specifically so TileLang can populate the bundled TVM path. If tvm.tirx is missing only in a lightweight/local environment outside the supported TileOps TileLang/runner setup, that is not a TileOps import contract bug. Please either provide a reproduction in the supported TileOps environment, or reframe the issue/PR away from Fixes #1669 as a narrower cleanup that reduces eager import surface. Also fix the PR title format; validate-pr-title currently fails because the title lacks the required [Type] / [Type][Scope] prefix.

superAngGao:
Closing this PR because the original Fixes #1669 premise does not hold for the supported TileOps runner/TileLang stack. The lazy import change may still be worth revisiting later as a narrower refactor to reduce import side effects, but it should not be framed as fixing a supported-environment import bug.

PR #1680, opened 2026-07-09
[CI] Parallelize GPU smoke tests with pytest-xdist to speed up the GPU smoke tests

RuneFang:
This PR speeds up the GPU smoke suite by enabling `pytest-xdist` in `.github/workflows/gpu-smoke.yml`, so tests are dispatched across multiple worker processes  instead of running strictly serially. To keep parallel workers from stomping on each other during tilelang compilation, `tests/conftest.py` gives each xdist worker its own `TILELANG_TMP_DIR` subdirectory (`gwN/`) . Worker count and distribution mode are controlled by the new `XDIST_N` / `XDIST_DIST_MODE` env vars (default `auto` + `loadfile`). Local testing shows the end-to-end smoke run drops to under 5 minutes.

### DIFF
--- a/.github/workflows/gpu-smoke.yml
+++ b/.github/workflows/gpu-smoke.yml
@@ -477,7 +477,7 @@ jobs:
           # call) fails as a timeout with a thread dump instead of wedging the job. --no-deps
           # tileops install doesn't bring this; pin via constraints.txt so the install is
           # deterministic (the runner image should bake it and drop this install — follow-up).
-          python3 -m pip install -q -c constraints.txt pytest-timeout
+          python3 -m pip install -q -c constraints.txt pytest-timeout pytest-xdist
           current_dir="$(pwd)"
           export PYTHONPATH="${current_dir}${PYTHONPATH:+:$PYTHONPATH}"
           echo "PYTHONPATH=$PYTHONPATH"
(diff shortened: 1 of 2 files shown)


zhen8838 (changes requested):
Requesting changes for the current head `43fab1f5da6ccc988542be2f81316b203c78ce35`.

The new xdist path defaults GPU smoke to CPU-count parallelism:

```bash
XDIST_N="${PYTEST_XDIST_N:-auto}"
python3 -m pytest ... -n "${XDIST_N}" --dist="${XDIST_DIST_MODE}"
```

For TileOps GPU smoke, `-n auto` is too aggressive as the CI default. It scales by CPU workers, not GPU capacity, while `tests/conftest.py` itself notes that several workers share the same GPU(s). The per-worker `TILELANG_TMP_DIR` partition is useful for compile-temp isolation, and `gc.collect()` / `torch.cuda.empty_cache()` after each test can reduce leftover cache pressure, but neither prevents concurrent cold compilation / kernel launch / peak tensor allocation across many workers on the same GPU. That makes the smoke job vulnerable to non-deterministic OOM/resource contention and can turn the signal flakier rather than faster.

Please make the default conservative: for example a fixed small worker count such as 2, or compute a cap from GPU count via `nvidia-smi -L` and keep `PYTEST_XDIST_N` as an explicit override. The important part is that CI should not default to CPU-count GPU concurrency.

zhen8838:
Please close this PR for now.
Our GPU smoke environment currently has a single GPU instance, so xdist workers would contend for the same device rather than provide meaningful test scale-out.
The earlier slow smoke runs were caused by permission isolation preventing the runner from reusing the TileLang JIT cache, not by a lack of pytest parallelism. We will fix cache access and reuse separately. Once that is resolved, we can re-evaluate the appropriate test execution model.

PR #1684, opened 2026-07-13
[Perf][Linear-Attn] Add DeltaNet decode raw fast path

superAngGao:
## Summary

- Add a Hopper raw TileLang fast path for `DeltaNetDecodeOp` when `dtype in {fp16, bf16}` and `DK=DV=128`.
- Reuse the Gated DeltaNet decode warp-per-Vtile strategy for the ungated DeltaNet recurrence:
  - one warp handles one `(batch, head, V tile)`
  - `v_tile=16`, `raw_group_size=2`
  - K/Q staged once in shared memory
  - each lane keeps its K slice in fp32 local storage
  - two-lane shuffle reductions produce each output value
- Keep dispatch conservative: fp32 stays on the no-TF32 FP32 kernel; non-128 shapes and non-Hopper devices stay on the existing TileLang fallback.
- Add fp16/bf16 `128x128` raw decode smoke/full coverage without changing manifest-owned workload definitions.

## Benchmark

Environment: H200, `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`, CUDA 12.9 runner stack.

Compared raw fast path against the current generic fallback, which already includes the 2026-07-10 Pass-1 shared-memory state read optimization.

| dtype | shape `(B,H,DK,DV)` | raw (ms) | generic (ms) | speedup |
| fp16 | `(1,32,128,128)` | `0.007456` | `0.015251` | `2.05x` |
| fp16 | `(8,32,128,128)` | `0.013632` | `0.017964` | `1.32x` |
| fp16 | `(32,32,128,128)` | `0.033224` | `0.039338` | `1.18x` |
| bf16 | `(1,32,128,128)` | `0.008390` | `0.015515` | `1.85x` |
| bf16 | `(8,32,128,128)` | `0.013002` | `0.019029` | `1.46x` |
| bf16 | `(32,32,128,128)` | `0.034677` | `0.053997` | `1.56x` |

## Test plan

- [x] pre-commit passed in CI.
- [x] GPU smoke passed: 19 tests.
- [x] Structural readiness passed: manifest validation and benchmark contract tests.
- [x] Test node delta: tests/ops/test_deltanet_recurrence.py 15 -> 32 (+17).

### Verification

All Docker validation used the prebuilt TileOpsGov runner image `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10` as-is, with no tilelang/torch/dependency install or upgrade. The image reports `torch 2.10.0+cu129` and `tilelang 0.1.11+cu129.git65dbc983`.

```bash
python scripts/validate_manifest.py   --levels schema,signature,shape,dtype,bench   --strict   --check-op DeltaNetDecodeOp
```

Result: passed.

```bash
python -m ruff check   tileops/kernels/deltanet_recurrence.py   tileops/ops/deltanet_recurrence.py   tests/ops/test_deltanet_recurrence.py   benchmarks/ops/bench_deltanet_recurrence.py
```

Result: passed.

```bash
git diff --check
```

Result: passed.

```bash
python -m pytest -q benchmarks/tests   tests/test_validate_manifest.py::TestIntegration::test_validator_passes_on_current_codebase   --tb=short -p no:cacheprovider
```

Result: `9 passed`.

```bash
LD_PRELOAD=/lib/x86_64-linux-gnu/libcuda.so.1 python -m pytest -q   tests/ops/test_deltanet_recurrence.py -m smoke --tb=short -s -p no:cacheprovider
```

Result: `19 passed, 13 deselected, 14 warnings`.

```bash
LD_PRELOAD=/lib/x86_64-linux-gnu/libcuda.so.1 python -m pytest -q   tests/ops/test_deltanet_recurrence.py --tb=short -s -p no:cacheprovider
```

Result: `32 passed, 14 warnings`.

### DIFF
--- a/tests/ops/test_deltanet_recurrence.py
+++ b/tests/ops/test_deltanet_recurrence.py
@@ -3,7 +3,10 @@
 import pytest
 import torch
 
+import tileops.ops.deltanet_recurrence as deltanet_ops
 from tests.test_base import FixtureBase, TestBase
+from tileops.kernels.deltanet_recurrence import DeltaNetDecodeRawCudaFlaStyleKernel
+from tileops.kernels.kernel_base import Kernel
 from tileops.ops import DeltaNetDecodeOp
 from workloads.deltanet import DeltaNetDecodeTest as _DeltaNetDecodeTestWorkload
 
@@ -152,6 +155,228 @@ def test_deltanet_decode_rejects_manifest_shape_mismatch() -> None:
 
     with pytest.raises(ValueError, match="q must have shape"):
         op.forward(q, k, v, beta, state)
+
+
+def _skip_unless_raw_cuda_decode_supported() -> None:
+    if not torch.cuda.is_available():
+        pytest.skip("CUDA required for raw DeltaNet decode smoke coverage")
+    try:
+        sm_version = deltanet_ops.get_sm_version()
+    except Exception as exc:
+        pytest.skip(f"could not query CUDA architecture: {exc}")
+    if sm_version not in DeltaNetDecodeRawCudaFlaStyleKernel.supported_archs:
+        pytest.skip(
+            f"raw DeltaNet decode requires SM90, got SM{sm_version}"
+        )
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
+def test_deltanet_decode_raw_cuda_real_128x128_smoke(dtype: torch.dtype) -> None:
+    """PR smoke must compile and execute the real raw CUDA 128x128 fast path."""
+    _skip_unless_raw_cuda_decode_supported()
+
+    torch.manual_seed(42)
+    test = DeltaNetDecodeTest(2, 4, 128, 128, dtype)
+    op = DeltaNetDecodeOp(2, 4, 128, 128, dtype, tune=False)
+
+    assert isinstance(op.kernel, DeltaNetDecodeRawCudaFlaStyleKernel)
+    test.check(op, *test.gen_inputs(), **_get_tolerances(dtype))
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16])
+def test_deltanet_decode_raw_cuda_real_128x128_multi_step_smoke(
+    dtype: torch.dtype,
+) -> None:
+    """PR smoke must exercise raw CUDA state propagation across decode steps."""
+    _skip_unless_raw_cuda_decode_supported()
+
+    torch.manual_seed(42)
+    num_steps = 8
+    B, H, DK, DV = 2, 4, 128, 128
+    op = DeltaNetDecodeOp(B, H, DK, DV, dtype, tune=False)
+    tols = _get_tolerances(dtype)
+
+    assert isinstance(op.kernel, DeltaNetDecodeRawCudaFlaStyleKernel)
+
+    state_op = torch.zeros(B, H, DK, DV, device="cuda", dtype=dtype)
+    state_ref = torch.zeros(B, H, DK, DV, device="cuda", dtype=dtype)
+
+    for _ in range(num_steps):
+        q = torch.randn(B, H, DK, device="cuda", dtype=dtype) * 0.1
+        k = torch.randn(B, H, DK, device="cuda", dtype=dtype) * 0.1
+        v = torch.randn(B, H, DV, device="cuda", dtype=dtype) * 0.1
+        beta = torch.rand(B, H, device="cuda", dtype=dtype) * 0.5
+
+        o_ref, state_ref = deltanet_decode_torch(q, k, v, beta, state_ref)
+        o_ref = o_ref.to(dtype)
+        state_ref = state_ref.to(dtype)
+
+        with torch.no_grad():
+            o_op, state_op = op(q, k, v, beta, state_op)
+
+        torch.testing.assert_close(o_op, o_ref, **tols)
+        torch.testing.assert_close(state_op, state_ref, **tols)
+
+
+class _DispatchMarkerKernel(Kernel):
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__()
+        self.args = args
+        self.kwargs = kwargs
+
+    def forward(self, *args, **kwargs):
+        raise NotImplementedError
+
+
+class _DefaultDispatchKernel(_DispatchMarkerKernel):
+    pass
+
+
+class _FP32DispatchKernel(_DispatchMarkerKernel):
+    pass
+
+
+class _RawDispatchKernel(_DispatchMarkerKernel):
+    pass
+
+
+def _dispatch_kernel_map() -> dict:
+    return {
+        "DeltaNetDecodeKernel": _DefaultDispatchKernel,
+        "DeltaNetDecodeFP32Kernel": _FP32DispatchKernel,
+        "DeltaNetDecodeRawCudaFlaStyleKernel": _RawDispatchKernel,
+    }
+
+
+def _mock_dispatch_arch(monkeypatch, sm_version: int) -> None:
+    monkeypatch.setattr(deltanet_ops, "get_sm_version", lambda: sm_version)
+
+
+@pytest.mark.parametrize(("dtype", "tune"), [
+    pytest.param(torch.float16, False, marks=pytest.mark.smoke, id="smoke-fp16"),
+    pytest.param(torch.bfloat16, False, marks=pytest.mark.smoke, id="smoke-bf16"),
+    pytest.param(torch.bfloat16, True, marks=pytest.mark.full, id="full-bf16-tuned"),
+])
+def test_deltanet_decode_raw_cuda_dispatch_selects_raw_on_supported_sm90(
+    monkeypatch,
+    dtype: torch.dtype,
+    tune: bool,
+) -> None:
+    _mock_dispatch_arch(monkeypatch, 90)
+
+    op = DeltaNetDecodeOp(
+        1,
+        32,
+        128,
+        128,
+        dtype=dtype,
+        kernel_map=_dispatch_kernel_map(),
+        tune=tune,
+    )
+
+    assert isinstance(op.kernel, _RawDispatchKernel)
+    assert op.kernel.kwargs["tune"] is tune
+
+
+@pytest.mark.smoke
+def test_deltanet_decode_raw_cuda_dispatch_falls_back_on_unsupported_sm(
+    monkeypatch,
+) -> None:
+    _mock_dispatch_arch(monkeypatch, 80)
+
+    op = DeltaNetDecodeOp(
+        1,
+        32,
+        128,
+        128,
+        dtype=torch.bfloat16,
+        kernel_map=_dispatch_kernel_map(),
+        tune=False,
+    )
+
+    assert isinstance(op.kernel, _DefaultDispatchKernel)
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize(("dim_k", "dim_v"), [(64, 128), (128, 64)])
+def test_deltanet_decode_raw_cuda_dispatch_falls_back_on_non_128_shapes(
+    monkeypatch,
+    dim_k: int,
+    dim_v: int,
+) -> None:
+    _mock_dispatch_arch(monkeypatch, 90)
+
+    op = DeltaNetDecodeOp(
+        1,
+        32,
+        dim_k,
+        dim_v,
+        dtype=torch.bfloat16,
+        kernel_map=_dispatch_kernel_map(),
+        tune=False,
+    )
+
+    assert isinstance(op.kernel, _DefaultDispatchKernel)
+    assert op.kernel.kwargs["tune"] is False
+
+
+@pytest.mark.smoke
+def test_deltanet_decode_raw_cuda_dispatch_uses_fp32_kernel_for_fp32(
+    monkeypatch,
+) -> None:
+    _mock_dispatch_arch(monkeypatch, 90)
+
+    op = DeltaNetDecodeOp(
+        1,
+        32,
+        128,
+        128,
+        dtype=torch.float32,
+        kernel_map=_dispatch_kernel_map(),
+        tune=False,
+    )
+
+    assert isinstance(op.kernel, _FP32DispatchKernel)
+    assert op.kernel.kwargs["tune"] is False
+
+
+@pytest.mark.smoke
+def test_deltanet_decode_raw_cuda_config_requires_full_warp_mapping() -> None:
+    with pytest.raises(ValueError, match="threads .* must equal raw_group_size \\* v_tile"):
+        DeltaNetDecodeRawCudaFlaStyleKernel(
+            1,
+            32,
+            128,
+            128,
+            dtype="bfloat16",
+            config={
+                "threads": 16,
+                "v_tile": 16,
+                "raw_group_size": 2,
+                "raw_maxrregcount": 146,
+            },
+        )
+
+
+@pytest.mark.smoke
+def test_deltanet_decode_raw_cuda_config_requires_two_lane_group() -> None:
+    with pytest.raises(ValueError, match="raw_group_size must equal 2"):
+        DeltaNetDecodeRawCudaFlaStyleKernel(
+            1,
+            32,
+            128,
+            128,
+            dtype="bfloat16",
+            config={
+                "threads": 32,
+                "v_tile": 8,
+                "raw_group_size": 4,
+                "raw_maxrregcount": 146,
+            },
+        )
 
 
 if __name__ == "__main__":
--- a/tests/ops/test_deltanet_recurrence.py
+++ b/tests/ops/test_deltanet_recurrence.py
@@ -74,6 +74,8 @@ class DeltaNetDecodeFixture(FixtureBase):
             pytest.param(1, 4, 64, 64, torch.bfloat16, False, marks=pytest.mark.smoke),
             pytest.param(2, 8, 64, 64, torch.float32, False, marks=pytest.mark.full),
             pytest.param(2, 4, 128, 128, torch.float32, False, marks=pytest.mark.full),
+            pytest.param(2, 4, 128, 128, torch.float16, False, marks=pytest.mark.full),
+            pytest.param(2, 4, 128, 128, torch.bfloat16, False, marks=pytest.mark.full),
             pytest.param(2, 8, 64, 64, torch.float16, False, marks=pytest.mark.full),
             pytest.param(2, 8, 64, 64, torch.bfloat16, False, marks=pytest.mark.full),
         ]),
--- a/tileops/kernels/deltanet_recurrence.py
+++ b/tileops/kernels/deltanet_recurrence.py
@@ -450,7 +450,8 @@ class DeltaNetDecodeRawCudaFlaStyleKernel(Kernel):
         from tilelang.profiler import do_bench
 
         best_time = float("inf")
-        best_config = self.default_config
+        best_config: Optional[dict] = None
+        failures: list[str] = []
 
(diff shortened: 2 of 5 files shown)


tests/ops/test_deltanet_recurrence.py
```diff
@@ -74,6 +74,8 @@ class DeltaNetDecodeFixture(FixtureBase):
             pytest.param(1, 4, 64, 64, torch.bfloat16, False, marks=pytest.mark.smoke),
             pytest.param(2, 8, 64, 64, torch.float32, False, marks=pytest.mark.full),
             pytest.param(2, 4, 128, 128, torch.float32, False, marks=pytest.mark.full),
+            pytest.param(2, 4, 128, 128, torch.float16, False, marks=pytest.mark.full),
```
Ibuki-wind, line 80: These new 128x128 fp16/bf16 correctness rows are marked `full`, but PR `gpu-smoke` runs only `smoke` on pull_request, so CI does not automatically exercise the new SM90 raw dispatch path. The PR body's H200 full-test result is not a regression guard for future changes. → Add a smoke-level targeted test that verifies or forces `DeltaNetDecodeOp` selects the raw kernel for low-precision `DK=DV=128` on supported SM90, while keeping the existing full rows for broader correctness.

tileops/ops/deltanet_recurrence.py
```diff
@@ -28,6 +30,25 @@ class DeltaNetDecodeOp(Op):
...
+        return sm_version in DeltaNetDecodeRawCudaFlaStyleKernel.supported_archs
+
+    @staticmethod
+    def _should_use_raw_cuda_decode(
+        dim_k: int,
+        dim_v: int,
+        dtype: torch.dtype,
+        tune: bool,
+    ) -> bool:
+        if tune or dtype not in (torch.float16, torch.bfloat16) or dim_k != 128 or dim_v != 128:
```
Ibuki-wind: `_should_use_raw_cuda_decode` adds architecture-gated raw dispatch, but there is no focused regression test for the selection contract. A future edit can accidentally route `tune=True`, fp32, non-128 shapes, or unsupported SMs into the raw kernel without a small unit test catching it. → Add dispatch tests mirroring the GatedDeltaNet raw path: supported SM90 + fp16/bf16 + 128x128 selects raw, and unsupported SM / `tune=True` / fp32 / non-128 shapes fall back.

tileops/kernels/deltanet_recurrence.py
```diff
@@ -273,6 +376,78 @@ def forward(
...
+        self.dim_v = dim_v
+        self.dtype = dtype
+        self.init_config(config, tune=False)
+        if self.config["raw_group_size"] != 2:
+            raise ValueError(
+                "raw_group_size must equal 2 because this kernel uses fixed "
+                "two-lane shuffle reductions."
+            )
+        required_threads = self.config["raw_group_size"] * self.config["v_tile"]
+        if self.config["threads"] != required_threads:
```
Ibuki-wind, line 416: This fixed warp-lane mapping depends on `threads == raw_group_size * v_tile`, but the new DeltaNet raw kernel does not add a regression test for that config invariant. The equivalent GatedDeltaNet raw path has focused tests for these failure modes. → Add tests that constructing `DeltaNetDecodeRawCudaFlaStyleKernel` raises for `raw_group_size != 2` and for `threads != raw_group_size * v_tile`, so this low-level lane mapping cannot be broken silently.

Ibuki-wind (changes requested):
### Overall
The raw fast path needs automated guards for the dispatch path and fixed lane-mapping invariants before this lands.

superAngGao:
Thanks for the careful review. I pushed follow-up commit `f09a2cb` to address the raw decode guard concerns.

What changed:

- Added smoke-level dispatch tests that force/mock SM90 and verify low-precision `DK=DV=128` selects `DeltaNetDecodeRawCudaFlaStyleKernel`.
- Added focused fallback tests for unsupported SM, fp32, and non-128 shapes.
- Added raw config invariant tests for `raw_group_size != 2` and `threads != raw_group_size * v_tile`, matching the GatedDeltaNet guard style.

One intentional behavior update versus the original comment: `tune=True` no longer falls back. We decided to support that path for DeltaNet raw decode. The raw kernel now has a narrow manual autotune space over `raw_maxrregcount` candidates `(0, 128, 132, 146, 160)`, while keeping the warp-lane mapping invariants fixed (`raw_group_size=2`, `v_tile=16`, `threads=32`). The dispatch test includes a full `tune=True` case to ensure the tune request reaches the raw kernel.

Local validation:

- `python -m ruff check tileops/kernels/deltanet_recurrence.py tileops/ops/deltanet_recurrence.py tests/ops/test_deltanet_recurrence.py`
- Docker smoke: `python3 -m pytest -q tests/ops/test_deltanet_recurrence.py -m smoke --tb=short -s` -> `14 passed, 13 deselected`
- Docker full file: `python3 -m pytest -q tests/ops/test_deltanet_recurrence.py --tb=short -s` -> `27 passed`
- Also verified a real `DeltaNetDecodeRawCudaFlaStyleKernel(..., tune=True)` construction runs through the raw autotune path successfully.

tileops/ops/deltanet_recurrence.py
```diff
@@ -28,6 +30,25 @@ class DeltaNetDecodeOp(Op):
...
+        except Exception:
+            return False
+        return sm_version in DeltaNetDecodeRawCudaFlaStyleKernel.supported_archs
+
+    @staticmethod
+    def _should_use_raw_cuda_decode(
+        dim_k: int,
+        dim_v: int,
+        dtype: torch.dtype,
+        tune: bool,
```
Ibuki-wind: `tune` is no longer used in `_should_use_raw_cuda_decode` now that the raw kernel supports autotune, so the signature no longer reflects the dispatch logic. → Remove the `tune` parameter from this helper and its call site, or add a short comment if the unused parameter is intentionally kept for symmetry.

tests/ops/test_deltanet_recurrence.py
```diff
@@ -133,5 +138,165 @@ def test_deltanet_decode_multi_step(
...
+    return {
+        "DeltaNetDecodeKernel": _DefaultDispatchKernel,
+        "DeltaNetDecodeFP32Kernel": _FP32DispatchKernel,
+        "DeltaNetDecodeRawCudaFlaStyleKernel": _RawDispatchKernel,
+    }
+
+
+def _mock_dispatch_arch(monkeypatch, sm_version: int) -> None:
+    monkeypatch.setattr(deltanet_ops, "get_sm_version", lambda: sm_version)
+    monkeypatch.setattr(op_base_ops, "get_sm_version", lambda: sm_version)
```
Ibuki-wind: `_mock_dispatch_arch` patches both `deltanet_ops.get_sm_version` and `op_base_ops.get_sm_version`, but the reason for the second patch is not obvious from this local dispatch test. → Add a short comment explaining which path in `dispatch_kernel` needs the `op_base_ops` patch, or drop it if only `deltanet_ops.get_sm_version` is required.

tests/ops/test_deltanet_recurrence.py
```diff
@@ -133,5 +138,165 @@ def test_deltanet_decode_multi_step(
         torch.testing.assert_close(state_op, state_ref, **tols)
 
 
+class _DispatchMarkerKernel:
+    supported_archs = None
```
Ibuki-wind: `supported_archs` is not used by these marker kernels, so this helper exposes more fake Kernel surface than the tests need. → Keep the marker minimal and remove unused interface fields unless a specific dispatch path requires them.

tileops/kernels/deltanet_recurrence.py
```diff
@@ -273,6 +376,128 @@ def forward(
...
+                )(config["threads"])
+                t = do_bench(
+                    lambda _fn=fn: _fn(q, k, v, beta, state),
+                    warmup=10,
+                    rep=20,
+                )
+                if t < best_time:
+                    best_time = t
+                    best_config = config
+            except Exception:
```
Ibuki-wind: `_autotune_raw_cuda` intentionally skips configs that fail to compile or benchmark, but swallowing every `Exception` without context makes tuning failures hard to diagnose if all candidates fail or one config is unexpectedly broken. → Add a short comment explaining that config failures are expected during autotune, or record enough failure context to report/debug an all-failed tuning run.

Ibuki-wind (commented):
Non-blocking cleanup: please also update the PR validation section. It still reports the earlier `18 passed`, but the PR now adds dispatch/config tests, so the validation evidence should reflect the current test set.

zhen8838 (changes requested):
The raw CUDA path is still not exercised by PR GPU Smoke.

The new SM90 tests in `tests/ops/test_deltanet_recurrence.py` replace the selected kernel with `_RawDispatchKernel`, so they verify only routing. The actual fp16/bf16 128x128 correctness and multi-step cases remain `full`; pull-request GPU Smoke runs `-m smoke`, so it neither compiles nor executes the new raw kernel.

Mark representative real fp16 and bf16 128x128 cases as `smoke` (including the multi-step state case). Keep the marker-kernel tests as focused dispatch coverage, but they cannot substitute for an end-to-end raw-kernel regression.

Ibuki-wind (dismissed):
### Overall
Clean on the current head; the raw-dispatch and lane-mapping coverage gaps are addressed, and the remaining cleanup comments are non-blocking.

tests/ops/test_deltanet_recurrence.py
superAngGao: Addressed in `5b82dab0`: I added smoke-level real raw-kernel coverage for fp16 and bf16 `DK=DV=128`. The new tests explicitly assert `DeltaNetDecodeOp` selected `DeltaNetDecodeRawCudaFlaStyleKernel`, then run correctness against the torch reference. The original full rows remain for broader matrix coverage.

tileops/ops/deltanet_recurrence.py
superAngGao: Addressed across `f09a2cb1` and `5b82dab0`: the dispatch tests cover supported SM90 low-precision 128x128 -> raw, unsupported SM -> default, fp32 -> FP32 kernel, non-128 shapes -> default, and the tuned raw path. `5b82dab0` also removes the now-unused `tune` argument from `_should_use_raw_cuda_decode()`.

tileops/kernels/deltanet_recurrence.py
superAngGao: Addressed in `f09a2cb1`: added focused config invariant tests for both `threads != raw_group_size * v_tile` and `raw_group_size != 2`, so the fixed two-lane warp mapping is guarded directly.

tileops/ops/deltanet_recurrence.py
superAngGao: Addressed in `5b82dab0`: removed the unused `tune` parameter from `_should_use_raw_cuda_decode()` and its call site. Raw dispatch is now based only on dtype, DK/DV, and supported architecture.

tests/ops/test_deltanet_recurrence.py
superAngGao: Addressed in `5b82dab0`: dropped the `op_base_ops.get_sm_version` patch and import. The marker dispatch tests only need to mock the DeltaNet op module's architecture query for raw/default selection.

tests/ops/test_deltanet_recurrence.py
superAngGao: Addressed in `5b82dab0`: the marker kernel now subclasses the normal `Kernel` base and no longer declares its own `supported_archs` field. That keeps the fake surface minimal while still satisfying the dispatch machinery through the base class.

tileops/kernels/deltanet_recurrence.py
superAngGao: Addressed in `5b82dab0`: `_autotune_raw_cuda()` now records failed candidate configs with exception type/message, reports skipped candidates, and prints an explicit all-failed fallback message before using the default config.

superAngGao:
Follow-up pushed in `5b82dab0`.

This addresses the remaining raw-kernel smoke coverage gap:

- Added smoke-level real raw CUDA correctness coverage for fp16 and bf16 `DK=DV=128`.
- Added smoke-level real raw CUDA multi-step state propagation coverage for fp16 and bf16 `DK=DV=128`.
- These tests explicitly assert that `DeltaNetDecodeOp` selected `DeltaNetDecodeRawCudaFlaStyleKernel`, so PR `gpu-smoke` now compiles and executes the actual raw kernel rather than only marker-kernel dispatch tests.
- Kept the marker-kernel dispatch tests as focused routing coverage.

Also cleaned up the non-blocking review items:

- Removed unused `tune` from `_should_use_raw_cuda_decode()`.
- Removed the extra `op_base_ops.get_sm_version` mock.
- Made marker kernels inherit the normal `Kernel` base instead of carrying a custom `supported_archs` field.
- Added raw autotune failure context for skipped/all-failed candidates.
- Updated the PR validation section.

Validation in the official runner image:

```bash
python -m ruff check \
  tileops/kernels/deltanet_recurrence.py \
  tileops/ops/deltanet_recurrence.py \
  tests/ops/test_deltanet_recurrence.py \
  benchmarks/ops/bench_deltanet_recurrence.py && git diff --check
```
Result: passed

```bash
LD_PRELOAD=/lib/x86_64-linux-gnu/libcuda.so.1 python -m pytest -q \
  tests/ops/test_deltanet_recurrence.py -m smoke --tb=short -s -p no:cacheprovider
```
Result: `18 passed, 13 deselected, 14 warnings in 13.88s`

```bash
LD_PRELOAD=/lib/x86_64-linux-gnu/libcuda.so.1 python -m pytest -q \
  tests/ops/test_deltanet_recurrence.py --tb=short -s -p no:cacheprovider
```
Result: `31 passed, 14 warnings in 5.16s`

superAngGao:
Resolved the merge conflict by rebasing onto current `upstream/main` in `7d73a8a7`.

Conflict notes:
- Kept the new manifest-driven DeltaNet decode benchmark fixture from `main`.
- Migrated the PR raw-path benchmark matrix into `tileops/manifest/linear_attention.yaml` workloads instead of restoring the old hand-written benchmark params.
- Added `DeltaNetDecodeRawCudaFlaStyleKernel` to the manifest `source.kernel_map` so the manifest matches the op/kernel surface.
- Preserved the new main-side manifest shape mismatch smoke test alongside the PR raw dispatch/config tests and real raw-kernel smoke tests.

Re-run in the official runner image:
- `ruff ... && git diff --check`: passed
- `tests/ops/test_deltanet_recurrence.py -m smoke`: `19 passed, 13 deselected`
- full `tests/ops/test_deltanet_recurrence.py`: `32 passed`
- `benchmarks/tests` + manifest validator target: `9 passed`

superAngGao:
@zhen8838 Thanks for catching this. You are right that the marker-kernel dispatch tests were not enough, since they only verified routing and did not compile or execute the actual raw CUDA kernel in PR GPU Smoke.

I rebased the PR and added representative real raw-kernel smoke coverage for both fp16 and bf16 `DK=DV=128`, including the multi-step state propagation case. These tests now explicitly assert that `DeltaNetDecodeOp` selected `DeltaNetDecodeRawCudaFlaStyleKernel`, then run correctness against the torch reference, so `-m smoke` exercises the actual raw path.

I also re-ran the official runner image locally after resolving the conflict:

- `tests/ops/test_deltanet_recurrence.py -m smoke`: `19 passed, 13 deselected`
- full `tests/ops/test_deltanet_recurrence.py`: `32 passed`
- benchmark contract + manifest validator target: `9 passed`

Appreciate the careful review.

tileops/manifest/linear_attention.yaml
```diff
@@ -138,8 +138,14 @@ DeltaNetDecodeOp:
 
   workloads:
     - {q_shape: [1, 4, 64], k_shape: [1, 4, 64], v_shape: [1, 4, 64], beta_shape: [1, 4], state_shape: [1, 4, 64, 64], dtypes: [float16, bfloat16, float32], label: "delta-decode-smoke-d64"}
-    - {q_shape: [1, 32, 128], k_shape: [1, 32, 128], v_shape: [1, 32, 128], beta_shape: [1, 32], state_shape: [1, 32, 128, 128], dtypes: [bfloat16, float32], label: "delta-decode-b1-h32-d128"}
-    - {q_shape: [8, 32, 128], k_shape: [8, 32, 128], v_shape: [8, 32, 128], beta_shape: [8, 32], state_shape: [8, 32, 128, 128], dtypes: [bfloat16, float32], label: "delta-decode-b8-h32-d128"}
+    - {q_shape: [1, 32, 128], k_shape: [1, 32, 128], v_shape: [1, 32, 128], beta_shape: [1, 32], state_shape: [1, 32, 128, 128], dtypes: [float16, bfloat16, float32], label: "delta-decode-b1-h32-d128"}
```
Ibuki-wind: `DeltaNetDecodeOp.workloads` is manifest-owned shape metadata, and this perf implementation PR is not flipping `DeltaNetDecodeOp` from `spec-only` to `implemented`; `.claude/rules/manifest-trust-model.md` only carves out workload edits for that promotion case. → Move the workload dtype/row additions to a manifest-only PR, and keep this PR's manifest change limited to `source.kernel_map.DeltaNetDecodeRawCudaFlaStyleKernel`.

Ibuki-wind (changes requested):
### Overall
The raw kernel changes look covered, but the PR still edits manifest-owned workload definitions; move that manifest scope out before approval.

Ibuki-wind (changes requested):
### Overall
The implementation and CI are clean; finish the approval-gate authoring cleanup before approval.

### Cross-cutting concerns
- Update the PR body to the current `.foundry/mold/pr-body-template.md` contract: use a Test plan and record pre-commit, structural readiness, and the test-node delta (`tests/ops/test_deltanet_recurrence.py: 15 -> 32, +17`).
- Condense the author follow-up/review replies to outcome-only one-liners such as `Done in <sha>.`; the current multi-paragraph process narratives do not meet the approval gate reply discipline.

tileops/manifest/linear_attention.yaml
superAngGao: Thanks, agreed. I pushed `f9d8bd20` to keep this PR manifest scope narrow.

What changed:
- Reverted the `DeltaNetDecodeOp.workloads` dtype widening and extra benchmark rows from this PR.
- Kept only `source.kernel_map.DeltaNetDecodeRawCudaFlaStyleKernel`, so the manifest now matches the new op/kernel surface without changing manifest-owned workload definitions.

I verified the resulting PR diff against `main`: `tileops/manifest/linear_attention.yaml` now has exactly one manifest change, the raw kernel-map entry.

Validation used the official runner image exactly as requested, without installing or upgrading dependencies: `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10` (`torch 2.10.0+cu129`, `tilelang 0.1.11+cu129.git65dbc983`).

Results:
- `python scripts/validate_manifest.py --levels schema,signature,shape,dtype,bench --strict --check-op DeltaNetDecodeOp` -> passed
- `python -m pytest -q benchmarks/tests tests/test_validate_manifest.py::TestIntegration::test_validator_passes_on_current_codebase --tb=short -p no:cacheprovider` -> `9 passed`
- `python -m ruff check tileops/kernels/deltanet_recurrence.py tileops/ops/deltanet_recurrence.py tests/ops/test_deltanet_recurrence.py benchmarks/ops/bench_deltanet_recurrence.py` -> passed
- Docker GPU smoke: `tests/ops/test_deltanet_recurrence.py -m smoke` -> `19 passed, 13 deselected`
- Docker full file: `tests/ops/test_deltanet_recurrence.py` -> `32 passed`

Ibuki-wind (approved):
Clean — no issues.

lcy-seso:
@zhen8838 This PR needs your approval. Could you please look into this? 😊

zhen8838 (approved):

PR #1688, opened 2026-07-13
[CI][Runner] unset RUNNER_TOKEN before the runner starts

lcy-seso:
## Problem

`entrypoint.sh` keeps `RUNNER_TOKEN` exported for the whole runner lifecycle (it was retained for the `config.sh remove` cleanup trap). Jobs run as children of the runner listener and inherit its environment, so **any job can read the registration token** and use it to register a rogue runner with trusted labels (e.g. `self-hosted,tile-ops,nightly`) while the token is valid (~1 hour). A rogue trusted runner would then receive trusted jobs carrying a write-scoped `GITHUB_TOKEN`.

This is exploitable today by any code that runs on the runner, and becomes the critical gap once the on-demand fork runner pool starts running untrusted external fork PRs on this image.

## Fix

Copy the token into an unexported shell variable, `unset RUNNER_TOKEN` immediately after validation, and use the variable in `config.sh` and the cleanup trap. Shell variables are not part of the environment, so neither the listener nor its job children can see the token.

## Verification

- `bash -n` + `shellcheck` on `entrypoint.sh` pass.
- Env-inheritance check: after `reg_token="$RUNNER_TOKEN"; unset RUNNER_TOKEN`, a child process sees the variable as absent while the parent shell variable still holds the value.
- pre-commit hooks pass.

## Deployment note

The fix takes effect on the next runner image rebuild; the CI host launchers pass the token via `--env-file` unchanged.

### DIFF
--- a/.github/runner/entrypoint.sh
+++ b/.github/runner/entrypoint.sh
@@ -16,6 +16,13 @@ fi
 : "${RUNNER_TOKEN:?Environment variable RUNNER_TOKEN is required}"
 : "${RUNNER_URL:?Environment variable RUNNER_URL is required}"
 
+# Jobs run as children of the listener and inherit its environment; a job that reads
+# RUNNER_TOKEN can register a rogue runner with trusted labels while the token is valid.
+# Hold the token in an unexported shell variable and drop it from the environment before
+# anything else runs.
+reg_token="${RUNNER_TOKEN}"
+unset RUNNER_TOKEN
+
 RUNNER_NAME="${RUNNER_NAME:-$(hostname)}"
 RUNNER_LABELS="${RUNNER_LABELS:-self-hosted,tile-ops,venv}"
 RUNNER_WORKDIR="${RUNNER_WORKDIR:-_work}"
@@ -23,7 +30,7 @@ RUNNER_WORKDIR="${RUNNER_WORKDIR:-_work}"
 # ── Cleanup function — deregister runner on exit ─────────────────────────────
 cleanup() {
     echo "Removing runner registration..."
-    ./config.sh remove --token "${RUNNER_TOKEN}" 2>/dev/null || true
+    ./config.sh remove --token "${reg_token}" 2>/dev/null || true
 }
 
 trap cleanup EXIT INT TERM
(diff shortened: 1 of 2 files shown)


.github/runner/entrypoint.sh
```diff
@@ -16,22 +16,29 @@ fi
 : "${RUNNER_TOKEN:?Environment variable RUNNER_TOKEN is required}"
 : "${RUNNER_URL:?Environment variable RUNNER_URL is required}"
 
+# Jobs run as children of the listener and inherit its environment; a job that reads
+# RUNNER_TOKEN can register a rogue runner with trusted labels while the token is valid.
+# Hold the token in an unexported shell variable and drop it from the environment before
+# anything else runs.
+reg_token="${RUNNER_TOKEN}"
```
Ibuki-wind, line 25: If `reg_token` is already present in the inherited environment, this assignment preserves its export attribute, so child jobs receive `reg_token=<registration-token>` even after `RUNNER_TOKEN` is unset → run `unset reg_token` before this assignment, then assign `reg_token="${RUNNER_TOKEN}"` and unset `RUNNER_TOKEN`.

Ibuki-wind (changes requested):
### Overall
The token is still reachable by job children when `reg_token` is inherited as an exported environment variable; clear that export bit before copying `RUNNER_TOKEN`.

.github/runner/entrypoint.sh
lcy-seso: Fixed in ab9bea20: `unset reg_token` now precedes the assignment, so an inherited export attribute is cleared. Verified with a pre-exported `reg_token=poison` in the parent environment — the child process sees neither `reg_token` nor `RUNNER_TOKEN`.

.github/runner/entrypoint.sh
```diff
@@ -16,22 +16,29 @@ fi
...
+unset RUNNER_TOKEN
```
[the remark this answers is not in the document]
lcy-seso: Fixed in ab9bea20 as suggested — `unset reg_token` before the assignment clears any inherited export attribute. Verified: with `reg_token` pre-exported in the parent environment, child processes see neither `reg_token` nor `RUNNER_TOKEN`.

Ibuki-wind (changes requested):
### Overall
Approval is blocked only on review hygiene: rewrite the PR body to the repository template and trim the review-thread replies to outcome-only one-liners.

Ibuki-wind (approved):
Clean — no code issues. The previous review-hygiene notes are non-blocking and should not prevent merging this security fix.

Issue #1674, opened 2026-07-08
[Tracking][API] Input-inferred Op API follow-up cleanup


superAngGao:
Follow-up umbrella for #1648.

## Summary

Track the remaining work to align TileOPs public Op APIs with the #1648 input-inferred shape metadata principle.

#1648 established the core API rule:

- values derivable from forward input tensor metadata should be inferred in `forward()`;
- kernels should be built lazily and cached by the resolved specialization signature;
- semantic/config parameters should remain explicit;
- scheduling metadata must not be inferred by reading CUDA tensor values.

This umbrella collects the follow-up inventory from #1648 and tracks scoped child issues by operator family.

## Current child / source issues

- #1648: original input-inferred Op API principle and follow-up inventory.
- #1673: Conv and AvgPool input-inferred API cleanup.

## Core implementation pattern

Each scoped child issue should update the full public contract, not only the Python constructor:

- Op implementation: preferred constructor signature, forward-time shape inference, validation, lazy kernel cache key, and dynamic specialization metadata used by `eval_roofline()`.
- Manifest: `signature.params`, `static_dims`, `shape_rules`, workloads, roofline bindings, and validator expectations should reflect the preferred input-inferred API.
- Tests: preferred API, compatibility or migration behavior for explicit-dimension call sites if retained, shape mismatch validation, cache reuse, and changing-shape behavior.
- Benchmarks and in-repo call sites: use the preferred API unless explicitly testing compatibility.
- Docs/docstrings: examples should not keep teaching redundant shape constructor arguments.

### Conv / Pool

Tracked by #1673.

Candidates:

- `AvgPool1dFwdOp`, `AvgPool2dFwdOp`, `AvgPool3dFwdOp`
- `Conv1dFwdOp`, `Conv1dBiasFwdOp`
- `Conv2dFwdOp`, `Conv2dBiasFwdOp`
- `Conv3dFwdOp`, `Conv3dBiasFwdOp`

Likely derivable metadata:

- input `n`, `c_in`, and spatial sizes from `input.shape`;
- convolution `c_out` and kernel extents from `weight.shape`.

Keep explicit:

- `kernel_size`, `stride`, `padding`, `dilation`, `groups`, `ceil_mode`, `count_include_pad`, `divisor_override`.

### Norm family

Candidates:

- `FusedAddLayerNormFwdOp`
- `FusedAddRMSNormFwdOp`
- `AdaLayerNormFwdOp`
[...]

PR #1690, opened 2026-07-13
[Feat][NORM_SOFTMAX] Infer shape metadata from inputs

### LINKED ISSUE #1686: [FEAT][NORM_SOFTMAX] infer normalization and softmax shape metadata from inputs
Parent tracking issue: #1674

### Symptom / Motivation
Upstream issue #1648 established the API principle that public Op constructors should not require users to repeat shape metadata that is directly derivable from forward input tensor metadata. #1673 applied that principle to Conv and AvgPool. This issue applies the same cleanup to the spatial / batch normalization and softmax-family Ops that still require shape metadata in their constructors.

The current spatial normalization constructors require values such as `N`, `C`, `spatial`, and `dtype`, even though those are available from the tensors passed to `forward()`:

- `BatchNormFwdOp(N, C, spatial=..., dtype=..., ...)`
- `BatchNormBwdOp(N, C, *spatial, dtype=..., ...)`
- `GroupNormFwdOp(N, C, spatial, num_groups, dtype, ...)`
- `GroupNormFwdOpNoAffine(N, C, spatial, num_groups, dtype, ...)`
- `InstanceNormFwdOp(N, C, spatial, dtype, ...)`
- `InstanceNormFwdOpNoAffine(N, C, spatial, dtype, ...)`

This makes the preferred API more verbose than necessary and can let constructor-provided dimensions drift from the actual tensors passed to `forward()`.

The softmax-family wrappers have a narrower version of the same problem:

- `SoftmaxFwdOp(N, dtype, dim=...)`
- `LogSoftmaxFwdOp(N, dtype, dim=...)`

Their shared `_SoftmaxBaseOp` already performs forward-time `M/N` derivation and lazy kernel lookup, but the public wrappers still require callers to pass `N` and `dtype` up front.

This issue intentionally focuses on spatial / batch normalization. Row-wise and layer-style normalization cleanup is already covered separately or has already moved toward input-inferred/lazy specialization patterns:

- `RMSNormFwdOp`
- `LayerNormFwdOp`
- `AdaLayerNormFwdOp`
- `AdaLayerNormZeroFwdOp`
- `FusedAddLayerNormFwdOp`
- `FusedAddRMSNormFwdOp`

### Root Cause Analysis
The current BatchNorm, GroupNorm, and InstanceNorm Ops build shape-specialized kernels eagerly in `__init__()`, so the constructors require all dimensions needed to instantiate the kernel before any tensor is seen. `forward()` then validates actual tensor shapes against constructor-committed shapes instead of deriving the specialization from the actual inputs.

For `SoftmaxFwdOp` and `LogSoftmaxFwdOp`, the base implementation is already mostly lazy, but the subclass constructors still force `N` and `dtype` as required arguments. This leaves the public API inconsistent with the input-inferred contract and with other reduction Ops that bind dimensions during `forward()`.

For GroupNorm and InstanceNorm, the Op layer also pre-allocates device constants such as unit weights / zero bias buffers at construction time, which couples the Op instance to the active CUDA device and to the constructor-provided specialization. These buffers should be created lazily as part of the resolved specialization and cached consistently with the kernel.

### Related Files
- `tileops/ops/norm/batch_norm.py`: `BatchNormFwdOp`, `BatchNormBwdOp`
- `tileops/ops/norm/group_norm.py`: `GroupNormFwdOp`, `GroupNormFwdOpNoAffine`
- `tileops/ops/norm/instance_norm.py`: `InstanceNormFwdOp`, `InstanceNormFwdOpNoAffine`
- `tileops/ops/reduction/softmax.py`: `SoftmaxFwdOp`
- `tileops/ops/reduction/log_softmax.py`: `LogSoftmaxFwdOp`
- `tileops/ops/reduction/_softmax_base.py`
- `tileops/kernels/norm/batch_norm.py`
- `tileops/kernels/norm/group_norm.py`
- `tileops/kernels/norm/instance_norm.py` if present for no-affine paths
- `tileops/kernels/reduction/softmax.py`
- `tileops/manifest/normalization.yaml`
- `tests/ops/test_batch_norm.py`
- `tests/ops/test_group_norm.py`
- `tests/ops/test_instance_norm.py`
- `tests/ops/test_softmax.py`
- `benchmarks/ops/bench_batch_norm.py`
- `benchmarks/ops/bench_group_norm.py`
- `benchmarks/ops/bench_instance_norm.py`
- `benchmarks/ops/bench_softmax.py`

## Goal
Align the spatial / batch normalization and softmax-family public APIs with the #1648 input-inferred shape metadata principle while preserving semantic/config parameters as explicit constructor arguments.

Preferred APIs should infer tensor-derived metadata in `forward()` and build kernels lazily:

- BatchNorm forward: infer `N`, `C`, spatial extents, and input dtype from `x`; validate `weight`, `bias`, `running_mean`, and `running_var` against inferred `C`.
- BatchNorm backward: infer `N`, `C`, spatial extents, and dtype from `grad_out` / `x`; validate `weight`, `mean`, and `rstd` against inferred `C`.
- GroupNorm affine: infer `N`, `C`, spatial extents, and dtype from `x`; validate `weight` and `bias` against inferred `C`.
- GroupNorm no-affine: infer `N`, `C`, spatial extents, and dtype from `x`.
- InstanceNorm affine: infer `N`, `C`, spatial extents, and dtype from `x`; validate `weight` and `bias` against inferred `C`.
- InstanceNorm no-affine: infer `N`, `C`, spatial extents, and dtype from `x`; validate running stats when the eval-mode path uses them.
- Softmax / LogSoftmax: infer `N`, `M`, dtype, and CUDA device from `x`; keep `dim` as the explicit semantic axis parameter.

Keep semantic/config parameters explicit:

- BatchNorm: `training`, `momentum`, `eps`
- GroupNorm: `num_groups`, `eps`
- InstanceNorm: `use_input_stats`, `momentum`, `eps`
- Softmax / LogSoftmax: `dim`
- Kernel override and tuning knobs: `kernel_map`, `tune`

## Plan

1. Update the scoped normalization Ops so preferred constructors no longer require `N`, `C`, `spatial`, or `dtype` when those values can be derived from forward tensors.
2. Replace eager kernel construction with lazy kernel caches keyed by the resolved specialization signature, including shape, dtype, CUDA device, affine/no-affine variant, training/inference path, `num_groups`, `eps`, and tuning-relevant config.
3. Move device-dependent pre-allocated constants in GroupNorm / InstanceNorm into the lazy specialization cache so a single Op instance is not accidentally bound to the CUDA device active at construction time.
4. Update `SoftmaxFwdOp` and `LogSoftmaxFwdOp` wrappers so preferred constructors no longer require `N` or `dtype`; reuse `_SoftmaxBaseOp` forward-time specialization.
5. Preserve practical compatibility paths where useful, but make the preferred API and all examples input-inferred.
6. Update `eval_roofline()` to use the active or most recent resolved specialization, or fail clearly if called before a specialization is available.
7. Update `tileops/manifest/normalization.yaml` and `tileops/manifest/reduction.yaml` so `signature.params`, `static_dims`, `shape_rules`, workloads, roofline bindings, and source metadata match the implemented preferred API.
8. Update unit tests to cover preferred input-inferred construction, legacy compatibility or migration behavior if retained, shape/dtype mismatch validation, lazy cache reuse, changing-shape behavior, and multi-device behavior where practical.
9. Update benchmarks and in-repo call sites to use the preferred input-inferred API unless they intentionally test compatibility.

## Kernel Attribute Contract
Follow the lazy-specialization style used by the closed Conv / Pool cleanup in #1673:

- `self._kernel_cache` is the authoritative cache of compiled/built kernel specializations.
- `self.kernel` may be retained for compatibility and debugging, but it must mean "the most recently used kernel", not "the only kernel owned by this Op".
- `self.kernel` should not be eagerly built in `__init__()`. Initialize it to `None` if the class exposes it, then assign it in `forward()` after resolving the specialization:

```python
kernel = self._get_kernel(resolved_specialization)
self.kernel = kernel
```

- If a later `forward()` call resolves to a different specialization, `self.kernel` should be overwritten with that specialization's cached kernel.
- A cache hit should not trigger JIT/build work; it should only do metadata validation, specialization-key construction, dictionary lookup, and the `self.kernel` last-used assignment.
- Core execution, roofline logic, tests, and benchmarks should rely on `_kernel_cache` / `_last_roofline_spec` rather than treating `self.kernel` as canonical state.

## Constraints
- Do not infer scheduling metadata by reading CUDA tensor values.
- Do not change normalization mathematical semantics relative to the existing PyTorch reference paths.
- Do not change softmax / log_softmax mathematical semantics or `dim=None` compatibility behavior relative to the existing PyTorch reference path.
- Keep row-wise/layer-style normalization cleanup out of this issue unless a follow-up explicitly expands the scope.
- Keep affine and no-affine paths distinct when the kernel contracts require different specializations.
- Preserve clear validation errors for rank, channel, affine parameter, running-stat, dtype, and device mismatches.
- Avoid binding an Op instance to a CUDA device at construction time when no tensor has been seen yet.

## Acceptance Criteria
- [ ] Preferred BatchNorm forward/backward APIs no longer require constructor arguments for `N`, `C`, `spatial`, or `dtype` derivable from forward tensors.
- [ ] Preferred GroupNorm affine/no-affine APIs no longer require constructor arguments for `N`, `C`, `spatial`, or `dtype` derivable from `x`.
- [ ] Preferred InstanceNorm affine/no-affine APIs no longer require constructor arguments for `N`, `C`, `spatial`, or `dtype` derivable from `x`.
- [ ] Preferred Softmax / LogSoftmax APIs no longer require constructor arguments for `N` or `dtype` derivable from `x`.
- [ ] Kernels and device-local helper buffers are lazily built and cached by the resolved specialization signature.
- [ ] `eval_roofline()` reports values for the resolved specialization and fails clearly if called before a specialization is available.
- [ ] Manifest entries for scoped normalization and softmax-family Ops match the implemented public contract, including params, shape rules, roofline metadata, and static dimension declarations.
- [ ] Unit tests cover correctness, validation, cache reuse, changing-shape behavior, and any retained compatibility behavior.
- [ ] Benchmarks and in-repo call sites use the preferred input-inferred API unless they are compatibility tests.
- [ ] Modified files pass targeted normalization / softmax tests and manifest validation.

superAngGao:
## Summary

- infer BatchNorm/GroupNorm/InstanceNorm and Softmax/LogSoftmax shape/dtype metadata from forward inputs instead of requiring construction-time `N/C/spatial/dtype` metadata
- lazily specialize and JIT kernels in `forward()` with `_kernel_cache` as the authoritative specialization cache
- retain `self.kernel` as the last-used kernel handle only, so repeated calls can overwrite it when dispatching a different specialization while cache hits avoid rebuild/JIT work
- update tests and benchmarks to use the preferred input-derived API while keeping explicit constructor metadata as optional compatibility guards

## Kernel Cache / `self.kernel` Contract

- `__init__` no longer eagerly builds kernels for these ops.
- `forward()` resolves the specialization from runtime tensors, looks up/builds the cached kernel, and assigns `self.kernel = kernel` for compatibility/debug visibility.
- `_kernel_cache` owns compiled specializations; `self.kernel` is not the cache and may change on later calls with different input shapes/devices/dtypes.
- For an already-JITed specialization, the hot path is metadata validation + cache key lookup + assignment, without another kernel build.

## Testing

- `python -m py_compile` on changed op/test/benchmark files
- `python -m ruff check` on changed op/test/benchmark files
- local Docker GPU smoke with `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10` on GPU 4:
  - `87 passed, 12 skipped, 138 deselected, 14 warnings`
- local Docker manifest checks for affected ops:
  - `SoftmaxFwdOp`, `LogSoftmaxFwdOp`, `BatchNormFwdOp`, `BatchNormBwdOp`, `GroupNormFwdOp`, `GroupNormFwdOpNoAffine`, `InstanceNormFwdOp`, `InstanceNormFwdOpNoAffine`
  - all passed; remaining output is existing advisory `_infer_output_shapes` warnings

Closes #1686
Refs #1674

### DIFF
--- a/tests/ops/test_norm_lazy_cache.py
+++ b/tests/ops/test_norm_lazy_cache.py
@@ -2,7 +2,18 @@ import pytest
 import torch
 
 from tileops.kernels.kernel_base import Kernel
-from tileops.ops.norm.batch_norm import BatchNormFwdOp
+from tileops.ops.norm.batch_norm import BatchNormBwdOp, BatchNormFwdOp
+
+
+def _to_cl(x: torch.Tensor) -> torch.Tensor:
+    return x.permute(1, 0, *range(2, x.ndim)).reshape(x.shape[1], -1).contiguous()
+
+
+def _from_cl(x_cl: torch.Tensor, orig_shape: torch.Size) -> torch.Tensor:
+    n = orig_shape[0]
+    c = orig_shape[1]
+    spatial = orig_shape[2:]
+    return x_cl.reshape(c, n, *spatial).permute(1, 0, *range(2, len(orig_shape))).contiguous()
 
 
 class _FakeBatchNormFwdInferKernel(Kernel):
@@ -66,6 +77,41 @@ class _FakeBatchNormFwdTrainKernel(_FakeBatchNormFwdInferKernel):
         return y.to(self.dtype), mean, rstd
 
 
+class _FakeBatchNormBwdKernel(Kernel):
+    def __init__(
+        self,
+        C: int,
+        L: int,
+        dtype: torch.dtype,
+        tune: bool = False,
+    ) -> None:
+        super().__init__()
+        self.C = C
+        self.L = L
+        self.dtype = dtype
+        self.tune = tune
+
+    def forward(
+        self,
+        grad_out_cl: torch.Tensor,
+        x_cl: torch.Tensor,
+        weight: torch.Tensor,
+        mean: torch.Tensor,
+        rstd: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        grad_out_f = grad_out_cl.float()
+        x_hat = (x_cl.float() - mean[:, None]) * rstd[:, None]
+        grad_bias = grad_out_f.sum(dim=1)
+        grad_weight = (grad_out_f * x_hat).sum(dim=1)
+        grad_x = (
+            weight[:, None]
+            * rstd[:, None]
+            * (self.L * grad_out_f - grad_bias[:, None] - x_hat * grad_weight[:, None])
+            / self.L
+        )
+        return grad_x.to(self.dtype), grad_weight, grad_bias
+
+
 def _batch_norm_infer_ref(
     x: torch.Tensor,
     running_mean: torch.Tensor,
@@ -79,6 +125,35 @@ def _batch_norm_infer_ref(
         running_var.reshape(affine_shape) + eps)
     y = y * weight.reshape(affine_shape) + bias.reshape(affine_shape)
     return y.to(x.dtype)
+
+
+def _batch_norm_train_ref(
+    x: torch.Tensor,
+    weight: torch.Tensor,
+    bias: torch.Tensor,
+    eps: float,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    x_cl = _to_cl(x)
+    mean = x_cl.float().mean(dim=1)
+    var = x_cl.float().var(dim=1, unbiased=False)
+    rstd = torch.rsqrt(var + eps)
+    y_cl = (x_cl.float() - mean[:, None]) * rstd[:, None]
+    y_cl = y_cl * weight[:, None] + bias[:, None]
+    return _from_cl(y_cl.to(x.dtype), x.shape), mean, rstd
+
+
+def _batch_norm_bwd_ref(
+    grad_out: torch.Tensor,
+    x: torch.Tensor,
+    weight: torch.Tensor,
+    mean: torch.Tensor,
+    rstd: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    grad_out_cl = _to_cl(grad_out)
+    x_cl = _to_cl(x)
+    kernel = _FakeBatchNormBwdKernel(x.shape[1], grad_out.numel() // x.shape[1], x.dtype)
+    grad_x_cl, grad_weight, grad_bias = kernel(grad_out_cl, x_cl, weight, mean, rstd)
+    return _from_cl(grad_x_cl, x.shape), grad_weight, grad_bias
 
 
 @pytest.mark.smoke
@@ -109,6 +184,7 @@ def test_batch_norm_fwd_lazy_cache_reuse_and_respecialization() -> None:
 
     run_case(2, 8, (4, 4), torch.float16)
     assert len(op._kernel_cache) == 1
+    first_kernel = op.kernel
     assert op.eval_roofline() == (
         10 * 8 * 32,
         2 * 8 * 32 * torch.float16.itemsize + 4 * 8 * 4,
@@ -116,10 +192,102 @@ def test_batch_norm_fwd_lazy_cache_reuse_and_respecialization() -> None:
 
     run_case(2, 8, (4, 4), torch.float16)
     assert len(op._kernel_cache) == 1
+    assert op.kernel is first_kernel
 
     run_case(3, 12, (2, 8), torch.bfloat16)
     assert len(op._kernel_cache) == 2
+    assert op.kernel is not first_kernel
     assert op.eval_roofline() == (
         10 * 12 * 48,
         2 * 12 * 48 * torch.bfloat16.itemsize + 4 * 12 * 4,
+    )
+
+
+@pytest.mark.smoke
+def test_batch_norm_training_fwd_lazy_cache_reuse_and_respecialization() -> None:
+    """Training BatchNorm forward cache path is executable under fake kernels."""
+    if not torch.cuda.is_available():
+        pytest.skip("CUDA required for forward call")
+
+    op = BatchNormFwdOp(
+        training=True,
+        kernel_map={
+            "fwd_infer_kernel": _FakeBatchNormFwdInferKernel,
+            "fwd_train_kernel": _FakeBatchNormFwdTrainKernel,
+        },
+    )
+
+    def run_case(N: int, C: int, spatial: tuple[int, ...], dtype: torch.dtype) -> None:
+        x = torch.randn((N, C, *spatial), device="cuda", dtype=dtype)
+        weight = torch.randn(C, device="cuda", dtype=torch.float32)
+        bias = torch.randn(C, device="cuda", dtype=torch.float32)
+        running_mean = torch.zeros(C, device="cuda", dtype=torch.float32)
+        running_var = torch.ones(C, device="cuda", dtype=torch.float32)
+
+        y = op(x, running_mean, running_var, weight, bias)
+        ref_y, _, _ = _batch_norm_train_ref(x, weight, bias, op.eps)
+        assert torch.allclose(y.float(), ref_y.float(), atol=0.0, rtol=0.0)
+
+    run_case(2, 8, (4, 4), torch.float16)
+    assert len(op._kernel_cache) == 1
+    first_kernel = op.kernel
+    assert op.eval_roofline() == (
+        10 * 8 * 32,
+        2 * 8 * 32 * torch.float16.itemsize + 4 * 8 * 4,
+    )
+
+    run_case(2, 8, (4, 4), torch.float16)
+    assert len(op._kernel_cache) == 1
+    assert op.kernel is first_kernel
+
+    run_case(3, 12, (2, 8), torch.bfloat16)
+    assert len(op._kernel_cache) == 2
+    assert op.kernel is not first_kernel
+    assert op.eval_roofline() == (
+        10 * 12 * 48,
+        2 * 12 * 48 * torch.bfloat16.itemsize + 4 * 12 * 4,
+    )
+
+
+@pytest.mark.smoke
+def test_batch_norm_bwd_lazy_cache_reuse_and_respecialization() -> None:
+    """BatchNorm backward cache path is executable under fake kernels."""
+    if not torch.cuda.is_available():
+        pytest.skip("CUDA required for backward call")
+
+    eps = 1e-5
+    op = BatchNormBwdOp(kernel_map={"bwd_kernel": _FakeBatchNormBwdKernel})
+
+    def run_case(N: int, C: int, spatial: tuple[int, ...], dtype: torch.dtype) -> None:
+        x = torch.randn((N, C, *spatial), device="cuda", dtype=dtype)
+        grad_out = torch.randn((N, C, *spatial), device="cuda", dtype=dtype)
+        weight = torch.randn(C, device="cuda", dtype=torch.float32)
+        _y, mean, rstd = _batch_norm_train_ref(
+            x, torch.ones_like(weight), torch.zeros_like(weight), eps)
+
+        grad_x, grad_weight, grad_bias = op(grad_out, x, weight, mean, rstd)
+        ref_grad_x, ref_grad_weight, ref_grad_bias = _batch_norm_bwd_ref(
+            grad_out, x, weight, mean, rstd)
+        assert torch.allclose(grad_x.float(), ref_grad_x.float(), atol=0.0, rtol=0.0)
+        assert torch.allclose(grad_weight, ref_grad_weight, atol=0.0, rtol=0.0)
+        assert torch.allclose(grad_bias, ref_grad_bias, atol=0.0, rtol=0.0)
+
+    run_case(2, 8, (4, 4), torch.float16)
+    assert len(op._kernel_cache) == 1
+    first_kernel = op.kernel
+    assert op.eval_roofline() == (
+        8 * 8 * 32,
+        3 * 8 * 32 * torch.float16.itemsize + 3 * 8 * 4,
+    )
+
+    run_case(2, 8, (4, 4), torch.float16)
+    assert len(op._kernel_cache) == 1
+    assert op.kernel is first_kernel
+
+    run_case(3, 12, (2, 8), torch.bfloat16)
+    assert len(op._kernel_cache) == 2
+    assert op.kernel is not first_kernel
+    assert op.eval_roofline() == (
+        8 * 12 * 48,
+        3 * 12 * 48 * torch.bfloat16.itemsize + 3 * 12 * 4,
     )
--- a/tests/ops/test_norm_lazy_cache.py
+++ b/tests/ops/test_norm_lazy_cache.py
@@ -1,0 +1,125 @@
+import pytest
+import torch
+
+from tileops.kernels.kernel_base import Kernel
+from tileops.ops.norm.batch_norm import BatchNormFwdOp
+
+
+class _FakeBatchNormFwdInferKernel(Kernel):
+    def __init__(
+        self,
+        C: int,
+        L: int,
+        dtype: torch.dtype,
+        eps: float,
+        tune: bool = False,
+    ) -> None:
+        super().__init__()
+        self.C = C
+        self.L = L
+        self.dtype = dtype
+        self.eps = eps
+        self.tune = tune
+
+    def forward(
+        self,
+        x_cl: torch.Tensor,
+        weight: torch.Tensor,
+        bias: torch.Tensor,
+        running_mean: torch.Tensor,
+        running_var: torch.Tensor,
+    ) -> torch.Tensor:
+        y = (x_cl.float() - running_mean[:, None]) * torch.rsqrt(
+            running_var[:, None] + self.eps)
+        y = y * weight[:, None] + bias[:, None]
+        return y.to(self.dtype)
+
+
+class _FakeBatchNormFwdTrainKernel(_FakeBatchNormFwdInferKernel):
+    def __init__(
+        self,
+        C: int,
+        L: int,
+        dtype: torch.dtype,
+        eps: float,
+        momentum: float,
+        tune: bool = False,
+    ) -> None:
+        super().__init__(C, L, dtype, eps, tune=tune)
+        self.momentum = momentum
+
+    def forward(
+        self,
+        x_cl: torch.Tensor,
+        weight: torch.Tensor,
+        bias: torch.Tensor,
+        running_mean: torch.Tensor,
+        running_var: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        mean = x_cl.float().mean(dim=1)
+        var = x_cl.float().var(dim=1, unbiased=False)
+        rstd = torch.rsqrt(var + self.eps)
+        running_mean.mul_(1 - self.momentum).add_(self.momentum * mean)
+        running_var.mul_(1 - self.momentum).add_(self.momentum * var)
+        y = (x_cl.float() - mean[:, None]) * rstd[:, None]
+        y = y * weight[:, None] + bias[:, None]
+        return y.to(self.dtype), mean, rstd
+
+
+def _batch_norm_infer_ref(
+    x: torch.Tensor,
+    running_mean: torch.Tensor,
+    running_var: torch.Tensor,
+    weight: torch.Tensor,
+    bias: torch.Tensor,
+    eps: float,
+) -> torch.Tensor:
+    affine_shape = (1, x.shape[1]) + (1,) * (x.ndim - 2)
+    y = (x.float() - running_mean.reshape(affine_shape)) * torch.rsqrt(
+        running_var.reshape(affine_shape) + eps)
+    y = y * weight.reshape(affine_shape) + bias.reshape(affine_shape)
+    return y.to(x.dtype)
+
+
+@pytest.mark.smoke
+def test_batch_norm_fwd_lazy_cache_reuse_and_respecialization() -> None:
+    """BatchNorm op-layer cache reuses identical specs and caches changed specs."""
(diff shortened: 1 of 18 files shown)


tileops/ops/norm/group_norm.py
```diff
@@ -70,66 +68,146 @@ class GroupNormFwdOp(Op):
...
         return (
-            5 * self.N * self.C * self.spatial_size,
-            (2 * self.N * self.C * self.spatial_size + 2 * self.C) * elem_bytes,
+            5 * N * C * spatial_size,
+            (2 * N * C * spatial_size + 2 * C) * elem_bytes,
         )
 
+    def _resolve_spec(
+        self, x: torch.Tensor
+    ) -> Tuple[int, int, Tuple[int, ...], int, int, int, int, torch.dtype]:
```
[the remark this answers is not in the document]
superAngGao: Thanks, good catch. Fixed in 69141a62 by aligning the `_resolve_spec` return type annotation with the actual tuple returned.

tileops/ops/norm/group_norm.py
```diff
@@ -264,51 +342,141 @@ class GroupNormFwdOpNoAffine(Op):
...
         return (
-            3 * self.N * self.C * self.spatial_size,
-            2 * self.N * self.C * self.spatial_size * elem_bytes,
+            3 * N * C * spatial_size,
+            2 * N * C * spatial_size * elem_bytes,
         )
 
+    def _resolve_spec(
+        self, x: torch.Tensor
+    ) -> Tuple[int, int, Tuple[int, ...], int, int, int, int, int, torch.dtype]:
```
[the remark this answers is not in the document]
superAngGao: Thanks, good catch. Fixed in 69141a62 by aligning the `_resolve_spec` return type annotation with the actual tuple returned.

tileops/ops/norm/instance_norm.py
```diff
@@ -156,17 +147,99 @@ def _validate_dtypes(
...
+            if t.dtype != expected_dtype:
                 raise ValueError(
-                    f"Expected {name}.dtype == self.dtype ({self.dtype}), "
+                    f"Expected {name}.dtype == {expected_dtype}, "
                     f"got {t.dtype}"
                 )
 
+    def _resolve_spec(
+        self, x: torch.Tensor
+    ) -> Tuple[int, int, Tuple[int, ...], int, int, int, torch.dtype]:
```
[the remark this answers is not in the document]
superAngGao: Thanks, good catch. Fixed in 69141a62 by aligning the `_resolve_spec` return type annotation with the actual tuple returned.

tileops/ops/norm/batch_norm.py
```diff
@@ -355,14 +569,20 @@ def _batchnorm_fwd_eager_forward(
     bias: torch.Tensor,
 ) -> torch.Tensor:
     """Direct kernel call (no torch.compile wrapping)."""
-    self._validate_inputs(x)
+    N, C, spatial, L, dtype = self._resolve_spec(x)
```
Ibuki-wind: This duplicates the lazy resolve/validate/bind/kernel path from the class `forward()` above, and the bwd wrapper repeats the same pattern below. Because this module later patches `BatchNormFwdOp.forward`, the two runtime paths can drift the next time validation or cache keys change. Fix by putting the shared body behind one private helper and have both the wrapped custom-op path and normal eager path call it.

Ibuki-wind (changes requested):
I found one review issue: the implementation looks functionally sound in the targeted run, but the tests still miss the central lazy-specialization contract from #1686.

Please add focused coverage for reusing a single op instance across specializations: same specialization should hit the cache without adding a kernel, while a different valid N/C/spatial/dtype specialization should add a new cached kernel and still produce correct output/roofline metadata. This is especially important for BatchNorm, which currently has no lazy-cache assertions, and for GroupNorm/InstanceNorm where the tests cover preferred construction and multi-device smoke but not changing-shape/cache-reuse behavior.

Local validation I ran on the PR worktree: `LD_PRELOAD=/lib/x86_64-linux-gnu/libcuda.so.1 pytest -q tests/ops/test_batch_norm.py tests/ops/test_group_norm.py tests/ops/test_instance_norm.py tests/ops/test_softmax.py` -> 208 passed, 19 skipped.

tileops/ops/norm/batch_norm.py
superAngGao: Thanks, agreed. Fixed in `64d2c171` by moving the shared lazy resolve/validate/bind/kernel body behind `_forward_impl()` for both `BatchNormFwdOp` and `BatchNormBwdOp`. The original class `forward()` and the custom-op eager path now both delegate to that helper, so future validation/cache-key changes have one implementation point.

superAngGao:
Addressed the lazy-specialization coverage request in `64d2c171`:

- Added a focused BatchNorm op-layer cache contract test in `tests/ops/test_norm_lazy_cache.py`. It uses lightweight fake kernels so the cache/re-specialization behavior runs in current CI instead of being hidden by the existing TileLang 0.1.9 skip on `tests/ops/test_batch_norm.py`.
- Added GroupNorm and InstanceNorm smoke tests that reuse one op instance across identical and changed `(N, C, spatial, dtype)` specs, verify cache size reuse/growth, check output correctness, and assert `eval_roofline()` reflects the latest specialization.
- Refactored BatchNorm fwd/bwd so the normal eager path and the custom-op wrapped path both call `_forward_impl()`.

Validation in the official runner image:

```bash
ruff check tileops/ops/norm/batch_norm.py tests/ops/test_batch_norm.py tests/ops/test_group_norm.py tests/ops/test_instance_norm.py tests/ops/test_norm_lazy_cache.py
```
Result: passed

```bash
python -m pytest -q \
  tests/ops/test_norm_lazy_cache.py::test_batch_norm_fwd_lazy_cache_reuse_and_respecialization \
  tests/ops/test_group_norm.py::test_group_norm_lazy_cache_reuse_and_respecialization \
  tests/ops/test_instance_norm.py::test_instance_norm_lazy_cache_reuse_and_respecialization \
  -rs --tb=short -p no:cacheprovider
```
Result: `3 passed`

```bash
LD_PRELOAD=/lib/x86_64-linux-gnu/libcuda.so.1 python -m pytest -q \
  tests/ops/test_norm_lazy_cache.py \
  tests/ops/test_batch_norm.py \
  tests/ops/test_group_norm.py \
  tests/ops/test_instance_norm.py \
  tests/ops/test_softmax.py \
  -rs --tb=short -p no:cacheprovider
```
Result: `206 passed, 24 skipped`

tileops/ops/norm/batch_norm.py
```diff
@@ -102,10 +99,10 @@ class BatchNormFwdOp(Op):
 
     def __init__(
         self,
-        N: int,
-        C: int,
+        N: Optional[int] = None,
```
RMLYC: `BatchNormFwdOp.__init__` still accepts `N`, `C`, `spatial`, and `dtype` as compatibility commitments, while `BatchNormBwdOp` repeats the same pattern. These values must come exclusively from `x`/`grad_out` in `forward()`, otherwise the API retains two shape/dtype sources of truth. → Remove those parameters from both constructors, delete the `_committed_*` fields and checks in `_resolve_spec()`, and remove the backward-compatible `_validate_inputs()` entry point; keep `self.N/C/spatial/dtype` only as last-resolved runtime state populated by `_bind_spec()`.

tileops/ops/norm/group_norm.py
```diff
@@ -70,66 +68,146 @@ class GroupNormFwdOp(Op):
 
     def __init__(
         self,
-        N: int,
-        C: int,
-        spatial: tuple,
         num_groups: int,
-        dtype: torch.dtype,
+        N: Optional[int] = None,
```
RMLYC: `GroupNormFwdOp` still exposes `N`, `C`, `spatial`, and `dtype` as optional constructor commitments, and `GroupNormFwdOpNoAffine` repeats the same compatibility interface. The migrated contract keeps only semantic/config parameters in the constructor and resolves tensor metadata from `x`. → Change both constructors to accept only `num_groups`, `eps`, `kernel_map`, and `tune`; remove the `_committed_*` state/checks and initialize introspection fields to `None` until `_bind_spec()` runs.

tileops/ops/norm/instance_norm.py
```diff
@@ -85,10 +85,10 @@ class InstanceNormFwdOp(Op):
 
     def __init__(
         self,
-        N: int,
-        C: int,
-        spatial: tuple,
-        dtype: torch.dtype,
+        N: Optional[int] = None,
```
RMLYC: `InstanceNormFwdOp` retains `N`, `C`, `spatial`, and `dtype` as constructor compatibility guards, and `InstanceNormFwdOpNoAffine` repeats the same interface. These are tensor-derived values and must not remain part of the migrated public constructor. → Remove all four parameters and the corresponding `_committed_*` validation from both variants; keep only `use_input_stats`, `momentum`, `eps`, `kernel_map`, and `tune` in `__init__`, and derive `N/C/spatial/dtype` from `x` on every forward call.

tileops/ops/reduction/softmax.py
```diff
@@ -48,8 +47,8 @@ class SoftmaxFwdOp(_SoftmaxBaseOp):
 
     def __init__(
         self,
-        N: int,
-        dtype: torch.dtype,
+        N: Optional[int] = None,
```
RMLYC: `SoftmaxFwdOp` still publishes `N` and `dtype` as optional compatibility guards even though both are derived from `x` and neither is a manifest parameter; this also leaves `dim` behind legacy positional arguments. → Change the constructor to `__init__(dim=None, *, kernel_map=None, tune=False)`, stop passing `N`/`dtype` commitments into `_SoftmaxBaseOp`, and rely on `forward()` to resolve `N`, `M`, dtype, and device.

tileops/ops/reduction/log_softmax.py
```diff
@@ -48,8 +47,8 @@ class LogSoftmaxFwdOp(_SoftmaxBaseOp):
 
     def __init__(
         self,
-        N: int,
-        dtype: torch.dtype,
+        N: Optional[int] = None,
```
RMLYC: `LogSoftmaxFwdOp` retains the same `N`/`dtype` compatibility constructor as `SoftmaxFwdOp`, leaving tensor-derived metadata in the public API. → Change the constructor to `__init__(dim=None, *, kernel_map=None, tune=False)` and infer `N/M/dtype/device` exclusively from `x` in `forward()`.

tests/ops/test_norm_lazy_cache.py
```diff
@@ -0,0 +1,125 @@
...
+) -> torch.Tensor:
+    affine_shape = (1, x.shape[1]) + (1,) * (x.ndim - 2)
+    y = (x.float() - running_mean.reshape(affine_shape)) * torch.rsqrt(
+        running_var.reshape(affine_shape) + eps)
+    y = y * weight.reshape(affine_shape) + bias.reshape(affine_shape)
+    return y.to(x.dtype)
+
+
+@pytest.mark.smoke
+def test_batch_norm_fwd_lazy_cache_reuse_and_respecialization() -> None:
```
RMLYC, line 160: New-path coverage is incomplete: this test constructs `BatchNormFwdOp` with `training=False`, so `_FakeBatchNormFwdTrainKernel` is registered but never instantiated or executed, and there is no fake-kernel coverage for the new `BatchNormBwdOp` lazy-specialization path. Because `tests/conftest.py` skips every test in `tests/ops/test_batch_norm.py` under the current TileLang configuration, CI does not execute the changed training-forward or backward resolve/cache paths. → Add executable fake-kernel tests for BatchNorm training forward and backward; verify output correctness, same-specialization cache reuse, changed shape/dtype cache growth, `self.kernel` replacement, and `eval_roofline()` metadata for the latest specialization.

RMLYC (changes requested):
### Overall

The lazy-specialization direction is correct, but the scoped public constructors still retain forbidden shape/dtype compatibility paths and the new BatchNorm training-forward/backward paths lack executable CI coverage; remove the constructor commitments and add the missing tests before merge.

Ibuki-wind (changes requested):
### Overall
BatchNorm training-forward and backward lazy specialization need executable coverage before approval; I am not blocking on the constructor-compatibility cleanup comments.

tileops/ops/norm/batch_norm.py
superAngGao: Addressed in `2013b6b6`: `BatchNormFwdOp` no longer accepts `N`, `C`, `spatial`, or `dtype`, and `BatchNormBwdOp` is now only `kernel_map`/`tune`. I removed the `_committed_*` fields/checks and the backward-compatible `_validate_inputs()` path; `self.N/C/spatial/L/dtype` are now only last-resolved runtime metadata populated through `_bind_spec()`. I also removed the stale `static_dims: C` manifest entries so the metadata matches the forward-derived contract.

tileops/ops/norm/group_norm.py
superAngGao: Addressed in `2013b6b6`: both GroupNorm constructors now expose only `num_groups`, `eps`, `kernel_map`, and `tune`. The compatibility `N/C/spatial/dtype` constructor guards and `_committed_*` checks are gone, and the introspection fields start as `None` until `_bind_spec()` records the latest forward-time specialization. The corresponding manifest `static_dims: C` entries were removed as well.

tileops/ops/norm/instance_norm.py
superAngGao: Addressed in `2013b6b6`: both InstanceNorm variants now keep only `use_input_stats`, `momentum`, `eps`, `kernel_map`, and `tune` in the public constructor. `N/C/spatial/dtype` are derived from `x` on each forward, `_committed_*` validation is removed, and dtype validation now uses the actual input dtype plus affine/running-stat constraints. The manifest no longer declares constructor-time `C` either.

tileops/ops/reduction/softmax.py
superAngGao: Addressed in `2013b6b6`: `SoftmaxFwdOp.__init__` is now `dim=None, *, kernel_map=None, tune=False`; it no longer publishes or passes `N`/`dtype` commitments. `forward()` remains the single source for `N`, `M`, dtype, device, and the cached kernel specialization.

tileops/ops/reduction/log_softmax.py
superAngGao: Addressed in `2013b6b6`: `LogSoftmaxFwdOp.__init__` now matches the requested `dim=None, *, kernel_map=None, tune=False` shape and does not expose `N`/`dtype` compatibility parameters. The runtime metadata is inferred from `x` during `forward()`.

tests/ops/test_norm_lazy_cache.py
superAngGao: Addressed in `2013b6b6`: `tests/ops/test_norm_lazy_cache.py` now has executable fake-kernel coverage for BatchNorm training forward and BatchNorm backward, in addition to the inference path. The tests check output correctness, same-specialization cache reuse, changed shape/dtype cache growth, `self.kernel` replacement, and `eval_roofline()` metadata after the latest specialization.

superAngGao:
Follow-up pushed in `2013b6b6`.

What changed:
- Removed the remaining constructor-time shape/dtype compatibility commitments from BatchNorm, GroupNorm, InstanceNorm, Softmax, and LogSoftmax public constructors.
- Removed the corresponding `_committed_*` state/checks in the norm ops; `N/C/spatial/L/dtype` are now last-resolved runtime metadata populated from forward-time inputs.
- Removed stale `static_dims: C` manifest entries for these norm ops so the manifest no longer advertises constructor-time channel commitments.
- Added executable fake-kernel tests for BatchNorm training forward and BatchNorm backward lazy specialization, covering output correctness, same-specialization cache reuse, changed shape/dtype cache growth, `self.kernel` replacement, and latest `eval_roofline()` metadata.

Validation in the official runner image:

```bash
ruff check tileops/ops/norm/batch_norm.py tileops/ops/norm/group_norm.py tileops/ops/norm/instance_norm.py tileops/ops/reduction/_softmax_base.py tileops/ops/reduction/softmax.py tileops/ops/reduction/log_softmax.py tests/ops/test_norm_lazy_cache.py tests/ops/test_group_norm.py tests/ops/test_instance_norm.py tests/ops/test_norm_ops.py && git diff --check
```
Result: passed

```bash
python -m pytest -q \
  tests/test_validate_manifest.py::TestIntegration::test_schema_validation_no_errors_on_real_manifest \
  tests/test_validate_manifest.py::TestIntegration::test_validator_passes_on_current_codebase \
  --tb=short -p no:cacheprovider
```
Result: `2 passed`

```bash
LD_PRELOAD=/lib/x86_64-linux-gnu/libcuda.so.1 python -m pytest -q \
  tests/ops/test_norm_lazy_cache.py \
  tests/ops/test_norm_ops.py \
  tests/ops/test_batch_norm.py \
  tests/ops/test_group_norm.py \
  tests/ops/test_instance_norm.py \
  tests/ops/test_softmax.py \
  -rs --tb=short -p no:cacheprovider
```
Result: `209 passed, 26 skipped`

Ibuki-wind (approved):
Clean — no issues.

RMLYC (approved):
LGTM

Issue #1643, opened 2026-07-01
[Docker] Missing mamba (official Triton version) installation in Docker images


stelladuyx:
## Problem

Our Docker images currently lack the installation of mamba (official Triton version). This affects both:

1. **TileOPs CI runner image** (`.github/runner/Dockerfile`)
2. **TileLang development images** (`tilelang/docker/Dockerfile.*`)

While the TileOPs runner image includes `triton` installed via pip (line 40), it does not include the official Triton mamba implementation which may be needed for:
- Mamba/SSM model benchmarking and testing
- Compatibility with official Triton reference implementations
- Performance comparison baselines

## Current State

**TileOPs runner Dockerfile** (`.github/runner/Dockerfile:40`):
```dockerfile
RUN python -m pip install -c /tmp/constraints.txt \
        setuptools wheel ninja scikit-build-core patchelf cmake \
        triton apache-tvm-ffi cloudpickle ml_dtypes numpy psutil tqdm \
        typing_extensions Cython z3-solver torch_c_dlpack_ext einops "PyYAML>=6.0"
```

**TileLang Dockerfiles** (e.g., `docker/Dockerfile.cu126`):
- No mamba installation present

## Proposed Solution

Add mamba installation from the official Triton repository to the relevant Docker images. Options:

1. **Install from PyPI** (if available):
   ```dockerfile
   RUN python -m pip install mamba-ssm
   ```

2. **Install from official Triton/mamba source**:
   ```dockerfile
   RUN git clone https://github.com/state-spaces/mamba.git /tmp/mamba \
       && cd /tmp/mamba \
       && pip install -e . \
       && rm -rf /tmp/mamba
   ```

3. **Install specific version for compatibility**:
   ```dockerfile
   RUN python -m pip install -c /tmp/constraints.txt mamba-ssm==<version>
   ```

## Files to Modify

- [ ] `TileOPs/.github/runner/Dockerfile` (add to `fullstack` or `post-fa3` stage)
- [ ] `tilelang/docker/Dockerfile.cu126` 
- [ ] `tilelang/docker/Dockerfile.cu128`
- [ ] Other tilelang CUDA variant Dockerfiles as needed

## Additional Context

This addition should be:
- **Best-effort** (non-fatal) for benchmark baselines, similar to FlashAttention-2 installation pattern
- Constrained to avoid re-resolving the CUDA-specific torch stack
- Documented in the Dockerfile comments regarding its purpose

[...]

PR #1710, opened 2026-07-15
[CI][Runner] Fix benchmark baseline dependencies

superAngGao:
## Summary
- fix the current nightly benchmark failure in the FP8 GEMM FlashInfer baseline
- make the `flashinfer-cubin` cubin directories writable by the unprivileged `ci-runner` user without touching packaged `.cubin` files
- add `mamba-ssm==2.3.1` for the official Mamba Triton benchmark baseline, including its import-time CUDA extension, addressing #1643 for the TileOps CI runner image
- add DeepGEMM `v2.1.1.post3` for the grouped-GEMM benchmark baseline requested in #1643
- document that runner image rollout is manual and happens outside this repository

## Problem
The nightly benchmark is currently failing in the FP8 GEMM benchmark cases. Those benchmark cases call the FlashInfer baseline through `flashinfer.mm_fp8(...)`. Before the benchmark can compare performance, FlashInfer tries to prepare TRTLLM GEMM cubin entries under the directory returned by `flashinfer_cubin.get_cubin_dir()`.

In the self-hosted runner image, `flashinfer-cubin` is installed by pip during the Docker build while the build is still running as root. As a result, its packaged cubin directories under site-packages are root-owned. Later, GitHub Actions jobs run as the unprivileged `ci-runner` user. When the nightly benchmark reaches the FlashInfer FP8 baseline, that user cannot create/populate cubin subdirectories, for example:

```text
/usr/local/lib/python3.12/dist-packages/flashinfer_cubin/cubins/flashinfer/trtllm/gemm
```

So the benchmark fails with a permission error before the actual performance comparison runs. A workflow-only cache/env change is not sufficient for this failure mode because when the `flashinfer-cubin` package is installed, FlashInfer uses that packaged cubin directory first. The root-owned directories therefore need to be fixed inside the runner image.

Separately, #1643 points out that the runner image does not include `mamba-ssm`, so the Mamba/Mamba-2 benchmarks cannot use the official `mamba_ssm.ops.triton.*` baseline. A follow-up comment on #1643 also requests DeepGEMM for the grouped-GEMM benchmark baseline. This PR handles the TileOPs CI runner image side of those dependency requests.

## Fix
In the final runner image stage, after creating `ci-runner`, resolve `flashinfer_cubin.get_cubin_dir()` inside the image, pre-create the TRTLLM GEMM cubin subdirectory, and chown only directory inodes under the cubin root. This gives FlashInfer writable directories for current and sibling cubin subtrees without recursively chowning packaged `.cubin` files, avoiding Docker layer bloat from OverlayFS copy-up.

For the Mamba benchmark baseline, install `mamba-ssm==2.3.1` in the `fullstack` stage as a benchmark dependency. The install uses `MAMBA_FORCE_BUILD=TRUE`, `--no-build-isolation`, `--no-deps`, and the existing torch cu129 constraints. `mamba_ssm` imports `selective_scan_cuda` at package import time, so the CUDA extension is built and included even though the TileOps benchmarks use the official `mamba_ssm.ops.triton.*` modules. I intentionally did not use the current latest `mamba-ssm==2.3.2.post1` because that release declares `tilelang==0.1.8` and `apache-tvm-ffi<=0.1.9`, which conflicts with this runner image's TileOps/tilelang runtime stack.

For the grouped-GEMM benchmark baseline, install DeepGEMM from the official `deepseek-ai/DeepGEMM` source as a benchmark dependency, pinned to commit `c9f8b34dcdacc20aa746b786f983492c51072870` (the `v2.1.1.post3` tag target). The install uses `--no-build-isolation --no-deps` and fetches only the required header submodules. The PyPI `deep_gemm==1.0.0` sdist is stale relative to the current upstream version used for this benchmark surface.

For TileLang source builds, explicitly initialize `3rdparty/composable_kernel` together with `3rdparty/cutlass` and `3rdparty/tvm`. Without that, CMake can trigger a hidden network fetch for the missing submodule during the wheel build. The wheel build also passes `CMAKE_BUILD_PARALLEL_LEVEL=${MAX_JOBS}` so the local image rebuild can use the intended build parallelism.

## Rollout note
This repository does not build the self-hosted runner image in CI. After this lands, a maintainer still needs to rebuild and push the runner image from a GPU host, then redeploy the self-hosted runner launcher to use the new tag outside this repository.

Merging this TileOPs PR only changes the image recipe. The live self-hosted runners keep using their existing image until that manual rollout is done.

## Test
- `pre-commit run mdformat --files .github/runner/README.md`
- `git diff --check`
- Local GPU-host runner image build completed from the updated recipe using pre-cloned source/tarball BuildKit contexts for network-heavy inputs; final image id `sha256:73768b2992f26cde6703b6eb15bba9f9438048a5a9e416154bf893275776f326`.
- Runtime probe inside the built image passed: torch `2.10.0+cu129`, CUDA `12.9`, tilelang `0.1.11+cu129.git65dbc983`, FlashInfer cubin write probe, `selective_scan_cuda`, mamba `2.3.1`, and DeepGEMM `2.1.1`.
- The verified image was retagged locally to `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`; `ci-runner-gpu2` and `ci-runner-gpu3` were recreated from that tag and now run image id `sha256:73768b2992f26cde6703b6eb15bba9f9438048a5a9e416154bf893275776f326`.

### DIFF
--- a/.github/runner/Dockerfile
+++ b/.github/runner/Dockerfile
@@ -159,12 +159,16 @@ ARG RUNNER_VERSION=2.334.0
 # hosts (it matches the majority of existing warm-cache entries; host normalization preserves them).
 # FlashInfer's flashinfer-cubin package keeps its packaged cubin tree under site-packages.
 # Some baselines materialize TRTLLM GEMM entries there at runtime, so the unprivileged runner
-# must own that tree instead of inheriting pip's root-owned directories.
+# must own the writable directories instead of inheriting pip's root-owned directories.
 RUN groupadd -g 1000 ci-runner && useradd -u 1000 -g 1000 -m -s /bin/bash ci-runner \
     && FLASHINFER_CUBIN_DIR="$(python -c 'import pathlib, flashinfer_cubin; print(pathlib.Path(flashinfer_cubin.get_cubin_dir()))' 2>/dev/null || true)" \
     && if [ -n "${FLASHINFER_CUBIN_DIR}" ]; then \
         mkdir -p "${FLASHINFER_CUBIN_DIR}/flashinfer/trtllm/gemm"; \
-        chown -R ci-runner:ci-runner "${FLASHINFER_CUBIN_DIR}"; \
+        chown ci-runner:ci-runner \
+            "${FLASHINFER_CUBIN_DIR}" \
+            "${FLASHINFER_CUBIN_DIR}/flashinfer" \
+            "${FLASHINFER_CUBIN_DIR}/flashinfer/trtllm" \
+            "${FLASHINFER_CUBIN_DIR}/flashinfer/trtllm/gemm"; \
     fi
 WORKDIR /home/ci-runner/runner
 RUN mkdir -p /home/ci-runner/runner/_work/_tool \
--- a/.github/runner/Dockerfile
+++ b/.github/runner/Dockerfile
@@ -96,9 +96,10 @@ RUN python -m pip freeze | grep -iE '^(torch|torchvision|torchaudio)==' > /tmp/t
     done
 # The Mamba benchmarks compare against the official mamba_ssm Triton kernels. Keep this
 # best-effort and dependency-free so it cannot re-resolve the CUDA-specific torch/triton stack
-# or pull the newer mamba_ssm tilelang/apache-tvm-ffi pins into the runner image. Skip the
-# unrelated CUDA extension build; these benchmarks import mamba_ssm.ops.triton.*.
-RUN MAMBA_FORCE_BUILD=TRUE MAMBA_SKIP_CUDA_BUILD=TRUE \
+# or pull the newer mamba_ssm tilelang/apache-tvm-ffi pins into the runner image. mamba_ssm's
+# package import loads selective_scan_cuda, so the CUDA extension must be built even when the
+# benchmark uses the Triton modules.
+RUN MAMBA_FORCE_BUILD=TRUE \
     python -m pip install --no-cache-dir --no-build-isolation --no-deps \
         -c /tmp/constraints.txt -c /tmp/torch-cu129.txt \
         "mamba-ssm==2.3.1" \
@@ -159,9 +160,11 @@ RUN if [ -n "${TILELANG_GIT_SHA}" ]; then \
         && git remote add origin https://github.com/tile-ai/tilelang.git \
         && git fetch --depth 1 origin "${TILELANG_GIT_SHA}" \
         && git checkout -q FETCH_HEAD \
-        && git submodule update --init --depth 1 --recursive 3rdparty/cutlass 3rdparty/tvm \
+        && git submodule update --init --depth 1 --recursive \
+            3rdparty/composable_kernel 3rdparty/cutlass 3rdparty/tvm \
         && mkdir -p /opt/tilelang-wheels \
-        && python -m pip wheel . --no-deps --no-build-isolation -w /opt/tilelang-wheels \
+        && CMAKE_BUILD_PARALLEL_LEVEL="${MAX_JOBS}" \
+            python -m pip wheel . --no-deps --no-build-isolation -w /opt/tilelang-wheels \
         && python -m pip install --no-deps /opt/tilelang-wheels/tilelang-*.whl \
         && rm -rf /tmp/tilelang; \
     elif [ -n "${TILELANG_VERSION}" ]; then \
--- a/.github/runner/Dockerfile
+++ b/.github/runner/Dockerfile
@@ -157,7 +157,15 @@ ARG RUNNER_VERSION=2.334.0
 # useradd before WORKDIR: home stays ci-runner-owned, and WORKDIR precedes the relative-path RUN (hadolint).
 # Pin UID/GID 1000 so the account deterministically owns the shared /ci-cache bind-mount across
 # hosts (it matches the majority of existing warm-cache entries; host normalization preserves them).
-RUN groupadd -g 1000 ci-runner && useradd -u 1000 -g 1000 -m -s /bin/bash ci-runner
+# FlashInfer's flashinfer-cubin package keeps its packaged cubin tree under site-packages.
+# Some baselines materialize TRTLLM GEMM entries there at runtime, so the unprivileged runner
+# must own that tree instead of inheriting pip's root-owned directories.
+RUN groupadd -g 1000 ci-runner && useradd -u 1000 -g 1000 -m -s /bin/bash ci-runner \
+    && FLASHINFER_CUBIN_DIR="$(python -c 'import pathlib, flashinfer_cubin; print(pathlib.Path(flashinfer_cubin.get_cubin_dir()))' 2>/dev/null || true)" \
+    && if [ -n "${FLASHINFER_CUBIN_DIR}" ]; then \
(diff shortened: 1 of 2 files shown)


.github/runner/Dockerfile
```diff
@@ -157,7 +157,15 @@ ARG RUNNER_VERSION=2.334.0
...
-RUN groupadd -g 1000 ci-runner && useradd -u 1000 -g 1000 -m -s /bin/bash ci-runner
+# FlashInfer's flashinfer-cubin package keeps its packaged cubin tree under site-packages.
+# Some baselines materialize TRTLLM GEMM entries there at runtime, so the unprivileged runner
+# must own that tree instead of inheriting pip's root-owned directories.
+RUN groupadd -g 1000 ci-runner && useradd -u 1000 -g 1000 -m -s /bin/bash ci-runner \
+    && FLASHINFER_CUBIN_DIR="$(python -c 'import pathlib, flashinfer_cubin; print(pathlib.Path(flashinfer_cubin.get_cubin_dir()))' 2>/dev/null || true)" \
+    && if [ -n "${FLASHINFER_CUBIN_DIR}" ]; then \
+        mkdir -p "${FLASHINFER_CUBIN_DIR}/flashinfer/trtllm/gemm"; \
+        chown -R ci-runner:ci-runner "${FLASHINFER_CUBIN_DIR}"; \
+    fi
```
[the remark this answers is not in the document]
superAngGao: Thanks, good catch. I agree that recursively chowning the whole packaged cubin tree can unnecessarily copy up the large pre-existing `.cubin` files into the final Docker layer.

I updated the Dockerfile to avoid `chown -R` here. The image now pre-creates the FlashInfer TRTLLM GEMM directory and only changes ownership for the writable directory chain (`cubins`, `flashinfer`, `trtllm`, and `gemm`). That should preserve the permission fix for the nightly FP8 GEMM benchmark while avoiding recursive ownership changes over the packaged cubin files.

.github/runner/README.md
```diff
@@ -37,6 +37,20 @@ DOCKER_BUILDKIT=1 docker build \
...
+Changes to this Dockerfile are not picked up by CI automatically. After a Dockerfile change
+lands, rebuild and tag the image from a GPU host using the build command above, then push it:
+
+```bash
+docker push ghcr.io/tile-ai/tileops-runner:<new-tag>
+```
+
+Then update the runner launcher configuration in TileOpsGov (`ci/runner-launcher/common.env`,
+the `IMAGE=` value) to point at the new tag, and restart/redeploy the local runner launcher.
+Merging the TileOPs PR only changes the image recipe; the live self-hosted runners keep using
```
lcy-seso: **Required:** `TileOpsGov` is a private repo — its name and internal paths (`ci/runner-launcher/common.env`, `IMAGE=`) must not appear in this public README. Keep only the public-relevant fact (rollout is manual, done by a maintainer outside this repo); the launcher steps belong in the internal ops docs.

```suggestion
Then redeploy the self-hosted runner launcher to use the new tag (maintainer task, done
outside this repository).
```

.github/runner/Dockerfile
```diff
@@ -157,7 +164,19 @@ ARG RUNNER_VERSION=2.334.0
...
+# must own the writable directories instead of inheriting pip's root-owned directories.
+RUN groupadd -g 1000 ci-runner && useradd -u 1000 -g 1000 -m -s /bin/bash ci-runner \
+    && FLASHINFER_CUBIN_DIR="$(python -c 'import pathlib, flashinfer_cubin; print(pathlib.Path(flashinfer_cubin.get_cubin_dir()))' 2>/dev/null || true)" \
+    && if [ -n "${FLASHINFER_CUBIN_DIR}" ]; then \
+        mkdir -p "${FLASHINFER_CUBIN_DIR}/flashinfer/trtllm/gemm"; \
+        chown ci-runner:ci-runner \
+            "${FLASHINFER_CUBIN_DIR}" \
+            "${FLASHINFER_CUBIN_DIR}/flashinfer" \
+            "${FLASHINFER_CUBIN_DIR}/flashinfer/trtllm" \
+            "${FLASHINFER_CUBIN_DIR}/flashinfer/trtllm/gemm"; \
```
lcy-seso: Non-blocking: this only covers the `trtllm/gemm` chain — another FlashInfer baseline writing a sibling subtree will hit the same root-owned-parent failure. Chowning directory inodes only covers the whole tree with no layer bloat (no `.cubin` files copied):

```suggestion
        find "${FLASHINFER_CUBIN_DIR}" -type d -exec chown ci-runner:ci-runner {} +; \
```

lcy-seso (commented):
Both fixes look correct (good root-cause analysis on the cubin permissions; pinning `mamba-ssm==2.3.1` over 2.3.2.post1 is the right call). Two inline comments below.

One more required change not tied to a code line: **remove the `TileOpsGov` references from the PR description as well** (the "Rollout note" names the private repo and its internal file path). Public PR bodies / commits / issues must not reference internal repos.

zhen8838 (dismissed):
Approved. Re-reviewed current head 37a2d9f. The Dockerfile changes remain consistent with the runner-image scope: mamba-ssm 2.3.1 uses the supported MAMBA_FORCE_BUILD and MAMBA_SKIP_CUDA_BUILD controls for the Triton-only baseline, DeepGEMM v2.1.1.post3 is pinned to its official source commit with the required header submodules, and the FlashInfer fix changes directory ownership without recursively chowning packaged .cubin files. The remaining verification requirement is operational rather than a code blocker: this image is built outside CI, so the final GPU-host build and ci-runner import/writeability checks must happen before rollout.

superAngGao:
Thanks for the earlier review notes. I pushed one more cleanup pass after the full local image rebuild:

- removed the internal rollout references from the public README/PR body, keeping only the maintainer/manual-rollout note
- kept the FlashInfer fix on directory ownership only, so packaged `.cubin` files are not recursively chowned
- changed the Mamba install to include the import-time CUDA extension because `mamba_ssm` imports `selective_scan_cuda`
- made the TileLang source build initialize `3rdparty/composable_kernel` explicitly and pass `CMAKE_BUILD_PARALLEL_LEVEL=${MAX_JOBS}`
- completed a local GPU-host build and runtime probe for torch/cu129, tilelang, FlashInfer cubin writeability, `selective_scan_cuda`, Mamba, and DeepGEMM; the live local runner containers were recreated from the verified image tag

superAngGao:
Additional rollout validation: after rebuilding and restarting the runner containers with the updated Docker image, the manually dispatched Nightly run passed on `main`:

- run: https://github.com/tile-ai/TileOPs/actions/runs/29403174158
- workflow: `Nightly`
- result: `success`
- jobs: `benchmark`, `op_test`, `publish-bench-data`, and `packaging` all completed successfully

This confirms the updated runner image covers both the benchmark dependency fixes and the op test environment for the current nightly workflow.

zhen8838 (approved):
Approved at current head d75ce92. Re-reviewed the incremental runner-image changes: mamba-ssm 2.3.1 is now built with the import-time selective_scan_cuda extension required by the official package, DeepGEMM fetch and submodule retries are bounded and remain best-effort, and the pinned TileLang composable_kernel submodule plus CMAKE_BUILD_PARALLEL_LEVEL wiring are consistent with the source tree. The runtime probe documents the relevant ci-runner checks. Completed CI checks are green; Dockerfile-related smoke/manifest/benchmark checks are skipped by change detection.

PR #1712, opened 2026-07-15
[CI][Nightly] Restore benchmark and op test coverage

### LINKED ISSUE #1696: [CI][SMOKE] Bring TileLang 0.1.11 skipped tests back into monitoring
## Goal

Bring the tests currently hidden behind the TileLang 0.1.11 skip guard back into normal smoke monitoring. The skip list should not become a permanent blind spot: every skipped area should either be fixed and re-enabled, or have a narrower tracked issue with an explicit owner/path.

## Sub-issues

- [x] #1694
- [x] #1695

### 1. MHA

MHA needs a separate path: route the MHA op/test coverage to the corresponding GQA kernel where the semantics match. This should restore MHA compile/autotune smoke coverage without keeping the current prefix skips indefinitely.

Tracked in #1694.

### 2. Non-MHA ops

The remaining skipped ops should be fixed or revalidated and then re-enter smoke:

- BatchNorm forward/backward
- Engram GateConv backward
- Engram decode, including multi-step decode
- BatchNorm `torch.compile` smoke tests in `tests/ops/test_norm_ops.py`

Tracked in #1695.

## Expected end state

- `TILELANG_019_KNOWN_FAILING_PATH_SUFFIXES` no longer hides non-MHA op tests.
- `TILELANG_019_KNOWN_FAILING_NODEIDS` no longer hides non-MHA op tests.
- MHA prefix skips are removed once MHA dispatch coverage is restored through the matching GQA path.
- CI smoke runs these tests in the official runner environment (`ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`, TileLang 0.1.11).

## Validation target

At minimum, the re-enabled non-MHA path should include:

```bash
python -m pytest -q tests/ops/test_batch_norm.py tests/ops/test_engram.py tests/ops/test_norm_ops.py -m smoke -rs --tb=short -p no:cacheprovider
```

MHA validation should be defined in #1694 based on the dispatch path selected there.

superAngGao:
Closes #1696

## Summary
- restore the nightly benchmark surface that was still hidden behind the old benchmark denylist
- restore MHA backward full op-test coverage by removing the runtime skip from `test_mha_bwd`
- route `MultiHeadAttentionBwdOp` through `GroupedQueryAttentionBwdOp` with `heads_kv == heads`, matching the existing MHA forward dispatch through GQA
- update the MHA backward workload to derive `lse` consistently with the GQA square-attention path
- update the MHA backward manifest source metadata to reference the `gqa_bwd_*` kernel map

## Problem
Earlier kernel-side fixes restored the relevant smoke coverage, but the broader nightly surfaces were not fully re-enabled afterward. In particular, benchmark cases were still hidden behind the old tilelang 0.1.9-era denylist, and MHA backward full op-test cases were still skipped at runtime.

As a result, the latest nightly run could pass while still reporting skipped benchmark and full-test coverage for code paths that should now be exercised again.

When the MHA backward skip is removed, the workload exposes a separate interface mismatch: input generation still assumes MHA forward returns exactly `(o, lse)`, while the maintained MHA forward path is backed by GQA.

## Fix
Remove the stale benchmark denylist from `benchmarks/conftest.py`, keeping the separate fp8 e4m3 benchmark skip because that still has an explicit unsupported/lowering reason.

`MultiHeadAttentionBwdOp` now delegates to `GroupedQueryAttentionBwdOp` with `heads_kv=heads`. The wrapper keeps the public MHA entry point and preserves the Op-level `dispatch_kernel(...)` contract. Legacy `mha_bwd_*` kernel-map override keys are now rejected explicitly because those kernels use the old MHA backward ABI and cannot be safely passed to the GQA backward op.

`MhaBwdTest.gen_inputs()` now extracts `o` defensively from the forward result and computes `lse` with the same square-attention helper used by the GQA backward workload.

### DIFF
--- a/benchmarks/conftest.py
+++ b/benchmarks/conftest.py
@@ -10,46 +10,6 @@ collect_ignore_glob = [
     "ops/attention/bench_deepseek_nsa*.py",
 ]
 
-TILELANG_019_BENCH_SKIP_REASON = (
-    "Skipped under tilelang 0.1.9: known benchmark regressions in autodiff/"
-    "codegen lowering produce incorrect numerics or compile failures; "
-    "re-enable when the benchmark runs cleanly against current tilelang."
-)
-
-TILELANG_019_BENCH_PATHS = {
-    # Linear attention / recurrent ops.
-    "benchmarks/ops/bench_engram.py",
-    # Attention.
-    "benchmarks/ops/attention/bench_gqa_decode.py",
-    "benchmarks/ops/attention/bench_gqa_decode_paged.py",
-    "benchmarks/ops/attention/bench_gqa_sliding_window.py",
-    "benchmarks/ops/attention/bench_gqa_sliding_window_varlen.py",
-    "benchmarks/ops/attention/bench_mha.py",
-    # Other operators.
-    "benchmarks/ops/bench_fft.py",
-    "benchmarks/ops/bench_mhc_post.py",
-    "benchmarks/ops/bench_mhc_pre.py",
-    "benchmarks/ops/bench_rope.py",
-    "benchmarks/ops/bench_topk_selector.py",
-    # Normalization.
-    "benchmarks/ops/bench_batch_norm.py",
-    "benchmarks/ops/bench_fused_add_layer_norm.py",
-    "benchmarks/ops/bench_instance_norm.py",
-    # MoE.
-    "benchmarks/ops/bench_moe_permute.py",
-    "benchmarks/ops/bench_moe_permute_align.py",
-    # GEMM.
-    "benchmarks/ops/bench_grouped_gemm.py",
-}
-
-TILELANG_019_BENCH_PREFIXES = ()
-
-TILELANG_019_BENCH_NODEIDS = {
-    "benchmarks/ops/bench_gemm.py::test_gemm_bench[wide-fp16]",
-    "benchmarks/ops/bench_group_norm.py::test_group_norm_bench[tail-spatial-g16-float16]",
-}
-
-
 def _normalized_benchmark_nodeid(item: pytest.Item) -> str:
     nodeid = item.nodeid
     if nodeid.startswith("benchmarks/"):
@@ -89,7 +49,6 @@ def pytest_sessionfinish(session, exitstatus):
 
 
 def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
-    tilelang_019_skip = pytest.mark.skip(reason=TILELANG_019_BENCH_SKIP_REASON)
     fp8_e4m3_skip = pytest.mark.skip(
         reason=(
             "Skipped under tilelang 0.1.9: fp8 e4m3 benchmark fails due to "
@@ -101,18 +60,6 @@ def pytest_collection_modifyitems(items: list[pytest.Item]) -> None:
     for item in items:
         nodeid = _normalized_benchmark_nodeid(item)
         path = nodeid.split("::", 1)[0]
-
-        if path in TILELANG_019_BENCH_PATHS:
-            item.add_marker(tilelang_019_skip)
-            continue
-
-        if nodeid.startswith(TILELANG_019_BENCH_PREFIXES):
-            item.add_marker(tilelang_019_skip)
-            continue
-
-        if nodeid in TILELANG_019_BENCH_NODEIDS:
-            item.add_marker(tilelang_019_skip)
-            continue
 
         if (
             path == "benchmarks/ops/bench_elementwise_fp8.py"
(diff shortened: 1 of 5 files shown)


tileops/ops/attention/mha.py
```diff
@@ -118,35 +126,54 @@ def __init__(self,
...
+            return None
+        translated = dict(kernel_map)
+        aliases = {
+            "mha_bwd_preprocess_kernel": "gqa_bwd_preprocess_kernel",
+            "mha_bwd_kernel": "gqa_bwd_kernel",
+            "mha_bwd_postprocess_kernel": "gqa_bwd_postprocess_kernel",
+        }
+        for old_key, new_key in aliases.items():
+            if old_key in translated and new_key not in translated:
+                translated[new_key] = translated.pop(old_key)
```
Ibuki-wind: `_gqa_kernel_map` renames legacy override keys without adapting the legacy kernel ABI. A pre-PR override such as `{"mha_bwd_kernel": MHABwdKernel}` is now passed to `GroupedQueryAttentionBwdOp`, which constructs it with the extra `heads_kv` argument; `MHABwdKernel` also has a different forward/output contract from `GQABwdKernel`. The claimed old-key compatibility therefore fails at construction or invocation. → Preserve legacy overrides through an adapter/legacy execution path, or explicitly reject the old ABI instead of silently translating it, and add a focused regression test using an MHA-signature fake kernel.

Ibuki-wind (changes requested):
### Overall
Resolve the blocking compatibility issue noted inline before enabling the restored coverage.

tileops/ops/attention/mha.py
superAngGao: Thanks, agreed. I removed the silent legacy MHA->GQA key translation and now explicitly reject legacy `mha_bwd_*` overrides with a clear error, since those kernels use the old MHA backward ABI and cannot be passed to `GroupedQueryAttentionBwdOp` safely. I also added a focused smoke regression test with an MHA-signature fake backward kernel to cover this case.

benchmarks/conftest.py
```diff
@@ -10,46 +10,6 @@
     "ops/attention/bench_deepseek_nsa*.py",
 ]
 
-TILELANG_019_BENCH_SKIP_REASON = (
-    "Skipped under tilelang 0.1.9: known benchmark regressions in autodiff/"
-    "codegen lowering produce incorrect numerics or compile failures; "
-    "re-enable when the benchmark runs cleanly against current tilelang."
-)
-
-TILELANG_019_BENCH_PATHS = {
```
bigSheep123, line 19: The old skip reason set the bar for re-enabling these: "re-enable when the benchmark runs cleanly against current tilelang". The PR body clears that bar for bench_mha (16 passed) plus single cases from grouped_gemm / gemm / group_norm, but the rest of the restored files are never named — engram, gqa_decode, gqa_decode_paged, both sliding_window variants, mhc_pre/post, batch_norm, fused_add_layer_norm, instance_norm, moe_permute_align. I ran bench_rope, bench_fft and bench_topk_selector against this head on an H200 and got 27 passed in ~5 min, so I'm fairly optimistic about the remainder — but if any of the unlisted files still trips on current tilelang, the first nightly after merge goes red and someone gets to bisect it back to this PR. Either paste one full green run of the restored set (the Docker runner you already used would do) into the PR body, or move the never-run files over to the #1714 follow-up alongside the other exclusions. Side note: bench_moe_permute.py in the old denylist doesn't exist in the tree anymore, so that entry was already dead weight.

bigSheep123 (commented):
### Overall
Delegation, manifest carve-out scope, and the lse derivation all check out; the one open point is how much of the restored benchmark surface has actually been run (inline).

benchmarks/conftest.py
superAngGao: Thanks, this is a fair bar. I ran the full restored benchmark set in the official runner image (`ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`) instead of keeping only the representative samples, and updated the PR body with the result.

Result: `144 passed, 3 skipped, 35 warnings in 5029.92s (1:23:49)`.

The command covered every existing file from the removed denylist plus the current nodeids for the old `bench_gemm` / `bench_group_norm` entries. The old `bench_moe_permute.py` entry is indeed dead because that file no longer exists. The only skips are the three `bench_gqa_decode_paged.py` `page_size=16` cases, which are already listed out of scope and tracked in #1714. No restored benchmark failed on current tilelang.

Ibuki-wind (approved):
Clean — no issues.

PR #1716, opened 2026-07-16
[Fix][NORM] Stop default_config from selecting a replicated fragment layout

### LINKED ISSUE #1715: [Perf][NORM] Row-reduction norm kernels collapse to a replicated layout on non-power-of-two rows (5-16x slowdown)
## Summary
The row-wise normalization kernels select a **replicated** fragment layout for
certain shapes and run **5–16× slower** than the partitioned layout. The output
is numerically correct, so no correctness test detects it — it only shows up as a
bad number in the benchmark.

## Affected ops
LayerNorm, RMSNorm, GroupNorm, FusedAddLayerNorm, FusedAddRMSNorm, AdaLayerNorm,
AdaLayerNorm-Zero — the `(block_m, N_padded)` + `reduce(dim=1)` family.
InstanceNorm aliases GroupNorm and is affected too. **BatchNorm is not** (different
kernel structure).

## When it happens
Rows whose padded length is **not a power of two** — e.g. the DiT hidden dim 1152
(pads to 1280), and conv-net GroupNorm rows `(C/G)·H·W` = 6272 / 7200. Power-of-two
rows (llama 4096 / 8192 / 16384) are unaffected. The trigger is *non-power-of-two
row length + block_m > 1*, not block_m > 1 by itself.

## Impact (measured, H200, M=1024, N=1152, fp16)
| op         | collapsed | healthy  | slowdown |
|------------|-----------|----------|----------|
| layer_norm | 289 us    | 17.8 us  | 16×      |
| rms_norm   | 70 us     | 13.8 us  | 5×       |

On the worst nightly case (AdaLN dit-xl-2) reported bandwidth drops to ~0.017 TB/s
(~0.35% of peak). Anyone running DiT or conv-nets hits this silently, since all
numeric tests stay green.

## Root cause (brief)
The kernels keep a `(block_m, N_padded)` tile in registers and reduce along the
row. For that reduction to lower to a cross-thread `AllReduce`, each row's elements
must be distributed uniformly across threads. When `block_m > 1` and TileLang's
greedy vector-width heuristic leaves the row misaligned across threads, every
partitioned layout candidate is rejected and layout inference falls back to a
replicated one: each thread declares a whole `N_padded`-element row, spills to
local memory, and the `AllReduce` degenerates into a per-thread serial loop.
`default_config`'s shared-memory-budget heuristic selected exactly the
`block_m ∈ {2,4,8,16}` values that trigger this.

## Fix direction
Pin `default_config` to `block_m = 1`: with one row per CTA the reduction's
per-row uniformity holds for any N/threads/dtype, so the collapse becomes
structurally impossible rather than avoided case-by-case.

michaelwithu:
Fixes #1715.

## Problem
The row-wise norm kernels (LayerNorm, RMSNorm, GroupNorm, FusedAdd{Layer,RMS}Norm,
AdaLayerNorm, AdaLayerNorm-Zero) select a **replicated** fragment layout for rows
whose padded length is not a power of two, running 5–38× slower than the
partitioned layout. Output is numerically correct, so no correctness test catches
it. `default_config`'s shared-memory-budget heuristic selected exactly the
`block_m ∈ {2,4,8,16}` values that trigger the collapse.

## Fix
Replace the five duplicated per-file heuristics with a shared `select_row_config()`
that pins `block_m = 1`. This is structural, not tuned: with one row per CTA the
layout heuristic's total-tile divisibility check coincides with the per-row
alignment the cross-thread `AllReduce` needs, so the collapse becomes impossible
for any N/threads/dtype. `autotune` keeps `block_m` as a swept knob
(`select_row_configs`) — the kernel interface is not narrowed; configs that
collapse simply lose on their measured runtime.

Scope: 7 row-reduction kernel classes (InstanceNorm aliases GroupNorm, so it is
covered). BatchNorm is excluded — it uses a different config schema and reduces
along a different axis.

### 1. The collapsed layout, in the generated CUDA (layer_norm, N=1152→1280)

Old default `block_m=16, threads=256` — each thread declares a **whole row** and
reduces serially; the cross-thread `AllReduce` is gone:

```cpp
half_t x_local[1280];   // whole row per thread (16× the 80-element fair share)
half_t x_f32[1280];
...
for (int rv = 0; rv < 1280; ++rv) { ... }   // intra-thread serial reduction
```

Fixed default `block_m=1, threads=128` — each thread holds its fair share and the
reduction is a cross-thread `AllReduce`:

```cpp
half_t x_local[10];     // fair share = block_m * N_padded / threads
...
acc[0] = tl::AllReduce<tl::SumOp, 128, ...>::run(acc[0], ...);
```

### 2. Before / after (H200, M=1024, N=1152, fp16, per-process)

| op                   | old default (collapsed) | new default (block_m=1) | speedup |
|----------------------|-------------------------|-------------------------|---------|
| layer_norm           | bm16,th256 — 310 us     | 16.3 us                 | 19×     |
| rms_norm             | bm8,th128 — 100 us      | 16.3 us                 | 6×      |
| group_norm           | bm16,th256 — 295 us     | 15.2 us                 | 19×     |
| ada_layer_norm       | bm16,th256 — 537 us     | 14.2 us                 | 38×     |
| fused_add_layer_norm | bm8,th256 — 840 us      | 23.0 us                 | 37×     |

### 3. `block_m > 1` brings no meaningful speedup, even where it is safe

On a power-of-two row (N=4096) where every `block_m` stays partitioned (no
collapse), sweeping `block_m` at threads=128, 3 runs each (µs):

| block_m | regs/thread (widest) | run1 | run2 | run3 |
|---------|----------------------|------|------|------|
| 1       | 32                   | 17.3 | 14.4 | 14.2 |
| 2       | 64                   | 21.7 | 27.7 | 15.8 |
| 4       | 128                  | 14.1 | 16.5 | 15.2 |
| 8       | 256                  | 15.9 | 15.2 | 15.9 |

`block_m=1` is within noise of (and often faster than) every larger `block_m`, and
the ordering is non-monotonic. Larger `block_m` only grows the per-thread register
footprint (32 → 256 elements) and risks the collapse on non-power-of-two rows,
with no upside. Pinning `block_m=1` by default sacrifices nothing measurable.

### 4. Healthy (power-of-two) cases: neutral to ~1.4× faster, no regression

The default changes for every shape, including the power-of-two rows that never
collapsed. Those already used a small `block_m`; the change to `threads=128`
lifts occupancy on wide rows (layer_norm, M=2048, fp16, per-process):

| N     | old default          | new default (bm1,th128) | effect                 |
|-------|----------------------|-------------------------|------------------------|
| 4096  | bm4,th256 — 14–21 us | 14–18 us                | neutral (within noise) |
| 8192  | bm2,th256 — 29.4 us  | 26.0 us                 | 1.13×                  |
| 16384 | bm1,th256 — 70.7 us  | 51.0 us                 | 1.4×                   |

No healthy case regresses; the gain grows with N and comes from `threads`, not
`block_m`.

## Verification
- norm numeric suite (smoke): 84 passed, 0 failed.
- pre-commit (incl. gitleaks): green.

### DIFF
--- a/tileops/kernels/norm/_config.py
+++ b/tileops/kernels/norm/_config.py
@@ -15,6 +15,8 @@ knob so the kernel interface is not narrowed; configs that collapse lose on thei
 measured runtime.
 """
 
+import torch
+
 __all__ = ["select_row_config", "select_row_configs"]
 
 # Powers of two only (tl::AllReduce is an XOR butterfly) that also divide
@@ -27,12 +29,17 @@ _CANDIDATE_BLOCK_M = (1, 2, 4, 8)
 _DEFAULT_THREADS = 128  # measured optimum; see the module docstring
 
 
-def _feasible_threads(n_padded: int) -> list[int]:
-    """Thread counts that divide the row and keep loads vectorizable (>=8/thread)."""
+def _feasible_threads(n_padded: int, dtype: torch.dtype = torch.float16) -> list[int]:
+    """Thread counts that divide the row and keep loads 128-bit vectorizable.
+
+    128-bit needs ``16 // element_size`` elements per thread (8 for fp16/bf16,
+    4 for fp32), so the floor is dtype-dependent.
+    """
+    min_elements = 16 // torch.tensor([], dtype=dtype).element_size()
     return [
         threads
         for threads in _CANDIDATE_THREADS
-        if n_padded % threads == 0 and n_padded // threads >= 8
+        if n_padded % threads == 0 and n_padded // threads >= min_elements
     ]
 
 
@@ -48,13 +55,13 @@ def select_row_config(n_padded: int) -> dict:
     return {"block_m": 1, "threads": threads}
 
 
-def select_row_configs(n_padded: int) -> list[dict]:
+def select_row_configs(n_padded: int, dtype: torch.dtype = torch.float16) -> list[dict]:
     """Autotune space: block_m x usable thread counts, default always included.
 
     block_m is swept (not pinned) so the interface is preserved; configs that
     collapse are rejected by the autotuner's own measured runtime.
     """
-    threads = _feasible_threads(n_padded)
+    threads = _feasible_threads(n_padded, dtype)
     configs = [
         {"block_m": block_m, "threads": t}
         for block_m in _CANDIDATE_BLOCK_M
--- a/tileops/kernels/norm/_config.py
+++ b/tileops/kernels/norm/_config.py
@@ -1,0 +1,66 @@
+"""Tile-config selection shared by the row-wise norm kernels.
+
+These kernels hold a ``(block_m, N_padded)`` row block in register fragments and
+reduce along the row. When ``N_padded`` is not a power of two, TileLang's layout
+inference places a partitioned layout for some ``block_m`` but falls back to a
+replicated one for others -- each thread owns a whole row, spills to local
+memory, and the cross-thread ``AllReduce`` degenerates into a serial loop. That
+is 5-16x slower yet numerically correct, so no correctness test catches it.
+
+``select_row_config`` pins ``block_m=1``: with one row per CTA the per-row
+uniformity the reduction needs always holds, so the collapse is impossible for
+any N/threads/dtype -- structural, not a tuned value. ``threads=128`` is the
+measured optimum. ``select_row_configs`` keeps ``block_m`` as a swept autotune
+knob so the kernel interface is not narrowed; configs that collapse lose on their
+measured runtime.
+"""
+
+__all__ = ["select_row_config", "select_row_configs"]
+
+# Powers of two only (tl::AllReduce is an XOR butterfly) that also divide
+# N_padded, or layout inference reports "no available layout". CUDA caps at 1024.
+_CANDIDATE_THREADS = (128, 256, 512, 1024)
+
+# block_m values offered to autotune; block_m=1 is always the safe default.
+_CANDIDATE_BLOCK_M = (1, 2, 4, 8)
+
+_DEFAULT_THREADS = 128  # measured optimum; see the module docstring
+
+
+def _feasible_threads(n_padded: int) -> list[int]:
+    """Thread counts that divide the row and keep loads vectorizable (>=8/thread)."""
+    return [
+        threads
+        for threads in _CANDIDATE_THREADS
+        if n_padded % threads == 0 and n_padded // threads >= 8
+    ]
+
+
+def select_row_config(n_padded: int) -> dict:
+    """Structurally collapse-free default ``{block_m, threads}`` for a row reduction."""
+    threads = _DEFAULT_THREADS
+    if n_padded % threads:
+        # N_padded is a multiple of the 256-element alignment, so 128 always
+        # divides it; this is a guard, not a live path.
+        threads = next(
+            (t for t in _CANDIDATE_THREADS if n_padded % t == 0), _DEFAULT_THREADS
+        )
+    return {"block_m": 1, "threads": threads}
+
+
+def select_row_configs(n_padded: int) -> list[dict]:
+    """Autotune space: block_m x usable thread counts, default always included.
+
+    block_m is swept (not pinned) so the interface is preserved; configs that
+    collapse are rejected by the autotuner's own measured runtime.
+    """
+    threads = _feasible_threads(n_padded)
+    configs = [
+        {"block_m": block_m, "threads": t}
+        for block_m in _CANDIDATE_BLOCK_M
+        for t in threads
+    ]
+    default = select_row_config(n_padded)
+    if default not in configs:
+        configs.insert(0, default)
+    return configs
--- a/tileops/kernels/norm/_config.py
+++ b/tileops/kernels/norm/_config.py
@@ -33,14 +33,13 @@ def _feasible_threads(n_padded: int, dtype: torch.dtype = torch.float16) -> list
     """Thread counts that divide the row and keep loads 128-bit vectorizable.
 
     128-bit needs ``16 // element_size`` elements per thread (8 for fp16/bf16,
-    4 for fp32), so the floor is dtype-dependent.
+    4 for fp32). If no candidate meets that floor (small rows), fall back to any
+    thread count that divides the row so the autotune space is never empty.
     """
-    min_elements = 16 // dtype.itemsize
-    return [
-        threads
-        for threads in _CANDIDATE_THREADS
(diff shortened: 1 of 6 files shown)


michaelwithu:
Thanks @gemini-code-assist — good catch. The `>= 8` elements/thread floor in `_feasible_threads` was fp16-centric (128-bit = 8 fp16 elements); fp32 only needs 4. Addressed in the follow-up commit: the floor is now `16 // element_size` and `dtype` is threaded through `select_row_configs` and every op's `autotune_configs`. Verified: `n_padded=1024` fp32 now offers threads {128, 256} (was {128}), fp16 unchanged, and the `default_config` path is untouched.

michaelwithu:
Round 2 addressed (both correct): (1) `_feasible_threads` now reads `dtype.itemsize` instead of allocating a temp tensor; (2) removed the dead `threads` fallback in `select_row_config` — every candidate thread count is a multiple of 128, so it never selected anything but 128, and `n_padded` is always a multiple of the 256 alignment anyway. Behavior unchanged (default `{block_m:1, threads:128}`, dtype-aware autotune space intact); ruff + numeric smoke green.

michaelwithu:
Round 3 — all 4 addressed:
1. `_feasible_threads`: reverted the element-size read to `torch.tensor([], dtype=dtype).element_size()` (version-safe and matches the existing per-file convention; FWIW `dtype.itemsize` does work on torch>=2.1, but the tensor form avoids the debate) **and** added a fallback to any dividing thread count so it is never empty for small rows.
2. `select_row_configs`: restored the 48 KB shared-memory budget via a `num_buffers` parameter that caps `block_m`; the default is always inserted, so the list is never empty even when the budget admits no larger block_m.
3+4. FusedAdd{Layer,RMS}Norm pass `num_buffers=2` (two shared buffers).

Verified: default unchanged (bm1/th128); dtype-aware threads intact (fp32 n=1024 → {128,256}); N=16384 correctly caps block_m (nb=1 → bm1 x {128..1024}, nb=2 → default only); ruff + numeric smoke green.

zhen8838 (approved):
Reviewed current head `b7534a687997da05097e242d8b9d92b63c934e12` against the previously reviewed `756903d`. The increment only tightens norm autotune candidate selection: dtype-aware element sizing, a fallback when vectorizable thread candidates are empty, and shared-memory budgeting for the two-buffer fused add norms. Current CI is green, including gpu-smoke and pre-commit. No blocking code issues found.

tileops/kernels/norm/_config.py
```diff
@@ -0,0 +1,73 @@
...
+    min_elements = 16 // torch.tensor([], dtype=dtype).element_size()
+    candidates = [t for t in _CANDIDATE_THREADS if n_padded % t == 0]
+    vectorizable = [t for t in candidates if n_padded // t >= min_elements]
+    return vectorizable or candidates
+
+
+def select_row_config(n_padded: int) -> dict:
+    """Structurally collapse-free default ``{block_m, threads}`` for a row reduction."""
+    # N_padded is a multiple of the 256-element alignment, so 128 always divides it.
+    return {"block_m": 1, "threads": _DEFAULT_THREADS}
```
bigSheep123, line 48: Pinning `block_m=1` makes `M % block_m == 0` for every M, which silently neutralizes all pre-existing tail-M coverage: the "Tail-M: M not divisible by block_m (proves T.copy partial block safety)" params in tests/ops/test_rms_norm.py, test_layer_norm.py, test_ada_layer_norm.py, test_ada_layer_norm_zero.py, test_fused_add_layer_norm.py, test_fused_add_rms_norm.py (e.g. M=1025), and tests/ops/test_group_norm.py::test_group_norm_no_affine_tail_block (M in {3, 15, 1}) all construct ops with tune=False and no explicit config, so they now run at block_m=1 and never reach the partial-block `T.copy` masking path. That path is still shipped — `select_row_configs` sweeps block_m in {2, 4, 8} and callers may pass `config=` — so it is now covered by zero tests (pre-review "Coverage set": cells neutralized while the code path is still present). → Add one tail-M case that pins `config={"block_m": 8, "threads": 128}` (mirroring the `_TAIL_BLOCK_M` pattern at tests/ops/test_vector_norm.py:114) for at least one of the seven row-norm kernel classes; issue #1715 declares no Constraints, so the trust model permits the test landing in this PR.

tileops/kernels/norm/_config.py
```diff
@@ -0,0 +1,73 @@
...
+    max_block_m = (48 * 1024) // (num_buffers * smem_per_row)
+    configs = [
+        {"block_m": block_m, "threads": t}
+        for block_m in _CANDIDATE_BLOCK_M
+        if block_m <= max_block_m
+        for t in threads
+    ]
+    default = select_row_config(n_padded)
+    if default not in configs:
+        configs.insert(0, default)
```
bigSheep123, line 72: `configs.insert(0, default)` can insert an invalid config: for an `n_padded` that is not a multiple of 128 (e.g. 192), `_feasible_threads` returns `[]` and the inserted default `{"block_m": 1, "threads": 128}` violates the threads-divides-row requirement, so autotune fails at kernel compile time instead of raising a clear error. Unreachable today (every caller aligns to 256 elements) but the invariant lives only in the callers. → `assert n_padded % _DEFAULT_THREADS == 0` at the top of `select_row_configs` to make the contract explicit at the boundary.

tileops/kernels/norm/_config.py
```diff
@@ -0,0 +1,73 @@
+"""Tile-config selection shared by the row-wise norm kernels.
+
+These kernels hold a ``(block_m, N_padded)`` row block in register fragments and
+reduce along the row. When ``N_padded`` is not a power of two, TileLang's layout
+inference places a partitioned layout for some ``block_m`` but falls back to a
+replicated one for others -- each thread owns a whole row, spills to local
+memory, and the cross-thread ``AllReduce`` degenerates into a serial loop. That
+is 5-16x slower yet numerically correct, so no correctness test catches it.
```
bigSheep123, line 8: The module docstring says "That is 5-16x slower", but the PR's own before/after table measures up to 38x (ada_layer_norm 537 → 14.2 us). → Update the range to 5-38x so the doc matches the measured evidence.

bigSheep123 (changes requested):
### Overall
Pinning the default to `block_m=1` leaves the partial-block (tail-M) `T.copy` masking path with zero test coverage while it remains reachable via `select_row_configs` and user-supplied configs — land one pinned-config tail-M test (issue #1715 declares no Constraints, so a joint impl+test PR is permitted) before merge.

Issue #1664, opened 2026-07-07
[FEAT][GEMM] fp16/bf16 dense GEMM: small-batch kernel-mode + GemmOp/GemvKernel perf tuning


bigSheep123:
# [FEAT][GEMM] fp16/bf16 dense GEMM: small-batch kernel-mode + GemmOp/GemvKernel perf tuning

Carved out of the GEMM category (#1653): the fp16/bf16 half. The FP8 half is tracked in
#1657. This issue is fp16/bf16-only.

## Scope — three nominal ops, one op class

| #400 nominal op | Realized as | State | Work here |
|---|---|---|---|
| gemm_fp16 | `GemmOp` (fp16 + bf16, warp-specialized Hopper kernel) | ✅ shipped | perf tuning |
| gemv_fp16 | kernel-mode of `GemmOp` (`GemvKernel`, m==1/n==1) | ✅ shipped | perf tuning |
| small_batch_fp16 | kernel-mode of `GemmOp` (small-M schedule) | ⬜ unimplemented | implement |

Design is settled by #1653: small-batch is a **kernel-mode of `GemmOp`** (same signature,
kernel picked at forward), not a separate op — supersedes the earlier separate-op proposal
in #370 (closed not-planned).

## Motivation — measured gaps

The only quantified dense-GEMM-vs-cuBLAS record is the H200 table in #1629, measuring the
pre-rewrite cp.async kernel: parity at m=4096 (0.92–1.02×), but **0.16–0.47× at m≤1024
decode shapes** (worst: 128×2112×7168 at 0.16×) and 0.69× on non-256-aligned N. The
warp-specialized kernel that shipped in #1638 targets exactly that regime but has **no
published numbers**, so a fresh baseline is deliverable zero. `GemvKernel` was last
measured in #244 (72–90% of peak BW on large n×k, 11% util at 1024×1024) and never
re-measured after unification into `GemmOp` (#245).

### 0. Baseline (prerequisite for everything below)

- [ ] Re-run `benchmarks/ops/bench_gemm.py` on H200 across all manifest workloads; post
      the WS-kernel-vs-cuBLAS table in this issue.
- [ ] Resolve the unanswered review flag on #1638: `T.warpgroup_fence_operand(c_local,
      num_regs=64)` is sized for 128×64 tiles while the default config is 128×128 —
      verify or fix before any tuning (correctness precedes performance).

### 1. small_batch_fp16 — new kernel-mode (2 ≤ m ≤ 16)

[...]

PR #1720, opened 2026-07-16
[Maintain][GEMM] Add mid-m band (m=16/32/64/96) NT bf16 workloads

bigSheep123:
## What

Add four mid-m (m ∈ {16, 32, 64, 96}) NT bf16 workloads to the `GemmOp` manifest entry, filling the coverage gap between the small-m fast paths (m ≤ 8) and the decode/prefill sweep (m ∈ {128, 1024, 4096}).

The n/k pairs reuse the DeepSeek-V3 projection shapes already present in the sweep:

| label | m | n | k |
|---|---|---|---|
| mid-m16-attn | 16 | 4096 | 7168 |
| mid-m32-attn | 32 | 4096 | 7168 |
| mid-m64-down | 64 | 7168 | 2048 |
| mid-m96-gate-up | 96 | 2112 | 7168 |

## Why

The m = 9..127 band has no manifest coverage today, and it is the target of the upcoming dense-GEMM work (analytic config selection, mid-m kernel structure). Acceptance for that work is gated on "all manifest workloads, no regression", which requires the band to be represented.

Baseline measured on H200 (CUPTI kernel-only timing with per-iteration L2 flush; per-rep interleaved ours-vs-baseline ratios; baseline = min(torch.matmul default, cuBLASLt searched-best)): the current generic path runs these four shapes at 0.48 / 0.48 / 0.67 / 0.35 of cuBLAS respectively.

## Scope

Manifest-only, per the manifest trust model (workloads edits on an implemented op require a standalone human-reviewed PR): one file, +7 lines (3 comment lines + 4 entries). No code, test, or benchmark files touched.

## Verification

- `scripts/validate_manifest.py --check-op GemmOp` passes (the only warning is the pre-existing `_infer_output_shapes` advisory, unrelated to workloads).
- `load_workloads("GemmOp")` returns 12 entries (was 8).
- `tests/ops/test_gemm.py` is hand-parametrized, so the test matrix is unchanged; `benchmarks/ops/bench_gemm.py` is manifest-driven and gains 4 bench rows.
- Correctness of the four shapes on the current generic kernel path verified locally (max rel err ≤ 3e-3 vs torch.matmul).

Refs #1664

### DIFF
--- a/tileops/manifest/gemm.yaml
+++ b/tileops/manifest/gemm.yaml
@@ -37,6 +37,13 @@ GemmOp:
   - {m: 4096, n: 4096, k: 7168, trans_a: false, trans_b: true, dtypes: [float16, bfloat16], label: "ds-v3-prefill-attn-proj"}
   - {m: 4096, n: 7168, k: 16384, trans_a: false, trans_b: true, dtypes: [bfloat16], label: "k-dominant-7168x16384"}
   - {m: 4096, n: 24576, k: 1536, trans_a: false, trans_b: true, dtypes: [bfloat16], label: "wide-n-24576"}
+  # Mid-m band: samples of m = 9..127, the token-count gap below the
+  # decode/prefill sweep (m in {128, 1024, 4096}); n/k reuse the
+  # DeepSeek-V3 projection shapes above.
+  - {m: 16, n: 4096, k: 7168, trans_a: false, trans_b: true, dtypes: [bfloat16], label: "mid-m16-attn"}
+  - {m: 32, n: 4096, k: 7168, trans_a: false, trans_b: true, dtypes: [bfloat16], label: "mid-m32-attn"}
+  - {m: 64, n: 7168, k: 2048, trans_a: false, trans_b: true, dtypes: [bfloat16], label: "mid-m64-down"}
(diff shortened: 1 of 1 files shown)


Ibuki-wind (approved):
Clean — no issues.

bigSheep123:
The two earlier `validate-pr-title` failures were evaluating a stale event payload, not the current title: the `synchronize` event for ccbff68 snapshotted the pre-rename title `[Manifest][GEMM]` (renamed 07:16:56Z, event created 07:16:58Z), and the 08:04 re-run reused that same payload, so it could never see the fix.

The title is now `[Maintain][GEMM]` — matching the manifest-only diff, per the `[Maintain][…]` convention of past manifest PRs — and an empty commit (41340a8) mints a fresh `synchronize` event whose payload carries the corrected title.

The new workflow runs are waiting at `action_required`; once a maintainer approves them, `validate-pr-title` should pass and the remaining required checks will run.

zhen8838 (approved):
Rechecked current head 41340a8. The only delta from the reviewed manifest is an empty preflight-retrigger commit; the four NT bf16 GemmOp workload rows remain schema- and contract-consistent, and git diff --check passes. No GitHub checks are currently reported.

PR #1728, opened 2026-07-19
[Fix][NORM] Fix global memory write-read data race in `BatchNormFwdTrainKernel`

### LINKED ISSUE #1727: [Bug] BatchNormFwdTrainKernel: intra-block global memory read-write race cause by missing barrier
### Description

The training-mode batch-norm forward kernel `BatchNormFwdTrainKernel` ([`tileops/kernels/norm/batch_norm.py` L175-L176](https://github.com/tile-ai/TileOPs/blob/48782d50e698c00d053e8cd0b2f2625e6ca54bdd/tileops/kernels/norm/batch_norm.py#L175-L176)) updates the **running statistics** — the per-channel `running_mean` / `running_var` that batch norm accumulates across batches and uses at inference — with a read-modify-write that **every thread of the block runs on the same global address**, with no barrier ordering one thread's write against another thread's read. 

This is an **intra-block, global-memory write-read race**: `running_mean` / `running_var` come out **non-deterministic** (numerically **wrong** once `running_mean` is nonzero). The bug is architecture-independent; reproduced here on `sm_120`.

### Environment
TileOPs `48782d5` (current main)
Python: 3.10.19 (Linux, x86_64) 
TileLang: 0.1.12 
torch: 2.10.0+cu128

GPU: NVIDIA GeForce RTX 5070 Ti (sm_120)
CUDA: 12.8 

### Source Code in TileLang DSL (buggy)

[`tileops/kernels/norm/batch_norm.py` L175-L176](https://github.com/tile-ai/TileOPs/blob/48782d50e698c00d053e8cd0b2f2625e6ca54bdd/tileops/kernels/norm/batch_norm.py#L175-L176):

```python
with T.Kernel(C, threads=threads) as (bc):          # one block per channel, `threads` threads
    ...
    running_mean[bc] = (1.0 - mom) * running_mean[bc] + mom * mean_val   # <-- read-modify-write
    running_var[bc]  = (1.0 - mom) * running_var[bc]  + mom * unbiased_var  # <-- read-modify-write
```

Because they sit at block scope (`with T.Kernel(C, threads=threads) as (bc)`), every thread of the block does the same read-modify-write to the one global memory `running_mean[bc]` (and `running_var[bc]`), unsynchronized. That is the race.

### Compiled TVM TIR

All threads (`threads=256` at the default config) do such global memory writes, same as the source code:

```python
tx = T.launch_thread("threadIdx.x", 256)
...
# executed by every thread, no `if tx == 0` guard:
running_mean_1[bx] = T.float32(0.9) * running_mean_1[bx] + T.float32(0.1) * mean_val  # <-- read-modify-write
running_var_1[bx]  = T.float32(0.9) * running_var_1[bx]  + T.float32(0.1) * (var_val * 4096.0 / 4095.0)  # <-- read-modify-write
```

When the block's warps desynchronize, a later warp loads `running_mean_1[bx]` after an earlier warp has already stored to it, so the `0.9*x + 0.1*mean` step is applied more than once — 1 up to `#warps` times, depending on the schedule. Hence, the running statistic is non-deterministic.

### Reproduction

This driver uses TileOPs' own test infrastructure. Run it from the TileOPs repo root directory.

```bash
git clone https://github.com/tile-ai/TileOPs.git
cd TileOPs && git checkout 48782d50e698c00d053e8cd0b2f2625e6ca54bdd
pip install -e .
pip install tilelang==0.1.12
PYTHONPATH=. python reproduce.py            # from the repo root
```

reproduce.py:
```python
import torch
from workloads.batch_norm import BatchNormFwdTest
from tileops.ops.norm.batch_norm import BatchNormFwdOp
from tests.ops.test_batch_norm import _ref_fwd
from tileops.kernels.norm.batch_norm import BatchNormFwdTrainKernel, BatchNormFwdInferKernel

# sm_120 architecture
for k in (BatchNormFwdTrainKernel, BatchNormFwdInferKernel):
    k.supported_archs += [120]

torch.manual_seed(0)
# C=1024 -> 1024 blocks (oversubscribe the ~70 SMs); L=N*512=8192 -> 8 warps/block.
N, C, spatial = 16, 1024, (512,)
x, w, b, rm0, rv0 = BatchNormFwdTest(N, C, spatial, torch.float16, True).gen_inputs()
op = BatchNormFwdOp(training=True)
_, ref_rm, _ = _ref_fwd(x, w, b, rm0.clone(), rv0.clone(), training=True)

nondet = fails = 0; first = None; err = 0.0
for _ in range(200):
    rm, rv = rm0.clone(), rv0.clone()
    op(x, rm, rv, w, b)
    if first is None: first = rm.clone()
    nondet += not torch.equal(rm, first)
    err = max(err, (rm - ref_rm).abs().max().item())
    fails += not torch.allclose(rm.float(), ref_rm.float(), atol=1e-2, rtol=1e-2)

print(f"running_mean nondeterministic: {nondet}/199   (race-free => 0)")
print(f"their allclose(atol=1e-2) fails: {fails}/200   (max err {err:.5f})")
```

### Expected behavior

A race-free kernel is deterministic:
```
running_mean nondeterministic: 0/199     (race-free => 0)
their allclose(atol=1e-2) fails: 0/200   (max err 0.00000)
```

Actual output on `sm_120` at `C=1024`:
```
running_mean nondeterministic: 199/199   (race-free => 0)
their allclose(atol=1e-2) fails: 0/200   (max err 0.00686)
```

The data race occurs every run and silently corrupts results (`running_mean` is bit-different each time, 199/199). The `allclose` still passes (0/200): the `running_mean=0` init keeps the error under `1e-2`. 

> With a realistic, nonzero `running_mean`, the same race errors can increase by orders of magnitude.

### Suggested Fix

Guard the two updates to a single thread — the standard `T.get_thread_binding()` idiom (as in
`topk_selector`, `grouped_gemm`) — at [`tileops/kernels/norm/batch_norm.py` L175-L176](https://github.com/tile-ai/TileOPs/blob/48782d50e698c00d053e8cd0b2f2625e6ca54bdd/tileops/kernels/norm/batch_norm.py#L175-L176):

```diff
@@ -174,3 +174,4 @@ with T.Kernel(C, threads=threads) as (bc):
                 unbiased_var = var_val * T.cast(L, accum_dtype) / (T.cast(L, accum_dtype) - T.cast(1.0, accum_dtype))
-                running_mean[bc] = (T.cast(1.0, accum_dtype) - mom) * running_mean[bc] + mom * mean_val
-                running_var[bc] = (T.cast(1.0, accum_dtype) - mom) * running_var[bc] + mom * unbiased_var
+                if T.get_thread_binding() == 0:
+                    running_mean[bc] = (T.cast(1.0, accum_dtype) - mom) * running_mean[bc] + mom * mean_val
+                    running_var[bc] = (T.cast(1.0, accum_dtype) - mom) * running_var[bc] + mom * unbiased_var
```

`mean_val` / `unbiased_var` are block-uniform, so restricting to one writer is behavior-preserving. (A barrier separating the reads from the writes — the "missing barrier" lens — would also fix it, but keeps 256 redundant stores; the single-thread guard is cleaner.)

### Other Notes

`test_batch_norm_fwd` has no config where the channel count `C` oversubscribes the SMs (all use `C <= 256`), so a block's warps never desynchronize and the race stays hidden. Add a high-`C` config and assert **determinism** — the `running_mean = 0` init keeps the error under `atol`, so the existing `allclose` alone stays green even when the race fires:

```diff
@@ -78,1 +78,2 @@ class BatchNormFwdFixture(FixtureBase):
             pytest.param(8, 64, (30, 30), torch.bfloat16, True, marks=pytest.mark.full),
+            pytest.param(16, 1024, (512,), torch.float16, True, marks=pytest.mark.full),
         ]),

@@ -131,1 +131,6 @@ def test_batch_norm_fwd(N, C, spatial, dtype, training):
     if training:
+        rm2, rv2 = running_mean_ref.clone(), running_var_ref.clone()
+        op(x, rm2, rv2, weight, bias)
+        det_err = (running_mean.float() - rm2.float()).abs().max()
+        assert torch.equal(running_mean, rm2) and torch.equal(running_var, rv2), \
+            f"running stats non-deterministic across runs: max_err={det_err:.4e}"
         rm_err = (running_mean.float() - ref_rm.float()).abs().max()
```

arxgy:
[`BatchNormFwdTrainKernel`](https://github.com/tile-ai/TileOPs/blob/48782d50e698c00d053e8cd0b2f2625e6ca54bdd/tileops/kernels/norm/batch_norm.py#L175-L176) updates `running_mean` / `running_var` with a read-modify-write run by every thread of the block, with no barrier between one thread's write and another's read — an intra-block global-memory write-read race on `running_mean[bc]`. Once the channel count oversubscribes the SMs, the running statistics come out non-deterministic (and wrong once `running_mean` is nonzero).

This PR guards the update to one thread (`T.get_thread_binding() == 0`). The stored values are
block-uniform, so it is behavior-preserving, with no measured perf change.

### Bug Fix

- Fixes #1727 

### Regression Test

Adds a `C=1024` config that exposes the race and a determinism assertion (a second run must match;
`allclose` is masked by the `running_mean=0` init). Fails on the current kernel, passes on the fix.
`scripts/test_node_delta.py`: 19 -> 20 (+1).

### DIFF
--- a/tests/ops/test_batch_norm.py
+++ b/tests/ops/test_batch_norm.py
@@ -76,6 +76,8 @@ class BatchNormFwdFixture(FixtureBase):
             # Non-aligned spatial: H*W=900, exercises partial-tile path
             pytest.param(8, 64, (30, 30), torch.float16, True, marks=pytest.mark.full),
             pytest.param(8, 64, (30, 30), torch.bfloat16, True, marks=pytest.mark.full),
+            # High channel count oversubscribes the SMs, exposing the running-stat update race.
+            pytest.param(16, 1024, (512,), torch.float16, True, marks=pytest.mark.full),
         ]),
     ]
 
@@ -129,6 +131,13 @@ def test_batch_norm_fwd(N, C, spatial, dtype, training):
         f"fwd mismatch (training={training}): max_err={max_err:.4e}"
 
     if training:
+        # allclose is masked when running_mean starts near the batch mean; check determinism.
+        rm2, rv2 = running_mean_ref.clone(), running_var_ref.clone()
+        op(x, rm2, rv2, weight, bias)
+        det_err = (running_mean.float() - rm2.float()).abs().max()
+        assert torch.equal(running_mean, rm2) and torch.equal(running_var, rv2), \
+            f"running stats non-deterministic across runs: max_err={det_err:.4e}"
+
         rm_err = (running_mean.float() - ref_rm.float()).abs().max()
         assert torch.allclose(running_mean.float(), ref_rm.float(), atol=atol, rtol=rtol), \
             f"running_mean mismatch: max_err={rm_err:.4e}"
--- a/tileops/kernels/norm/batch_norm.py
+++ b/tileops/kernels/norm/batch_norm.py
@@ -172,8 +172,10 @@ def _batch_norm_fwd_train_kernel(
                 # (Bessel's correction: biased_var * L / (L - 1)).
                 mom = T.cast(momentum, accum_dtype)
                 unbiased_var = var_val * T.cast(L, accum_dtype) / (T.cast(L, accum_dtype) - T.cast(1.0, accum_dtype))
-                running_mean[bc] = (T.cast(1.0, accum_dtype) - mom) * running_mean[bc] + mom * mean_val
-                running_var[bc] = (T.cast(1.0, accum_dtype) - mom) * running_var[bc] + mom * unbiased_var
+                # One writer per block: this running-stat RMW races if every thread runs it.
+                if T.get_thread_binding() == 0:
+                    running_mean[bc] = (T.cast(1.0, accum_dtype) - mom) * running_mean[bc] + mom * mean_val
+                    running_var[bc] = (T.cast(1.0, accum_dtype) - mom) * running_var[bc] + mom * unbiased_var
 
                 # Pass 2 – normalize.
                 if block_l >= L:


Ibuki-wind (approved):
Clean — no issues. The single-writer fix is correct and the regression test is well targeted. Thank you for the thoughtful contribution!

PR #1735, opened 2026-07-21
[Fix][Benchmark] Skip FlashInfer FP8 per-tensor baseline

### LINKED ISSUE #1733: [BUG][CI][Benchmark] Nightly FP8 GEMM benchmark segfaults in FlashInfer per-tensor baseline
## Problem

The `Nightly / benchmark` job is currently failing repeatedly with a native segfault, exit code 139. This is not a pytest assertion failure or a performance threshold regression; the pytest process crashes.

Example failing Actions job: https://github.com/tile-ai/TileOPs/actions/runs/29699980125/job/88227280432

The observed crash point is:

```text
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[small-batch-down-m8-per-tensor-float8_e4m3fn]
```

The native stack points into the FlashInfer / TRTLLM low-latency GEMM path:

```text
cuModuleGetFunction
gemm::gemm::GemmInterface::run(...)
flashinfer::trtllm_low_latency_gemm(...)
__tvm_ffi_trtllm_low_latency_gemm
Segmentation fault (core dumped)
```

## Environment

Reproduced locally with the official runner image and the same cache mount used by the runner service:

```text
image: ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10
user: ci-runner, uid=1000:1000
source path: /home/ci-runner/runner/_work/TileOPs/TileOPs
cache mount: -v /data3/shared/ci-cache:/ci-cache
torch: 2.10.0+cu129
tilelang: 0.1.11+cu129.git65dbc983
flashinfer: 0.6.11.post2
GPU: H200
GPU clock: 1500 MHz
```

Relevant env:

```text
TILELANG_CACHE_DIR=/ci-cache/tilelang
TILELANG_TMP_DIR=/ci-cache/tilelang/tmp
TRITON_CACHE_DIR=/ci-cache/triton
PIP_CACHE_DIR=/ci-cache/pip
MAX_JOBS=64
PYTORCH_ALLOC_CONF=expandable_segments:True
```

No tilelang, torch, or dependency install/upgrade was performed inside the container.

## Reproduction

From a clean latest `tile-ai/TileOPs` clone, run the GEMM FP8 benchmark cases around the nightly crash point. The relevant range is the 10 cases before the crash plus the crash case:

```text
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[ds-v3-decode-down-per-tensor-float8_e4m3fn]
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[ds-v3-prefill-gate-up-per-tensor-float8_e4m3fn]
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[ds-v3-prefill-down-per-tensor-float8_e4m3fn]
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[ds-v3-decode-gate-up-block128-float8_e4m3fn]
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[ds-v3-decode-down-block128-float8_e4m3fn]
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[ds-v3-prefill-gate-up-block128-float8_e4m3fn]
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[ds-v3-prefill-down-block128-float8_e4m3fn]
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[ds-v3-prefill-attn-proj-block128-float8_e4m3fn]
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[k-dominant-7168x16384-block128-float8_e4m3fn]
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[wide-n-24576-block128-float8_e4m3fn]
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[small-batch-down-m8-per-tensor-float8_e4m3fn]
```

Observed result:

```text
... first 10 cases PASSED
benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[small-batch-down-m8-per-tensor-float8_e4m3fn]
!!!!!!! Segfault encountered !!!!!!!
...
Segmentation fault (core dumped)
exit=139
```

## What We Know

- Calling the final `small-batch-down-m8-per-tensor` FlashInfer path by itself does not segfault. It raises a catchable `RuntimeError` / `InternalError`.
- Running FlashInfer-only over the same FP8 GEMM 953-963 shape sequence does not segfault. Per-tensor cases raise catchable errors; block128 cases complete.
- Running TileOps + torch baselines for the prefix, then calling the final small-batch FlashInfer path separately, does not segfault.
- A script that mirrors the benchmark order without `BenchmarkReport.record(...)` also did not segfault in our local check.
- A script that mirrors the benchmark order and includes `BenchmarkReport.record(...)` reproduced the native segfault at the final small-batch FlashInfer per-tensor baseline.

## Unknown

Root cause is not known yet.

We have identified a reproducible trigger, but not the underlying memory/lifetime issue. This may be a FlashInfer/TRTLLM low-latency GEMM module cache issue, a benchmark harness interaction, a lifetime/state interaction caused by recorded benchmark objects, or some other native memory corruption that is only exposed by the benchmark recording path.

## Proposed Nightly Unblock

Disable only the optional FlashInfer per-tensor FP8 GEMM baseline in `benchmarks/ops/bench_gemm.py`.

This keeps benchmark coverage for:

```text
tileops
torch-scaled-mm
flashinfer-fp8-blockscale-sm90
```

and skips only:

```text
flashinfer-mm-fp8
```

for `scale_mode == "per_tensor"`.

This should unblock nightly without removing TileOps benchmark coverage or the torch reference baseline. The skipped FlashInfer per-tensor baseline is already unusable in this runner environment because it raises an error even when run alone; under benchmark state it can escalate to a native segfault.

## Longer-Term Follow-Up

Build a smaller reproducer and investigate with native tooling, e.g. cuda-memcheck / ASAN / debug symbols, to determine whether the underlying issue is in FlashInfer low-latency GEMM module caching, CUDA module lookup, benchmark object lifetime, or a TileOps/benchmark interaction.

zhen8838:
## Summary

- Skip the optional FlashInfer FP8 per-tensor GEMM baseline before it enters the TRTLLM low-latency GEMM path.
- Keep the TileOps and torch-scaled-mm baselines, and keep the FlashInfer block128 baseline enabled.
- Remove the now-unused per-tensor FlashInfer preparation helper.

## Why

FlashInfer 0.6.11.post2 can retain invalid thread-local TRTLLM GEMM module-cache state after an unsupported per-tensor shape fails. A later benchmark case can then segfault in `cuModuleGetFunction` instead of returning a catchable error. Skipping this optional baseline prevents the process from entering the failing path while preserving the other benchmark coverage.

Fixes #1733

## Validation

- `ruff check benchmarks/ops/bench_gemm.py`
- `python -m compileall -q benchmarks/ops/bench_gemm.py`
- `git diff --check`
- `top-exec 'CUDA_VISIBLE_DEVICES=0 pytest benchmarks/ops/bench_gemm.py -vv -s -k "ds-v3-prefill-gate-up-per-tensor-float8_e4m3fn or small-batch-down-m8-per-tensor-float8_e4m3fn"'` (2 passed, no segfault)

### DIFF
--- a/benchmarks/ops/bench_gemm.py
+++ b/benchmarks/ops/bench_gemm.py
@@ -10,6 +10,7 @@ from tileops.ops import GemmFp8Op, GemmOp
 
 _OP_NAME = "GemmOp"
 _FP8_OP_NAME = "GemmFp8Op"
+_FLASHINFER_FP8_PER_TENSOR_SKIP_SHAPES = {(4096, 2112, 7168)}
 
 _DTYPE_MAP = {
     "bfloat16": torch.bfloat16,
@@ -64,6 +65,28 @@ def _flashinfer_fp8_blockscale_ref(test: GemmFp8Test, *inputs: torch.Tensor) -> 
             f"got {tuple(scale_a.shape)} and {tuple(scale_b.shape)}"
         )
     return fp8_blockscale_gemm_sm90(a, b, scale_a, scale_b, out_dtype=test.out_dtype)
+
+
+def _prepare_flashinfer_fp8_per_tensor(
+    test: GemmFp8Test, *inputs: torch.Tensor
+) -> tuple[torch.Tensor, torch.Tensor]:
+    import flashinfer
+
+    a, b, scale_a, scale_b = inputs[:4]
+    if len(inputs) == 5:
+        raise ValueError("FlashInfer FP8 per-tensor GEMM baseline does not support bias.")
+    if a.dtype != torch.float8_e4m3fn or b.dtype != torch.float8_e4m3fn:
+        raise ValueError("FlashInfer FP8 per-tensor GEMM baseline requires float8_e4m3fn.")
+    if test.out_dtype != torch.bfloat16:
+        raise ValueError("FlashInfer FP8 per-tensor GEMM baseline requires bfloat16 output.")
+    if scale_a.shape != (1, 1) or scale_b.shape != (1, 1):
+        raise ValueError(
+            "FlashInfer FP8 per-tensor GEMM baseline requires (1, 1) scales, "
+            f"got {tuple(scale_a.shape)} and {tuple(scale_b.shape)}"
+        )
+    prepared_b = flashinfer.prepare_low_latency_gemm_weights(b, {})
+    alpha = (scale_a * scale_b).reshape(())
+    return prepared_b, alpha
 
 
 class GemmFp8Benchmark(BenchmarkBase[GemmFp8Test]):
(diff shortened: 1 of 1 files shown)


superAngGao (approved):
LGTM. The sm100+ gate for the optional FlashInfer per-tensor FP8 baseline makes sense given that the low-latency path is not supported on Hopper, and it avoids entering a failing native path that can poison later benchmark cases. TileOps, torch-scaled-mm, and the FlashInfer block128 baseline remain covered.

lcy-seso (approved):

PR #1737, opened 2026-07-23
[Perf] Remove host padding from LayerNorm kernels

michaelwithu:
## Summary

- Move non-aligned padding and boundary handling from the PyTorch Op layer into LayerNorm kernels.
- Return natural-shape outputs without host `F.pad` or slicing.
- Add predicated `cp.async` prefetch for AdaLayerNorm modulation tensors.
- Preserve the aligned vectorized path and support multi-row tail tiles.

## Performance

H200 CUPTI: 4.3–6.3x faster on the reported LayerNorm, AdaLayerNorm, and AdaLayerNormZero workloads.

## Validation

- `74 passed`; Compute Sanitizer reports 0 errors.
- Test nodes: `56 -> 74`, covering natural-shape I/O, async policy boundaries, and multi-row tails.

### DIFF
--- a/tests/ops/test_ada_layer_norm.py
+++ b/tests/ops/test_ada_layer_norm.py
@@ -3,7 +3,10 @@ import torch
 import torch.nn.functional as F
 
 from tests.test_base import FixtureBase, TestBase
-from tileops.kernels.norm.ada_layer_norm import AdaLayerNormKernel
+from tileops.kernels.norm.ada_layer_norm import (
+    AdaLayerNormKernel,
+    _should_use_cp_async,
+)
 from tileops.ops.norm.ada_layer_norm import AdaLayerNormFwdOp
 from workloads.normalization import AdaLayerNormTest as _AdaLayerNormTestWorkload
 
@@ -94,6 +97,67 @@ def test_ada_layer_norm_async_copy_handles_row_tail(block_m: int) -> None:
         has_gate=False,
         config={"block_m": block_m, "threads": 128},
     )
+    actual = kernel(*inputs)
+    expected = test.ref_program(*inputs)
+    atol, rtol = _get_tolerances(dtype)
+    torch.testing.assert_close(actual, expected, atol=atol, rtol=rtol)
+
+
+@pytest.mark.smoke
+def test_ada_layer_norm_async_policy_edges() -> None:
+    cases = [
+        (511, torch.float16, False),
+        (512, torch.float16, False),
+        (513, torch.float16, False),
+        (514, torch.float16, True),
+        (1918, torch.float16, True),
+        (1919, torch.float16, False),
+        (1920, torch.float16, True),
+        (513, torch.float32, True),
+        (1919, torch.float32, True),
+    ]
+    for n, dtype, expected_async in cases:
+        assert (
+            _should_use_cp_async(n, dtype, has_gate=False)
+            is expected_async
+        )
+
+
+@pytest.mark.smoke
+def test_ada_layer_norm_async_policy_shared_memory_limit() -> None:
+    cases = [
+        (8190, torch.float16, False, True),
+        (8194, torch.float16, False, False),
+        (6142, torch.float16, True, True),
+        (6146, torch.float16, True, False),
+        (4094, torch.float32, False, True),
+        (4098, torch.float32, False, False),
+    ]
+    for n, dtype, has_gate, expected_async in cases:
+        assert (
+            _should_use_cp_async(n, dtype, has_gate)
+            is expected_async
+        )
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize(
+    "n, dtype",
+    [
+        pytest.param(511, torch.float16, id="fp16-below"),
+        pytest.param(514, torch.float16, id="fp16-lower-inside"),
+        pytest.param(1919, torch.float16, id="fp16-row-bytes-outside"),
+        pytest.param(1920, torch.float16, id="fp16-former-upper-bound"),
+        pytest.param(513, torch.float32, id="fp32-lower-inside"),
+    ],
+)
+def test_ada_layer_norm_async_policy_edge_correctness(
+    n: int, dtype: torch.dtype,
+) -> None:
+    m = 4
+    test = AdaLayerNormTest(m, n, dtype)
+    inputs = test.gen_inputs()
+    kernel = AdaLayerNormKernel(m, n, test.eps, dtype, has_gate=False)
     actual = kernel(*inputs)
     expected = test.ref_program(*inputs)
     atol, rtol = _get_tolerances(dtype)
--- a/tests/ops/test_ada_layer_norm.py
+++ b/tests/ops/test_ada_layer_norm.py
@@ -3,6 +3,7 @@ import torch
 import torch.nn.functional as F
 
 from tests.test_base import FixtureBase, TestBase
+from tileops.kernels.norm.ada_layer_norm import AdaLayerNormKernel
 from tileops.ops.norm.ada_layer_norm import AdaLayerNormFwdOp
 from workloads.normalization import AdaLayerNormTest as _AdaLayerNormTestWorkload
 
@@ -59,6 +60,22 @@ def test_ada_layer_norm_op(m: int, n: int, dtype: torch.dtype) -> None:
     op = AdaLayerNormFwdOp(dtype=dtype)
     atol, rtol = _get_tolerances(dtype)
     test.check(op, *test.gen_inputs(), atol=atol, rtol=rtol)
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16, torch.float32])
+def test_ada_layer_norm_kernel_handles_natural_unaligned_shape(
+    dtype: torch.dtype,
+) -> None:
+    m, n = 16, 1152
+    test = AdaLayerNormTest(m, n, dtype)
+    inputs = test.gen_inputs()
+    kernel = AdaLayerNormKernel(m, n, test.eps, dtype, has_gate=False)
+    actual = kernel(*inputs)
+    expected = test.ref_program(*inputs)
+    assert actual.shape == (m, n)
+    atol, rtol = _get_tolerances(dtype)
+    torch.testing.assert_close(actual, expected, atol=atol, rtol=rtol)
 
 
 class AdaLayerNorm3DFixture(FixtureBase):
--- a/tests/ops/test_ada_layer_norm.py
+++ b/tests/ops/test_ada_layer_norm.py
@@ -83,9 +83,12 @@ def test_ada_layer_norm_kernel_handles_natural_unaligned_shape(
 
 @pytest.mark.smoke
 @pytest.mark.parametrize("block_m", [2, 4])
-def test_ada_layer_norm_async_copy_handles_row_tail(block_m: int) -> None:
+@pytest.mark.parametrize("n", [514, 1152])
+def test_ada_layer_norm_async_copy_handles_row_tail(
+    block_m: int, n: int,
+) -> None:
     """Regression: the async 2-D tile must support block_m > 1 and tail rows."""
-    m, n = 17, 1152
+    m = 17
     dtype = torch.float16
     test = AdaLayerNormTest(m, n, dtype)
     inputs = test.gen_inputs()
@@ -97,6 +100,7 @@ def test_ada_layer_norm_async_copy_handles_row_tail(block_m: int) -> None:
         has_gate=False,
         config={"block_m": block_m, "threads": 128},
     )
+    assert kernel.use_cp_async
     actual = kernel(*inputs)
     expected = test.ref_program(*inputs)
     atol, rtol = _get_tolerances(dtype)
@@ -158,6 +162,8 @@ def test_ada_layer_norm_async_policy_edge_correctness(
     test = AdaLayerNormTest(m, n, dtype)
     inputs = test.gen_inputs()
     kernel = AdaLayerNormKernel(m, n, test.eps, dtype, has_gate=False)
+    if n == 514:
+        assert kernel.use_cp_async
     actual = kernel(*inputs)
     expected = test.ref_program(*inputs)
     atol, rtol = _get_tolerances(dtype)
--- a/tests/ops/test_ada_layer_norm.py
+++ b/tests/ops/test_ada_layer_norm.py
@@ -12,7 +12,9 @@ from workloads.normalization import AdaLayerNormTest as _AdaLayerNormTestWorkloa
 
 
 class AdaLayerNormTest(_AdaLayerNormTestWorkload, TestBase):
-    def ref_program(self, x: torch.Tensor, scale: torch.Tensor, shift: torch.Tensor) -> torch.Tensor:
+    def ref_program(
+        self, x: torch.Tensor, scale: torch.Tensor, shift: torch.Tensor
+    ) -> torch.Tensor:
         # AdaLN: y = scale * LayerNorm(x) + shift
         normed = F.layer_norm(
             x.float(),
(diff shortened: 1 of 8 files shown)


tileops/kernels/norm/ada_layer_norm.py
```diff
@@ -92,37 +119,53 @@ def main(
...
+                            )
+                        T.copy(x_local, shared_buf)
+                        T.copy(shared_buf, y[pid_m * block_m, 0])
 
             return main
 
         else:
 
             @T.prim_func
             def main_gated(
```
Ibuki-wind: - `main` / `main_gated` are ~107 near-identical lines, differing only by `gate`.
- Each copy spells out all four `needs_pad` x `use_cp_async` combinations — every branch written twice.
- Use `@T.macro` for load+cast, the two-pass reduction, and the modulation load. `layer_norm.py` holds a
  third copy of that reduction.

tileops/kernels/norm/ada_layer_norm.py
```diff
@@ -273,8 +387,16 @@ def __init__(
         self.dtype = dtype
         self.has_gate = has_gate
         self.N_padded = _align_up(N, ALIGNMENT)
+        # Async modulation prefetch pays off for medium unaligned rows. Wider
+        # rows stay on the direct-load path to avoid shared-memory pressure.
+        self.use_cp_async = self.N_padded != N and 512 <= N < 1920
```
Ibuki-wind: - `1920` rests on one benchmarked point (N=1152).
- The shared-memory justification doesn't hold: footprint is `block_m * N_padded * sizeof(dtype) *
  num_buffers`, and this predicate sees neither `dtype` nor `block_m` (the latter comes from
  `select_row_config` later).
- `use_cp_async` is baked into `__init__` and the cache key, so autotune can't explore it — even though it
  changes `num_buffers`.

tileops/kernels/norm/layer_norm.py
```diff
@@ -51,13 +54,22 @@ def main(
...
                 rstd = T.alloc_fragment((block_m,), "float32")
 
-                # Load input row block via shared memory
-                T.copy(x[pid_m * block_m, 0], shared_buf)
-                T.copy(shared_buf, x_local)
-
-                # Cast to fp32 once — reused across all passes
-                for i, j in T.Parallel(block_m, N_padded):
-                    x_f32[i, j] = T.cast(x_local[i, j], "float32")
+                if needs_pad:
```
Ibuki-wind, line 57: - LayerNorm's non-aligned path changed (vectorized `T.copy` → predicated element-wise), but the table only
  reports the Ada kernels.
- My measurement (H200, `M=1024` fp16, end-to-end `op.forward`, not CUPTI):

  | N | `main` | PR |
  |---:|---:|---:|
  | 1152 | 133.0 us | 45.8 us |
  | 3000 | 181.5 us | 43.7 us |

- Direction is good — please add CUPTI numbers so it's on record as not a regression.

tileops/kernels/norm/ada_layer_norm.py
```diff
@@ -66,14 +67,40 @@ def main(
                     rstd = T.alloc_fragment((block_m,), "float32")
                     scale_local = T.alloc_fragment((block_m, N_padded), dtype)
```
Ibuki-wind: - `scale_local` / `shift_local` / `gate_local` are dead when `use_cp_async`; `x_local` when `needs_pad`.
- Guard them like `scale_shared` below.

tileops/kernels/norm/ada_layer_norm.py
```diff
@@ -37,8 +37,9 @@ def _align_up(n: int, alignment: int) -> int:
 
 
 @functools.lru_cache(maxsize=32)
-def _ada_layer_norm_kernel(M, N, eps, dtype, has_gate=False):
+def _ada_layer_norm_kernel(M, N, eps, dtype, has_gate=False, use_cp_async=False):
     N_padded = _align_up(N, ALIGNMENT)
+    needs_pad = N_padded != N
```
Ibuki-wind, line 58: - L13-16 still describes the old contract — no mention that boundary handling moved into the kernel.
- `layer_norm.py`'s header was updated; this one wasn't.

tests/ops/test_ada_layer_norm.py
```diff
@@ -61,6 +62,44 @@ def test_ada_layer_norm_op(m: int, n: int, dtype: torch.dtype) -> None:
     test.check(op, *test.gen_inputs(), atol=atol, rtol=rtol)
 
 
+@pytest.mark.smoke
+@pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16, torch.float32])
+def test_ada_layer_norm_kernel_handles_natural_unaligned_shape(
+    dtype: torch.dtype,
+) -> None:
+    m, n = 16, 1152
```
Ibuki-wind, line 78: - Async is only tested at `N=1152`; the window is `[512, 1920)` — add cases at both edges and just outside.
- L85: `select_row_config` pins `block_m=1`, so multi-row async is unreachable by default. Does autotune ever
  pick `block_m>1` here?

Ibuki-wind (commented):
**Verified:** the `T.async_copy` regions slice `0:N_padded` on a tensor declared `(M, N)`, but `compute-sanitizer --tool memcheck` reports 0 errors at `block_m=1` and at `M=17, block_m=4` — TileLang predicates both dimensions. The padding correction is also right.

Dropping `F.pad` from three Op layers is a clean win, as is sizing the autotune smem cap by `num_buffers`. No correctness objections below.

tileops/kernels/norm/ada_layer_norm.py
michaelwithu: Refactored the duplicated load/cast, centered two-pass reduction, async prefetch, and aligned modulation loads into local `T.macro`s. `main` and `main_gated` remain separate because their signatures and output formulas differ.

tileops/kernels/norm/ada_layer_norm.py
michaelwithu: Replaced the fixed `[512, 1920)` heuristic. The policy now checks the actual lowering/resource constraints: non-aligned N, 4-byte aligned source rows (dtype-aware), and all 3/4 shared buffers fitting the 48 KB budget. An H200 sync/async CUPTI sweep sampled eligible FP16/BF16 shapes from N=2 through N=3000; async improved every sampled case by about 1-26%. `use_cp_async` remains a shape/dtype strategy selected before block/thread autotuning; the smem-aware config cap is retained.

tileops/kernels/norm/layer_norm.py
michaelwithu: Added H200 CUPTI measurements to the PR description. LayerNorm FP16 at M=1024: N=1152 is 18.76 -> 4.38 us (4.28x), and N=3000 is 25.56 -> 6.71 us (3.81x).

tileops/kernels/norm/ada_layer_norm.py
michaelwithu: Guarded the allocations by compile-time path: aligned-only fragments are allocated only for the aligned path, while modulation shared buffers are allocated only for the async path.

tileops/kernels/norm/ada_layer_norm.py
michaelwithu: Updated the module header to describe natural-shape inputs/outputs, in-kernel boundary zero-fill/stores, padded variance correction, and the aligned vectorized path.

tests/ops/test_ada_layer_norm.py
michaelwithu: Added policy tests at the former bounds, row-byte alignment boundaries, and shared-memory limits, plus kernel correctness cases on both sync and async sides. The tests exposed that FP16/BF16 odd N cannot lower to cp.async, which is now excluded. For N=1152 FP16, tune=True evaluated block_m=1/2/4 and selected block_m=1; block_m>1 remains a valid candidate and is covered for multi-row/tail correctness.

zhen8838 (changes requested):
Blocking issue:

- `tileops/kernels/norm/ada_layer_norm.py:108-115` still issues `T.async_copy` for `src[..., 0:N_padded]` even though `src` is declared with natural shape `(M, N)`. The new `_should_use_cp_async()` policy at `:42-54` only checks `row_bytes % 4` and shared-memory size; it does not make the padded source extent valid. For the newly accepted `N=514` FP16 case, the async path requests 768 columns from a 514-column source, so a transfer-level predicate can still read beyond the natural row. Apply an OOB-safe load/zero-fill strategy or constrain the async path to transfer-safe widths, and retain a non-aligned regression that exercises the actual async kernel under a sanitizer or equivalent memory-safety check.

michaelwithu:
Verified the newly admitted minimal-alignment case directly under Compute Sanitizer and pushed regression coverage in `14460fe`.

The retained tests now exercise `N=514` FP16 on the actual async path (`assert kernel.use_cp_async`) for both AdaLayerNorm and AdaLayerNormZero, including `M=17, block_m=2/4` to cover multi-row copies and the final row tail. The existing `N=1152` coverage is retained.

Sanitizer command:

```bash
compute-sanitizer --tool memcheck --error-exitcode=99 \
  python -m pytest -q \
  'tests/ops/test_ada_layer_norm.py::test_ada_layer_norm_async_copy_handles_row_tail[514-4]' \
  'tests/ops/test_ada_layer_norm_zero.py::test_ada_layer_norm_zero_async_copy_handles_row_tail[514-4]'
```

Result:

```text
2 passed in 5.16s
========= ERROR SUMMARY: 0 errors
```

I also ran the default `block_m=1` `N=514` FP16 async correctness case independently under memcheck:

```text
1 passed in 13.67s
========= ERROR SUMMARY: 0 errors
```

This confirms that, with the TileLang build used by the project (`0.1.11+cuda.git65dbc983`), safe-memory legalization predicates the padded source region and the predicated-off `cp.async` transactions do not produce OOB reads.

tests/ops/test_ada_layer_norm.py
```diff
@@ -61,6 +65,111 @@ def test_ada_layer_norm_op(m: int, n: int, dtype: torch.dtype) -> None:
...
+    kernel = AdaLayerNormKernel(m, n, test.eps, dtype, has_gate=False)
+    actual = kernel(*inputs)
+    expected = test.ref_program(*inputs)
+    assert actual.shape == (m, n)
+    atol, rtol = _get_tolerances(dtype)
+    torch.testing.assert_close(actual, expected, atol=atol, rtol=rtol)
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize("block_m", [2, 4])
```
Ibuki-wind: This `block_m × n` Cartesian product runs four expensive smoke cases through the same async implementation; `block_m=4, n=514` already covers the widest multi-row tail and the newly admitted minimal padded transfer, while `n=1152` is covered by the natural-shape test above. → Keep one row-tail regression instead of all four combinations.

tests/ops/test_ada_layer_norm.py
```diff
@@ -61,6 +65,111 @@ def test_ada_layer_norm_op(m: int, n: int, dtype: torch.dtype) -> None:
...
+            _should_use_cp_async(n, dtype, has_gate)
+            is expected_async
+        )
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize(
+    "n, dtype",
+    [
+        pytest.param(511, torch.float16, id="fp16-below"),
```
Ibuki-wind, line 146: These five GPU cases duplicate the predicate truth table immediately above: `511`/`1919` execute the same synchronous padded branch, while `514`/`1920`/FP32 `513` execute the same async padded branch. → Keep one representative synchronous case and one representative async case; leave exact policy boundaries to the cheap `_should_use_cp_async` unit table.

tests/ops/test_ada_layer_norm_zero.py
```diff
@@ -63,6 +64,48 @@ def test_ada_layer_norm_zero_op(m: int, n: int, dtype: torch.dtype) -> None:
...
+    kernel = AdaLayerNormKernel(m, n, test.eps, dtype, has_gate=True)
+    actual = kernel(*inputs)
+    expected = test.ref_program(*inputs)
+    assert actual.shape == (m, n)
+    atol, rtol = _get_tolerances(dtype)
+    torch.testing.assert_close(actual, expected, atol=atol, rtol=rtol)
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize("block_m", [2, 4])
```
Ibuki-wind: The gated variant repeats the same four-case `block_m × n` async matrix even though neither axis changes the gated code path. → Retain one maximal row-tail case (for example `block_m=4, n=514`) and remove the three same-failure-mode duplicates.

Ibuki-wind (changes requested):
### Overall
The kernel change is correct and the sanitizer evidence closes the memory-safety concern; shrink the duplicated GPU matrix before merge.

tileops/kernels/norm/ada_layer_norm.py
Ibuki-wind: The extracted leaf macros are useful, but they do not resolve the issue in this thread: `main` and `main_gated` still each spell out the complete allocation plus `needs_pad × use_cp_async` control tree. The extra ABI argument and final gate multiply justify two thin `prim_func` signatures, not two kernel bodies. Please put the common body in one `T.macro` (or a Python factory that emits it once), pass `gate`/`_dummy`, and use the closure-time `has_gate` boolean only to inject the gate allocation/prefetch/load/multiply. That remains compile-time specialization—no runtime `has_gate` branch—and leaves each `prim_func` as a signature plus one body call. Acceptance criterion: changes to padding, async synchronization, reduction, or stores are made in one place.

Ibuki-wind (changes requested):
### Overall

The sanitizer evidence closes the memory-safety question, but the gated/ungated body duplication in the existing thread remains blocking. Keep two ABI signatures and generate one shared kernel body; this is compile-time consolidation, not a runtime mode branch.

Ibuki-wind (dismissed):
No functional blockers; the remaining source/test deduplication comments are non-blocking.

tileops/kernels/norm/ada_layer_norm.py
michaelwithu: Addressed the follow-up in `5a3b7b5`. `main` and `main_gated` now retain only their distinct ABI signatures and each invokes one shared `kernel_body` `T.macro`. The closure-time `has_gate` specialization controls only the gate allocation/prefetch/load/multiply, so padding, async synchronization, reduction, and stores are maintained in one place with no runtime mode branch. Verified gated and ungated aligned, synchronous-padded, and async-padded paths on H200; the retained async tail regressions also pass Compute Sanitizer with 0 errors.

tests/ops/test_ada_layer_norm.py
michaelwithu: Addressed in `5a3b7b5`: collapsed the `block_m × n` matrix to the single maximal representative `M=17, block_m=4, n=514`. The natural-shape test continues to cover `n=1152`.

tests/ops/test_ada_layer_norm.py
michaelwithu: Addressed in `5a3b7b5`: the GPU correctness coverage now keeps one synchronous padded case (`n=511`) and one async padded case (`n=514`). Exact row-alignment and shared-memory boundaries remain in the lightweight `_should_use_cp_async` policy tables.

tests/ops/test_ada_layer_norm_zero.py
michaelwithu: Addressed in `5a3b7b5`: the gated async matrix is reduced to the single maximal row-tail case `M=17, block_m=4, n=514`. It passes correctness testing and Compute Sanitizer with 0 errors.

lcy-seso:
@zhen8838 This PR needs your approval. Could you please look into the modifications.

zhen8838 (changes requested):
Request changes on current head 5a3b7b5.

The AdaLN/AdaLN-Zero async path still calls `T.async_copy(src[..., 0:N_padded], dst)` while the external modulation source has natural shape `(M, N)`. TileLang's `T.async_copy(src, dst)` is global-to-shared: this reads the external source into the larger shared destination; it is not a shared-to-global copy. Padding the shared destination does not extend the source allocation. The transfer therefore needs an explicit source-tail-safe implementation (or a proven lowering-level mask), or the async path must be restricted to widths safe for the actual transfer. The `N=514` correctness tests exercise the path but do not establish that the transfer itself is OOB-safe.

Please fix the source-tail handling and retain the non-aligned AdaLN and AdaLN-Zero regression coverage before approval.

michaelwithu:
Addressed the async source-tail concern in `93f0034`.

The modulation prefetch now emits explicit 4-byte `T.ptx_cp_async` transactions with a source predicate covering both dimensions: `row < M && col + async_copy_elems <= N`. Predicated-off transactions use `cp.async` zero-fill semantics, so no source address beyond the natural `(M, N)` allocation is dereferenced while the padded shared tail is initialized to zero.

For the retained `M=17, N=514, block_m=4` regression, generated CUDA contains `tl::cp_async_gs_conditional<4>` for scale/shift (and gate in AdaLN-Zero), with the row and vector-end guards present. Both non-aligned async regressions remain:
- `test_ada_layer_norm_async_copy_handles_row_tail`
- `test_ada_layer_norm_zero_async_copy_handles_row_tail`

Validation on H200:
- both retained correctness regressions pass;
- FP16/BF16/FP32 natural-shape async coverage passes;
- Compute Sanitizer memcheck: 2 passed, `ERROR SUMMARY: 0 errors`.

zhen8838 (approved):
Approved on current head 93f0034.

The async modulation prefetch now performs fixed-width 4-byte source transactions with a predicate requiring `col + transfer_width <= N`. Tail transactions are predicated off and zero-fill the shared destination, so the natural `(M, N)` source is not read through `N_padded`. The non-aligned AdaLN and AdaLN-Zero coverage remains in place, and the current GitHub checks including `gpu-smoke` pass.

PR #1738, opened 2026-07-23
[Refactor][MoE] Separate dispatched expert compute

michaelwithu:
## Summary

- Extract communication-independent dispatched-expert MLP compute for tight expert-major batches.
- Compose routed expert execution from local dispatch, expert compute, and combine while preserving the public API.
- Add GPU-resident dynamic row counts, static capacity, and CUDA Graph replay.
- Add local dispatch and an optional DeepEP V2 adapter without introducing a runtime DeepEP dependency.
- Keep routing weights and opaque combine handles outside expert compute.

## Review follow-up

- Resolve one effective grouped-GEMM kernel before constructing gate-up/down sub-ops, so an explicit override cannot undo the alignment fallback.
- Consolidate normal and device-bounded fused-gated paths around one column mapping and one pointwise epilogue while keeping distinct compiled signatures and no runtime row-policy branch.
- Remove exact-signature/dataclass-default tests and the duplicate hand-written mock dispatch/combine test; retain behavioral reference, capacity-tail, graph-replay, and real local-dispatch integration coverage.
- Consolidate benchmark timing, routing setup, expert distributions, FP32 reference checks, and TFLOPS calculation in a 167-line shared helper.
- Reduce the four benchmark entry points plus shared helper from roughly 915 to 778 lines, and remove the redundant pure/full compute benchmark while retaining the dynamic-capacity sweep and TileOps-vs-DeepGEMM comparison.

## Validation

- 24 dispatched-expert and dispatch-adapter smoke tests pass.
- Normal large-row and device-bounded direct/explicit fused-gated tests pass.
- Local dispatch benchmark smoke passes.
- TileOps-vs-DeepGEMM `M=1` smoke passes with FP32 correctness checks.
- DeepEP V2 2-rank fresh and cached-handle dispatch pass on 2× H200 using the non-Gin path.
- Ruff, Ruff format for benchmark files, Python compilation, and `git diff --check` pass.

## Scope

- BF16 tight/no-pad expert compute only; FP8/FP4 and padded-direct compute remain out of scope.
- TileOps does not acquire a runtime DeepEP dependency; callers inject the DeepEP buffer.
- Capacity-tail values remain unspecified; consumers use the GPU-resident `expert_offsets[-1:]` valid-row view.

### DIFF
--- a/tests/ops/test_dispatched_expert.py
+++ b/tests/ops/test_dispatched_expert.py
@@ -1,0 +1,214 @@
+"""Tests for the communication-independent dispatched expert MLP."""
+
+import inspect
+
+import pytest
+import torch
+import torch.nn.functional as F
+
+from tileops.ops.moe import (
+    DispatchedExpertMLPFwdOp,
+    ExpertBatch,
+    ExpertBatchOutput,
+)
+
+
+def _reference(hidden, w_gate_up, w_down, sizes):
+    outputs = []
+    start = 0
+    ffn_size = w_down.shape[-1]
+    for expert, size in enumerate(sizes):
+        rows = hidden[start : start + size].float()
+        gate_up = rows @ w_gate_up[expert].float().t()
+        act = F.silu(gate_up[:, :ffn_size]) * gate_up[:, ffn_size:]
+        outputs.append(act @ w_down[expert].float().t())
+        start += size
+    return torch.cat(outputs).to(hidden.dtype)
+
+
+@pytest.mark.smoke
+def test_public_interface_has_no_routing_or_communication_inputs():
+    parameters = inspect.signature(DispatchedExpertMLPFwdOp.forward).parameters
+    assert tuple(parameters) == (
+        "self",
+        "expert_input",
+        "w_gate_up",
+        "w_down",
+        "true_sizes",
+        "true_offsets",
+    )
+    forbidden = {"topk_ids", "topk_weights", "rank", "world_size", "handle"}
+    assert forbidden.isdisjoint(parameters)
+    batch_parameters = inspect.signature(
+        DispatchedExpertMLPFwdOp.forward_batch
+    ).parameters
+    assert tuple(batch_parameters) == ("self", "batch", "w_gate_up", "w_down")
+    assert forbidden.isdisjoint(batch_parameters)
+
+
+@pytest.mark.smoke
+def test_expert_batch_contract_rejects_non_tight_layout():
+    with pytest.raises(ValueError, match="layout='tight'"):
+        ExpertBatch(
+            hidden=torch.empty(4, 8),
+            expert_offsets=torch.tensor([0, 2, 4], dtype=torch.int32),
+            layout="aligned",
+        )
+
+
+@pytest.mark.smoke
+def test_expert_batch_output_contract_defaults():
+    hidden = torch.empty(4, 8)
+    output = ExpertBatchOutput(hidden=hidden)
+    assert output.hidden is hidden
+    assert output.row_order_preserved
+    assert not output.routing_weights_applied
+
+
+@pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float16])
+@pytest.mark.parametrize("sizes", [[4, 4, 4, 4], [0, 2, 13, 1]])
+@pytest.mark.smoke
+def test_dispatched_expert_matches_reference_and_preserves_rows(dtype, sizes):
+    if not torch.cuda.is_available():
+        pytest.skip("CUDA is required")
+
+    torch.manual_seed(0)
+    num_experts = len(sizes)
+    num_pairs = sum(sizes)
+    hidden_size, ffn_size = 128, 96
+    hidden = torch.randn(
+        num_pairs, hidden_size, device="cuda", dtype=dtype
+    ) * 0.1
+    w_gate_up = torch.randn(
+        num_experts, 2 * ffn_size, hidden_size, device="cuda", dtype=dtype
+    ) * 0.02
+    w_down = torch.randn(
+        num_experts, hidden_size, ffn_size, device="cuda", dtype=dtype
+    ) * 0.02
+    true_sizes = torch.tensor(sizes, device="cuda", dtype=torch.int32)
+    true_offsets = torch.tensor(
+        [sum(sizes[:expert]) for expert in range(num_experts)],
+        device="cuda",
+        dtype=torch.int32,
+    )
+
+    op = DispatchedExpertMLPFwdOp(
+        num_pairs=num_pairs,
+        num_experts=num_experts,
+        hidden_size=hidden_size,
+        ffn_size=ffn_size,
+        dtype=dtype,
+    )
+    output = op(hidden, w_gate_up, w_down, true_sizes, true_offsets)
+    reference = _reference(hidden, w_gate_up, w_down, sizes)
+
+    assert output.shape == hidden.shape
+    assert output.dtype == dtype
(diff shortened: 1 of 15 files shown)


tileops/ops/moe/routed_expert/dispatched_expert.py
```diff
@@ -0,0 +1,237 @@
...
+            )
+            self._activation_op = None
+        else:
+            self._gemm_gate_up = MoeGroupedGemmNopadFwdOp(
+                numel=num_pairs,
+                num_experts=num_experts,
+                n=ffn_size * 2,
+                k=hidden_size,
+                dtype=dtype,
+                kernel_map={"moe_grouped_gemm_kernel": kernel_cls, **(kernel_map or {})},
```
Ibuki-wind: The alignment fallback can be silently undone here and again for `_gemm_down`: after `kernel_cls` is changed to `MoeGroupedGemmNopadKernel`, a caller-supplied `kernel_map['moe_grouped_gemm_kernel']` is expanded last and reinstates the misaligned 3WG kernel that the preceding check rejected. → Resolve one effective GEMM class first and either reject a conflicting override or build the sub-op maps without re-merging that key.

tests/ops/test_dispatched_expert.py
```diff
@@ -0,0 +1,391 @@
...
+        rows = hidden[start : start + size].float()
+        gate_up = rows @ w_gate_up[expert].float().t()
+        act = F.silu(gate_up[:, :ffn_size]) * gate_up[:, ffn_size:]
+        outputs.append(act @ w_down[expert].float().t())
+        start += size
+    return torch.cat(outputs).to(hidden.dtype)
+
+
+@pytest.mark.smoke
+def test_public_interface_has_no_routing_or_communication_inputs():
```
Ibuki-wind: This exact-signature introspection test (together with the adjacent dataclass-default checks) duplicates declarations instead of exercising behavior and will fail harmless parameter/default refactors. → Delete these contract-shape tests and keep the behavioral reference, capacity-tail, and graph-replay tests.

tests/ops/test_dispatched_expert.py
```diff
@@ -0,0 +1,391 @@
...
+    output = op(hidden, w_gate_up, w_down, true_sizes, true_offsets)
+    reference = _reference(hidden, w_gate_up, w_down, sizes)
+
+    assert output.shape == hidden.shape
+    assert output.dtype == dtype
+    assert torch.allclose(output.float(), reference.float(), atol=1e-2, rtol=1e-2)
+
+
+@pytest.mark.smoke
+def test_expert_batch_mock_dispatch_combine_applies_weights_once():
```
Ibuki-wind: This 100-line mock test never calls a dispatch or combine implementation: it manually builds the sorted batch, runs the expert op already covered above, then compares a manual weighted `index_add_` against the same manual reference. → Delete it; the local-dispatch integration test in `test_expert_dispatch.py` covers the real composition with far less duplicate setup.

Ibuki-wind (changes requested):
### Overall
One kernel-selection bug remains, and the draft needs a substantial size reduction before it is reviewable as a maintainable refactor.

### Cross-cutting concerns
- The four new benchmark scripts repeat timing loops, tensor/model setup, output formatting, and correctness machinery across roughly 900 lines; consolidate the common harness and keep each entry point to scenario definitions plus backend adapters.

tileops/kernels/elementwise.py
```diff
@@ -628,12 +639,91 @@ def _make_fused_gated_explicit(M, N, dtype, op_func, threads=256, num_per_thread
...
+                        value = x[row, N + col]
+                        y[row, col] = op_func(gate, value)
+
+        return main
+
+    return kernel
+
+
+@functools.lru_cache(maxsize=32)
+def _make_fused_gated_direct_bounded(
```
Ibuki-wind: These two bounded factories add another direct/explicit pair that repeats the same column mapping and `gate/value/op_func/store` template already owned by `_make_fused_gated_direct` and `_make_fused_gated_explicit`. The device-side row loop/guard is the real specialization; the pointwise strategy is not new. → Keep distinct compiled signatures (`valid_rows` must stay device-side), but use one source generator per strategy with a compile-time row policy: the normal policy emits the existing 2-D/flattened row mapping, and the bounded policy emits `grid_rows`, the persistent row loop, and `row < valid_rows[0]`. Alternatively, make one bounded generator parameterized by static `num_per_thread` (`1` for direct). Do not add a runtime bounded/unbounded branch. Acceptance criterion: adding or changing the fused-gated epilogue or column mapping requires editing one source implementation, not four factories.

Ibuki-wind (changes requested):
### Overall

The bounded-row kernel is a justified compiled specialization because `valid_rows` must remain device-side. Its source generation should be consolidated with the existing FusedGated strategies.

RuneFang:
Does it now support the Megakernel MoE ?

lcy-seso:
I rebased this branch onto `main` (`fb95f0a`) and force-pushed it, so the old head `a1c9aff` is gone. All five commits are preserved with their original authorship; the last one carries a few extra fixes described below. The branch was 84 commits behind and some of the APIs it builds on changed, so please review what landed.

**Conflicts and how I resolved them**

- `fused_routed_expert.py` — took your refactor (delegate to `DispatchedExpertMLPFwdOp`). But `main` meanwhile changed the fused-activation path from "warn and silently downgrade" to "raise, one condition at a time". I moved `main`'s validation into `DispatchedExpertMLPFwdOp`, keeping your `effective_gemm_kernel_map` fix from the review commit (the caller's override must not be re-merged after an alignment fallback).
- `main` removed the `dtype` constructor argument from `MoeGroupedGemmNopadFwdOp`, `MoeGroupedGemmNopad3WGFusedActFwdOp` and `build_activation_op` — they bind dtype from the runtime tensors now. I dropped every `dtype=` forwarded to those. `DispatchedExpertMLPFwdOp` keeps its own `dtype` parameter, since the tests pass it.
- `elementwise.py` — `BinaryKernel` and its subclasses took `main`'s current signature (`a_shape, b_shape, dtype, config=, tune=`); the branch still had the older `N_total, coalesced_shape, a_strides, ...` form. Your `_make_fused_gated_bounded` consolidation and `forward_rows` are unchanged.
- `routed_expert/__init__.py` — `__all__` conflict from sorting only; `DeepEPDispatchAdapter` is in.

**Three fixes I had to make on top — these are changes to your code**

1. `test_silu_and_mul_supports_more_than_65535_rows` called `SiluAndMulFwdOp(M=m, N=n, dtype=dtype, strategy=strategy)`. `main`'s op layer has neither `dtype` nor `strategy`, so I dropped the `strategy` parametrization and call `SiluAndMulFwdOp(M=m, N=n)`. If you want both strategies covered at the op layer, that needs a separate change.
2. `test_silu_and_mul_device_bounded_rows` passed `strategy=` to `SiluAndMulFwdKernel`; on `main` the kernel takes it as `config={"strategy": ...}`.
3. The four MoE bench files used `from _moe_bench_utils import ...`. `benchmarks/ops/` has an `__init__.py`, so the bare import fails at collection — changed to `from benchmarks.ops._moe_bench_utils import ...`.

**Verification**

`ruff` clean across `src/`, `tests/`, `benchmarks/`, `scripts/`. All 57 tests in the affected suites collect. I did **not** run the GPU tests — CI is the real check. `bench_dispatched_expert_deepgemm.py` still imports `deep_gemm` at module scope, so it fails collection wherever that package is absent; unchanged from before, but worth guarding if CI does not have it.

PR #1739, opened 2026-07-24
[Feat][GEMM] Add W4A16 GEMM op and decode benchmarks

superAngGao:
## Summary

- Add `GemmW4A16Op` for FP16 activation/output GEMM with affine UINT4 packed weights and group-128 scale/zero metadata.
- Add a TileLang W4A16 kernel that depacks/dequantizes W4 tiles and computes FP16-output Tensor Core GEMM.
- Add manifest, workload, roofline, correctness tests, error-path tests, and decode-oriented benchmark coverage.
- Add dense dequantized PyTorch and optional vLLM Marlin baselines for W4A16 benchmark reporting.

## Validation

Validated in the TileOpsGov runner image `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10` without installing or upgrading dependencies.

- `pytest -q tests/ops/test_gemm.py -m "smoke or full" --tb=short` -> 38 passed.
- `python3 scripts/validate_manifest.py --check-op GemmW4A16Op --verbose` passed with advisory input-precondition warnings only.
- `pytest -q benchmarks/tests --tb=short` -> 17 passed.
- `pytest -q benchmarks/ops/bench_gemm.py -k w4a16 --tb=short` -> 6 passed.


superAngGao:
Resolved the benchmark-infrastructure blocker by rebasing onto current `main` and resolving `bench_gemm.py` against the ManifestBenchmark architecture.

Changes after rebase:

- `GemmW4A16Op` now uses `ManifestBenchmark(_W4A16_OP_NAME, op, test)` in `bench_gemm.py`.
- Did not restore `GemmBenchmark`, `GemmFp8Benchmark`, or `GemmW4A16Benchmark`.
- Kept `StaticGemmW4A16Benchmark` only for dense/Marlin baselines with specialized memory accounting.
- The merge conflict with #1764 is gone.

Validation in `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`:

- W4A16 correctness/error tests: 5 passed.
- Manifest-driven W4A16 benchmark: 2 passed.
- Manifest and benchmark contract tests: 24 passed.
- `scripts/validate_manifest.py --check-op GemmW4A16Op --verbose`: passed (advisory input-precondition warnings only).
- Ruff check and format check passed after resolving the file.

workloads/gemm.py
```diff
@@ -52,16 +59,80 @@ def _scale_shapes(self) -> tuple[tuple[int, int], tuple[int, int]]:
...
+    n, k = weight.shape
+    if k % group_size != 0:
+        raise ValueError(f"K must be divisible by group_size={group_size}, got {k}")
+    if k % 2 != 0:
+        raise ValueError(f"K must be even for nibble packing, got {k}")
+
+    grouped = weight.float().reshape(n, k // group_size, group_size)
+    group_min = grouped.amin(dim=-1)
+    group_max = grouped.amax(dim=-1)
+    scale = ((group_max - group_min) / 15.0).clamp_min(1e-12)
```
Ibuki-wind, line 93: This affine quantizer fails for one-sided groups because the range excludes zero before the integer zero-point is clamped. For an all-positive group, `zero` clamps to 0 but `scale=(max-min)/15`, so the representable maximum is only `max-min` and most values saturate; the all-negative case is symmetric. → Compute the range from `min(group_min, 0)` and `max(group_max, 0)` (or use a representable floating zero-point), and add positive-only/negative-only regression groups.

tests/ops/test_gemm.py
```diff
@@ -36,143 +35,334 @@ def ref_program(self, *inputs: torch.Tensor) -> torch.Tensor:
...
-            pytest.param(
-                18432, 1, 7168, torch.bfloat16, False, False, False,
-                marks=pytest.mark.full,
-                id="full-bf16-tuned-thin-n-alt",
-            ),
-        ]),
+        (
+            "m, n, k, dtype, trans_a, trans_b, tune",
+            [
+                pytest.param(
```
Ibuki-wind: The existing GEMM and FP8 fixture tables were mechanically rewrapped from compact rows into hundreds of lines even though their behavior is unchanged; this accounts for most of the file's 335 additions and obscures the actual W4 tests. → Revert formatting-only churn in pre-existing tests and keep the diff limited to the W4A16 additions.

benchmarks/ops/bench_gemm_w4a16_feasibility.py
```diff
@@ -0,0 +1,52 @@
...
+)
+from tests.ops.test_gemm import GemmW4A16Test
+from tileops.ops import GemmW4A16Op
+
+
+@pytest.mark.parametrize(
+    "case",
+    [pytest.param(case, id=case.label) for case in FEASIBILITY_CASES],
+)
+def test_gemm_w4a16_feasibility_bench(case: W4A16BenchmarkCase) -> None:
```
Ibuki-wind: This file preserves an already-concluded exploratory experiment (`M=128` was unfavorable and decode was selected), duplicating the maintained W4 benchmark path plus four more shapes and their shared case machinery. → Remove the historical feasibility suite from the product tree; keep only the current manifest benchmark and the active decode comparison cases.

Ibuki-wind (changes requested):
### Overall
The W4 kernel smoke cases pass, but the input generator is wrong for one-sided groups and the PR carries substantial historical/formatting bulk that should not land.

tests/ops/test_gemm.py
superAngGao: Fixed in 606865cc. I restored tests/ops/test_gemm.py back to the upstream/main layout for the pre-existing GEMM and FP8 fixture tables, then re-applied only the W4A16 test additions plus the new quantizer regression. The file diff is now limited to W4A16-related changes.

benchmarks/ops/bench_gemm_w4a16_feasibility.py
superAngGao: Fixed in 606865cc. I removed benchmarks/ops/bench_gemm_w4a16_feasibility.py and dropped the now-unused FEASIBILITY_CASES from the shared W4A16 benchmark utils, leaving the manifest benchmark and the active decode AKO suite.

workloads/gemm.py
superAngGao: Fixed in 606865cc. The affine INT4 helper now builds each group range from min(group_min, 0) to max(group_max, 0), so positive-only groups keep zero=0 and negative-only groups keep zero=15 without clipping the endpoint. I also added a positive-only/negative-only regression in test_quantize_weight_int4_keeps_one_sided_groups_in_range.

superAngGao:
Follow-up: I tightened the W4A16 benchmark layering as well. The W4A16 bench now uses a benchmark-local workload wrapper over workloads.gemm.GemmW4A16Workload and no longer imports GemmW4A16Test or profiles test.ref_program. It remains manifest-driven and decode-oriented.

zhen8838 (changes requested):
Request changes on current head 545548a.

The earlier code-level findings around one-sided affine quantization, the feasibility benchmark, and unrelated test-file formatting have been addressed. The current W4A16 kernel, tail predicates, metadata shapes, nibble packing, and D128 decode workload are statically coherent.

The remaining blockers are review-readiness and validation evidence:

- The PR body is missing the required `## Test plan` section and the kernel/op-specific `## Benchmark` section. It also does not provide the required test-node delta for the added GEMM tests.
- The current `gpu-smoke` run fails in the fork-PR checkout step, and all GPU tests are skipped as a result. Please rerun a valid GPU smoke after the checkout/runner issue is recovered; the checkout failure is infrastructure evidence, not code validation.

Please update the PR body with the required sections and current results, then rerun GPU smoke before requesting approval again.

superAngGao:
Thanks for the review. I updated the PR body with the required `## Test plan` and `## Benchmark` sections, including the GEMM test-node delta.

I also investigated the `gpu-smoke` failure:

- The earlier GitHub run on `545548a` failed before running GPU tests. It hit repeated checkout/trusted-actions checkout network failures (`GnuTLS recv error`, `github.com:443` timeouts, then an incomplete trusted-action blob).
- I then reproduced locally with the TileOpsGov CI runner image exactly as requested, using `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`, mounted source/cache volumes, and no dependency installs/upgrades.
- That surfaced one real collection blocker: `test_quantize_weight_int4_keeps_one_sided_groups_in_range` had no explicit test-tier marker. Fixed in `72c11b6` by marking it `smoke`.

Local official-runner validation after the fix:

- `python3 -m pytest -q tests/ops/test_gemm.py -m "smoke or full" --tb=short`: `38 passed in 122.10s`
- `python3 scripts/validate_manifest.py --check-op GemmW4A16Op --verbose`: passed, advisory input-precondition warnings only
- `python3 -m pytest -q benchmarks/tests --tb=short`: `17 passed`
- `python3 -m pytest -q benchmarks/ops/bench_gemm.py -k w4a16 --tb=short`: `6 passed, 28 deselected`

A new CI run has been triggered from `72c11b6`; non-GPU checks are passing and `gpu-smoke` is currently still pending in the runner queue.

zhen8838 (approved):
Approve current head 72c11b6.

The prior review blockers are resolved: the W4A16 one-sided affine quantization edge case is covered, the benchmark is in the GEMM benchmark path without test-oracle imports, and the PR body now includes the required `## Benchmark`, `## Test plan`, and test-node-delta metadata. The final delta adds the missing `smoke` marker to the quantization regression. Title validation and all required CI checks, including `gpu-smoke`, pass.

Ibuki-wind (approved):

Issue #1744, opened 2026-07-25
[META][CI] run nightly benchmarks one process per file: bound native-failure loss, capture hang forensics


lcy-seso:
### Symptom / Motivation

Nightly 30120597863 (07-24): one benchmark stalled >900s; `--timeout-method=thread` killed the process before junit write — all 1328 completed results lost. 07-19: a FlashInfer segfault killed the suite the same way (#1733/#1735).

Measurement code shares one process with pytest, so process-wide resources keep colliding: tilelang `dup2`s `/dev/null` over pytest's capture fd (the `EBADF` that destroyed the 07-24 stack dump; guarded in #1745), CUPTI is per-process, and a native hang or segfault ends the whole session. Each incident so far got its own point workaround.

### Root Cause Analysis

`--junit-xml` writes only at session end; any early process death discards everything. A Python-level timeout cannot interrupt a blocked native call. Only process-per-file isolation bounds the loss and lets the parent kill (and stack-dump) a hung child.

### Related Files

- .github/workflows/nightly.yml
- benchmarks/benchmark_base.py (`bench_kernel`, `_native_output_suppressor`)

## Goal

One native failure (hang / segfault / OOM) costs one file's results: `bench_results.xml` always contains every completed test, and a hung child leaves a stack dump in the artifacts.

## Plan

1. One pytest process per `benchmarks/ops/bench_*.py`; per-file junit fragments, merged at the end.
2. Parent-side timeout: on expiry `py-spy dump` the child (native stack), SIGKILL, write a synthetic failure entry.
3. Map child exit 139/137 to the same synthetic entry + log tail.
4. Verify wall-clock overhead vs single-process.
5. Upstream the fd check to tilelang's `suppress_stdout_stderr`; then delete `_native_output_suppressor`.

## Constraints

- `bench_kernel` latency methodology, GPU clock validation, and `profile_run.log` reporting unchanged — process structure and result durability only.

## Acceptance Criteria

- [ ] Modified files pass unit tests
- [ ] Injected hang in one file: job still uploads `bench_results.xml` with all other results + stack dump of the hung child in artifacts
- [ ] Injected `os.abort()` in one file: same containment
- [ ] Isolation overhead <10% wall-clock vs current single-process run
- [ ] PR states the removal condition for `_native_output_suppressor` and links the upstream tilelang change

PR #1745, opened 2026-07-25
[Fix][Benchmark] stop losing results on native failures; derive workload keys from signature

### LINKED ISSUE #1740: [BUG][MOE_GROUPED_GEMM] nightly bench hangs >900s on deepseek-v3-decode-down
### Symptom / Motivation

Nightly run [30120597863](https://github.com/tile-ai/TileOPs/actions/runs/30120597863) (2026-07-24, main @ 76f35fdf) failed in the `benchmark` job. The suite stalled after 1328/1652 tests, and after 900s `pytest-timeout` (`--timeout=900 --timeout-method=thread`) fired and the pytest process died without a test summary or `bench_results.xml` (the timeout handler crashed with `OSError: [Errno 9] Bad file descriptor` reading the capture fd).

Mapping the progress-bar character count against `pytest --collect-only -q benchmarks/ops` at the same commit pinpoints the hanging test:

```
benchmarks/ops/bench_moe_grouped_gemm_nopad.py::test_moe_grouped_gemm_nopad_bench[deepseek-v3-decode-down-bfloat16]
```

Workload (from the manifest entry for `MoeGroupedGemmNopadFwdOp`): `numel=4096, num_experts=256, n=7168, k=2048, bfloat16`. The two gate-up cases (`n=4096, k=7168`) immediately before it completed normally; the suite hung on the first `decode-down` case and never reached `prefill-down`.

Timeline evidence: progress line `[ 78%]` flushed at 19:05:38Z (test #1296); tests #1297–#1328 completed in ~154s (32 buffered status chars flushed post-mortem); test #1329 then ran >900s until the 20:23:12Z timeout kill.

This is likely the first nightly to reach this test in days, so it may not be a fresh regression: the 07-19 and 07-21 nightlies both segfaulted (exit 139) earlier in the suite at the FlashInfer FP8 per-tensor baseline (#1733, fixed by #1735), and the 07-22/07-23 nightlies died at checkout due to runner network failures. With those unmasked, the MoE grouped-GEMM hang is now the blocking failure.

### Root Cause Analysis

Not yet isolated. The hang is inside `test_moe_grouped_gemm_nopad_bench[deepseek-v3-decode-down-bfloat16]`, which executes in order: op construction, JIT warmup call, `bm.profile(...)`, then a per-expert PyTorch baseline loop (256 small NT matmuls). Candidate phases:

- TileLang JIT compilation hang for the `n=7168, k=2048` tile configuration (cold-cache path; the nightly cache repopulates on cold keys).
- GPU-side kernel hang (e.g. tile-scheduler loop not terminating for the decode shape where per-expert token count is 16 = 4096/256).
- Profiler/baseline phase hang.

Secondary issue: with `--timeout-method=thread`, a single hung test kills the whole pytest process, so all results from the 1328 completed benchmarks are lost (no `bench_results.xml`, 276-byte log artifact).

### Related Files

- benchmarks/ops/bench_moe_grouped_gemm_nopad.py
- tileops/ops/moe/
- tileops/kernels/moe/
- .github/workflows/nightly.yml

## Goal

Nightly `benchmark` job runs `bench_moe_grouped_gemm_nopad.py` to completion: the `deepseek-v3-decode-down-bfloat16` case finishes well under the 900s budget, and the job produces `bench_results.xml`.

## Plan

1. Reproduce on the nightly runner environment: `pytest benchmarks/ops/bench_moe_grouped_gemm_nopad.py -k deepseek-v3-decode-down -x --timeout=900`, with both warm and cold TileLang cache.
2. Isolate the hanging phase (op ctor / JIT warmup / profile / torch baseline) with stack dumps (`py-spy dump` or faulthandler) and `cuda-gdb` if the hang is GPU-side.
3. Fix the root cause in the kernel/op or its tile-config selection for the `n=7168, k=2048` decode shape.
4. Evaluate hardening so one hung benchmark cannot destroy the whole suite's results (e.g. per-test timeout with subprocess isolation for benchmarks, or emitting partial junit XML) — may be split into a follow-up CI issue if non-trivial.

## Constraints

- Fix must not regress the passing gate-up / prefill workloads of `MoeGroupedGemmNopadFwdOp` (latency parity within normal run-to-run variance).

## Acceptance Criteria

- [ ] `pytest benchmarks/ops/bench_moe_grouped_gemm_nopad.py` passes all four deepseek-v3 workloads on the nightly runner, each case completing in <900s
- [ ] Root cause of the hang is documented in the fixing PR: the hanging phase is identified with first-hand evidence; the mechanism is identified if reproducible — otherwise the PR must explicitly state the mechanism is unidentified, enumerate the reproduction attempts, and record the remaining hypotheses (no fabricated mechanism)
- [ ] Post-merge verification (not a pre-merge review gate): the first subsequent nightly `benchmark` job progresses past `bench_moe_grouped_gemm_nopad.py` and uploads a non-empty `bench_results.xml`; the PR body must carry this as an explicit post-merge checklist item

### LINKED ISSUE #1746: [REFACTOR][BENCHMARK] bind workload shapes from manifest signature; drop harness shape-key allowlist
### Symptom / Motivation

`benchmarks/benchmark_base.py` hardcodes `_WORKLOAD_SHAPE_KEYS = ("x_shape", "input_shape")` and classifies workload keys by exclusion. The manifest already declares each op's tensor input name in `signature.inputs`; the harness ignores it. #1734 broke `bench_dropout`/`bench_fft` collection exactly because of this, and #1745 patched it by growing the allowlist — the next differently-named input breaks collection again, and misspelled workload keys are silently forwarded as op kwargs.

An audit shows the manifest itself drifted: 24 `elementwise_unary_math` ops declare input `input` but their workloads use `x_shape`; `DropoutOp`/`FFTC2COp` have it backwards (signature says `x`, PyTorch and their workloads say `input`).

### Root Cause Analysis

The workload→harness binding is name-convention-based instead of signature-derived, and no validator rule ties workload keys to the signature, so the two drift independently.

### Related Files

- benchmarks/benchmark_base.py (`workloads_to_params`, `_WORKLOAD_META_KEYS`)
- tileops/manifest/elementwise_unary_math.yaml, regularization.yaml, sequence_modeling.yaml
- scripts/validate_manifest.py, docs/design/manifest.md

## Goal

For single-tensor-input ops the workload shape key is derived from `signature.inputs` (exactly `{input}_shape`), unknown workload keys are rejected by the validator, and the harness contains no hardcoded input-name list.

## Plan

1. Manifest: rename `x_shape` → `input_shape` in the 24 `elementwise_unary_math` workloads; rename signature input `x` → `input` for `DropoutOp` and `FFTC2COp` (PyTorch mirror rule), including `same_as`/roofline references.
2. `workloads_to_params`: look up `signature.inputs`, require exactly one tensor input and the `{input}_shape` key; validate remaining workload keys against `signature.params` ∪ {`dtypes`, `label`}; delete `_WORKLOAD_SHAPE_KEYS`.
3. Validator: for single-input ops whose workloads carry any `*_shape` key, enforce the same contract (multi-input family conventions like `q_shape`/`kv_shape` stay out of scope).
4. Update `docs/design/manifest.md` workloads section; adjust unit tests.

## Constraints

- Benchmark measurement methodology unchanged; parametrize contract of existing bench files (`shape, dtype[, extra]`) unchanged.
- Manifest signature/workload edits in this PR are human-authorized (repo owner) as part of the schema alignment.

## Acceptance Criteria

- [ ] `pytest --collect-only benchmarks/ops` succeeds with zero collection errors
- [ ] `scripts/validate_manifest.py` passes on the aligned manifest and fails on a workload with a misspelled param key or wrong shape key (covered by `tests/test_validate_manifest.py`)
- [ ] `grep -n '_WORKLOAD_SHAPE_KEYS' benchmarks/` returns nothing

lcy-seso:
Closes #1740. Closes #1746.

## Summary

- **One failure no longer destroys the session's results.** tilelang's fd-level output suppression corrupted pytest's capture stream; when the 07-24 nightly stalled >900s, the timeout handler crashed and all 1328 completed results were lost. Suppression now applies only when stdout/stderr are the real fds 1/2. The stall itself did not reproduce (7 attempts, latency parity with the last green nightly); mechanism unidentified. Process isolation: #1744.
- **Measurement never degrades silently.** CUPTI projection failure raises a dedicated sentinel — genuine CUDA errors/OOM propagate; the CUDA-events fallback and the >1 GiB clone-skip are logged and marked in results.
- **Workload keys derive from the manifest signature.** The hardcoded shape-key allowlist (which let #1734 break dropout/fft collection) is gone; the contract lives once in `tileops.manifest`, consumed by both `workloads_to_params` and validator rule R21. 26 manifest entries aligned to their signatures (`x_shape`→`input_shape`; `DropoutOp`/`FFTC2COp` input `x`→`input` incl. `forward()`). Unknown signature/entry/workload keys are rejected; the validator stays total over malformed YAML.
- **Cleanup.** "harness" jargon removed repo-wide; benchmark_base tests folded into one file; 24 redundant/tombstone validator tests deleted; `init_dims`/`parity_opt_out` special cases replaced by generic unknown-key rejection. Net −350 lines.

## Test plan

- `validate_manifest.py` full pass; 207 validator + 17 benchmark unit tests; `--collect-only benchmarks/ops` 1638, zero errors
- GPU (H200): dropout/fft tests 37 passed, benches 6 passed; `bench_moe_grouped_gemm_nopad` 4/4; fallback + error-propagation contract tests

## Benchmark

No kernel change and the timing protocol (CUPTI kernel-only, warmup/repeats/trials, L2 flush) is untouched; this PR changes suppression selection and fallback/error semantics around it. Latency parity: `bench_moe_grouped_gemm_nopad` decode-down 2.3061 ms vs 07-17's 2.3088 ms.

## Regression

- Bench files keep the `(shape, dtype[, extra])` parametrize contract
- Post-merge check: next nightly progresses past `bench_moe_grouped_gemm_nopad.py` with a non-empty `bench_results.xml`


benchmarks/benchmark_base.py
```diff
@@ -19,18 +21,25 @@
...
+# (input named ``x`` or ``input``; see :data:`_WORKLOAD_SHAPE_KEYS`).
+# Multi-input ops (e.g. attention families that declare ``q_shape`` /
+# ``kv_shape``) are not supported: :func:`workloads_to_params` will raise
+# ``KeyError`` if no accepted shape key is present. Extending to
+# signature-aware tensor binding is tracked as a follow-up and must also
+# update ``docs/design/manifest.md``.
 _WORKLOAD_META_KEYS: frozenset[str] = frozenset(
-    {"x_shape", "dtypes", "label"}
+    {*_WORKLOAD_SHAPE_KEYS, "dtypes", "label"}
 )
```
lcy-seso, line 27: Remove this hotfix.

tests/test_bench_native_suppression.py
```diff
@@ -0,0 +1,65 @@
...
+
+``bench_kernel`` silences C++-level profiler chatter with an fd-level
+redirect (tilelang's ``suppress_stdout_stderr``). That redirect targets
+``sys.stdout.fileno()`` / ``sys.stderr.fileno()``; when a test harness has
+already replaced those streams (pytest fd capture), redirecting the
+underlying descriptors corrupts the harness's capture stream. The helper
+must therefore return the fd-level suppressor only when stdout/stderr still
+point at the process-level descriptors 1 and 2, and a no-op context
+otherwise.
+"""
```
Ibuki-wind: why this is necessary?

Ibuki-wind (changes requested):

benchmarks/benchmark_base.py
```diff
@@ -424,10 +459,10 @@ def _workload_extra_params(w: dict) -> dict[str, Any]:
     ``correction``). These are forwarded to the op constructor by benchmark
     files that opt into ``include_extra=True``.
 
-    Only the reserved meta keys (``x_shape``, ``dtypes``, ``label``) and
-    dunder-style metadata keys are stripped; everything else — including
-    any other ``*_shape`` keys — is surfaced as an op param. This matches
-    the single-input ``x_shape``-only harness contract documented in
+    Only the reserved meta keys (the accepted shape keys, ``dtypes``,
```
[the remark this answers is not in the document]
lcy-seso: Superseded: the x_shape/input_shape allowlist and its docstring were removed; the shape key now derives from signature.inputs (_workload_contract, commit cb7a68f6).

benchmarks/benchmark_base.py
lcy-seso: Removed: the shape-key allowlist is gone; binding derives from the manifest signature and unknown keys are rejected (commit cb7a68f6).

tests/test_bench_native_suppression.py
lcy-seso: File deleted in 2ad71a10's parent: the stub-based tests were removed; the suppression rationale lives in _native_output_suppressor's docstring and behavior is exercised by the real benchmark runs.

benchmarks/benchmark_base.py
```diff
@@ -416,25 +453,16 @@ def _build_result(self, latency: float) -> dict:
...
-    ``kv_shape`` are out of scope and would need a dedicated harness.
+    Strips the shape key, the harness keys, and dunder metadata.
     """
+    reserved = _WORKLOAD_HARNESS_KEYS | {shape_key}
     return {
         k: v
         for k, v in w.items()
-        if k not in _WORKLOAD_META_KEYS and not k.startswith("__")
+        if k not in reserved and not k.startswith("__")
     }
```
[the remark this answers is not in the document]
lcy-seso: Fixed in 2ad71a10: non-string keys are rejected by workloads_to_params (repr'd in the KeyError) and filtered in _workload_extra_params; validate_manifest reports them as schema errors.

benchmarks/tests/test_benchmark_base.py
```diff
@@ -192,16 +192,7 @@ def test_workloads_to_params_include_extra_propagates_dim():
...
-
-    # Direct unit test on the helper (no manifest mutation required).
-    assert _workload_extra_params(
-        {"x_shape": [4, 4], "dtypes": ["float16"], "label": "t",
-         "dim": 0, "keepdim": True}
-    ) == {"dim": 0, "keepdim": True}
+    from benchmarks.benchmark_base import workloads_to_params
 
     # End-to-end with the manifest: include_extra=True must still yield
     # well-formed triples with the (shape, dtype, extra) mapping. The
```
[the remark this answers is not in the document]
lcy-seso: Fixed in 2ad71a10: the test now asserts at least one SumFwdOp workload propagates dim in its extras.

benchmarks/benchmark_base.py
```diff
@@ -71,9 +89,6 @@ class BenchmarkWorkload(ShapeDtypeWorkload, InputGeneratingWorkload, Protocol):
     ...
 
 
-# Backward-compatible alias
-RooflineWorkload = ShapeDtypeWorkload
-
 W = TypeVar("W")
```
Ibuki-wind, line 80: `RooflineWorkload` is explicitly a backward-compatible public alias, but this PR removes it even though `ShapeDtypeWorkload` still exists and neither root-cause fix requires the deletion; `from benchmarks.benchmark_base import RooflineWorkload` now raises `ImportError`. → Restore `RooflineWorkload = ShapeDtypeWorkload` and keep the compatibility surface unchanged.

Ibuki-wind (changes requested):
### Overall
The two root-cause fixes are sound; remove the unrelated compatibility regression before merge.

Ibuki-wind (dismissed):
Clean — no issues.

benchmarks/benchmark_base.py
lcy-seso: Deletion stands by repo-owner ruling: the project is pre-release with no compatibility surface; unused aliases are removed rather than preserved. No consumer of RooflineWorkload exists in the repo.

benchmarks/benchmark_base.py
```diff
@@ -679,7 +747,12 @@ def dump(path: str) -> None:
                     row = [str(entry["params"].get(k, "")) for k in param_keys]
                     for rk in result_keys:
                         val = entry["result"].get(rk)
-                        row.append(f"{val:.4f}" if val is not None else "N/A")
+                        if val is None:
+                            row.append("N/A")
+                        elif isinstance(val, (int, float)):
+                            row.append(f"{val:.4f}")
+                        else:
+                            row.append(str(val))
```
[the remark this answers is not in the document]
lcy-seso: Fixed in 7444e1e3: bool is excluded from the numeric format branch; inputs_cloned=False renders as False.

docs/design/manifest.md
```diff
@@ -478,10 +480,18 @@ consumption.
...
 Each entry under `workloads:` is a mapping. Shape keys take the form
-`<tensor_name>_shape` (e.g. `x_shape`, `q_shape`, `kv_shape`) — one per
-tensor input in the workload. `dtypes` and `label` are also reserved.
-Any other key becomes an **op-call parameter** forwarded to the op's
-`__init__`.
+`<tensor_name>_shape`, where `<tensor_name>` MUST be a `signature.inputs`
+key — the workload shape key is derived from the signature, never chosen
+independently. `dtypes` and `label` are also reserved. Any other key
+becomes an **op-call parameter** forwarded to the op's `__init__` and MUST
+be a declared `signature.params` name.
```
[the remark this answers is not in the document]
lcy-seso: Fixed in 7444e1e3: the MUST language is scoped to single-tensor-input ops (R21); aggregate multi-input conventions are stated as the explicit exception.

benchmarks/benchmark_base.py
```diff
@@ -145,11 +169,37 @@ def _get_l2_flush_cache() -> torch.Tensor:
...
+def _native_output_suppressor():
+    """Return an output suppressor that cannot clobber a redirected stream.
+
+    tilelang's ``suppress_stdout_stderr`` silences C++ profiler chatter by ``dup2``-ing
+    ``/dev/null`` over ``sys.stdout.fileno()``. Under pytest's fd capture that
+    fileno is the capture tmpfile, so the redirect corrupts the capture stream
+    (later reads fail with ``EBADF``, losing timeout stack dumps and results).
+    Under capture a no-op context is returned; capture already hides the
+    chatter.
+    """
```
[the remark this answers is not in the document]
lcy-seso: Fixed in 7444e1e3: docstring now states the actual invariant — suppression applies when stdout/stderr are process fds 1/2 (including OS-level redirection, where silencing is intended); the guard targets pytest fd capture.

benchmarks/benchmark_base.py
```diff
@@ -272,16 +324,25 @@ def _run(i):
...
+                # failures (CUDA errors, OOM) propagate to the caller.
                 if n_regions != n_repeat:
-                    raise RuntimeError
+                    raise _CuptiProjectionError(
+                        f"{n_regions}/{n_repeat} annotation windows projected"
+                    )
                 trial_means.append((total_us / n_repeat) * 1e-3)
-    except RuntimeError:
+        _bench_meta.timing = "cupti"
+    except _CuptiProjectionError as exc:
```
Ibuki-wind, line 305: `bench_kernel` now relies on `_CuptiProjectionError` to keep genuine CUDA/OOM `RuntimeError`s from being swallowed by the CUDA-events fallback, and it adds output-distinguishing fallback metadata, but no test forces either the projection-mismatch branch or a kernel `RuntimeError`; the passing GPU smoke only exercises the successful CUPTI path. → Add benchmark contract tests that force `n_regions != n_repeat` and assert CUDA-events fallback with `timing="cuda-events"`, then force `fn` to raise `RuntimeError` and assert that it propagates.

Ibuki-wind (changes requested):
### Overall
The cleanup is sound; cover the new benchmark fallback/error-propagation contract before merge.

benchmarks/benchmark_base.py
```diff
@@ -145,11 +166,38 @@ def _get_l2_flush_cache() -> torch.Tensor:
...
+    """Return an fd-level output suppressor that is safe under pytest capture.
+
+    tilelang's ``suppress_stdout_stderr`` silences C++ profiler chatter by
+    ``dup2``-ing ``/dev/null`` over ``sys.stdout.fileno()``. Under pytest's fd
+    capture that fileno is the capture tmpfile, so the redirect corrupts the
+    capture stream (later reads fail with ``EBADF``, losing timeout stack
+    dumps and results). The suppressor is applied only when stdout/stderr are
+    the process-level fds 1/2 (including OS-level ``> file`` redirection,
+    where silencing is intended); under capture a no-op context is returned.
+    """
```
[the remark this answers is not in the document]
lcy-seso: Rejected by owner decision: stub-based unit tests for this guard were deliberately removed (streams faked via _FakeStream/StringIO test nothing real). The guard executes on every pytest-run benchmark; a regression corrupts pytest capture and fails loudly. Rationale documented in the function docstring.

benchmarks/benchmark_base.py
lcy-seso: Fixed in 7217db7c: benchmarks/tests/test_bench_kernel_contract.py adds two real-execution tests — a kernel-free callable naturally yields zero projected windows and asserts the CUDA-events fallback with timing=cuda-events; a raising callable asserts RuntimeError propagates. No mocks involved.

benchmarks/tests/test_benchmark_base.py
```diff
@@ -231,3 +226,129 @@ def eval_roofline(self):
...
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA required")
+def test_projection_failure_falls_back_to_cuda_events():
+    """A callable launching no CUDA kernel projects no annotation windows;
+    bench_kernel must fall back and mark the deviating timing method."""
+    latency = bench_kernel(lambda: sum(range(64)), n_warmup=1, n_repeat=2, n_trials=1)
+    assert latency >= 0.0
+    assert _bench_meta.timing == "cuda-events"
+
+
+def test_kernel_runtime_error_propagates():
```
Ibuki-wind, line 249: `test_kernel_runtime_error_propagates` is collected by the CPU-only `benchmark-contract-tests` job, but `bench_kernel` initializes `_get_l2_flush_cache()` before calling `boom`, so the test fails on CPU with a CUDA initialization error instead of observing `kernel failure` (reproduced with `CUDA_VISIBLE_DEVICES=''`: 1 failed, 15 passed, 1 skipped; the GitHub job now fails too). The sibling projection test is skipped on CPU, and neither new test is marked `smoke`, so the PR GPU job does not supply the missing execution. → Make both contracts execute in required CI, preferably with targeted CPU fakes around the CUDA/profiler plumbing, while asserting CUDA-events fallback metadata and propagation of the callable's `RuntimeError`.

docs/design/manifest.md
```diff
@@ -102,6 +102,8 @@ dtype_combos:
 
 **R20. `static_dims`.** For arbitrary-rank ops (no `shape` declaration), `static_dims` declares values the user commits to at Op construction time. Each entry maps an `__init__` keyword name to a single-axis shape expression `<tensor>.shape[<const_or_param>]`. See [`static_dims`](#static_dims) for full semantics, rules, and examples.
```
Ibuki-wind, line 104: R21 now requires `{input}_shape` for every single-tensor-input op, but `_check_single_input_workload_keys` intentionally returns early when workloads have no `*_shape` key and `test_out_of_scope_shapes_pass` locks in that shapeless-workload behavior; the normative text is therefore stricter than the implementation. → Restore the “whose workloads carry a shape” scope, or change the validator and its test consistently if shapeless single-input workloads are no longer allowed.

Ibuki-wind (changes requested):
### Overall
The cleanup is mostly sound; align the new tests with CI and restore the intended R21 scope before merge.

tileops/ops/dropout.py
```diff
@@ -112,15 +112,15 @@ def _eager_forward(self, x: torch.Tensor) -> torch.Tensor:
...
 
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
-        if not x.is_cuda:
+    def forward(self, input: torch.Tensor) -> torch.Tensor:  # noqa: A002
+        if not input.is_cuda:
             raise ValueError("Input must be a CUDA tensor")
-        if x.dtype not in (torch.float16, torch.bfloat16, torch.float32):
-            raise ValueError(f"x.dtype must be float16, bfloat16, or float32, got {x.dtype}")
+        if input.dtype not in (torch.float16, torch.bfloat16, torch.float32):
+            raise ValueError(f"input.dtype must be float16, bfloat16, or float32, got {input.dtype}")
```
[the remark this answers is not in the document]
lcy-seso: Fixed: message lowercased to match the parameter name.

benchmarks/tests/test_benchmark_base.py
lcy-seso: Fixed: both contract tests are smoke-tier with a CUDA skipif — the GPU gpu-smoke gate (required) executes them on every PR; the CPU contract-tests job skips them. Verified: CUDA_VISIBLE_DEVICES='' gives 15 passed/2 skipped; GPU gives 17 passed. CPU fakes around the CUDA plumbing are excluded by owner ruling (no stub-based tests).

docs/design/manifest.md
lcy-seso: Fixed: restored the 'workloads carrying a *_shape key' qualifier so R21's normative text matches the validator's scope; shapeless single-input workloads remain allowed.

tileops/manifest/__init__.py
```diff
@@ -77,8 +77,33 @@ def load_workloads(op_name: str) -> list[dict[str, Any]]:
...
+
+    The shape key is ``{input}_shape``; allowed keys are the shape key,
+    :data:`WORKLOAD_RESERVED_KEYS`, and the declared ``params`` names.
+    """
+    inputs = list((signature or {}).get("inputs") or {})
+    if len(inputs) != 1:
+        return None
+    shape_key = f"{inputs[0]}_shape"
+    params = (signature or {}).get("params") or {}
+    allowed = frozenset(params) | WORKLOAD_RESERVED_KEYS | {shape_key}
```
Ibuki-wind: `single_input_workload_contract` assumes `signature.params` is iterable: a shaped single-input entry with `params: 7` makes `check_l0` crash here with `TypeError` instead of returning the existing schema error. This remains even after applying the unresolved `inputs` guard. → Build param names only from a mapping with string keys (otherwise leave them empty so L0 reports the malformed field), and add scalar/non-string-param regression cases.

scripts/validate_manifest.py
```diff
@@ -347,14 +403,13 @@ def check_l0(
                         _check_shape_rule_callables(op_name, i, rule)
                     )
 
-        # Reject the deprecated `init_dims` key explicitly (R20 rename).
-        # L0 doesn't flag unknown signature keys, so without this check an
-        # accidental reintroduction would silently pass and be ignored by L1.
-        if "init_dims" in sig:
+        # Unknown signature keys are silently ignored by L1+, so reject
+        # anything outside the schema here.
+        unknown_sig = sorted(set(sig) - _VALID_SIGNATURE_KEYS)
```
Ibuki-wind: `sorted(set(sig) - ...)` crashes when malformed YAML contains mixed unknown key types (for example `1:` and `zzz:`), because Python cannot order `int` and `str`; the identical `unknown_top` expression has the same failure. A schema validator must report malformed input rather than die while formatting it. → Reject non-string keys explicitly or sort a normalized representation, and cover both signature and top-level mixed-key cases.

docs/design/manifest.md
```diff
@@ -102,6 +102,8 @@ dtype_combos:
...
 
+**R21. Workload keys derive from the signature.** Single-tensor-input ops whose workloads carry a `*_shape` key: the shape key MUST be `{input}_shape`; every other key MUST be a `signature.params` name or reserved `dtypes` / `label`. Enforced by the validator and `workloads_to_params`. Multi-input aggregate keys (`kv_shape`) are family bench-file conventions, out of scope.
```
Ibuki-wind, line 105: R21 exempts multi-input aggregate conventions such as `kv_shape`, while the workload table below still requires one `<tensor_name>_shape` key per tensor input. That leaves two incompatible authoring rules and no clear extension point for multi-input families. → Scope the table’s shape row to the single-input/generic convention and state where each multi-input family defines its aggregate shape-key mapping.

Ibuki-wind (changes requested):
### Overall
The consolidation is directionally good, but the new schema path still crashes on malformed YAML and the workload contract remains internally ambiguous. Please make the validator total over malformed input and leave one clear, extensible workload rule.

tileops/manifest/__init__.py
```diff
@@ -77,8 +77,33 @@ def load_workloads(op_name: str) -> list[dict[str, Any]]:
...
+    return shape_key, allowed
```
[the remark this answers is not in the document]
lcy-seso: Fixed in 83bade6b: the contract derives keys only from a dict inputs with exactly one string key; anything else returns None and L0 reports the malformed field.

tileops/manifest/__init__.py
lcy-seso: Fixed in 83bade6b: param names are taken only from a dict with string keys, otherwise left empty so L0's schema error stands; regression covers params=7 and inputs=7 through check_l0.

scripts/validate_manifest.py
lcy-seso: Fixed in 83bade6b: both unknown-key reporters sort repr(key), so mixed int/str keys format instead of raising; the regression exercises mixed-type unknown keys at signature and top level.

docs/design/manifest.md
lcy-seso: Fixed in 83bade6b: the table's shape row is scoped to the single-input R21 convention and states that multi-input families define their aggregate shape keys in their family bench files.

lcy-seso:
Correction: the four fixes referenced in the just-resolved threads landed in 4bb47f09, not 83bade6b (a formatting hook aborted the first commit attempt after the replies were drafted).

tileops/manifest/__init__.py
```diff
@@ -77,8 +77,41 @@ def load_workloads(op_name: str) -> list[dict[str, Any]]:
...
+    The shape key is ``{input}_shape``; allowed keys are the shape key,
+    :data:`WORKLOAD_RESERVED_KEYS`, and the declared ``params`` names.
+    """
+    sig = signature if isinstance(signature, dict) else {}
+    inputs = sig.get("inputs")
+    if not isinstance(inputs, dict) or len(inputs) != 1:
+        return None
+    (input_name,) = inputs
+    if not isinstance(input_name, str):
+        return None
```
Ibuki-wind, line 100: `None` currently conflates “out-of-scope arity” with a malformed one-input schema, and the filtering below similarly hides malformed param names: `check_l0` returns `[]` for both `inputs: {1: {dtype: float16}}` and `params: {3: {type: int}}`, so R21 is silently skipped despite the claim that L0 reports these fields. The same schema-name invariant also applies to outputs. → Make L0 explicitly reject non-string tensor/param names, keep this helper total for downstream consumers, and add exact non-string input/param regression assertions rather than only scalar-container cases.

Ibuki-wind (changes requested):
### Overall
The crash fixes and R21 documentation now hold, but malformed signature names can still silently bypass L0 and the shared contract; reject them explicitly before treating the schema as valid.

tileops/manifest/__init__.py
lcy-seso: Fixed in 4f4435d5: L0 now rejects non-string names in signature.inputs/outputs/params explicitly, so the shared contract's None stays a pure arity signal; regression asserts cover inputs {1: ...} and params {3: ...} exactly.

benchmarks/benchmark_base.py
```diff
@@ -416,59 +454,48 @@ def _build_result(self, latency: float) -> dict:
...
-    ``extra_params`` is a dict of op-call params declared on the workload
-    entry (e.g. ``{"dim": 0, "keepdim": False}``). Use this when the
-    benchmark needs to drive op calls from manifest-declared workload params.
+    Each entry becomes ``pytest.param(shape, dtype, id=...)``; with
+    ``include_extra=True`` a third element carries the op-call params
+    declared on the workload entry (e.g. ``{"dim": 0}``).
     """
     workloads = load_workloads(op_name)
+    shape_key, allowed = _workload_contract(op_name)
     params = []
```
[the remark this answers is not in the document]
lcy-seso: Fixed in 4f4435d5: the contract is resolved before load_workloads, so the op-not-found message is reachable.

docs/design/manifest.md
```diff
@@ -477,18 +479,15 @@ consumption.
...
+| --------------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `{input}_shape` | yes\*    | Shape for the single tensor input (list of ints), named per R21. \*Multi-input families define their own aggregate shape keys (e.g. `q_shape`/`kv_shape`) in their family bench files. |
+| `dtypes`        | yes      | List of dtype strings (`["float16", "bfloat16"]`).                                                                                                                                     |
+| `label`         | no       | Human-readable id used in the pytest param id and report tables.                                                                                                                       |
+| *any other key* | no       | Op param value (`dim`, `keepdim`, `correction`, …). Overrides the manifest's `signature.params` default.                                                                               |
```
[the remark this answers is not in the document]
lcy-seso: Fixed in 4f4435d5: the table's any-other-key row states the R21 constraint (MUST be a declared signature.params name).

benchmarks/benchmark_base.py
```diff
@@ -416,59 +454,48 @@ def _build_result(self, latency: float) -> dict:
...
-    With ``include_extra=True`` each entry becomes
-    ``pytest.param(shape, dtype, extra_params, id=...)`` where
-    ``extra_params`` is a dict of op-call params declared on the workload
-    entry (e.g. ``{"dim": 0, "keepdim": False}``). Use this when the
-    benchmark needs to drive op calls from manifest-declared workload params.
+    Each entry becomes ``pytest.param(shape, dtype, id=...)``; with
+    ``include_extra=True`` a third element carries the op-call params
+    declared on the workload entry (e.g. ``{"dim": 0}``).
     """
+    shape_key, allowed = _workload_contract(op_name)
```
Ibuki-wind, line 473: Moving `_workload_contract` before `load_workloads` makes its unknown-op branch observable, but no contract test calls `workloads_to_params` with an unknown op, so the newly selected error path and message are unguarded. → Add a focused `KeyError` test for an unknown op, or remove this order-only change and retain `load_workloads` as the single not-found contract.

Ibuki-wind (changes requested):
### Overall
The validator cleanup is correct, but the final benchmark contract coverage and PR description still need to match the reachable behavior.

### Cross-cutting concerns
- The PR body says “no kernel or measurement-path change,” although this diff changes fd suppression selection, CUPTI projection failure handling, CUDA-events fallback, and emitted timing metadata. Rewrite it as final-state-only verification, remove the incident/process narrative, and describe the measurement-path changes accurately.

benchmarks/benchmark_base.py
lcy-seso: Took the second option in ca8abf77: the duplicate not-found branch is deleted and load_workloads stays the single not-found contract (called first); _workload_contract only classifies arity for ops that exist. No new error path to guard.

benchmarks/benchmark_base.py
```diff
@@ -194,8 +231,6 @@ def bench_kernel(
             "Check that gen_inputs() returns a tuple."
         )
```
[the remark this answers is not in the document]
lcy-seso: Rejected: bench_kernel is an internal helper and no caller passes non-positive counts — the defaults are the contract. Speculative input validation is the kind of bloat this PR removes; a future misuse fails loudly at the call site.

Ibuki-wind (approved):
Clean — no issues.

Issue #1748, opened 2026-07-25
[FEAT][MANIFEST] declare per-op torch.compile fullgraph support


lcy-seso:
### Symptom / Motivation

TileOps registers many kernels/ops via `torch.library.custom_op` + `register_fake` for `torch.compile` traceability, and carries `torch.compile` call sites in four test files — but **which of the 171 manifest ops actually promise fullgraph compile support is written nowhere**. The promise exists only implicitly as "whichever op happens to have a compile test", with no drift protection. Meanwhile `docs/design/architecture.md` claims every L2 Op is `torch.compile` compatible, which the evidence contradicts:

- `tests/ops/test_elementwise_compile.py`: 63 op classes exercised with `fullgraph=True`, cold call, numeric checks — real evidence.
- `tests/test_compile.py`: `MultiHeadAttentionFwdOp`, `fullgraph=True`, cold call, numeric checks — real evidence.
- `tests/ops/test_norm_ops.py`: BatchNorm fwd/bwd compiled with `fullgraph=False` — does NOT back a fullgraph promise.
- `tests/ops/test_pool.py`: max_pool compile tests warm the kernel cache with an eager `op(x)` before compiling — proves warm-state capture only, not the user-facing cold-call contract.
- avg_pool and dozens of other registered ops: no compile tests at all.

This repo is design-first with the manifest as the authoritative spec, yet this user-visible capability is not in the manifest. Recent maintenance incidents on the compile test suite (dynamo recompile-limit) had no declared contract to assign ownership against.

### Root Cause Analysis

The manifest schema has no field for user-visible capabilities; `source.*` records file locations and `kernel_map` dispatch, none of which implies graph capturability. Registration (`custom_op` + `register_fake`) is necessary but not sufficient (BatchNorm has both yet only passes with graph breaks allowed). With no declared set, compile tests and the implicit promise drift independently.

### Related Files

- tileops/manifest/ (per-family YAML entries)
- scripts/validate_manifest.py (L0 schema checks)
- tests/test_validate_manifest.py
- tests/ops/test_elementwise_compile.py
- tests/test_compile.py
- .github/workflows/preflight.yml (always-on CPU job wiring)
- docs/design/manifest.md
- docs/design/architecture.md
- docs/design/trust-model.md
- docs/design/ops-design.md
- .claude/domain-rules/manifest-spec.md

## Goal

[...]

PR #1755, opened 2026-07-25
[Feat][POOL] cold-call torch.compile support for max_pool ops

lcy-seso:
## Summary

Follow-up deferred from #1748: upgrade max_pool to the cold-call contract and declare the 6 ops.

- Root cause of the old eager warm-up: `forward` constructs the Kernel object outside the custom-op boundary on cache miss, and dynamo cannot trace the eager tilelang build (`inspect`/`id()`) nor `init_config`'s `print`. The eager build only feeds autotuning — now built only when `tune=True`; the print is skipped under `torch.compiler.is_compiling()`.
- Compile tests drop the warm-up (verified cold on H200 for all six ops, exact match vs `F.max_pool*`), register their case tables as contract evidence, and `pool.yaml` declares `torch_compile_fullgraph: true` on the 6 max_pool ops.
- Declarations + evidence in one PR by design: the always-on equality gate rejects either half alone.

## Test plan

- [x] 6 cold `compile_fullgraph` cases pass on GPU
- [x] Equality gate: 70 declared == 70 registered; `validate_manifest.py` passes
- [x] `tests/ops/test_pool.py -m smoke` — 83 passed; autotune (`tune=True`) still builds the kernel
- [x] Test node delta: 0 (`tests/ops/test_pool.py` 205 → 205)

### DIFF
--- a/tileops/manifest/pool.yaml
+++ b/tileops/manifest/pool.yaml
@@ -191,6 +191,7 @@ MaxPool1dFwdOp:
   # Equivalent to torch.nn.functional.max_pool1d (return_indices=False)
   family: pool
   status: implemented
+  torch_compile_fullgraph: true
 
   signature:
     inputs:
@@ -249,6 +250,7 @@ MaxPool1dIndicesFwdOp:
   # Equivalent to torch.nn.functional.max_pool1d (return_indices=True)
   family: pool
   status: implemented
+  torch_compile_fullgraph: true
   variant_of: MaxPool1dFwdOp
 
   signature:
@@ -307,6 +309,7 @@ MaxPool2dFwdOp:
   # Equivalent to torch.nn.functional.max_pool2d (return_indices=False)
   family: pool
   status: implemented
+  torch_compile_fullgraph: true
 
   signature:
     inputs:
@@ -376,6 +379,7 @@ MaxPool2dIndicesFwdOp:
   # Equivalent to torch.nn.functional.max_pool2d (return_indices=True)
   family: pool
   status: implemented
+  torch_compile_fullgraph: true
   variant_of: MaxPool2dFwdOp
 
   signature:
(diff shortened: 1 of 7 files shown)


tileops/manifest/pool.yaml
```diff
@@ -191,6 +191,7 @@ MaxPool1dFwdOp:
   # Equivalent to torch.nn.functional.max_pool1d (return_indices=False)
   family: pool
   status: implemented
+  torch_compile_fullgraph: true
```
Ibuki-wind, line 194: The `torch_compile_fullgraph` additions for all six MaxPool entries violate the Manifest stage contract: capability declarations require a separate manifest-only PR, and the feature checklist forbids modifying `tileops/manifest/` here. → Remove these six declarations from this implementation/test PR and land them in a separate manifest-only PR.

Ibuki-wind (changes requested):
### Overall
One blocking trust-boundary violation must be resolved before landing.

tileops/manifest/pool.yaml
Ibuki-wind, line 194: The six `torch_compile_fullgraph` declarations are unchanged and still violate the explicit stage boundary in `docs/design/trust-model.md:21-28`: user-visible capability declarations are manifest-owned, require a separate human-reviewed PR, and are specifically excluded from the implementation carve-out. The exact-equality gate creates a sequencing conflict, but that does not authorize bypassing the documented boundary in this feature PR. → Remove these declarations from this implementation/test PR and resolve declaration/evidence staging through a separate human-reviewed manifest-policy change.

Ibuki-wind (changes requested):
### Overall
One blocking stage-contract issue remains attached inline.

lcy-seso:
Closing per design review: the per-class lazy-build approach is a family-local patch (cold AvgPool1d still fails on this branch) of a repo-wide pattern (~95 eager builder sites across 49 files). Redoing with the systemic architecture: a shared opaque custom-op dispatch boundary (BatchNorm's pattern generalized) so traced Op.forward never constructs Kernels, applied to the whole pool family. Cold-call tests and evidence registration carry over to the redo.

PR #1761, opened 2026-07-26
[Feat][CI] run nightly benchmarks one process per bench file

### LINKED ISSUE #1744: [META][CI] run nightly benchmarks one process per file: bound native-failure loss, capture hang forensics

lcy-seso:
Closes #1744.

## Summary

- `scripts/ci/run_benchmarks.py`: one pytest process per bench file (filesystem discovery, no imports); fragments merged into `bench_results.xml`. A hang/segfault/OOM — at import, in a test, or in teardown — costs one file, not the session (the 2026-07-24 failure mode).
- On timeout: `py-spy dump`, SIGKILL, synthetic junit entry with log tail; dumps uploaded. Signal deaths get the same entry exactly once; teardown-stuck children are killed on a rolling deadline while later files run.
- The parent never imports torch; queued children pre-import Python only and hold no CUDA context until granted the GPU (NVML-verified).
- `nightly.yml` calls the runner (900s/file, job limit 150 min); clock validation and `profile_run.log` unchanged. `pytest-timeout`/`py-spy` are baked into the runner image (Dockerfile) with all tool pins centralized in `constraints.txt`; until the next manual image rebuild the job-time `py-spy` install is best-effort (a network failure only disables stack dumps).

## Test plan

- [x] `pytest benchmarks/tests/` — 22 passed (real injection: sleeping file, `os.abort()` in test / at import / in atexit, teardown deadline expiring mid-file, fragment + profile-log merge)
- [x] `pre-commit run --all-files` — clean
- [x] Full-suite run on H200 via the runner: 61 files executed end-to-end; three timeouts contained with stack dumps (#1784, #1786)

## Benchmark

No kernel or timing-protocol change. Back-to-back on a warm H200, 58 files (three pathological files excluded symmetrically, see #1784/#1786):

| pass | wall |
|---|---|
| single process | 6323 s |
| per-file runner | 7371 s raw (+16.6%) / 5618 s (−11.2%) excluding one contained hang |

The raw/adjusted delta is one event: the 2026-07-24 kernel hang recurred mid-pass and was contained at the per-file cap with its stack captured (#1784). Owner decision: AC-4 judged on the adjusted number.

## Regression

`_native_output_suppressor` stays until tilelang's `suppress_stdout_stderr` gains the fd-1/2 check upstream (tilelang PR not yet proposed; bare `pytest benchmarks/ops` still relies on the guard).


scripts/ci/run_benchmarks.py
```diff
@@ -0,0 +1,339 @@
...
+        errors.append(proc.stderr.strip())
+    dump_path.write_text("py-spy dump failed:\n" + "\n".join(errors) + "\n")
+
+
+def _collect_bench_files(
+    targets: list[str], log_path: Path, lingering: list[Child]
+) -> list[str]:
+    """Return the absolute bench file paths pytest collects under targets."""
+    out_file = log_path.with_suffix(".files")
+    child = Child(_COLLECT, [str(out_file), *targets], log_path)
```
Ibuki-wind: `_collect_bench_files` imports every benchmark module in one shared `_COLLECT` process before per-file isolation starts, so a native crash or hang during module import/collection aborts discovery and leaves no `bench_results.xml`. I reproduced this with a `bench_crash.py` containing module-level `os.abort()`: the runner exited and the output XML was missing, while the current injection only aborts inside the test body. → Discover candidates without importing all modules in one process, or isolate/fallback collection per file, and add module-import abort/hang regression coverage.

.github/workflows/nightly.yml
```diff
@@ -81,9 +81,15 @@ jobs:
...
+          # One pytest process per bench file: a native hang / segfault / OOM
+          # kill costs only that file's results. py-spy captures the stack of
+          # a hung child before the runner kills it.
           set -o pipefail
-          python3 -m pip install -q -c constraints.txt pytest-timeout
-          python3 -m pytest -q benchmarks/ops --timeout=900 --timeout-method=thread --junit-xml=bench_results.xml | tee tileops_benchmarks.log
+          python3 -m pip install -q -c constraints.txt py-spy
+          python3 scripts/ci/run_benchmarks.py benchmarks/ops \
+            --junit-xml bench_results.xml \
+            --timeout-per-file 1800 \
```
Ibuki-wind: The 1800-second per-file timeout cannot fit inside this job’s existing 120-minute limit: the PR reports a 5618-second adjusted runner time, so one full timeout already reaches about 7418 seconds, before up to 240 seconds of `py-spy` attempts and setup/upload time; the reported raw run was 7371 seconds. GitHub can therefore terminate the job before the runner writes the merged XML, reproducing the all-results-lost failure this change is meant to prevent. → Reduce the per-file timeout/overall runtime or raise the job timeout with enough headroom for one timeout, stack capture, teardown, and artifact upload.

scripts/ci/run_benchmarks.py
```diff
@@ -0,0 +1,339 @@
...
+                    sys.stdout.write(log_path.read_text(errors="replace"))
+                    message = (
+                        f"timed out after {args.timeout_per_file:.0f}s; "
+                        f"killed, stack dump at {dump_path}"
+                    )
+                    print(f"TIMEOUT: {rel}: {message}", flush=True)
+                    suites.append(_synthetic_suite(bench_file, message, _log_tail(log_path)))
+                    failed.append(rel)
+                else:
+                    lingering.append(child)
```
Ibuki-wind: Appending every reported child to `lingering` defers both closing `status_fd` and enforcing `TEARDOWN_TIMEOUT_S` until the entire suite finishes. A child stuck in interpreter/native teardown can therefore retain its CUDA context for the rest of the nightly and interfere with later files, and multiple stuck children add serial 120-second waits only after all tests; the accumulated descriptors/zombies also grow with the file count. → Close each status fd immediately and reap completed children concurrently or poll them during the loop, enforcing the teardown deadline from the time each child reports its result.

Ibuki-wind (changes requested):
### Overall
The runner still has failure paths that bypass its durability boundary; address the blocking inline findings before merge.

scripts/ci/run_benchmarks.py
lcy-seso: Done in 15eff99a.

.github/workflows/nightly.yml
lcy-seso: Done in 15eff99a.

scripts/ci/run_benchmarks.py
lcy-seso: Done in 15eff99a.

scripts/ci/run_benchmarks.py
```diff
@@ -0,0 +1,320 @@
...
+        data = os.read(self.status_fd, 16)
+        os.close(self.status_fd)
+        if data:
+            return int(data)
+        # EOF without a status report: the child died; reap the reason.
+        return self.proc.wait()
+
+    def poll_teardown(self, deadline: float) -> bool:
+        """Return True once the child is gone; kill it past its deadline."""
+        if self.proc.poll() is not None:
```
Ibuki-wind: `Child.poll_teardown` drops the process return code whenever `proc.poll()` is non-None, and the deadline-kill path also returns only a boolean, so failures after the status-pipe write are silently converted to success. I reproduced this with a passing test that registers `atexit.register(os.abort)`: the child dies by SIGABRT during teardown, but the runner prints `all benchmark files passed` and emits no JUnit error; a forced kill for stuck teardown is likewise only a warning. → Propagate each teardown termination outcome back to `main`, fail/synthesize an error for an unexpected signal or deadline kill, and add teardown-crash/hang regression coverage.

scripts/ci/run_benchmarks.py
```diff
@@ -0,0 +1,320 @@
...
+                child = pending.popleft()
+                top_up()
+                rel = os.path.relpath(bench_file)
+                print(f"\n=== [{index + 1}/{len(bench_files)}] {rel} ===", flush=True)
+                fragment = work_dir / f"{index:03d}.xml"
+                log_path = work_dir / f"{index:03d}.log"
+
+                start = time.monotonic()
+                child.release()
+                rc = child.wait_result(args.timeout_per_file)
```
Ibuki-wind: `_reap_lingering` is called only after `child.wait_result(args.timeout_per_file)` returns, so no teardown deadline is enforced while the next benchmark blocks here for up to 900 seconds. A previous child can therefore retain its CUDA context far beyond `TEARDOWN_TIMEOUT_S=120` and interfere with the entire next file; an accelerated integration reproduction with a 0.1-second teardown limit left the old process alive throughout a 0.5-second next test and killed it only afterward. → Supervise the current status pipe and lingering teardown deadlines in the same polling loop (or use a concurrent reaper) so each child is killed at its actual deadline, then cover a teardown deadline expiring while the next file is still running.

Ibuki-wind (changes requested):
### Overall
Child lifecycle supervision is still not reliable enough to preserve a truthful nightly result.

scripts/ci/run_benchmarks.py
lcy-seso: Done in 7680edf6.

scripts/ci/run_benchmarks.py
lcy-seso: Done in 7680edf6.

scripts/ci/run_benchmarks.py
```diff
@@ -0,0 +1,357 @@
...
+                    sys.stdout.write(log_path.read_text(errors="replace"))
+                    message = (
+                        f"timed out after {args.timeout_per_file:.0f}s; "
+                        f"killed, stack dump at {dump_path}"
+                    )
+                    print(f"TIMEOUT: {rel}: {message}", flush=True)
+                    suites.append(_synthetic_suite(bench_file, message, _log_tail(log_path)))
+                    failed.append(rel)
+                else:
+                    lingering.append((child, time.monotonic() + args.teardown_timeout, bench_file))
```
Ibuki-wind: Every non-timeout child is appended to `lingering`, including an `rc < 0` child that `wait_result` has already reaped and that the branch below already records as a runtime crash. The next `_reap_lingering` sees the same negative return code and records it again as “died in teardown”; a real `os.abort()` therefore produces two synthetic JUnit errors and duplicates the file in the failure summary. The current test misses this because `_cases` collapses duplicate `classname::name` entries into a dict; directly counting XML error nodes reproduces 2 instead of 1. → Do not enqueue a child whose pre-status signal death is already handled (while still enqueueing status-reported children so post-status deaths remain observable), and make the crash regression assert the raw XML contains exactly one error.

Ibuki-wind (changes requested):
### Overall
Failure bookkeeping still has one correctness regression that corrupts the merged report.

scripts/ci/run_benchmarks.py
lcy-seso: Done in 0d42aa55.

Ibuki-wind (changes requested):
### Overall
The implementation is clean; approval is blocked only on the repository’s authoring-discipline gate.

### Cross-cutting concerns
- Bring the PR body to final state: remove the stale “Draft until the full-suite overhead numbers are posted” sentence; replace the stale `20/20` verification fact with the current accurate result (22 benchmark tests are now collected, or state the exact H200 command/result actually rerun); remove the obsolete claim that collection-ignore paths are tested; and format the test plan as the template’s completed checkboxes for pre-commit/pytest.
- Edit the six reviewer-thread replies to outcome-only `Done in <sha>.` lines. The current replies restate implementation/design details, which the approval gate reserves for commits and review history.

Ibuki-wind (dismissed):
Clean — no issues.

.github/runner/Dockerfile
```diff
@@ -51,7 +51,7 @@ ENV MAX_JOBS=${MAX_JOBS} NVCC_THREADS=${NVCC_THREADS} TORCH_CUDA_ARCH_LIST=9.0
 # the final post-bench stack.
 FROM runtime AS post-fa3
 RUN python -m pip install --no-cache-dir -c /tmp/constraints.txt \
-        "pytest==9.0.2" "pytest-xdist>=3.0" "ruff==0.14.13"
+        pytest pytest-xdist ruff pytest-timeout py-spy
```
Ibuki-wind, line 54: The `post-fa3` stage now also contains `pytest-timeout` and `py-spy`, while `.github/runner/README.md` still documents this stage as containing only pytest / pytest-xdist / ruff; the manual runner-image documentation is therefore stale. → Update the `post-fa3` stage row in the README to include both newly baked tools.

constraints.txt
```diff
@@ -21,6 +21,10 @@ typing_extensions==4.15.0
 Cython==3.2.4
 z3-solver==4.15.4.0
 torch_c_dlpack_ext==0.1.5
-# CI test plugin: per-test hang guard for the GPU jobs. Pinned here so the runtime install is
-# deterministic; the runner image should bake it (follow-up) and drop the per-job install.
+# CI tooling: versions locked here; the runner image install lists carry bare names.
```
Ibuki-wind, line 25: The new runner-only block contains `py-spy` and `pytest-timeout`, neither of which is a member of `pyproject.toml`, so it makes the file header’s “membership lives in pyproject” contract false. → Amend the header or this block to distinguish project dependency membership from runner-only tooling membership.

Ibuki-wind (changes requested):
### Overall
The dependency move resolves cleanly, but the final-state documentation must be synchronized before approval.

### Cross-cutting concerns
- Update the PR Summary for the latest two commits: state that `pytest-timeout` and `py-spy` are now baked into the manually rebuilt/deployed runner image and that the tool pins were centralized in `constraints.txt`; the current Summary only says `py-spy` is pinned and does not describe the Dockerfile change.

.github/runner/Dockerfile
lcy-seso: Done in a30363cc.

constraints.txt
lcy-seso: Done in a30363cc.

Ibuki-wind (approved):
Clean — no issues.

Issue #1760, opened 2026-07-26
[BUG][MANIFEST] author real rooflines for six linear-attention placeholder entries


lcy-seso:
### Symptom / Motivation

GLAFwdOp, GLABwdOp, DeltaNetFwdOp, DeltaNetBwdOp, GatedDeltaNetFwdOp, and GatedDeltaNetBwdOp carry `roofline: {flops: "0", bytes: "0"}` — structural placeholders (normalized from integer zeros when L0 gained type checks). A zero roofline makes every efficiency ratio on these ops meaningless.

### Root Cause Analysis

The entries predate roofline authoring for the linear-attention family; sibling ops in the same file already use `func:` formulas in `tileops.perf.formulas`.

### Related Files

- tileops/manifest/linear_attention.yaml
- tileops/perf/formulas.py

## Goal

All six entries reference real roofline formulas; no `"0"` placeholders remain in the manifest.

## Plan

1. Derive flops/bytes formulas for GLA and DeltaNet fwd/bwd from the kernel math (chunked linear attention), mirroring the existing prefill/decode formulas in tileops/perf/formulas.py.
2. Add the functions to tileops/perf/formulas.py and point the six entries at them via `func:`.
3. Verify with the manifest-driven benchmarks that reported efficiency becomes finite and plausible.

## Constraints

- Manifest edits per docs/design/trust-model.md (human-reviewed).

## Acceptance Criteria

- [ ] Modified files pass unit tests
- [ ] No `"0"` roofline placeholders remain; benchmarks report finite efficiency for all six ops

PR #1757, opened 2026-07-26
[Feat][OPS] compile dispatch boundary; pool declarations; design-doc audit

lcy-seso:
## Summary

- New contract: a traced `Op.forward` never constructs Kernels. Shared registry (`tileops/ops/compile_boundary.py`) + opaque custom-op dispatch; auto-registration via `Op.dispatch_kernel`. Recorded in ops-design.md.
- Pool adopts it: all 9 ops compile cold under `fullgraph=True`; kernel files untouched; `pool.yaml` declares them. BatchNorm migrates to the shared registry.
- Design-doc audit (all 7 docs): stale facts fixed, implementation snapshots and non-decisions removed; MoE class tour moved to its module docstring; scaffold example made fully fictional. testing.md needed nothing.

## Test plan

- [x] 9 pool ops cold-compile on H200, match `F.*_pool*`; eager regression clean
- [x] Equality gate 73 == 73; validator + 189 structural tests pass; pool smoke 86, norm 5
- [x] Node delta +3 (new avg_pool compile cases); pre-commit clean

PR #1762, opened 2026-07-26
[Fix][Manifest] normalize six placeholder rooflines to the string form

lcy-seso:
## Summary

Manifest-only. `GLA*/DeltaNet*/GatedDeltaNet*` fwd/bwd carry `roofline: {flops: 0, bytes: 0}`; the spec requires formula strings. `"0"` evaluates identically — consumer behavior unchanged. Real formulas: #1760. Prerequisite for the stricter roofline type checks in #1757.

## Test plan

- [x] `validate_manifest.py` passes (with #1757's checks applied locally)
- [x] Six-line diff, values semantically identical

### DIFF
--- a/tileops/manifest/linear_attention.yaml
+++ b/tileops/manifest/linear_attention.yaml
@@ -243,7 +243,7 @@ GatedDeltaNetFwdOp:
       - "S % chunk_size == 0"
       - "NC == S // chunk_size"
   workloads: []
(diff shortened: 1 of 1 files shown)


lcy-seso:
Folding into #1757 per maintainer instruction — single PR carries the check and the normalization it requires.

PR #1777, opened 2026-07-26
[Refactor][POOL] collapse pool op class bodies into ndim-generic family bases

### LINKED ISSUE #1766: [REFACTOR][POOL] collapse per-dim op clones into generic pool bases
### Symptom / Motivation
`tileops/ops/pool.py` (1431 lines) carries nine near-identical per-dimension op classes — `AvgPool{1,2,3}dFwdOp` as three standalone ~170–230-line bodies, and `_MaxPool{1,2,3}dFwdOpBase` each with a `Fwd` + `Indices` subclass. The 1d/2d/3d bodies differ only in ndim mechanics (ctor normalization, `validate_pool_params`, per-axis `pool_output_dim` math, kernel cache keys, custom_op registration). Copy-paste structure triples every maintenance change.

### Root Cause Analysis
Per-op scaffolding predates the shared helpers in `tileops/kernels/pool/common.py`; dimension count was cloned into class bodies instead of being a parameter. Design review (Codex-assisted) confirms a two-generic-base collapse is conditionally sound; estimated net reduction 500–650 lines (35–45%).

### Related Files
- tileops/ops/pool.py
- tileops/kernels/pool/common.py
- tests/ops/test_pool.py
- docs/design/ops-design.md

## Goal
Collapse the nine pool op class bodies into two generic bases (`_AvgPoolFwdOpBase`, `_MaxPoolFwdOpBase`) parametrized by class-attribute `ndim`, keeping the nine public classes as thin subclasses with unchanged names, signatures, and per-op `kernel_map`.

## Plan

1. Add contract snapshot tests first: ctor signatures, `kernel_map` precedence, kernel-cache-key separation, `register_fake` shapes/dtypes, error messages, compile behavior.
2. Collapse the three AvgPool ops — preserve the three fast-path policies, AvgPool1d scalar state, and the 1d ctor's absence of `divisor_override`.
3. Collapse the six MaxPool variants with explicit `returns_indices` + `_kernel_slot`; indices keep int64 fake output and their roofline bytes.
4. Update `docs/design/ops-design.md`; gate on `validate_manifest.py`, CPU signature/shape parity, and CUDA eager + `torch.compile` parity.

## Constraints
- Public class names and `__init__`/`forward` signatures are pinned by the manifest — no `tileops/manifest/` edits in this work.
- Do NOT place `eval_roofline` / `_validate_dtypes` on the intermediate bases: manifest codegen installs them per concrete class; a base override silently bypasses per-op generation.
- Keep the existing explicit kernel-cache keys; `Op._cache_key` is not equivalent (pool has empty `_static_axes`).
- Preserve the avg-pool 1d/3d-vs-2d fast-path asymmetry as-is; behavior fixes are out of scope.
- Trust model: [docs/design/trust-model.md](docs/design/trust-model.md).

## Acceptance Criteria
- [ ] AC-1: contract snapshot suite green before and after each collapse step
- [ ] AC-2: `tileops/ops/pool.py` shrinks by >=500 lines with all nine public classes intact
- [ ] AC-3: `python scripts/validate_manifest.py` passes with zero manifest edits
- [ ] AC-4: CUDA eager and `torch.compile(fullgraph=True)` smoke parity for all nine ops
- [ ] AC-5: `docs/design/ops-design.md` documents the generic-base + `ndim` protocol

lcy-seso:
Closes #1766

## Summary

- Collapse the nine pool op class bodies into two ndim-parametrized generic bases (`_AvgPoolFwdOpBase`, `_MaxPoolFwdOpBase`); `tileops/ops/pool.py` shrinks 1431 -> 918 lines (-513).
- All nine public classes stay as thin subclasses with unchanged names, rank-specific ctor/forward signatures, and per-op `kernel_map`.
- Add a cross-family contract snapshot suite: ctor signatures, `kernel_map` precedence, kernel-cache-key separation, `register_fake` shapes/dtypes, error messages, compile behavior.
- Document the dimension-parametrized family-base + `ndim` protocol in `docs/design/ops-design.md`.
- Zero `tileops/manifest/` edits; `scripts/validate_manifest.py` passes unchanged.

## Test plan

- [x] Contract snapshot suite green before and after each collapse step
- [x] `tileops/ops/pool.py` shrinks by >=500 lines with all nine public classes intact
- [x] `python scripts/validate_manifest.py` passes with zero manifest edits
- [x] CUDA eager and `torch.compile(fullgraph=True)` smoke parity for all nine ops
- [x] `docs/design/ops-design.md` documents the generic-base + `ndim` protocol

## Regression

- Contract snapshots freeze the public surface: `inspect.signature` for all nine ctors is byte-identical to the pre-collapse baseline; fake-output shapes/dtypes (incl. int64 MaxPool indices) and error messages are snapshotted.
- Avg-pool 1d/3d-vs-2d fast-path asymmetry and the explicit kernel-cache keys are preserved as-is.

## Test node delta

```
File                      Base    HEAD    Delta
-----------------------------------------------
tests/ops/test_pool.py     194     244      +50
-----------------------------------------------
TOTAL                      194     244      +50

Growth: +25.8%
```

**Justification:** the added nodes are the contract snapshot suite that guards the collapse — each case pins a distinct public class or preservation path (ctor signature, cache-key separation, fake registration, error message, fullgraph compile), so the refactor cannot silently change the public surface.

### DIFF
--- a/tileops/ops/pool.py
+++ b/tileops/ops/pool.py
@@ -374,13 +374,6 @@ class AvgPool2dFwdOp(_AvgPoolFwdOpBase):
         return flops, bytes_
 
 
-_MAX_POOL_PARAM_SUFFIXES: Dict[int, Tuple[str, ...]] = {
-    1: ("w",),
-    2: ("h", "w"),
-    3: ("d", "h", "w"),
-}
-
-
 def _max_pool_roofline(op: "_MaxPoolFwdOpBase", *, indices: bool) -> tuple[int, int]:
     """Shared max-pool roofline: flops = out_elems * prod(kernel); bytes in+out."""
     if op._last_roofline_spec is None:
--- a/tileops/ops/pool.py
+++ b/tileops/ops/pool.py
@@ -43,204 +43,65 @@ def _device_index(tensor: torch.Tensor) -> int | None:
     return tensor.device.index
 
 
-class AvgPool1dFwdOp(Op):
-    """Average pooling over PyTorch-compatible NCL inputs."""
+# Layout token and per-axis name suffixes, indexed by spatial dimensionality.
+_POOL_LAYOUTS: Dict[int, str] = {1: "NCL", 2: "NCHW", 3: "NCDHW"}
+_POOL_DIM_NAMES: Dict[int, Tuple[str, ...]] = {1: ("l",), 2: ("h", "w"), 3: ("d", "h", "w")}
+# Kernel-kwarg suffixes for kernel_size/stride/padding(/dilation).
+# Why: the 1d max-pool kernels historically name their pooling axis `w`.
+_AVG_POOL_PARAM_SUFFIXES: Dict[int, Tuple[str, ...]] = _POOL_DIM_NAMES
+_MAX_POOL_PARAM_SUFFIXES: Dict[int, Tuple[str, ...]] = {
+    1: ("w",),
+    2: ("h", "w"),
+    3: ("d", "h", "w"),
+}
+
+
+def _validate_pool_input_dtypes(self, input: torch.Tensor) -> None:
+    """Shared pool-family dtype validator (bound per concrete class)."""
+    if input.dtype not in {torch.float16, torch.bfloat16, torch.float32}:
+        raise ValueError(
+            f"input.dtype must be float16, bfloat16, or float32, got {input.dtype}"
+        )
+
+
+class _AvgPoolFwdOpBase(Op):
+    """Generic average-pooling forward, parametrized by class-attribute ``ndim``.
+
+    Concrete subclasses set ``ndim``, supply ``default_kernel_map``, and keep
+    ``eval_roofline`` / ``_validate_dtypes`` in their own class body so
+    manifest codegen resolves them per concrete class.
+    """
+
+    ndim: ClassVar[int]
 
     def __init__(
         self,
-        kernel_size: int | Tuple[int],
-        stride: Optional[int | Tuple[int]] = None,
-        padding: int | Tuple[int] = 0,
-        ceil_mode: bool = False,
-        count_include_pad: bool = True,
-        kernel_map: Optional[Dict[str, Kernel]] = None,
-        tune: bool = False,
-    ) -> None:
-        self.n = None
-        self.c_in = None
-        self.l_in = None
-        self.kernel_size = _normalize_pool_dims("kernel_size", kernel_size, 1)[0]
-        self.stride = (
-            (self.kernel_size,) if stride is None else _normalize_pool_dims("stride", stride, 1)
-        )[0]
-        self.padding = _normalize_pool_dims("padding", padding, 1)[0]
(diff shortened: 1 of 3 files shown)


Ibuki-wind (changes requested):
### Overall
The family-base refactor needs its remaining contract divergences resolved before merge.

tileops/ops/pool.py
```diff
@@ -496,385 +576,239 @@ def _get_kernel_1d(
...
+            for k, size in enumerate(in_dims)
+        )
+        full = (n, c_in, *out_dims)
+        if self._returns_indices:
+            return {"output": full, "indices": full}
+        return {"output": full}
+
+    def forward(
+        self, input: torch.Tensor,
+    ) -> torch.Tensor | Tuple[torch.Tensor, torch.Tensor]:
```
Ibuki-wind: `_MaxPoolFwdOpBase.forward` now exposes the same union return annotation through every concrete class, so plain leaves no longer advertise `torch.Tensor` and indexed leaves no longer advertise `Tuple[torch.Tensor, torch.Tensor]`; this violates `docs/design/ops-design-reference.md:146`, which requires each return annotation to match its manifest outputs → Preserve per-concrete-class return annotations without duplicating the runtime method bodies, and add a signature snapshot covering one plain and one indexed leaf.

Ibuki-wind (changes requested):
### Overall
Resolve the inline contract mismatch before merge.

tileops/ops/pool.py
lcy-seso: Done in ea890427.

tileops/ops/pool.py
```diff
@@ -496,385 +587,240 @@ def _get_kernel_1d(
...
+        )
+        full = (n, c_in, *out_dims)
+        if self._returns_indices:
+            return {"output": full, "indices": full}
+        return {"output": full}
+
+    def __init_subclass__(cls, **kwargs) -> None:
+        # _returns_indices selects the forward variant at class-definition
+        # time so every concrete class carries the exact return annotation
+        # its manifest outputs declare (Tensor vs Tuple[Tensor, Tensor]).
```
Ibuki-wind, line 633: Review reply `3652412426` includes implementation narration (`A _make_max_pool_forward factory + __init_subclass__...`) instead of an allowed outcome-only status, so the approval-gate Replies check fails. → Edit it to exactly `Done in ea890427.`

Ibuki-wind (changes requested):
### Overall
Resolve the remaining inline gate failure.

tileops/ops/pool.py
lcy-seso: Done in ea890427.

Ibuki-wind (approved):
Divergence introspection: reaffirmed — `docs/design/ops-design-reference.md` Slot S15 requires concrete `forward` annotations to match manifest outputs, and the current implementation satisfies it.

Clean — no issues.

PR #1779, opened 2026-07-26
[Refactor][POOL] strip over-design and scaffolding from the merged cleanup PRs

lcy-seso:
## Summary

- Remove over-engineered pool hooks and private snapshot tests; make indexed max-pool forwards explicit.
- Deduplicate GLA, Mamba, and formula tests; remove dead helpers, commented benchmark rows, and process metadata.
- Consolidate benchmarks onto manifest workloads and `ManifestBenchmark`; delete obsolete benchmark modules and factor redundant sweep axes.
- Require implemented ops to declare `kernel_map` and manifest-driven benchmark coverage, filling the corresponding manifest gaps.
- Reduce the repository by 2,415 net lines without changing runtime operator behavior.

## Test plan

- [x] pre-commit passed
- [x] Pool tests passed: 246
- [x] Perf/formula and validator tests passed: 128
- [x] Changed benchmark modules collected: 538 nodes
- [x] Test node delta: 476 → 391 (-85)
- [x] `python scripts/validate_manifest.py --strict` passed


lcy-seso:
Two validator gates stay advisory after this PR because they are blocked on implementation, not policy — measured and tracked in #1782: shape parity is skipped on 141/165 implemented ops (no `_infer_output_shapes`), and 201 `shape_rules` cannot be evaluated under mock inputs (conv/pool internals). The gate flips are the last step there, after the implementations land.

benchmarks/ops/bench_binary_elementwise.py
```diff
@@ -352,28 +375,36 @@ def _gelu_tanh_and_mul_baseline(x: torch.Tensor) -> torch.Tensor:
...
+def _profile_fused_gated(bm: ManifestBenchmark, op, test, baseline_key: str) -> None:
     inputs = test.gen_inputs()
-
-    # The output shape (M, N) is the model-relevant geometry; the input
-    # carries the gate/value-concatenated trailing axis (2*N).
-    shape = (M, N)
-    op = op_cls(M=M, N=N, dtype=dtype)
     result = bm.profile(op, *inputs)
-    BenchmarkReport.record(op_name, locals(), result, tag="tileops")
+    BenchmarkReport.record(op, locals(), result, tag="tileops")
```
Ibuki-wind: `_profile_fused_gated` calls `BenchmarkReport.record(..., locals())` inside the helper, so the caller geometry is gone: a direct invocation records only `baseline_key` and drops `M`, `N`, and `dtype`, making workload rows indistinguishable. → Pass an explicit serializable params dict containing `M`, `N`, and `dtype` into the helper and use it for both records.

benchmarks/ops/bench_binary_elementwise.py
```diff
@@ -389,17 +420,25 @@ def test_fused_gated_bench(
...
 
 def _strategy_params():
-    """3 ops × 3 shapes × 3 dtypes × 2 strategies = 54 rows."""
+    """Default-strategy sentinel: both strategies across the shape axis and the
+    dtype axis on one kernel.
+
+    The direct/explicit_parallel trade-off is a property of the shared fused-gated
+    kernel body, so sweeping all three ops re-measures one conclusion three times.
+    """
+    op_name, kernel_cls = _STRATEGY_KERNELS[0]
```
Ibuki-wind: `_strategy_params` always selects `_STRATEGY_KERNELS[0]`, leaving the Gelu and GeluTanh entries dead and removing all validation of their default strategy. The subclasses share the wrapper but bind different activation bodies, whose instruction and register costs can change the direct-vs-explicit result. → Keep the full shape/dtype sweep on Silu, add one reference-shape/reference-dtype direct-vs-explicit sentinel for each Gelu kernel, and iterate the list instead of indexing element zero.

tileops/manifest/reduction.yaml
```diff
@@ -585,6 +593,7 @@ ArgmaxFwdOp:
     op: tileops/ops/reduction/argreduce.py
     test: tests/ops/test_argreduce.py
     bench: benchmarks/ops/bench_argreduce.py
+    bench_manifest_driven: true
```
Ibuki-wind, line 599: Deleting `bench_reduce_multidim.py` removed the 3D `dim=0/1` argreduce performance cases, while both canonical argreduce workload sets still contain only 2D `dim=-1`; the non-last-axis path is now absent from benchmarks. → Add a minimal 3D non-last-axis sentinel to the canonical argreduce manifest workload consumed by `bench_argreduce.py` (one shared-path sentinel is sufficient if argmax/argmin share the mechanics); do not restore the fragmented file.

Ibuki-wind (changes requested):
### Overall
The cleanup is directionally sound, but the benchmark migration needs the minimal correctness and coverage fixes called out inline.

benchmarks/ops/bench_binary_elementwise.py
lcy-seso: Adopted. Each caller now passes an explicit {M, N, dtype} dict into `_profile_fused_gated`, used for both the tileops and torch-ref records; verified the three columns are back in profile_run.log. See 3d587f55.

benchmarks/ops/bench_binary_elementwise.py
lcy-seso: Adopted. `_strategy_params` iterates `_STRATEGY_KERNELS` instead of indexing element zero: the shape/dtype sweep stays on Silu, and each Gelu kernel gets one reference-shape/reference-dtype sentinel per strategy (4 added cases). See 3d587f55.

tileops/manifest/reduction.yaml
lcy-seso: Adopted. Added a 3D `dim: 0` sentinel to the canonical ArgmaxFwdOp workloads; ArgminFwdOp dispatches the same ArgreduceKernel, so one op carries it. `bench_argreduce.py` already threads dim from workload extras. The fragmented file stays deleted. See 3d587f55.

lcy-seso:
Addressed in 3d587f55. See thread replies for specifics.

### Cross-cutting
- All three findings were coverage or reporting regressions introduced by this PR's own bench consolidation, not pre-existing gaps; each fix restores the lost axis inside the manifest-driven path rather than reinstating a file.
- Same class as the multidim reduction regression fixed in 5e7aa642: a consolidation dropped an axis the op signature declares but no manifest workload exercised.

Ibuki-wind (approved):
Clean — no issues.

PR #1780, opened 2026-07-26
[Refactor][Tooling] revert the retired-ctor-params validator mechanism

lcy-seso:
Dedicated validator-infrastructure PR (per the manifest-validator domain rule) reverting a mechanism that #1778 smuggled in:

- `_CTOR_RETIRED_PARAMS` frozenset + its check loop in `check_c3_ctor_signature_parity`, and `test_retired_ctor_param_fails`.

Why revert: (1) PR-scoping violation — validator infrastructure inside an elementwise refactor; (2) speculative enforcement generalized from exactly one name; (3) triple redundancy — the no-`strategy`-kwarg guarantee is already pinned by the inspect-based signature guards `test_..._ops_do_not_expose_strategy_kwarg` / `..._kernels_...` shipped in the same PR. The legitimate 1-line `_CTOR_INFRA_PARAMS` cleanup from #1778 is kept.

Verification: validator output unchanged on the full manifest; `tests/test_validate_manifest.py` 120 passed.

### DIFF
--- a/scripts/validate_manifest.py
+++ b/scripts/validate_manifest.py
@@ -3121,15 +3121,6 @@ def check_l4_benchmark(
 # Op interface contract.
 _CTOR_INFRA_PARAMS = frozenset({"self", "kernel_map", "tune"})
 
-# Ctor parameter names whose mechanism has been removed from the codebase
-# (e.g. elementwise ``strategy``, folded into the kernel config dict). A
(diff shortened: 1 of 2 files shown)


lcy-seso:
Folded into #1779 per owner's direction to reduce PR count; the validator change remains documented in that PR's body.

PR #1783, opened 2026-07-26
[Perf][Attention] Add page-aligned Hopper bs1 paged GQA dispatch

superAngGao:
## Summary

- Add a Hopper batch-1 TMA/WGMMA fast path for paged GQA decode.
- Resolve logical KV tiles through `block_table` while keeping fast-path TMA tiles page-contained.
- Share the context-split schedule, buffers/barriers, consumer/combine math, tier policy, and partial allocation with contiguous batch-1 decode.
- Add dispatch and numerical coverage for both runtime tiers, including reversed page tables.

## Validation

Official runner image: `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`.

- `pytest -q tests/ops/attention/test_gqa_decode_paged.py tests/ops/attention/test_gqa_decode.py` -> 36 passed.
- Coverage includes no-split, context-split, reversed pages, paged dispatch/fallback contracts, and contiguous decode regression coverage.
- H200 smoke: 4K KV TileOps 16.8 us vs FlashInfer 26.4 us; 32K KV TileOps 63.4 us vs FlashInfer 58.6 us.

### DIFF
--- a/tests/ops/attention/test_gqa_decode_paged.py
+++ b/tests/ops/attention/test_gqa_decode_paged.py
@@ -238,7 +238,7 @@ def test_gqa_decode_paged_bs1_dispatch() -> None:
         pytest.param(1, torch.bfloat16, 128, 256, None, id="bf16"),
         pytest.param(1, torch.float16, 64, 256, None, id="head-dim"),
         pytest.param(1, torch.float16, 128, 16, None, id="small-page"),
-        pytest.param(1, torch.float16, 128, 96, None, id="unsupported-wgmma-tile"),
+        pytest.param(1, torch.float16, 128, 192, None, id="generic-page-tile"),
         pytest.param(1, torch.float16, 128, 256, 2.0, id="softcap"),
     ],
 )
@@ -250,20 +250,11 @@ def test_gqa_decode_paged_bs1_dispatch_fallbacks(
     softcap: float | None,
 ) -> None:
     """Unsupported shapes and features stay on the generic paged kernel."""
-    seqlen_kv = 8064 if page_size == 96 else 8192
+    seqlen_kv = 8064 if page_size == 192 else 8192
     op = GroupedQueryAttentionDecodePagedWithKVCacheFwdOp(
         batch, 32, 4, seqlen_kv, dim, page_size, dtype, softcap=softcap)
     assert not op._uses_bs1_fast_path()
     assert op.kernel.__class__.__name__ == "GQADecodePagedKernel"
-
-
-@pytest.mark.smoke
-def test_gqa_decode_paged_rejects_non_divisible_page() -> None:
-    """Reject page layouts that neither current paged kernel can address correctly."""
-    with pytest.raises(ValueError, match="page_size must be divisible"):
-        GroupedQueryAttentionDecodePagedWithKVCacheFwdOp(
-            1, 32, 4, 8256, 128, 192, torch.float16)
-
 
 if __name__ == "__main__":
     pytest.main([__file__, "-vvs"])
--- a/tests/ops/attention/test_gqa_decode_paged.py
+++ b/tests/ops/attention/test_gqa_decode_paged.py
@@ -9,6 +9,7 @@ import torch.nn.functional as F
 
 from tests.test_base import FixtureBase, TestBase
 from tileops.ops import GroupedQueryAttentionDecodePagedWithKVCacheFwdOp
+from tileops.utils import is_hopper
 from workloads.attention.gqa import (
     GroupedQueryAttentionDecodePagedTest as _GroupedQueryAttentionDecodePagedTestWorkload,
 )
@@ -73,6 +74,7 @@ class GroupedQueryAttentionDecodePagedFixture(FixtureBase):
             pytest.param(1, 8, 4, 1024, 64, 256, torch.float16, False, marks=pytest.mark.full),
             pytest.param(2, 16, 8, 512, 128, 128, torch.float16, False, marks=pytest.mark.full),
             pytest.param(1, 16, 4, 2048, 128, 512, torch.float16, False, marks=pytest.mark.full),
+            pytest.param(1, 32, 4, 4096, 128, 256, torch.float16, False, marks=pytest.mark.full),
             pytest.param(1, 32, 16, 512, 64, 128, torch.float16, False, marks=pytest.mark.full),
         ]),
     ]
@@ -166,6 +168,47 @@ def test_gqa_decode_paged_op_softmax_controls(
         softcap=softcap,
     )
     test.check(op, *test.gen_inputs(), compare=test._maxdiff_cosine_compare)
+
+
+@pytest.mark.smoke
+def test_gqa_decode_paged_bs1_dispatch() -> None:
+    """Eligible Hopper requests select the paged TMA/WGMMA kernel."""
+    if not is_hopper():
+        pytest.skip("batch=1 warp-specialized paged decode requires Hopper")
+    op = GroupedQueryAttentionDecodePagedWithKVCacheFwdOp(
+        1, 32, 4, 8192, 128, 256, torch.float16)
+    assert op._uses_bs1_fast_path()
+    assert op.kernel.__class__.__name__ == "GQADecodePagedBs1Kernel"
+    assert op.kernel._select_tier(1024) == "ctx"
+    assert op.kernel._select_tier(512) == "no_split"
+    assert op.kernel._ctx_splits_for(8192) == 32
+    assert op.kernel._ctx_splits_for(2048) == 16
+    assert op.kernel._ctx_splits_for(3072) == 8
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize(
+    "batch, dtype, dim, page_size, softcap",
+    [
+        pytest.param(2, torch.float16, 128, 256, None, id="batched"),
+        pytest.param(1, torch.bfloat16, 128, 256, None, id="bf16"),
+        pytest.param(1, torch.float16, 64, 256, None, id="head-dim"),
+        pytest.param(1, torch.float16, 128, 16, None, id="small-page"),
+        pytest.param(1, torch.float16, 128, 256, 2.0, id="softcap"),
+    ],
+)
+def test_gqa_decode_paged_bs1_dispatch_fallbacks(
+    batch: int,
+    dtype: torch.dtype,
+    dim: int,
+    page_size: int,
+    softcap: float | None,
+) -> None:
+    """Unsupported shapes and features stay on the generic paged kernel."""
+    op = GroupedQueryAttentionDecodePagedWithKVCacheFwdOp(
+        batch, 32, 4, 8192, dim, page_size, dtype, softcap=softcap)
+    assert not op._uses_bs1_fast_path()
+    assert op.kernel.__class__.__name__ == "GQADecodePagedKernel"
 
 
 if __name__ == "__main__":
--- a/tests/ops/attention/test_gqa_decode_paged.py
+++ b/tests/ops/attention/test_gqa_decode_paged.py
@@ -74,7 +74,6 @@ class GroupedQueryAttentionDecodePagedFixture(FixtureBase):
             pytest.param(1, 8, 4, 1024, 64, 256, torch.float16, False, marks=pytest.mark.full),
             pytest.param(2, 16, 8, 512, 128, 128, torch.float16, False, marks=pytest.mark.full),
             pytest.param(1, 16, 4, 2048, 128, 512, torch.float16, False, marks=pytest.mark.full),
-            pytest.param(1, 32, 4, 4096, 128, 256, torch.float16, False, marks=pytest.mark.full),
             pytest.param(1, 32, 16, 512, 64, 128, torch.float16, False, marks=pytest.mark.full),
         ]),
     ]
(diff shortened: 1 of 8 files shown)


tests/ops/attention/test_gqa_decode_paged.py
```diff
@@ -73,6 +74,7 @@ class GroupedQueryAttentionDecodePagedFixture(FixtureBase):
             pytest.param(1, 8, 4, 1024, 64, 256, torch.float16, False, marks=pytest.mark.full),
             pytest.param(2, 16, 8, 512, 128, 128, torch.float16, False, marks=pytest.mark.full),
             pytest.param(1, 16, 4, 2048, 128, 512, torch.float16, False, marks=pytest.mark.full),
+            pytest.param(1, 32, 4, 4096, 128, 256, torch.float16, False, marks=pytest.mark.full),
```
Ibuki-wind: - This is the only case that runs `GQADecodePagedBs1Kernel` numerically, and it does so nondeterministically:
  `gen_inputs()` draws `real_seqlen_kv` uniformly over `{256, 512, ..., 4096}`, and 6 of those 16 values fall
  below `_MIN_CTX = 1024` and take the `no_split` fallback instead of the new ctx kernel.
- The row is `pytest.mark.full` and the PR gate runs `-m smoke`, so on this PR the new kernel's output is never
  compared to the reference; the three added smoke tests only assert predicate return values.
- Add a smoke-marked correctness test with `real_seqlen_kv` fixed above `_MIN_CTX` (e.g. `fill_(4096)`) and one
  below (e.g. `fill_(512)`), so both branches are checked numerically every run.

tests/ops/attention/test_gqa_decode_paged.py
```diff
@@ -139,5 +141,54 @@ def test_gqa_decode_paged_op_softmax_controls(
     test.check(op, *test.gen_inputs(), compare=test._maxdiff_cosine_compare)
 
 
+@pytest.mark.smoke
+def test_gqa_decode_paged_bs1_dispatch() -> None:
```
Ibuki-wind, line 218: - The page translation this PR introduces —
  `physical_block = block_table[bid, page_idx] * blocks_per_page + block_in_page` — is never exercised on an
  output-distinguishing input.
- `gen_inputs()` builds an identity `block_table` (`arange(num_pages)`), so `physical_block == logical_block`
  everywhere. A kernel that ignored `block_table` and read K/V linearly would pass the whole suite.
- #1787 uses `arange(...).flip(0)` for exactly this reason — add a case with a reversed or shuffled page table
  on a Hopper fast-path shape.

tileops/manifest/attention.yaml
```diff
@@ -518,6 +518,9 @@ GroupedQueryAttentionDecodePagedWithKVCacheFwdOp:
       - "H % H_kv == 0"
 
   workloads:
+    # Qwen3-30B-A3B batch=1 Hopper fast path.
+    - {q_shape: [1, 32, 128], kv_shape: [4096, 4, 128], page_size: 256, dtypes: [float16], label: "qwen3-30b-a3b-bs1-4k-p256"}
```
Ibuki-wind: - Non-blocking: these two `workloads` rows fall outside the manifest carve-out for an implementation PR
  (the op is already `status: implemented`), so strictly they'd land in a manifest-only PR.
- Not worth holding a human-authored perf PR over — noting it only so the split isn't a surprise later. The
  `source.kernel_map` addition below is explicitly permitted.

tileops/ops/attention/gqa.py
```diff
@@ -1484,7 +1490,28 @@ def __init__(self,
...
+
+    def _uses_bs1_fast_path(self) -> bool:
+        if not (self.batch == 1 and is_hopper() and self.dtype == torch.float16
+                and self.dim == 128 and self.softcap == 0.0):
+            return False
+        if self.heads % self.heads_kv != 0:
+            return False
+        # The TMA producer must load a complete WGMMA N tile without crossing
+        # a potentially non-contiguous page boundary.
+        block_n = min(128, self.page_size)
```
Ibuki-wind: - Non-blocking: the gate admits page sizes whose derived tile width no test compiles. `block_n =
  min(128, page_size)` with `page_size % block_n == 0` accepts any `page_size` in `[64, 128)`.
- `page_size=96` passes all five conditions and drives `default_config` to `block_N=96` — a WGMMA N tile
  outside the `{64, 128}` set every test and bench uses — and changes which `_CTX_SPLIT_CANDIDATES` entry
  `_ctx_splits_for` picks.
- Restrict to `page_size % 128 == 0 or page_size == 64` so the admitted region equals the tested region.

Ibuki-wind (changes requested):
### Overall
Top risk: a 403-line new kernel with a brand-new page-address translation lands with effectively zero numerical coverage of that translation. Fix the two coverage gaps below and move the `workloads:` rows to a manifest-only PR; the kernel body itself (exp2-domain online softmax, `glse = log2(logsum) + sm*scale`, and the `_combine` reweighting `exp2(glse[s] - lse_logsum)`) reconciles correctly with `acc_o /= logsum`, and the `_split` length redistribution covers `[0, real_seqlen_kv)` exactly for every `ctx_splits` the selector can return.

### Cross-cutting concerns
- Merge order with #1787 is load-bearing and currently unstated in a machine-checkable place. The `page_size` guard added at `gqa.py:1470` rejects at construction for **both** the fast path and the generic path, so after #1787 makes generic decode safe for `page_size=192`, leaving this guard in place turns a fixed layout back into a hard error. Whichever PR lands second must delete the guard in the same commit — a `FIXME(staged-rollout)` block per `.claude/rules/code-style.md` at `gqa.py:1470` would make that discoverable by `grep` instead of by reading a PR description.

tileops/kernels/attention/gqa_decode_bs1_paged.py
```diff
@@ -0,0 +1,403 @@
...
+    "-U__CUDA_NO_HALF2_OPERATORS__",
+    "-U__CUDA_NO_BFLOAT16_CONVERSIONS__",
+    "--expt-relaxed-constexpr",
+    "--expt-extended-lambda",
+    "-DNDEBUG",
+]
+
+
+@functools.lru_cache(maxsize=32)
+def _gqa_decode_paged_bs1_ctx_kernel(
```
Ibuki-wind, line 37: - This is a fork of `gqa_decode_bs1.py` (281 lines), not a new kernel: 179 lines are identical after
  whitespace normalisation.
- The real delta is ~18 lines of page-table translation in the TMA producer, plus `real_seqlen_kv` becoming
  a tensor and KV losing its batch axis.
- Prefer an optional `page_size` / `block_table` on the existing bs1 kernel, or hoist the shared consumer
  and `_combine` into macros both files use.
- The fork was also re-wrapped in a different style from the original (trailing commas, re-indented calls),
  which inflates the diff and hides how small the real change is. If the fork stays, keeping the original
  formatting makes it reviewable.

tileops/kernels/attention/gqa_decode_bs1_paged.py
```diff
@@ -0,0 +1,403 @@
...
+                        if i < kv_group_num:
+                            glse[bid, hid * kv_group_num + i, sid] = (
+                                T.log2(logsum[i]) + sm[i] * scale
+                            )
+                    for i, j in T.Parallel(block_M, dim):
+                        if i < kv_group_num:
+                            Output_partial[bid, hid * kv_group_num + i, sid, j] = acc_o[i, j]
+
+        @T.macro
+        def _combine(
```
Ibuki-wind: - `_combine` is byte-identical to `gqa_decode_bs1.py`'s modulo whitespace (27 lines).
- It's paging-agnostic — import it instead of copying.

tileops/ops/attention/gqa.py
```diff
@@ -1484,7 +1490,28 @@ def __init__(self,
 
     @property
     def default_kernel_map(self) -> Dict[str, Kernel]:
-        return {"gqa_decode_paged_kernel": GQADecodePagedKernel}
+        kernel_map: Dict[str, Kernel] = {"gqa_decode_paged_kernel": GQADecodePagedKernel}
+        if is_hopper():
+            kernel_map["gqa_decode_paged_bs1_kernel"] = GQADecodePagedBs1Kernel
+        return kernel_map
+
+    def _uses_bs1_fast_path(self) -> bool:
```
Ibuki-wind, line 1519: - `_uses_bs1_fast_path` and `_select_decode_kernel_key` are copied from the non-paged Op at L1413 with four
  lines of page check inserted.
- The shared predicate (batch / Hopper / fp16 / dim / softcap / group size) should be one module-level
  helper used by both Ops.
- The copy also dropped the original's docstring.

tileops/ops/attention/gqa.py
```diff
@@ -1464,12 +1465,17 @@ def __init__(self,
         self.seqlen_kv = seqlen_kv
         self.dim = dim
         self.page_size = page_size
+        if page_size <= 0:
+            raise ValueError("page_size must be positive")
+        block_n = min(128, page_size)
+        if page_size % block_n != 0:
+            raise ValueError("page_size must be divisible by the paged decode block_N")
```
Ibuki-wind: - Third place encoding `min(128, page_size)` — with `_uses_bs1_fast_path` below and
  `GQADecodePagedKernel.default_config`. Same helper should serve all three.
- This also rejects `page_size=192`, which #1787 adds a passing test for.

Ibuki-wind (commented):
Cleanliness / bloat pass. The new kernel is largely a fork of `gqa_decode_bs1.py` rather than a
parameterisation of it, and the dispatch predicate is a fork of the non-paged Op's. Details inline.

Also: the ctor check added here raises for `page_size=192`, which #1787 adds a passing test for — the two
PRs contradict each other.

tests/ops/attention/test_gqa_decode_paged.py
superAngGao: Fixed in `dc2d8d66`. The smoke suite now fixes `real_seqlen_kv` at 512 and 4096 and compares both runtime tiers numerically against the reference on every run.

tests/ops/attention/test_gqa_decode_paged.py
superAngGao: Fixed in `dc2d8d66`. The 4096-token ctx case reverses `block_table`, so a kernel that ignores logical-to-physical page translation now produces a different output and fails correctness.

tileops/manifest/attention.yaml
superAngGao: Fixed in `dc2d8d66`. I removed the two Qwen3 workload rows; this implementation PR now changes only the permitted `source.kernel_map` entry in the manifest.

tileops/ops/attention/gqa.py
superAngGao: Fixed across `0df6e9cb`, `dc2d8d66`, and `140f10ab`. The fast path now accepts only `page_size == 64` or multiples of 128; page size 96 has an explicit generic-fallback test. Tile selection has one shared source of truth.

tileops/kernels/attention/gqa_decode_bs1_paged.py
superAngGao: Fixed in `0df6e9cb`. The paging-independent WGMMA consumer, split-output combine, compile flags, and ring depth now live in `gqa_decode_bs1_common.py`; the paged file retains the page-table-aware TMA producer while the non-paged producer remains unchanged.

tileops/kernels/attention/gqa_decode_bs1_paged.py
superAngGao: Fixed in `0df6e9cb`. Both kernels now use the same `make_gqa_decode_bs1_combine` macro instead of carrying separate copies.

tileops/ops/attention/gqa.py
superAngGao: Fixed in `0df6e9cb`. The common Hopper/batch/dtype/dim/softcap/group contract is now `_supports_gqa_decode_bs1`, used by both Ops. The paged method adds only its documented page-contained TMA-tile condition.

tileops/ops/attention/gqa.py
superAngGao: Fixed in `140f10ab`. Generic config, constructor validation, and bs1 eligibility now derive the N tile from `gqa_decode_paged_block_n`. Because #1787 is still open, the temporary 192-page constructor rejection remains, now with the required `FIXME(staged-rollout)` invariant/why/cleanup block so the second landing removes it.

superAngGao:
Addressed the requested changes in three focused commits:

- `0df6e9cb`: share the paging-independent Hopper bs1 consumer/combine machinery and common dispatch predicate; add the staged-rollout guard metadata.
- `dc2d8d66`: add deterministic 512/4096 numerical smoke coverage, reverse the long-case page table, test page-size 96 fallback, and remove the two manifest workload rows.
- `140f10ab`: centralize paged `block_N` selection across generic config, constructor validation, and bs1 eligibility.

Validation in `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10` on GPU4/H200: paged plus non-paged decode tests are `33 passed`; pre-commit passes on all touched files. I also updated the PR body with the exact test-node delta and current #1787 landing dependency.

Ibuki-wind (approved):
Clean — no issues.

tileops/kernels/attention/gqa_decode_bs1_paged.py
```diff
@@ -0,0 +1,372 @@
...
+    RING_DEPTH,
+    make_gqa_decode_bs1_combine,
+    make_gqa_decode_bs1_consumer,
+)
+
+__all__ = ["GQADecodePagedBs1Kernel"]
+
+
+@functools.lru_cache(maxsize=32)
+def _gqa_decode_paged_bs1_ctx_kernel(
```
Ibuki-wind, line 37: The paged specialization is valid; the duplicated 372-line implementation is not.

- **Keep separate:** paged/contiguous `prim_func` signatures and custom-op schemas.
- **Share:** scheduling, buffers/barriers, consumer/combine math, tier/split selection, and partial-output allocation.
- **Paged file owns:** page validation, block-N eligibility, logical→physical block lookup, paged schema, and thin wrapper.
- **Implementation:** put the shared builder in `gqa_decode_bs1_common.py`; inject only the K/V address resolver.
- **Do not:** add a runtime layout branch in TIR.
- **Done when:** the paged file no longer copies context setup/consumer/combine logic, and the 512/4096 plus reversed-page tests still pass.

Ibuki-wind (changes requested):
### Overall
The Hopper paged specialization is justified; the remaining duplicated source implementation is not. Consolidate it according to the inline plan before merge.

tileops/kernels/attention/gqa_decode_bs1_paged.py
superAngGao: Addressed in 829178aa after rebasing onto #1787 / db4494c5.

The refactor now follows the ownership split in this review:

- `gqa_decode_bs1_common.py` owns the context-split builder, shared-memory buffers/layouts, barriers, producer/consumer schedule, WGMMA consumer and combine math, tier/split policy, and fp32 partial allocation.
- `gqa_decode_bs1.py` injects only the contiguous K/V loader and keeps its contiguous custom-op schema.
- `gqa_decode_bs1_paged.py` injects only logical-page to physical-page K/V address resolution and keeps page eligibility, paged schema, and the thin wrapper.
- There is no runtime layout branch in TIR; the two loaders are bound while each kernel is built.

I also removed the staged page-layout guard now that #1787 is merged. `page_size=192` falls back to generic paged decode with its 64-token page-contained tile, while the bs1 WGMMA path remains limited to page 64 or block-N 128 layouts.

Validation in the official runner image: `36 passed` for `test_gqa_decode_paged.py` plus `test_gqa_decode.py`, including 512 no-split, 4096 context split with reversed pages, page-192 generic fallback, and contiguous regressions.

Ibuki-wind (approved):

PR #1787, opened 2026-07-27
[Perf][Attention] Pipeline KV loads and fix page tiling in generic paged GQA decode

superAngGao:
## Summary

- issue both K and V shared-memory loads before the QK GEMM in generic paged GQA decode, for both no-split and split paths
- derive default, autotune, and custom-config `block_N` legality from one page-coverage helper
- add output-correctness coverage for a 3072-token split request with a reversed page table, plus construction-time rejection coverage for unsupported page layouts

## Correctness

The old address translation used `num_blockn_in_page = page_size // block_N`, so `page_size=192` with `block_N=128` skipped the last 64 tokens of every physical page. The generic default now selects the widest supported tile that divides the page exactly: `block_N=64` for `page_size=192`.

Layouts such as page sizes 96, 160, and 224 match neither supported tile width and now fail explicitly before JIT instead of silently dropping page tails. The same supported-tile set drives default config, autotune candidates, and custom-config validation.

This is the generic-path prerequisite for #1783. Landing order should be #1787 first, then rebase #1783 and remove its temporary `FIXME(staged-rollout)` constructor rejection in the same update.

## H200 benchmark

Image: `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`; 300 warmup iterations, 13 samples × 500 iterations, median latency.

| Shape / path | Before | After | Improvement |
|---|---:|---:|---:|
| B=8, page=256, real KV=512 (no-split) | 68.07 us | 63.27 us | 7.1% |
| B=8, page=256, real KV=4096 (split) | 192.76 us | 180.14 us | 6.5% |
| B=32, page=64, real KV=4096 (split) | 276.93 us | 266.77 us | 3.7% |

## Testing

Official runner image, GPU4/H200, repository source plus CI caches only; no dependency installation or upgrade:

```bash
python -m pytest -q tests/ops/attention/test_gqa_decode_paged.py
```

Result: `13 passed`.

The reversed `page_size=192` case checks the split path numerically. Separate smoke nodes assert that page sizes 96, 160, and 224 fail construction because no supported N tile covers each page exactly. `pre-commit` passes on both touched files.

### DIFF
--- a/tests/ops/attention/test_gqa_decode_paged.py
+++ b/tests/ops/attention/test_gqa_decode_paged.py
@@ -104,6 +104,24 @@ def test_gqa_decode_paged_op(
 
 
 @pytest.mark.smoke
+def test_gqa_decode_paged_non_divisible_128_page_split() -> None:
+    """page_size=192 uses 64-token tiles without skipping page tails."""
+    batch, heads, heads_kv, seqlen_kv, dim, page_size = 1, 16, 4, 3072, 128, 192
+    test = GroupedQueryAttentionDecodePagedTest(
+        batch, heads, heads_kv, seqlen_kv, dim, page_size, torch.float16)
+    q, k, v, real_seqlen_kv, block_table = test.gen_inputs()
+    real_seqlen_kv.fill_(seqlen_kv)
+    block_table.copy_(torch.arange(
+        seqlen_kv // page_size, device="cuda", dtype=torch.int32).flip(0).unsqueeze(0))
+    op = GroupedQueryAttentionDecodePagedWithKVCacheFwdOp(
+        batch, heads, heads_kv, seqlen_kv, dim, page_size, torch.float16)
+    assert op.kernel.config["block_N"] == 64
+    test.check(
+        op, q, k, v, real_seqlen_kv, block_table,
+        compare=test._maxdiff_cosine_compare)
+
+
+@pytest.mark.smoke
 @pytest.mark.parametrize("sm_scale, softcap", [
     pytest.param(0.25, None, id="custom-sm-scale"),
     pytest.param(None, 2.0, id="softcap"),
--- a/tests/ops/attention/test_gqa_decode_paged.py
+++ b/tests/ops/attention/test_gqa_decode_paged.py
@@ -115,10 +115,21 @@ def test_gqa_decode_paged_non_divisible_128_page_split() -> None:
         seqlen_kv // page_size, device="cuda", dtype=torch.int32).flip(0).unsqueeze(0))
     op = GroupedQueryAttentionDecodePagedWithKVCacheFwdOp(
         batch, heads, heads_kv, seqlen_kv, dim, page_size, torch.float16)
-    assert op.kernel.config["block_N"] == 64
+    assert page_size % op.kernel.config["block_N"] == 0
+    assert {config["block_N"] for config in op.kernel.autotune_configs} == {64}
     test.check(
         op, q, k, v, real_seqlen_kv, block_table,
         compare=test._maxdiff_cosine_compare)
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize("page_size", [96, 160, 224])
+def test_gqa_decode_paged_rejects_unsupported_page_tile(page_size: int) -> None:
+    """Reject page layouts that no supported generic N tile can cover exactly."""
+    seqlen_kv = page_size * 16
+    with pytest.raises(ValueError, match="matches no supported block_N"):
+        GroupedQueryAttentionDecodePagedWithKVCacheFwdOp(
+            1, 16, 4, seqlen_kv, 128, page_size, torch.float16)
 
 
 @pytest.mark.smoke
--- a/tileops/kernels/attention/gqa_decode_paged.py
+++ b/tileops/kernels/attention/gqa_decode_paged.py
@@ -16,6 +16,29 @@ from tileops.kernels.online_softmax import (
 
 __all__ = ["GQADecodePagedKernel"]
 
+
+def gqa_decode_paged_block_ns(page_size: int) -> tuple[int, ...]:
+    """Return page-contained N tiles supported by generic paged decode."""
+    if page_size <= 0:
+        raise ValueError("page_size must be positive")
+
+    block_ns = tuple(
+        block_n
+        for block_n in (128, 64)
+        if block_n <= page_size and page_size % block_n == 0
+    )
+    if block_ns:
+        return block_ns
+    if page_size < 64:
+        return (page_size,)
+    raise ValueError(f"page_size={page_size} matches no supported block_N")
+
+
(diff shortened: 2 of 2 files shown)


tileops/kernels/attention/gqa_decode_paged.py
```diff
@@ -459,8 +462,11 @@ def autotune_supply_prog(self):
 
     @property
     def default_config(self) -> dict:
-        # block_N must be <= page_size so num_blockn_in_page = page_size // block_N >= 1 (no div by zero)
+        # Each KV tile must stay within one physical page. Prefer the widest
+        # supported tile that divides the page exactly.
         block_N = min(128, self.page_size)
+        if self.page_size >= 64 and self.page_size % block_N != 0:
+            block_N = 64
```
Ibuki-wind: - The fallback to 64 is never checked against `page_size`, so the invariant in the comment above doesn't hold.
- `page_size=160`: `min(128,160)=128`, `160%128!=0` → 64, but `160%64=32` too. `num_blockn_in_page=160//64=2`
  (L85/L223), so tokens 128..159 of every page are never addressed — the same silent drop this PR fixes for
  192, still live for 160 / 224 / 288 / 352.
- Select explicitly and reject when nothing fits, so a wrong-answer layout fails at construction:

```python
block_N = next(
    (c for c in (128, 64) if c <= self.page_size and self.page_size % c == 0),
    self.page_size if self.page_size < 64 else None,
)
if block_N is None:
    raise ValueError(f"page_size={self.page_size} matches no supported block_N")
```

tileops/kernels/attention/gqa_decode_paged.py
```diff
@@ -478,7 +484,8 @@ def autotune_configs(self) -> list[dict]:
             'num_split': c[2],
             'num_stages': c[3],
             'threads': c[4]
-        } for c in _configs if c[0] <= self.page_size]
+        } for c in _configs
+                   if c[0] <= self.page_size and self.page_size % c[0] == 0]
```
Ibuki-wind: - The filter is right but can empty the list, and nothing guards that: `page_size=96` excludes 128 (`>96`) and
  64 (`96%64!=0`), so `configs == []` and `tune=True` autotunes over zero candidates.
- The same 96 drives `default_config` (L467) to `block_N=96`, a width outside the supported set that no test
  compiles.
- Derive both this filter and `default_config` from one `_supported_block_n()` helper so they can't disagree
  about which page sizes are legal.

tests/ops/attention/test_gqa_decode_paged.py
```diff
@@ -103,6 +103,24 @@ def test_gqa_decode_paged_op(
     test.check(op, *test.gen_inputs(), compare=test._maxdiff_cosine_compare)
 
 
+@pytest.mark.smoke
```
Ibuki-wind, line 106: - `page_size=192` is the only layout covered, and it's the one the L469 fallback happens to get right — a
  `page_size=160` regression passes this suite unchanged.
- The reversed `block_table` and the `real_seqlen_kv.fill_` are the right shape; the axis is too narrow.
- Parametrize `page_size` over `[192, 160]`, and replace `assert ... == 64` with
  `assert page_size % op.kernel.config["block_N"] == 0` so the test states the invariant, not one instance.

Ibuki-wind (changes requested):
### Overall
Top risk: the page-divisibility invariant is asserted in the comment but not enforced by the code — `page_size=160` still reproduces the exact tail-skip this PR fixes for 192. Make `default_config` select from the supported tile set and fail loudly when nothing divides the page, then extend the new test beyond the single 192 case.

The K/V producer-stage reorder in both `gqa_decode_no_split` and `_gqa_decode_split` is correct: `V_shared` is written and consumed inside the same `T.Pipelined` iteration, so hoisting the copy next to the `K` copy only changes prefetch scheduling, not the data read by `T.gemm(acc_s_cast, V_shared, acc_o)`.

Ibuki-wind (commented):
Cleanliness pass only. The pipelining change is fine; two things on how the tiling rule is expressed.

Heads-up: #1783 adds a ctor check that raises for `page_size=192` — the exact case this PR adds a passing
test for. The two PRs contradict each other; whichever lands second breaks the other.

tileops/kernels/attention/gqa_decode_paged.py
```diff
@@ -93,6 +93,11 @@ def gqa_decode_no_split(
                     T.copy(
                         K[blockn_num_offset * block_N:(blockn_num_offset + 1) * block_N,
                           cur_kv_head, :], K_shared)
+                    # Match prefill's pipeline: issue both KV loads before QK
```
Ibuki-wind: - Two different comments for the identical edit (here and L236). Same wording in both, or drop them.

tileops/kernels/attention/gqa_decode_paged.py
superAngGao: Fixed in `5987efab`. Generic paged decode now selects only page-contained tiles from `{128, 64}` (with the existing exact-width path for pages below 64). Page sizes 160, 224, and similar layouts fail at construction instead of falling back to an unsafe 64-token tile.

tileops/kernels/attention/gqa_decode_paged.py
superAngGao: Fixed in `5987efab`. `gqa_decode_paged_block_ns()` is now the single source for default config, autotune candidates, and custom-config validation. Unsupported page size 96 fails before autotuning, so an empty candidate sweep is no longer possible.

tests/ops/attention/test_gqa_decode_paged.py
superAngGao: Fixed in `5987efab`. The page-192 reversed-table numerical test now asserts the divisibility invariant and verifies that autotune exposes only block_N=64. Page 160 has no supported exact-cover tile, so it is covered by a construction-time rejection test alongside 96 and 224 rather than by an unsafe numerical path.

tileops/kernels/attention/gqa_decode_paged.py
superAngGao: Fixed in `5987efab`. Both no-split and split loops now use the same comment: issue both K/V loads before QK so copies can overlap the GEMMs.

superAngGao:
Addressed all open review threads in `5987efab`:

- one supported-tile helper now governs default, autotune, and custom configs;
- page 192 uses an exact 64-token tile and retains reversed-page numerical coverage;
- page sizes 96, 160, and 224 fail loudly before JIT;
- the duplicated pipeline comments are now identical.

Validation in `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10` on GPU4/H200: `13 passed`; pre-commit passes. The PR body now states the landing order explicitly: merge #1787 first, then rebase #1783 and remove its staged constructor guard.

Ibuki-wind (approved):
Clean — no issues.

PR #1789, opened 2026-07-27
[Refactor] Move mean pooling to pool family

superAngGao:
## Problem

`MeanPoolingForwardOp` sits under the attention family, but it computes no attention: it chunks a token sequence and mean-reduces each chunk. NSA/GQA merely consume its output. Path-based family inference therefore misclassifies it as attention.

## Solution

Move the whole MeanPooling surface from attention to pool:

- `MeanPoolingFwdKernel`: `tileops/kernels/attention` → `tileops/kernels/pool`
- `MeanPoolingForwardOp`: `attention/deepseek_nsa.py` → `tileops/ops/pool.py`
- workload / test / benchmark moved out of their attention directories
- exports updated so `from tileops.ops import MeanPoolingForwardOp` and `from tileops.kernels import MeanPoolingFwdKernel` are unchanged

Kernel implementation, constructor, and runtime behavior are untouched — this is family ownership and file layout only.

## Validation

- `python -m compileall` over the moved files; `git diff --check`
- grepped for stale `workloads.attention.mean_pooling` / `deepseek_nsa_mean_pooling_fwd` references

pytest/ruff/CUDA tests not run locally: the available env lacks `torch`, `pytest`, `ruff`.


tileops/ops/pool.py
```diff
@@ -36,9 +37,47 @@
     "MaxPool2dIndicesFwdOp",
     "MaxPool3dFwdOp",
     "MaxPool3dIndicesFwdOp",
+    "MeanPoolingForwardOp",
 ]
 
 
+class MeanPoolingForwardOp(Op):
```
Ibuki-wind, line 44: Advisory (non-blocking): `pool.py` is laid out as imports -> `__all__` -> module-level helpers and constants (`_device_index`, `_POOL_LAYOUTS`, `_POOL_DIM_NAMES`, `_validate_pool_input_dtypes`) -> op classes. Inserting `MeanPoolingForwardOp` between `__all__` and that helper block splits the shared-helper section away from the classes that use it. → move the class below `_validate_pool_input_dtypes`, alongside `_AvgPoolFwdOpBase` and the other op classes.

tileops/ops/pool.py
```diff
@@ -36,9 +37,47 @@
...
+
+    def __init__(
+        self,
+        batch_size: int,
+        seq_len: int,
+        heads: int,
+        dim: int,
```
Ibuki-wind, line 51: Advisory (non-blocking): two naming defects come across verbatim with the move — the ctor parameter is misspelled `chunks_per_bacth` (should be `chunks_per_batch`), and the class is `MeanPoolingForwardOp` while every other op in this file is `*FwdOp` (`AvgPool1dFwdOp`, `MaxPool3dIndicesFwdOp`). Both are cheap to correct in this PR specifically because there is no `tileops/manifest/` entry pinning either name; once one is added, the spelling becomes contractual and the fix needs a manifest-only PR. → rename here, or state in the PR body that both are deliberately deferred.

Ibuki-wind (approved):
### Overall
Pure relocation — verified at head `6133067`: no surviving reference to `deepseek_nsa_mean_pooling_fwd`, `workloads.attention.mean_pooling`, or `tileops.kernels.attention.MeanPoolingFwdKernel`; `tileops/ops/__init__.py` and `tileops/kernels/__init__.py` keep both public symbols importable from the same top-level paths; `tileops/manifest/` has no mean-pooling entry, so no `source.kernel` / `source.op` pointer goes stale. Approving; two advisory nits inline.

### Cross-cutting concerns
- The diff spans implementation, test/`workloads/`, and benchmark layers at once, which `refactor.md §Trust-layer scope` normally splits. Justified here: a family rename is not expressible in one layer. Judged against `pre-review.md` — oracle origin unchanged, coverage set unchanged (moves only, no case added or removed), no new code path — so the cross-layer touch passes.

PR #1790, opened 2026-07-27
[Fix][Bench] Vectorize DeltaNet WY reference programs

### LINKED ISSUE #1786: [BUG][BENCH] deltanet reference programs exceed 30 min under profiler (micro-op storm)
### Symptom / Motivation

`bench_deltanet.py` and `bench_gated_deltanet.py` exceed a 1800 s per-file budget on H200. py-spy stacks show the process alive inside `torch::profiler::impl::ThreadLocalSubqueue::begin_op` while executing the chunked WY-representation reference (`prepare_wy_repr_deltanet_torch`): the reference issues millions of tiny view ops (`slice`/`as_strided`), each paying RecordFunction profiler overhead.

### Root Cause Analysis

The torch reference baseline runs inside the profiled timing loop; per-op profiler bookkeeping dominates and scales with the huge micro-op count of the chunked loop. Slow, not hung.

### Related Files

- benchmarks/ops/bench_deltanet.py, benchmarks/ops/bench_gated_deltanet.py (reference programs)

## Goal

Both bench files complete in minutes: vectorize the reference's chunk loop (or reduce baseline repeats) so the profiled micro-op count drops by orders of magnitude, without changing what the baseline computes.

## Plan

1. Vectorize/batch the WY-representation chunk loop in the reference programs.
2. Verify baseline numerics unchanged and file wall time < 5 min per file.

## Constraints

- Baseline semantics unchanged; measurement methodology unchanged.

## Acceptance Criteria

- [ ] Both files complete under 300 s on H200 with unchanged baseline results

superAngGao:
## Summary

- batch the DeltaNet WY-representation reference over batch, head, and chunk: the Python `(batch, head, chunk)` loop becomes batched Gram products + batched matrix inversion
- same vectorization for gated DeltaNet, preserving its gate-difference orientation and independent `Aw` / `Au` outputs

The profiler now sees a constant-size op sequence instead of millions of tiny `slice` / `as_strided` ops as sequence length grows.

## Numerical validation

Vectorized helpers vs. the original loop, on the benchmark workload distribution:

| Reference | WY max diff | Final output max diff |
|---|---:|---:|
| DeltaNet | 3.73e-8 | 2.98e-8 |
| Gated DeltaNet | 1.86e-8 | 1.86e-9 |

## H200 wall time

Image `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`, standard cache mounts, no dependency changes.

| Benchmark file | Result | Pytest time | Container wall time |
|---|---:|---:|---:|
| `bench_deltanet.py` | 19 passed | 17.36 s | 23.98 s |
| `bench_gated_deltanet.py` | 19 passed | 21.13 s | 31.02 s |
| both, final code | 38 passed | 32.45 s | 41.28 s |

Well under the 300 s threshold; profiler methodology unchanged.

Fixes #1786

### DIFF
--- a/benchmarks/ops/bench_deltanet.py
+++ b/benchmarks/ops/bench_deltanet.py
@@ -115,23 +115,15 @@ def prepare_wy_repr_deltanet_torch(k, beta, chunk_size):
     B, H, S, DK = k.shape
     assert S % chunk_size == 0
     BC = chunk_size
-    Aw = torch.empty(B, H, S, BC, dtype=torch.float32, device=k.device)
-    Au = torch.empty(B, H, S, BC, dtype=torch.float32, device=k.device)
-
-    for b in range(B):
-        for h in range(H):
-            for c in range(S // BC):
-                i0, i1 = c * BC, (c + 1) * BC
-                kc = k[b, h, i0:i1, :].float()
-                bc = beta[b, h, i0:i1].float()
-                Gram = kc @ kc.T
-                M = bc.unsqueeze(-1) * Gram
-                A = torch.eye(BC, device=k.device) + torch.tril(M, diagonal=-1)
-                A_inv = torch.linalg.inv(A)
-                Aw[b, h, i0:i1, :] = A_inv
-                Au[b, h, i0:i1, :] = A_inv
-
-    return Aw, Au
+    NC = S // BC
+    kc = k.float().reshape(B, H, NC, BC, DK)
+    bc = beta.float().reshape(B, H, NC, BC)
+    gram = kc @ kc.transpose(-2, -1)
+    m = bc.unsqueeze(-1) * gram
+    eye = torch.eye(BC, dtype=torch.float32, device=k.device)
+    a = eye + torch.tril(m, diagonal=-1)
+    a_inv = torch.linalg.inv(a).reshape(B, H, S, BC)
+    return a_inv, a_inv.clone()
 
 
 class _DeltaNetFwdTestBaseline(DeltaNetFwdTest):
--- a/benchmarks/ops/bench_gated_deltanet.py
+++ b/benchmarks/ops/bench_gated_deltanet.py
@@ -130,25 +130,17 @@ def prepare_wy_repr_gated_torch(k, g_cum, beta, chunk_size):
     B, H, S, DK = k.shape
     assert S % chunk_size == 0
     BC = chunk_size
-    Aw = torch.empty(B, H, S, BC, dtype=torch.float32, device=k.device)
-    Au = torch.empty(B, H, S, BC, dtype=torch.float32, device=k.device)
-
-    for b in range(B):
-        for h in range(H):
-            for c in range(S // BC):
-                i0, i1 = c * BC, (c + 1) * BC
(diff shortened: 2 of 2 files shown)


benchmarks/ops/bench_deltanet.py
```diff
@@ -115,23 +115,15 @@ def prepare_wy_repr_deltanet_torch(k, beta, chunk_size):
...
-    return Aw, Au
+    NC = S // BC
+    kc = k.float().reshape(B, H, NC, BC, DK)
+    bc = beta.float().reshape(B, H, NC, BC)
+    gram = kc @ kc.transpose(-2, -1)
+    m = bc.unsqueeze(-1) * gram
+    eye = torch.eye(BC, dtype=torch.float32, device=k.device)
+    a = eye + torch.tril(m, diagonal=-1)
+    a_inv = torch.linalg.inv(a).reshape(B, H, S, BC)
+    return a_inv, a_inv.clone()
```
Ibuki-wind, line 126: Advisory (non-blocking): the only consumer of these two returns is `compute_w_u_torch`, which reshapes and einsums them read-only — nothing mutates `Aw` or `Au`. The `.clone()` therefore allocates and copies a full `B*H*S*BC` fp32 tensor (67 MB at the nightly `S=32768, BC=64` case) on every `ref_program` call inside the profiling loop, i.e. on the exact path this PR speeds up. → return `a_inv, a_inv` unless a consumer is expected to write in place.

benchmarks/ops/bench_gated_deltanet.py
```diff
@@ -130,25 +130,17 @@ def prepare_wy_repr_gated_torch(k, g_cum, beta, chunk_size):
...
+    kc = k.float().reshape(B, H, NC, BC, DK)
+    gc = g_cum.float().reshape(B, H, NC, BC)
+    bc = beta.float().reshape(B, H, NC, BC)
+    gram = kc @ kc.transpose(-2, -1)
+    gamma = torch.exp(gc.unsqueeze(-1) - gc.unsqueeze(-2))
+    m = bc.unsqueeze(-1) * (gamma * gram)
+    eye = torch.eye(BC, dtype=torch.float32, device=k.device)
+    a_g = eye + torch.tril(m, diagonal=-1)
+    a_g_inv = torch.linalg.inv(a_g).reshape(B, H, S, BC)
+    return a_g_inv, a_g_inv.clone()
```
Ibuki-wind, line 143: Advisory (non-blocking): same redundant copy as `bench_deltanet.py:126` — `compute_w_u_torch` reads `Aw`/`Au` only. → return `a_g_inv, a_g_inv`.

Ibuki-wind (approved):
### Overall
Vectorization is numerically faithful — batched `gram`, row-scaling via `bc.unsqueeze(-1)`, `tril(m, -1)`, and the `(B,H,NC,BC,BC) -> (B,H,S,BC)` reshape reproduce the per-`(b,h,c)` loop exactly, and the gated `gamma[i,j] = exp(gc[i]-gc[j])` orientation matches `_differentiable_fwd` and `kernel2_gated_deltanet_torch`. Approving; one advisory nit inline.

PR #1791, opened 2026-07-27
benchmarks: Add CUPTI fallback control and improved diagnostics

stelladuyx:
## Problem

In nightly CI, CUPTI profiling occasionally fails, causing fallback to CUDA events which include ~50-60μs launch overhead per call. For fast kernels (<10μs), this inflates latency measurements by **6-7x**.

**Example: L2 Norm kernel**
- CUPTI (accurate): 0.009 ms, 37% roofline ✅
- CUDA events (inflated): 0.067 ms, 5% roofline ❌

This was discovered when investigating why nightly reported anomalously low roofline for L2 Norm operations.

## Solution

This PR adds controls and diagnostics to prevent generating inaccurate benchmark data:

### 1. Environment Variable Control

```bash
# Disable fallback (recommended for CI)
export TILEOPS_ALLOW_CUDA_EVENTS_FALLBACK=0

# Allow fallback (default, for debugging)
export TILEOPS_ALLOW_CUDA_EVENTS_FALLBACK=1
```

When fallback is disabled, benchmarks will fail rather than produce misleading data.

### 2. Enhanced Diagnostics

When CUPTI projection fails, log:
- Number of annotation windows vs expected
- Number of CUDA kernels captured
- Helps diagnose root cause

### 3. Improved Warnings

Clearly state the impact when fallback occurs:
```
CUPTI projection failed; falling back to CUDA-events timing, 
which includes ~50-60us launch overhead per call.
Latency will be inflated by ~6-7x for fast kernels (<10us).
Set TILEOPS_ALLOW_CUDA_EVENTS_FALLBACK=0 to prevent fallback.
```

### 4. CUDA Events Fix

Ensure L2 flush is properly drained before measurement window:
```python
cache.zero_()
torch.cuda.synchronize()  # Drain flush before measurement
start_event.record()
_run(i)
end_event.record()
```

This matches CUPTI's behavior of excluding flush overhead.

## Testing

Tested locally with forced CUPTI failure:
- ✅ Fallback behavior works correctly
- ✅ Warning messages are clear
- ✅ Environment variable control works
- ✅ Diagnostic logs provide useful information

## Recommended Next Steps

1. **Set `TILEOPS_ALLOW_CUDA_EVENTS_FALLBACK=0` in nightly CI**
   - Forces use of accurate CUPTI profiling
   - Prevents misleading benchmark data
   - Provides clear signal when CUPTI fails

2. **Investigate why CUPTI projection fails in nightly**
   - Check diagnostic logs from failed runs
   - Possible causes: pytest-timeout threading, PyTorch 2.10 issues
   - Fix root cause rather than relying on fallback

## Impact

- **No behavior change by default** (backward compatible)
- **Prevents data quality issues** when configured appropriately
- **Better diagnostics** for troubleshooting CUPTI issues

---

Related issue: nightly benchmark data showing anomalously low roofline percentages

### DIFF
--- a/benchmarks/benchmark_base.py
+++ b/benchmarks/benchmark_base.py
@@ -1,5 +1,6 @@
 import contextlib
 import logging
+import os
 import subprocess
 import sys
 import threading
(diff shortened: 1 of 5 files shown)


stelladuyx:
Closing - created from wrong base branch (perf/cumsum). Will recreate from main with only CUPTI-related changes.

PR #1792, opened 2026-07-27
[Bench] Add CUPTI fallback control and improved diagnostics

stelladuyx:
## Summary

- `TILEOPS_ALLOW_CUDA_EVENTS_FALLBACK=0` turns a CUPTI projection failure into a hard error instead of a silent fallback to CUDA-events timing; the default `1` keeps today's behavior.
- Fallback warning and `_CuptiProjectionError` now carry the captured CUDA-kernel count and state the distortion: CUDA events add ~50-60us launch overhead per call, ~6-7x inflation for kernels under 10us.
- CUDA-events path synchronizes after `cache.zero_()`, so the L2 flush drains outside the measurement window as it does under CUPTI.

When CUPTI fails in nightly CI, the inflated numbers publish as if they were real — L2 norm reads 0.067 ms / 5% of roofline under CUDA events vs 0.009 ms / 37% under CUPTI. A missing row beats a wrong one.

## Test plan

- [x] pre-commit passed
- [x] forced CUPTI failure locally: warns and falls back with the variable unset, raises with it set to `0`


zhen8838 (approved):
Reviewed current head `093848e34d5c605dcc4bd8287f5c1ca4f58354bf`. The CUPTI projection diagnostics, explicit fallback control, and L2-flush synchronization are scoped to the benchmark harness; static review found no correctness blocker. GPU smoke, compile-contract, benchmark-contract, pre-commit, actionlint, gitleaks, and security checks are green.

Non-blocking note: the diagnostic wording for captured CUDA kernels and the `~50-60us` / `~6-7x` fallback impact is approximate and could be framed as environment-observed data in a follow-up.

PR #1793, opened 2026-07-27
[Perf][Linear-Attn] Complete and optimize Gated DeltaNet backward

superAngGao:
## Summary

- Complete Gated DeltaNet chunkwise backward gradients.
- Add segmented affine carry for long sequences while keeping the short-sequence sequential specialization.
- Share correction GEMMs, reductions, and stores through a compile-time macro factory.
- Reuse carry intermediates for `dw`, combine `dk` / `dg` / `dw` reductions, and remove obsolete carry code.

## Validation

Official runner image: `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10` on H200.

- `pytest -q tests/ops/test_gated_deltanet_chunkwise_bwd.py tests/ops/test_gated_deltanet_fwd.py` -> 23 passed.
- `pre-commit`, `ruff format --check`, `ruff check`, and `git diff --check` passed.
- Benchmarked against FLA 0.5.1 on H200; TileOps is faster for the listed 4K/8K/16K fp16 d128 cases.


Ibuki-wind (commented):
Read the whole diff and hand-derived the new math — **it checks out**:

- `dL = -A^T dA A^T` with the strict-lower mask.
- `dk`'s Gram term as `dGram @ k + dGram^T @ k`; `dbeta` / `dg` as the row/col reductions.
- Segment composition `X[left] = B + A*X[right]`, including `alpha_acc` spanning the whole segment and
  `carry_after[cid]` written before the update.
- `dw_corr` reuse is valid since `dw` is linear in `du`.

Comments below cover coverage, dead code, and precision. Only the first is worth blocking on.

tests/ops/test_gated_deltanet_chunkwise_bwd.py
```diff
@@ -79,6 +80,7 @@ class GatedDeltaNetBwdFixture(FixtureBase):
             pytest.param(1, 128, 4, 64, 64, 32, torch.float32, False, marks=pytest.mark.full),
             pytest.param(1, 128, 4, 64, 64, 32, torch.float16, False, marks=pytest.mark.full),
             pytest.param(1, 128, 4, 64, 64, 32, torch.bfloat16, False, marks=pytest.mark.full),
+            pytest.param(1, 128, 2, 128, 128, 64, torch.float16, False, marks=pytest.mark.full),
```
Ibuki-wind: - `num_chunks=2` here → `segment_chunks=2` → `num_segments=1`, where the boundary scan degenerates to
  "write a zero carry and exit".
- Every test does this, so the segment composition this PR adds never executes. The 4K/8K/16K numbers run at
  `num_segments=8/16/32`.
- Please add `S>=1024` (`num_segments>=2`), ideally with a bf16 case.

tileops/kernels/gated_deltanet/compute_w_u_bwd.py
```diff
@@ -15,7 +15,10 @@
 import tilelang
 import tilelang.language as T
 
-__all__ = ["compute_w_u_bwd_tl"]
+__all__ = [
+    "compute_w_u_bwd_full_tl",
```
Ibuki-wind: - `compute_w_u_bwd_tl` below has no callers after this PR (~130 lines).
- Please delete it instead of adding it to `__all__`. `deltanet/compute_w_u_bwd.py` is a separate file and
  unaffected.

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
```diff
@@ -364,50 +406,683 @@ def dh_recurrence_bwd_kernel(
...
-    dk = dk_partial + dk_corr + dk_wu
-    dg_cum = dg_partial + dg_corr
+    if recurrence_split_carry == 0:
+        dh_recurrence_bwd_fn = _dh_recurrence_bwd_tl(
+            batch, head, seq_len, chunk_size, dim_k, dim_v, dtype,
+            block_v=recurrence_block_v,
+        )(num_stages, recurrence_threads)
+        dk_corr_partial, du_corr, dg_corr_partial, dw_corr_partial = \
+            dh_recurrence_bwd_fn(g_cum, k, v_new, S, dh_local)
+    elif recurrence_split_carry == 1:
```
Ibuki-wind: - `default_config` only produces `0` or `2`, so this branch and `_dh_carry_after_scan_tl` (~65 lines) are dead
  outside the new test — looks like a development intermediate.
- That leaves three implementations of the same recurrence. Branch `0` still serves d<=64; this one serves
  nothing. Drop it?

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
```diff
@@ -364,50 +406,683 @@ def dh_recurrence_bwd_kernel(
...
+                        dh_loc,
+                        disable_tma=True,
+                    )
+                    alpha = T.exp2(g_c[block_C - 1] * _LOG2E)
+                    for i, j in T.Parallel(dim_k, BV):
+                        summary_frag[i, j] = dh_loc[i, j] + summary[i, j] * alpha
+                    T.copy(summary_frag, summary)
+                    alpha_acc = alpha * alpha_acc
+
+                segment_alpha[bid, hid, vid, sid] = T.cast(alpha_acc, dtype)
```
Ibuki-wind: - `segment_alpha` should be fp32: it's a product of `segment_chunks` factors `exp2(g_last*log2e)` with
  `g_cum<=0`, so ~`1e-7` at 8 chunks — fp16 subnormal, and it flushes to zero a bit deeper.
- bf16 avoids underflow but carries ~0.4% error per factor.
- It's one scalar per segment, so fp32 is free.
- `alpha` doesn't depend on the V tile, so the `num_v_tiles` axis is redundant.

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
```diff
@@ -468,12 +1144,21 @@ def __init__(
 
     @property
     def default_config(self) -> dict:
-        threads = 256 if self.chunk_size >= 64 else 128
+        use_split_carry = self.chunk_size >= 64 and self.dim_v > 64 and self.dim_v % 64 == 0
+        threads = 128 if use_split_carry else (256 if self.chunk_size >= 64 else 128)
+        recurrence_threads = 128 if use_split_carry else threads
+        recurrence_block_v = (
+            (128 if self.dim_v == 128 else 64) if use_split_carry
```
Ibuki-wind: - `dim_v=128` → `recurrence_block_v=128` → `num_v_tiles == 1`: the three `*_partial` tensors gain a size-1
  axis and `_reduce_dh_recurrence_partials_tl` becomes a pure copy.
- That's one extra launch plus two `[B,H,S,DK]` round trips for nothing, on the shape you benchmark.
- It also contradicts the `_dh_recurrence_bwd_tl` docstring, which says the V split exists to keep per-CTA
  state small "for d128".
- Default to `BV=64`, or skip the reduce when `num_v_tiles == 1`.

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
```diff
@@ -364,50 +406,683 @@ def dh_recurrence_bwd_kernel(
...
+        dh_correction_from_carry_fn = _dh_correction_from_carry_tl(
+            batch, head, seq_len, chunk_size, dim_k, dim_v, dtype,
+            block_v=recurrence_block_v,
+        )(recurrence_threads)
+        dh_carry_after = dh_carry_after_scan_fn(g_cum, dh_local)
+        dk_corr_partial, du_corr, dg_corr_partial, dw_corr_partial = \
+            dh_correction_from_carry_fn(g_cum, k, v_new, S, dh_carry_after)
+    else:
+        num_chunks = seq_len // chunk_size
+        segment_chunks = 8 if num_chunks % 8 == 0 else (
```
Ibuki-wind: - Odd `num_chunks` (S=192 or 448 at chunk 64) falls through to `segment_chunks=1`: summary becomes a copy and
  the boundary scan becomes fully sequential — 4 launches doing more work than path `0`.
- A ragged final segment would avoid it.
- This is also a tuning knob buried in the `custom_op`; every other knob lives in `default_config`.

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
```diff
@@ -364,50 +406,683 @@ def dh_recurrence_bwd_kernel(
...
-    from .compute_w_u_bwd import compute_w_u_bwd_tl
+    from .compute_w_u_bwd import compute_w_u_bwd_full_tl
     from .fused_prepare_compute_w_u import fused_prepare_compute_w_u_tl
     from .gated_deltanet_fwd import _chunk_local_cumsum
 
     g_cum = _chunk_local_cumsum(g.float(), chunk_size).to(g.dtype)
 
     fused_fn = fused_prepare_compute_w_u_tl(
         batch, head, seq_len, chunk_size, dim_k, dim_v, dtype,
+        write_duplicate_A=False,
```
Ibuki-wind, line 1024: - `Au` is still in `out_idx`, so it's still allocated and returned — now holding uninitialized memory.
- Only the write is saved, and only a docstring protects the next caller. Make it a conditional output?

tileops/kernels/gated_deltanet/compute_w_u_bwd.py
```diff
@@ -149,3 +152,156 @@ def compute_w_u_bwd(
...
+                T.gemm(A_s, matrix_b_s, matrix_frag, transpose_A=True)
+                for i, j in T.Parallel(block_C, block_C):
+                    matrix_a_s[i, j] = T.if_then_else(
+                        i > j, -matrix_frag[i, j], T.float32(0.0)
+                    )
+
+                # dGram = dL * beta_i * exp(g_i - g_j).
+                for i, j in T.Parallel(block_C, block_C):
+                    matrix_b_s[i, j] = matrix_a_s[i, j] * beta_s[i] * T.exp2(
+                        (g_s[i] - g_s[j]) * 1.4426950408889634
```
Ibuki-wind: - `_LOG2E` already exists in this package (`fused_prepare_compute_w_u.py`, `gated_deltanet_fwd.py`) and is
  already duplicated across the repo.
- Please import it rather than adding two more literals (also below).

tileops/kernels/gated_deltanet/compute_w_u_bwd.py
```diff
@@ -149,3 +152,156 @@ def compute_w_u_bwd(
...
+                T.copy(vector_frag, k_work_s)
+                T.clear(dv_frag)
+                T.gemm(A_s, du_s, dv_frag, transpose_A=True)
+                T.copy(dv_frag, v_work_s)
+
+                for i, j in T.Parallel(block_C, dim_k):
+                    dw_s[i, j] = k_work_s[i, j] * k_s[i, j]
+                T.reduce_sum(dw_s, row_s, dim=1)
+                dbeta_v_s = T.alloc_shared([block_C], accum_dtype)
+                for i, j in T.Parallel(block_C, dim_v):
```
Ibuki-wind, line 114: - `T.alloc_shared` mid-body, also `dbeta_A_s` / `dg_row_s` / `dg_col_s` below.
- The rest of the file declares buffers in one block at the top.

tests/ops/test_gated_deltanet_chunkwise_bwd.py
```diff
@@ -123,6 +125,67 @@ def test_gated_deltanet_bwd(
...
+    for name, expected, actual in zip(
+        ["dq", "dk", "dv", "dg", "dbeta"],
+        baseline_outputs,
+        split_outputs,
+        strict=True,
+    ):
+        torch.testing.assert_close(
+            actual,
+            expected,
+            atol=1e-2,
```
Ibuki-wind: - Too loose here: both paths compute identical math and differ only in accumulation order.
- At 1e-2 on values of magnitude ~0.1, an entire dropped remote carry would still pass. 1e-3 or tighter.

tests/ops/test_gated_deltanet_chunkwise_bwd.py
```diff
@@ -123,6 +125,67 @@ def test_gated_deltanet_bwd(
...
+
+    fwd_op = GatedDeltaNetFwdOp(chunk_size=BC)
+    _o, S_fwd, _Aw, _Au = fwd_op.forward(q, k, v, g, beta)
+
+    common_config = {
+        "num_stages": 2,
+        "threads": 128,
+        "parallel_threads": 256,
+        "recurrence_threads": 128,
+        "recurrence_block_v": 64,
```
Ibuki-wind: - Pins `recurrence_block_v=64`, so the shipped d128 default (`128`, i.e. `num_v_tiles==1`) is never compared.
- Worth parametrizing over both.

tests/ops/test_gated_deltanet_chunkwise_bwd.py
```diff
@@ -123,6 +125,67 @@ def test_gated_deltanet_bwd(
...
+            )
+
+
+@pytest.mark.smoke
+@pytest.mark.parametrize("split_mode", [1, 2])
+def test_gated_deltanet_bwd_split_carry_matches_default_d128(split_mode: int) -> None:
+    torch.manual_seed(456)
+    B, H, S, DK, DV, BC = 1, 2, 128, 128, 128, 64
+    dtype = torch.float16
+    q = torch.randn(B, H, S, DK, device="cuda", dtype=dtype) * 0.1
```
Ibuki-wind, line 162: - No CUDA availability guard here, unlike the `FixtureBase` tests above.

tests/ops/test_gated_deltanet_chunkwise_bwd.py
superAngGao: Addressed in b74096e2. The carry-equivalence test now uses S=1024 (16 chunks, 2 segments at segment_chunks=8) instead of the degenerate S=128 case. It covers fp16 and bf16 and compares all five gradients between segmented and sequential paths at atol=rtol=1e-3.

tileops/kernels/gated_deltanet/compute_w_u_bwd.py
superAngGao: Addressed in b74096e2. Removed the unused compute_w_u_bwd_tl implementation and reduced __all__ to compute_w_u_bwd_full_tl.

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
superAngGao: Addressed in b74096e2. segment_alpha is now [B,H,num_segments] fp32, written once by vid==0, and the redundant V-tile axis was removed. The boundary scan consumes that fp32 scalar for every V tile.

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
superAngGao: Addressed in b74096e2. Removed _dh_carry_after_scan_tl and the development-intermediate mode. The implementation now has only mode 0 (sequential fallback) and mode 1 (segmented affine carry).

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
superAngGao: Addressed in b74096e2 by skipping _reduce_dh_recurrence_partials_tl when num_v_tiles==1 and viewing the single partial directly. This keeps BV=128 for the measured d128 segmented path without the extra launch or size-one copy.

tileops/kernels/gated_deltanet/compute_w_u_bwd.py
superAngGao: Addressed in b74096e2. compute_w_u_bwd.py now imports the package _LOG2E constant and uses it for both exponent sites.

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
superAngGao: Addressed in b74096e2. segment_chunks is now an explicit config field. Default dispatch uses segmented carry only when segment_chunks>1 and the chunk count is divisible by 2/4/8; odd chunk counts retain the sequential fallback rather than running four strictly worse launches. Ragged segmentation can be added later without changing the current correctness fallback.

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
superAngGao: Addressed in b74096e2 with a true conditional output. Forward retains the public (Aw, Au, w, u) ABI, while backward recompute compiles a single-A prim_func with out_idx=[-3,-2,-1] and receives (A,w,u). No Au tensor is allocated, returned, or left uninitialized on the backward path.

tileops/kernels/gated_deltanet/compute_w_u_bwd.py
superAngGao: Addressed in b74096e2. dbeta_v_s, dbeta_A_s, dg_row_s, and dg_col_s are declared with the other shared buffers at the top of the kernel body.

tests/ops/test_gated_deltanet_chunkwise_bwd.py
superAngGao: Addressed in b74096e2. The segmented-vs-sequential comparison now uses atol=rtol=1e-3 for dq/dk/dv/dg/dbeta.

tests/ops/test_gated_deltanet_chunkwise_bwd.py
superAngGao: Addressed in b74096e2. The test covers segmented BV=64 and the shipped d128 segmented default BV=128. The sequential reference stays at BV=64 because compiling the sequential recurrence with BV=128 requests about 313 KB dynamic shared memory; the comparison is between mathematically equivalent paths, not identical CTA tiling.

tests/ops/test_gated_deltanet_chunkwise_bwd.py
superAngGao: Addressed in b74096e2. Added an explicit torch.cuda.is_available() skip guard to the direct kernel comparison test.

zhen8838 (changes requested):
Blocking issue:

- The current head expands `tests/ops/test_gated_deltanet_chunkwise_bwd.py` with the segmented-vs-sequential `S=1024` coverage and a six-cell default-dispatch matrix, but the PR body still has no `## Test node delta` or per-case rationale. The approval gate requires the before/after node count and a distinct-path justification for each added parametrization cell when an existing test file grows materially. Add that final verification fact and rationale before approval.

superAngGao:
Reviewer fixes are in b74096e2. Local verification in the unmodified official runner image passed the focused backward+forward suite (24 passed), pre-commit, ruff, and diff checks. The first remote gpu-smoke attempt failed before installation/tests because actions/checkout hit repeated GitHub TLS timeouts; a rerun is currently in progress.

superAngGao:
Addressed the test-node approval gate in f10aae04 and the PR body. The backward test file now has an explicit 6 -> 14 node delta with a per-node distinct-path rationale. I also removed the redundant B=2 dispatch cell, retaining the B=4/H=16 case as the aggregate B*H invariance check. The focused forward/backward suite passes in the official runner image: 23 passed.

Ibuki-wind (approved):
Clean — no issues.

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
```diff
@@ -364,61 +440,718 @@ def dh_recurrence_bwd_kernel(
...
+                    v_new_c,
+                    disable_tma=True,
+                )
+                T.copy(S[bid, hid, cid, :, v_offset : v_offset + BV], h_c, disable_tma=True)
+                T.copy(dh_carry_after[bid, hid, vid, cid, :, :], dh_buf, disable_tma=True)
+
+                g_last = g_c[block_C - 1]
+                exp_g_last = T.exp2(g_last * _LOG2E)
+
+                for pn, sk in T.Parallel(block_C, dim_k):
```
Ibuki-wind: `_dh_correction_from_carry_tl` repeats the correction half of `_dh_recurrence_bwd_tl` almost statement-for-statement: `k_scaled`, `du_corr`, `dw_corr`, `dk_corr`, and both `dg` terms are now maintained in two places. The kernels need different grids because one produces the carry inline while the other consumes `dh_carry_after`; that does not require duplicating the equations after the incoming `dh_buf` exists. → Keep both compiled kernels and extract a compile-time `T.macro` factory for “correction from incoming carry”. Pass the resident buffers plus `(bid, hid, vid, cid)`/offset into it; let each caller own only carry production/loading and scheduling. Do not replace this with a runtime sequential/segmented branch. Acceptance criterion: the correction GEMMs, reductions, and stores have one source definition used by both paths.

Ibuki-wind (changes requested):
### Overall

The sequential fallback and segmented carry path are valid compiled specializations. The correction equations must have one source implementation shared by both.

tileops/kernels/gated_deltanet/gated_deltanet_bwd.py
superAngGao: Addressed in `ae93e92f`. The sequential and segmented paths remain separate compiled kernels with their existing grids and dispatch. Both now instantiate `_make_dh_correction_from_carry_macro`, which is the single source for `k_scaled`, `du_corr`, `dw_corr`, `dk_corr`, both `dg` reductions, and all correction stores. The callers retain only carry production/loading and scheduling. Source scan shows the three correction GEMMs only in the macro; the official runner backward+forward suite is `23 passed`.

superAngGao:
Addressed the latest correction-duplication review in `ae93e92f`.

- Kept sequential and segmented carry as separate compiled specializations; no runtime branch was introduced.
- Extracted one compile-time `T.macro` factory for the correction GEMMs, reductions, and stores after the incoming carry is resident.
- Sequential owns reverse carry propagation; segmented owns precomputed carry loading and parallel chunk scheduling.
- Verified by source scan that the correction GEMMs now have one definition.

Validation in the unmodified official runner image on GPU4/H200:

```text
python -m pytest -q tests/ops/test_gated_deltanet_chunkwise_bwd.py tests/ops/test_gated_deltanet_fwd.py
23 passed
```

`pre-commit`, `ruff format --check`, `ruff check`, and `git diff --check` all pass.

Ibuki-wind (approved):
Clean — both scheduling specializations now share one correction implementation without changing the math.

zhen8838 (approved):

PR #1794, opened 2026-07-28
[Bench] Force CUPTI profiling for L1/L2 norm

stelladuyx:
## Summary

- Require CUPTI timing for L1/L2 norm benchmarks.
- Disable CUDA-event fallback for these sub-10-us kernels.

## Rationale

CUDA events measured the reported 9–10 us kernels at about 67 us, reducing the apparent roofline from 37% to 5%.

## Validation

Local L1/L2 benchmark profiling completed with CUPTI enabled. Depends on #1792.

### DIFF
--- a/benchmarks/ops/bench_vector_norm.py
+++ b/benchmarks/ops/bench_vector_norm.py
@@ -4,6 +4,8 @@ Measures latency, TFLOPS, and DRAM bandwidth against PyTorch baselines.
 Workload shapes and roofline formulas are loaded from the ops manifest (tileops/manifest/).
 """
 
+import os
+
 import pytest
 import torch
 
@@ -28,6 +30,9 @@ _INF_NORM_OP = "InfNormFwdOp"
 def test_l1_norm_bench(
     shape: tuple, dtype: torch.dtype, op_params: dict
 ) -> None:
+    # Force CUPTI profiling for L1 norm (fast kernel, CUDA events too inaccurate)
+    os.environ["TILEOPS_ALLOW_CUDA_EVENTS_FALLBACK"] = "0"
+
     test = L1NormTest(shape, dtype)
     inputs = test.gen_inputs()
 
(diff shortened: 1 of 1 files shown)


zhen8838 (changes requested):
Blocking issues:

- `benchmarks/ops/bench_vector_norm.py:34` and `:73` mutate the process-wide `TILEOPS_ALLOW_CUDA_EVENTS_FALLBACK=0` without restoring it. After the L1/L2 parametrized tests run, `test_inf_norm_bench` and later benchmark modules inherit the forced CUPTI policy, so unrelated benchmarks can fail instead of using the configured fallback. Scope the override with `monkeypatch.setenv` or a context manager and restore the prior value after each test.
- The PR title `benchmarks: Force CUPTI profiling for L1/L2 norm` fails `validate-pr-title`. Rename it to the repository format, for example `[Bench] Force CUPTI profiling for L1/L2 norm`.
- Update the PR body to follow `.foundry/mold/pr-body-template.md`: provide the required `## Test plan`/verification structure and remove process-history narration.

benchmarks/ops/bench_vector_norm.py
```diff
@@ -28,6 +30,9 @@
 def test_l1_norm_bench(
     shape: tuple, dtype: torch.dtype, op_params: dict
 ) -> None:
+    # Force CUPTI profiling for L1 norm (fast kernel, CUDA events too inaccurate)
+    os.environ["TILEOPS_ALLOW_CUDA_EVENTS_FALLBACK"] = "0"
```
Ibuki-wind, line 34: `test_l1_norm_bench` and `test_l2_norm_bench` permanently mutate the process environment, so later cases—including `test_inf_norm_bench` and benchmarks collected from other modules—inherit disabled CUDA-event fallback. → Scope the override to each test invocation with `monkeypatch.setenv` or a restoring context manager instead of assigning `os.environ` directly.

Ibuki-wind (changes requested):
### Overall
One blocking runtime-isolation issue remains.

Ibuki-wind:
Closing this approach after the framework-level review:

- The timing failure is global, not specific to L1/L2 norm.
- A per-test environment override only hard-fails two symptoms; it does not fix call-to-kernel attribution.
- Keep CUPTI policy in the benchmark framework rather than individual op benchmarks.

The underlying problem remains valid, but this PR should not continue.

PR #1795, opened 2026-07-28
fix(bench): sample kernels from CUPTI collection window

superAngGao:
## Summary

- stop using Kineto projected annotation windows as the CUPTI benchmark timing/counting boundary
- sample business CUDA kernels directly from one continuous CUPTI CUDA activity window per trial
- filter the benchmark L2 flush fill kernel from CUDA activity summation
- keep CUDA-events fallback only for empty CUPTI business-kernel samples

## Previous Timing Model

Before this patch, `bench_kernel()` used two different concepts of a window:

- **CPU window**: every repeat wrapped `_run(i)` in `torch.profiler.record_function(_KERNEL_REGION)`. This created one CPU-side annotation scope per repeat.
- **GPU timing/counting window**: Kineto was expected to project those CPU annotation scopes onto the CUDA timeline. `_sum_kernel_time_us()` then summed CUDA kernels whose start timestamp fell inside a projected CUDA annotation window, and required the number of projected windows to equal `n_repeat`.

So the benchmark was not only asking CUPTI for CUDA kernel activity. It was also requiring Kineto projection to produce exactly one CUDA-side annotation window per repeat.

The L2 flush was intended to stay outside the timed call by doing:

```text
cache.zero_()
torch.cuda.synchronize()
record_function(_KERNEL_REGION):
    _run(i)
torch.cuda.synchronize()
```

That part drains CUDA work, but it does not provide a synchronization primitive for Kineto projection itself.

## What Went Wrong

The failure mode is a CPU/GPU window mismatch:

- CPU issued `n_repeat` annotation scopes.
- CUDA kernels were often present in the CUPTI activity trace.
- Kineto sometimes produced fewer CUDA-side projected annotation windows than repeats, e.g. `49/50`, `48/50`, `47/50`, or `0/50`.
- The benchmark treated this as an untrustworthy CUPTI trace and fell back to CUDA events.

The missing piece is a projection-ready synchronization signal. `torch.cuda.synchronize()` only waits for GPU work to finish. It does not guarantee that Kineto has materialized one projected CUDA annotation window for every CPU `record_function` scope. PyTorch/Kineto does not expose a public GPU-side semaphore for "projection is ready for this repeat".

Because of that, the previous implementation tied benchmark correctness to Kineto projection stability instead of CUPTI CUDA activity sampling itself.

## Observed Incidence

This was not a rare single-case issue.

From the previous nightly artifact we inspected, [run 30170991562](https://github.com/tile-ai/TileOPs/actions/runs/30170991562):

- all benchmark rows: `88/2923` used CUPTI, `2835/2923` used CUDA events (`96.99%` fallback)
- TileOps rows only: `26/1280` used CUPTI, `1254/1280` used CUDA events (`97.97%` fallback)

In local official-runner reproduction on clean `main`, the fallback rate was lower but still material:

- all rows: `208/2178` CUDA-events fallback (`9.55%`)
- TileOps-like rows: `90/1068` CUDA-events fallback (`8.43%`)

Focused reruns also showed repeated projection mismatches such as `49/50`, `48/50`, `47/50`, and `0/50`. Immediate retry did not reliably recover the same case, so simply retrying after fallback is not a principled fix.

GLA `0/50` projection behavior is intentionally left out of this patch; it appears to be a separate Kineto projection issue and should be handled independently.

## Fix

This patch removes projected annotations from the main timing/counting path.

The new CUPTI path uses:

- CUDA-only profiler activity collection
- one continuous CUPTI collection window per trial
- repeated `cache.zero_ -> synchronize -> run -> synchronize` inside that collection window
- direct summation of CUDA business-kernel activity samples
- name-based filtering of the benchmark L2 flush fill kernel (`vectorized_elementwise` + `FillFunctor`)

The benchmark now uses the CUPTI collection window itself as the sampling window. CPU annotations and Kineto projected CUDA annotations are no longer part of the primary timing/counting path.

If CUPTI captures no business CUDA kernel samples, we still fall back to CUDA events using the existing fallback policy. If CUPTI captures samples, we compute latency from the sampled CUDA kernel durations.

## Validation

Ran in `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10` without installing/upgrading tilelang, torch, or dependencies:

```text
python py_compile benchmarks/benchmark_base.py benchmarks/tests/test_benchmark_base.py
pytest -q benchmarks/tests/test_benchmark_base.py::test_empty_cupti_sample_falls_back_to_cuda_events benchmarks/tests/test_benchmark_base.py::test_kernel_runtime_error_propagates
pytest -q -s benchmarks/ops/bench_gemm.py::test_gemm_fp8_bench[ds-v3-decode-gate-up-block128-float8_e4m3fn]
pytest -q -s benchmarks/ops/attention/bench_gqa_sliding_window.py::test_gqa_sliding_window_fwd_bench[llama-3.1-8b-short-w256-float16]
```

### DIFF
--- a/benchmarks/benchmark_base.py
+++ b/benchmarks/benchmark_base.py
@@ -90,57 +90,38 @@ _bench_results = threading.local()
 _bench_meta = threading.local()
 
 
-class _CuptiProjectionError(Exception):
-    """CUPTI trace lacked a projected annotation window for every repeat."""
-
-
-# Name of the ``record_function`` annotation wrapping the timed call. Kineto
-# projects this scope onto the device timeline. The L2-flush ``cache.zero_()``
-# is synchronized to completion before the window opens (see ``bench_kernel``),
-# so its device event cannot fall inside a window regardless of how the
-# projection behaves; kernels the timed call launches do.
-_KERNEL_REGION = "tileops_bench_kernel"
+class _CuptiSamplingError(Exception):
+    """CUPTI trace lacked CUDA kernel samples for the measured window."""
 
 
 def _sum_kernel_time_us(kineto_results):
-    """Sum device time of the kernels the timed call launched.
+    """Sum sampled business-kernel device time from a CUPTI collection window.
 
-    Sums only kernels inside a :data:`_KERNEL_REGION` annotation window, so the
-    L2-flush fill is excluded and the kernel under test is counted regardless of
-    its name. A call launching several kernels contributes all of them.
+    The measured CUPTI window is continuous across the repeat loop and includes
+    the L2 flush launches. Filter those fill kernels by name, then sum every
+    remaining CUDA kernel sample. A call launching several kernels contributes
+    all of them.
 
     Iterates the C++ Kineto events directly to bypass ``key_averages()``, which
     is ~16x slower (~130ms of Python parsing/tree-building) for large traces.
 
     Returns:
-        ``(total_us, n_regions)``: summed kernel time in microseconds and the
-        number of annotation windows. The caller checks ``n_regions ==
-        n_repeat`` to confirm the scope projected on every iteration.
+        ``(total_us, n_kernels)``: summed kernel time in microseconds and the
+        number of sampled business CUDA kernels.
     """
-    import bisect
-
-    windows: list[tuple[int, int]] = []
-    kernels: list[tuple[int, int]] = []  # (start_ns, duration_ns)
+    total_us = 0.0
+    n_kernels = 0
     for evt in kineto_results.events():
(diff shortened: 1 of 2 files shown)


superAngGao:
Superseded by #1797, which absorbs this direction into a narrower CUPTI sampling-window fix and leaves non-integral multi-kernel samples to follow-up issue #1796.

superAngGao:
Closing in favor of #1797, which carries the narrowed CUPTI sampling-window fix. The remaining non-integral multi-kernel cases are tracked separately in #1796.

PR #1797, opened 2026-07-28
[Bench] Classify single-kernel CUPTI samples before timing

### LINKED ISSUE #1796: [Bench] Follow up non-integral multi-kernel CUPTI samples
## Background

The CUPTI sampling-window fix keeps CUDA-events fallback for traces where one measured trial window captures more CUDA kernels than `n_repeat`, but the kernel count is not an integer multiple of `n_repeat`.

That class is intentionally out of scope for the sampling-window PR: a non-integral count means we cannot safely infer how many logical benchmark iterations contributed to the summed GPU time.

## Observed Data

Using the official runner image and shared caches:

- image: `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`
- repo: clean `origin/main` plus the sampling-window patch
- run: `scripts/ci/run_benchmarks.py benchmarks/ops --timeout-per-file 900`
- result: `59/59` benchmark files passed
- report rows: `2308`
- CUPTI rows after the sampling-window fix: `2230 / 2308 = 96.62%`
- CUDA-events fallback rows: `78 / 2308 = 3.38%`

Probe summary by trial:

```text
n_repeat=3: 8 trials, all complete/integral
n_repeat=50: 6744 trials
  full or multi-kernel integral: 6601
  kernels < n_repeat, accepted as sampled CUPTI: 65
  kernels > n_repeat and non-integral: 77
  bad projected regions: 1
```

The main remaining fallback class is `kernels > n_repeat && kernels % n_repeat != 0`. Examples:

```text
GQA bwd torch baseline: 295 / 50
MHA bwd torch baseline: 197 / 50
BatchNorm bwd torch baseline: 248 / 50, 397 / 50
Conv1d TileOps: 99 / 50
DeltaNet/GatedDeltaNet/GLA bwd/fwdbwd baselines: 343 / 50, 593 / 50, 539 / 50, 839 / 50, 294 / 50, 544 / 50
```

There was also one projected-region outlier:

```text
bench_convolution.py torch conv baseline: regions=9, kernels=1600, n_repeat=50
```

## Problem

For `kernels < n_repeat`, the sampling-window fix can treat the measured kernels as a valid GPU sample and divide by the sampled kernel count.

For `kernels > n_repeat` and non-integral counts, that denominator is ambiguous. The captured kernels may include valid multi-kernel work, occasional extra kernels, or missing kernels from some logical iterations. Dividing by either `n_repeat` or raw kernel count can produce misleading latency.

## Follow-up Work

Find a robust GPU-side way to identify logical benchmark iterations for multi-kernel callables, or otherwise classify which extra/missing kernels are safe to account for. Until then, these cases should continue to fall back rather than silently report a potentially wrong CUPTI latency.

superAngGao:
Closes #1796

- Classify profiled calls before CUPTI timing.
- Accept partial CUPTI samples only for verified single-kernel callables.
- Fall back to CUDA events for multi-kernel or ambiguous traces.

Validation: benchmark timing/fallback tests and representative CI benchmark runs pass.

### DIFF
--- a/benchmarks/benchmark_base.py
+++ b/benchmarks/benchmark_base.py
@@ -5,6 +5,7 @@ import subprocess
 import sys
 import threading
 from abc import ABC, abstractmethod
+from collections import Counter
 from datetime import datetime
 from typing import (
     Any,
@@ -90,8 +91,8 @@ _bench_results = threading.local()
 _bench_meta = threading.local()
 
 
-class _CuptiProjectionError(Exception):
-    """CUPTI trace lacked a projected annotation window for every repeat."""
+class _CuptiSamplingError(Exception):
+    """CUPTI trace lacked a usable measured-window CUDA sample."""
 
 
 # Name of the ``record_function`` annotation wrapping the timed call. Kineto
@@ -102,25 +103,12 @@ class _CuptiProjectionError(Exception):
 _KERNEL_REGION = "tileops_bench_kernel"
 
 
-def _sum_kernel_time_us(kineto_results):
-    """Sum device time of the kernels the timed call launched.
-
-    Sums only kernels inside a :data:`_KERNEL_REGION` annotation window, so the
-    L2-flush fill is excluded and the kernel under test is counted regardless of
-    its name. A call launching several kernels contributes all of them.
-
-    Iterates the C++ Kineto events directly to bypass ``key_averages()``, which
-    is ~16x slower (~130ms of Python parsing/tree-building) for large traces.
-
-    Returns:
-        ``(total_us, n_regions)``: summed kernel time in microseconds and the
-        number of annotation windows. The caller checks ``n_regions ==
-        n_repeat`` to confirm the scope projected on every iteration.
-    """
+def _window_kernel_stats(kineto_results):
+    """Return per-annotation-window business CUDA kernel time and names."""
     import bisect
 
     windows: list[tuple[int, int]] = []
-    kernels: list[tuple[int, int]] = []  # (start_ns, duration_ns)
+    kernels: list[tuple[int, int, str]] = []  # (start_ns, duration_ns, name)
     for evt in kineto_results.events():
         if evt.device_type() != DeviceType.CUDA:
             continue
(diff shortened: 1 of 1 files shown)


zhen8838 (changes requested):
Blocking issues:

- `benchmarks/benchmark_base.py:147-155` treats `n_kernels < n_repeat` as the number of sampled calls. The same helper explicitly supports one call launching several business CUDA kernels, so a partial trace cannot infer the per-call denominator from the aggregate kernel count. For example, 50 calls that each launch two kernels can produce 50 captured kernels and be accepted as 50 calls, under-reporting latency by about 2x. Even an integer multiple of `n_repeat` does not prove the sample is complete. Keep incomplete/ambiguous samples on the fallback path, or establish and enforce a verifiable one-kernel-per-call contract/ per-call sampling boundary, with regression coverage for multi-kernel partial samples.
- Update the PR body to follow `.foundry/mold/pr-body-template.md` and the approval-gate authoring discipline. The current body replaces the required `## Test plan` structure with `## Problem`/`## Fix`/`## Validation` and includes process-history narration such as the reference to absorbing part of #1795.

superAngGao:
Addressed the blocking feedback:

- CUPTI samples now require exactly one business CUDA kernel name in the measured window and `count <= n_repeat`.
- Multi-kernel-name samples fall back to CUDA events instead of inferring a denominator from aggregate kernel count.
- Added regression coverage for single-kernel partial samples, multi-kernel rejection, and over-repeat rejection.
- Rewrote the PR body to follow `.foundry/mold/pr-body-template.md`.

superAngGao:
Updated again: removed the `test_benchmark_base.py` changes from the PR. The branch now only changes `benchmarks/benchmark_base.py`.

I kept the helper validation as a temporary check and listed the result in the PR body instead of adding repo tests.

benchmarks/benchmark_base.py
```diff
@@ -113,34 +112,54 @@ def _sum_kernel_time_us(kineto_results):
...
+def _sampled_runs(kernel_counts: Counter[str], n_repeat: int) -> int:
+    n_kernels = sum(kernel_counts.values())
+    if n_kernels == 0:
+        raise _CuptiSamplingError("no business CUDA kernels captured")
+    if len(kernel_counts) != 1:
+        raise _CuptiSamplingError(
+            f"{n_kernels} business CUDA kernels across "
+            f"{len(kernel_counts)} kernel names captured for {n_repeat} repeats"
+        )
+    if n_kernels <= n_repeat:
```
Ibuki-wind, line 189: `_sampled_runs` equates kernel count with completed call count, but one call can launch the same business kernel name more than once. If 25 of 50 calls are captured and each launches that kernel twice, `kernel_counts` has one name and count 50, so this accepts the trace and divides by 50, reporting half the per-call latency. → Either establish per-call sampling boundaries/count completed calls, or reject partial traces unless a verifiable one-kernel-per-call contract is available.

Ibuki-wind (changes requested):
### Overall
The sampling denominator is still ambiguous for repeated same-name kernel launches; make it represent completed calls before merge.

superAngGao:
Updated the PR again to match the narrower direction we discussed:

- Restored the original per-repeat CUPTI annotation/profiler flow; no trial-level big window or profiler lifecycle change remains.
- Removed the `regions == n_repeat` gate from the denominator decision.
- Count CUDA activity by business kernel name inside the projected windows.
- Accept CUPTI only when the sample has exactly one kernel name and `1 <= count <= n_repeat`; use the sampled count as the denominator.
- Fall back to CUDA events for multi-kernel-name, empty, or over-repeat samples.

The PR now only changes `benchmarks/benchmark_base.py`.

superAngGao:
Updated based on the denominator concerns:

- Kept the original per-repeat CUPTI annotation/profiler flow; no trial-level big window.
- Added a small profiled classification pass before timing (`min(2, n_repeat)`).
- Classification requires every projected classified call window to contain exactly one business CUDA kernel, and all classified windows must use one kernel name.
- The timed CUPTI path can then accept partial sampled counts such as `49/50` for this single-kernel callable contract.
- Multi-kernel, same-call multi-launch, empty, over-repeat, or failed-classification samples stay on CUDA-events fallback.

This is meant to address both the aggregate-count ambiguity and the same-name multi-kernel case: kernel names are still used in the timed sample, but only after classification proves one business kernel per projected call.

Ibuki-wind (approved):
Clean — the classification pass establishes the one-kernel-per-call contract before sampled CUPTI timing.

Ibuki-wind:
Re-reviewed from the framework perspective:

- This is a global Kineto projection problem, not a norm-specific issue.
- The classification pass verifies one business CUDA kernel per call before partial CUPTI samples are accepted.
- Multi-kernel or ambiguous traces remain on CUDA events, so the denominator blocker is resolved.
- Per-op environment overrides are not part of this PR.

Decision: approve this PR as the global stabilization fix. Separating kernel time from operator latency is useful follow-up work, not a blocker here.

Ibuki-wind:
Closing the current implementation so we stop layering inference on an unreliable signal:

- Two-call classification still depends on projected GPU annotations.
- It cannot establish attribution for later or multi-kernel calls.
- The replacement should map CPU call annotations to CUDA kernels through correlation IDs, then report kernel time separately from operator latency.

The problem remains valid; this closes only the current patch direction.

PR #1800, opened 2026-07-28
[Fix][Bench] Revert CUPTI fallback control from #1792

lcy-seso:
- Reverts the per-sample synchronization that introduces a CPU launch bubble inside CUDA-event timing.
- Restores the pre-#1792 fallback behavior while CUPTI attribution is redesigned.

### DIFF
--- a/benchmarks/benchmark_base.py
+++ b/benchmarks/benchmark_base.py
@@ -1,6 +1,5 @@
 import contextlib
 import logging
-import os
 import subprocess
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


haruhi55 (approved):
Exact revert of #1792. Restores the previous CUDA-event ordering and removes the per-sample synchronization that introduced a launch bubble into the measured interval.

Ibuki-wind (approved):

Ibuki-wind (approved):

PR #1810, opened 2026-07-30
[Bench] Add external baselines for common ops

### LINKED ISSUE #1806: [Bench] Add external baselines for roofline-only common ops
## Background

The benchmark dashboard currently has many common operators whose only performance reference is the roofline. Roofline should remain as the hardware-efficiency signal, but these operators also need competitive third-party baselines so nightly results are easier to interpret.

This issue covers common operator families only:

- Elementwise
- Reduction
- Normalization
- Convolution
- Pooling
- RoPE / positional encoding
- FFT

Out of scope for this issue:

- MoE
- Quantization
- MHC
- TopK
- Engram
- DeltaNet / SSD decode follow-ups that already have separate intended references

## Proposed baseline tags

Use a small dependency surface first, prioritizing coverage over per-family specialist libraries.

| Tag | Implementation path | Intended coverage |
| --- | --- | --- |
| `torch_inductor` | Compile benchmark-local PyTorch reference functions with `torch.compile` | Elementwise, reductions, normalization fallback, mean pooling |
| `torch_cudnn` | Use PyTorch APIs that dispatch to cuDNN where applicable | Convolution, pooling, BatchNorm |
| `flashinfer` | Call FlashInfer Python APIs directly | RoPE, RMSNorm, fused add RMSNorm |
| `torch_cufft` | Use `torch.fft` as the cuFFT-backed reference | FFTC2C |

## Coverage map

| Family | Operators | Baseline |
| --- | --- | --- |
| Elementwise | arithmetic, compare, logical/bitwise, activation, broadcast, dropout, gated activation strategy variants | `torch_inductor` |
| Reduction | sum/mean/prod/min/max/all/any/count_nonzero/cumsum/cumprod/norm/var/std/softmax/logsoftmax/logsumexp | `torch_inductor` |
| Normalization | LayerNorm, AdaLayerNorm, GroupNorm, InstanceNorm | `torch_inductor` |
| Normalization | RMSNorm, FusedAddRMSNorm | `flashinfer` |
| Normalization | BatchNorm fwd/bwd | `torch_cudnn` |
| Convolution | Conv1d, Conv2d, Conv2dBias | `torch_cudnn` |
| Pooling | AvgPool, MaxPool, MaxPool-with-indices variants | `torch_cudnn` |
| Pooling | MeanPoolingForward | `torch_inductor` |
| RoPE / positional encoding | Neox, non-Neox, Llama 3.1, Yarn, LongRoPE, position-id variants | `flashinfer` |
| FFT | FFTC2C | `torch_cufft` |

## Implementation notes

- Keep baseline functions benchmark-local. Do not import correctness references or workload oracle functions from tests.
- Preserve existing roofline metrics; these baselines should add competitive comparison data, not replace `% roof`.
- Record tag-prefixed metrics, for example `torch_inductor_latency_ms`, `flashinfer_latency_ms`, and corresponding ratios.
- If an optional third-party dependency is missing, skip only that baseline and keep the benchmark runnable.
- Dashboard/report generation should continue choosing the strongest competitive baseline for `vs baseline`.

## Acceptance criteria

- The listed common operator families have at least one third-party competitive baseline where they currently fall back to roofline-only comparison.
- Nightly benchmark output retains TileOPs latency, roofline metadata, baseline latency, and baseline ratio fields.
- Environments without FlashInfer skip FlashInfer-specific baselines without failing the whole benchmark suite.
- The dashboard no longer shows roofline as the only reference for the common operator families covered by this issue.

superAngGao:
## Summary

Draft PR for #1806.

- Add external baseline tags for common-op benchmarks: `torch_inductor`, `torch_cudnn`, `torch_cufft`, `torch_native`, and `flashinfer`.
- Add a shared `torch_inductor_baseline()` helper for benchmark-local PyTorch baselines compiled with `torch.compile(..., fullgraph=True)`.
- Add FlashInfer baselines for RMSNorm, fused add RMSNorm, and semantic-equivalent RoPE variants.
- Keep optional third-party baselines local to benchmark files and preserve existing roofline reporting.
- Fix nightly report coloring to use the strongest competitive baseline when multiple baselines are present.

## Validation

Official runner image: `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`.

- `compileall` and `ruff check` on changed files.
- `pytest -q tests/test_nightly_report.py benchmarks/tests/test_benchmark_base.py --tb=short` -> 19 passed.
- RoPE/Pool benchmark collect-only -> 52 cases collected.
- Targeted manifest checks for RoPE, RMSNorm, Conv/Pool, and FFT passed with advisory warnings only.
- TorchInductor and FlashInfer smoke checks passed.

Closes #1806.

### DIFF
--- a/benchmarks/benchmark_base.py
+++ b/benchmarks/benchmark_base.py
@@ -6,6 +6,7 @@ are re-exported here, so a bench file keeps importing what it always did.
 
 import statistics
 from abc import ABC, abstractmethod
+from collections.abc import Callable
 from typing import Any, Generic, Optional, TypeVar
 
 import pytest
@@ -259,6 +260,16 @@ def workload_field_params(workloads: list, keys: tuple) -> list:
             )
         )
     return params
+
+
+def torch_inductor_baseline(fn: Callable) -> Callable:
+    """Compile a benchmark-local PyTorch baseline with TorchInductor.
+
+    The baseline implementation itself stays in the benchmark file so tests do
+    not become performance oracles. This helper only centralizes the compile
+    policy used by common-op benchmark baselines.
+    """
+    return torch.compile(fn, fullgraph=True)
 
 
 class ManifestBenchmark(BenchmarkBase[Any]):
--- a/benchmarks/ops/bench_ada_layer_norm.py
+++ b/benchmarks/ops/bench_ada_layer_norm.py
@@ -36,7 +36,12 @@ def test_ada_layer_norm_bench(m: int, n: int, dtype: torch.dtype) -> None:
         normed = F.layer_norm(x, (n,), weight=None, bias=None, eps=test.eps)
         return scale * normed + shift
 
-    bm.compare({"tileops": op, "torch-inductor": torch_inductor_baseline(baseline_fn)}, *inputs, record_as=op, params=locals())
+    bm.compare(
+        {"tileops": op, "torch-inductor": torch_inductor_baseline(baseline_fn)},
+        *inputs,
+        record_as=op,
+        params=locals(),
+    )
 
 
 @pytest.mark.parametrize("m, n, dtype", _to_params(load_workloads(_ADA_ZERO_OP_NAME)))
@@ -52,4 +57,9 @@ def test_ada_layer_norm_zero_bench(m: int, n: int, dtype: torch.dtype) -> None:
         normed = F.layer_norm(x, (n,), weight=None, bias=None, eps=test.eps)
         return gate * (scale * normed + shift)
 
-    bm.compare({"tileops": op, "torch-inductor": torch_inductor_baseline(baseline_fn)}, *inputs, record_as=op, params=locals())
+    bm.compare(
+        {"tileops": op, "torch-inductor": torch_inductor_baseline(baseline_fn)},
+        *inputs,
+        record_as=op,
+        params=locals(),
+    )
--- a/benchmarks/ops/bench_ada_layer_norm.py
+++ b/benchmarks/ops/bench_ada_layer_norm.py
@@ -2,7 +2,7 @@ import pytest
 import torch
 import torch.nn.functional as F
 
-from benchmarks.benchmark_base import BenchmarkReport, ManifestBenchmark
+from benchmarks.benchmark_base import BenchmarkReport, ManifestBenchmark, torch_inductor_baseline
 from tileops.manifest import load_workloads
 from tileops.ops.norm.ada_layer_norm import AdaLayerNormFwdOp
 from tileops.ops.norm.ada_layer_norm_zero import AdaLayerNormZeroFwdOp
@@ -36,7 +36,7 @@ def test_ada_layer_norm_bench(m: int, n: int, dtype: torch.dtype) -> None:
         normed = F.layer_norm(x, (n,), weight=None, bias=None, eps=test.eps)
         return scale * normed + shift
 
-    bm.compare({"tileops": op, "torch-ref": baseline_fn}, *inputs, record_as=op, params=locals())
+    bm.compare({"tileops": op, "torch-inductor": torch_inductor_baseline(baseline_fn)}, *inputs, record_as=op, params=locals())
 
 
 @pytest.mark.parametrize("m, n, dtype", _to_params(load_workloads(_ADA_ZERO_OP_NAME)))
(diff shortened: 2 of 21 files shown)


superAngGao:
Thanks for the careful review. I pushed a follow-up commit (`1a28c9d`) addressing all three points.

- Fixed RoPE FlashInfer semantics: cached FlashInfer baselines now receive variant-specific cos/sin caches. YaRN uses the benchmark-local YaRN interpolation formula, and the benchmark also checks FlashInfer output against the benchmark-local semantic reference once before timing. I also tightened the torch reference path so Neox, Non-Neox, Llama 3.1, YaRN, and LongRoPE all use semantic-equivalent references instead of geometry-only rotation.
- Renamed pooling baselines from `torch_cudnn` to `torch_native`, matching the kernels observed in the official runner. Convolution and BatchNorm remain tagged `torch_cudnn`.
- Fixed nightly report coloring for multi-baseline configs. Since `ratio = baseline_latency / tileops_latency`, the strongest competitor is the smallest ratio, not the largest. Added a small regression test for the FlashInfer=0.5 / Inductor=2.0 case.

Validation in `ghcr.io/tile-ai/tileops-runner:65dbc98-torch2.10`:

- `compileall` for the changed files
- `pytest -q tests/test_nightly_report.py benchmarks/tests/test_benchmark_base.py --tb=short` -> 19 passed
- `pytest --collect-only -q benchmarks/ops/bench_rope.py benchmarks/ops/bench_pool.py --tb=short` -> 52 benchmark cases collected
- `ruff check` on the changed files
- `validate_manifest.py --check-op RopeYarnOp`
- `validate_manifest.py --check-op AvgPool2dFwdOp`
- FlashInfer RoPE semantic smoke: Neox, Non-Neox, Llama 3.1, YaRN, and LongRoPE all matched the benchmark-local reference within fp16 tolerance; YaRN/Llama/LongRoPE local tables also matched the production formula helpers.

lcy-seso:
I rebased this branch onto `main` (`fb95f0a`) and force-pushed it, so the old head `1a28c9d` is gone. The branch was 56 commits behind and the benchmark API changed underneath it, so this was more than a mechanical conflict resolution — please review what landed.

**What changed and why**

The branch was written against `bm.profile(...)` + `BenchmarkReport.record(..., tag=...)`. `main` has since replaced that with `bm.compare({tag: functor, ...}, *inputs, record_as=, params=)`, which times every functor in interleaved passes and records each one under its tag. I re-expressed the two commits on the new API rather than restoring the old calls:

- `torch_inductor_baseline()` is kept in `benchmark_base.py` (plus a `Callable` import), and its unit test is kept.
- Every site that wrapped a baseline in TorchInductor now passes it as a `compare` entry: `{"tileops": op, "torch-inductor": torch_inductor_baseline(fn)}`. I used a hyphen to match the existing `torch-ref` / `torch-cudnn` keys on `main`; the nightly report keys off the tag string, so the name is free — say the word if you want `torch_inductor`.
- flashinfer baselines (rmsnorm, fused_add_rmsnorm, rope) are kept, recorded via `bm.profile` + `BenchmarkReport.record(tag="flashinfer")` alongside the `compare` call.
- `bench_mean_pooling` keeps your condition: the offsets path is not fullgraph-compilable, so it stays on the eager baseline.
- `bench_rope` keeps the variant-correct baselines and the flashinfer correctness assertions from the second commit; the torch baseline goes through `compare` as `_rope_reference(..., variant=, interleave=)`.
- `scripts/nightly_report.py` and `tests/test_nightly_report.py` applied cleanly.

**What I dropped**

- The `torch_cudnn` → `torch_native` tag rename in `bench_pool`: `main` already names that baseline `torch-ref`, so there was nothing left to rename.
- The `AvgPool2dBenchCase` / `MaxPool*BenchCase` classes and the `if __name__ == "__main__"` blocks the diff re-added — `main` deleted both.
- The six `BenchmarkReport` config tests in `benchmarks/tests/test_benchmark_base.py` — `main` removed them when reporting moved to `benchmarks/report.py`. Only `test_torch_inductor_baseline_uses_fullgraph_compile` was carried over.

**Verification**

`ruff` clean. `tests/test_nightly_report.py` (1 passed), `benchmarks/tests/test_benchmark_base.py` (17 passed), `scripts/validate_manifest.py` all pass. I did **not** run the GPU benchmarks — CI is the real check.

**One thing worth a look**

`flashinfer.fused_add_rmsnorm` mutates `x` and `residual` in place. That was already true before, but `compare` runs the functors over several passes, so the tileops and torch entries will see mutated inputs. Timing-wise it may be fine; flagging it since the execution shape changed.

PR #1812, opened 2026-07-31
[Refactor] Consolidate duplicated input construction into the workloads layer

lcy-seso:
Input construction was duplicated across the two stages: 33 `gen_inputs` under `benchmarks/`, 33 more inlined in `tests/`, unreachable from each other because a benchmark may neither write `workloads/` nor import `tests/`.

All of it moves to `workloads/<family>.py`, named for the **input shape** rather than for an op — the shape `workloads/gemm.py` already had. That is what collapses the duplication: 18 op-named classes become 5.

| | before | after |
|---|---|---|
| `gen_inputs` in `benchmarks/` / inlined in `tests/` | 33 / 33 | 0 / 0 |
| workload classes, of which distinct `gen_inputs` | 129 / 113 | 115 / 111 |

Also deletes three protocol classes with no consumer, adds three AST gates for the invariant, and fixes `bench_gla_chunkwise.py::test_gla_bwd_bench`, which timed an autograd fallback through `profile()` (i.e. under `no_grad`) and raised on all five cases. Reasoning is in the commit messages.

## Validation

H200, local — no PR-level CI job runs `benchmarks/ops`, so that evidence cannot come from CI: `pytest tests/ -m smoke` 2617 passed / 0 failed; the touched benchmarks 417 + 358 + 177 passed; gates 23 passed; `ruff check .` clean.

Full 60-file sweep on a dedicated GPU: 3 files failed, against 5 on the baseline taken from unmodified `main` — a strict subset, so no regression. `bench_convolution.py` and `bench_grouped_gemm.py` failed on the baseline only because that run competed for the GPU; here they finish in 21s and 7s. The remaining three hit the 900s per-file cap, two of them already on the baseline.

## Follow-up

tile-ai/TileOPs#1813 — `bench_gla_chunkwise.py` is the third file at that cap. Its torch fallback used to raise immediately and never run; with the fix it executes, and the file crosses 900s. The cap is a hang watchdog rather than a runtime budget, and it can no longer tell a hang from a slow file. Pre-existing calibration issue, not introduced here.


lcy-seso (commented):
(withdrawn)

Ibuki-wind (approved):

Issue #1835, opened 2026-08-05
[FEATURE][GEMM] decide whether GroupedGemmFwdOp needs a nopad persistent path


lcy-seso:
### Symptom / Motivation

`GroupedGemmFwdOp` dispatches only `GroupedGemmKernel`. A second, working and tested implementation — `GroupedGemmPersistentKernel`, single-kernel persistent grouped GEMM with global atomic work-stealing — was reachable from no op and was deleted as part of the unreachable-kernel sweep. It is recoverable at `82535c3d:tileops/kernels/grouped_gemm/grouped_gemm_persistent.py`.

### Root Cause Analysis

The kernel could not be wired into the `grouped_gemm_kernel` dispatch slot, so its absence from `default_kernel_map` was not an oversight to fix by adding a line:

| | `GroupedGemmKernel` (dispatched) | `GroupedGemmPersistentKernel` (deleted) |
| --- | --- | --- |
| `forward` | `(A, B, batch_sizes, batch_offsets, batch_padded_offsets)` | `(A, B, true_sizes, true_offsets)` |
| Row packing | padded | tight |
| Transpose | `transpose_a` / `transpose_b` | none — NT only |

`GroupedGemmFwdOp`'s manifest workloads cover `nn` / `tn` / `nt`. The persistent kernel serves the `nt` + tight-packing subset and needs the op to prepare different metadata.

The MoE side is unaffected: `FusedMoEExpertsNopadPersistent3WGFwdOp` dispatches `GroupedGemmPersistent3WGKernel` and falls back to `MoeGroupedGemmNopadKernel` on unaligned dims, never to the deleted kernel.

## Goal

A recorded decision on whether the persistent path belongs at the op layer, and the corresponding code state.

## Plan

1. Benchmark the persistent kernel against `GroupedGemmKernel` on the manifest's `nt` workloads to establish whether the path is worth having.
2. If worth it: define the dispatch predicate, extend the op to prepare tight metadata, declare the kernel in the manifest `source.kernel_map`, restore the kernel and its correctness test.
3. If not: record the measurement and close.

## Constraints

- Restoring the kernel without an op that dispatches it re-creates the condition the sweep removed.
- `GroupedGemmPersistent3WGKernel`'s K-alignment guard previously pointed callers at the deleted kernel; that pointer is gone. If the persistent path returns, re-establish the guidance at the op layer rather than in a kernel error string.

## Acceptance Criteria

[...]

PR #1836, opened 2026-08-05
[Maintain][Attention] remove kernel classes no op dispatches to

### LINKED ISSUE #1830: [MAINTAIN][ATTENTION] remove attention kernels no op dispatches to
### Symptom / Motivation

Seven attention kernel classes are dispatched to by no op. They are exported from the package `__init__`, so they look live, but nothing reaches them.

### Root Cause Analysis

`MultiHeadAttention*Op` delegates to a wrapped GQA op rather than to MHA-specific kernels, leaving the whole MHA kernel set stranded — including its Hopper variants, so this is not a leftover architecture-compatibility path. The three `GQAPrefillWithKVCache*` kernels were superseded by their paged counterparts.

Zero references outside the `__init__` exports: no op, no test, no benchmark, no manifest `source.kernel_map` entry.

| Kernel | File |
| --- | --- |
| `MHAFwdKernel`, `MHAFwdWgmmaPipelinedKernel` | `tileops/kernels/attention/gqa_fwd.py` |
| `MHABwdKernel`, `MHABwdWgmmaPipelinedKernel` | `tileops/kernels/attention/gqa_bwd.py` |
| `GQAPrefillWithKVCacheFwdKernel`, `GQAPrefillWithKVCacheRopeFwdKernel`, `GQAPrefillWithKVCacheRopeAppendKernel` | `tileops/kernels/attention/gqa_fwd.py` |

### Second group: non-Hopper fallbacks

Removing `is_hopper()` (the library targets H200 only) leaves six more kernels with no default path selecting them. They differ from the seven above: each is still exported and still passes the arch check on H200, so a caller can reach one through a `kernel_map` override. The open question is whether an overridable non-Hopper fallback is worth keeping at all.

| Kernel | Hopper counterpart now dispatched |
| --- | --- |
| `GQAFwdKernel` | `GQAFwdWgmmaPipelinedKernel` / warp-specialized variants |
| `GQABwdKernel` | `GQABwdWgmmaPipelinedKernel` |
| `FlashAttnBwdPostprocessKernel` | none — the Hopper path casts `dq` instead |
| `GQASlidingWindowFwdKernel` | `GQASlidingWindowFwdWgmmaPipelinedKernel` |
| `GQASlidingWindowVarlenFwdKernel` | `GQASlidingWindowVarlenFwdWgmmaPipelinedKernel` |
| `MLADecodeKernel` | `MLADecodeWsKernel` |

`supported_archs == [80, 89, 90]` does not identify this group: `GQADecodeKernel`, `GQAPrefillFwdKernel` and others carry the same declaration while being the only implementation of their slot. The test is whether a Hopper counterpart exists and the kernel was reachable only through the non-Hopper branch.

### Related Files

- `tileops/kernels/attention/gqa_fwd.py`, `gqa_bwd.py`, `mla.py`, `sliding_window.py`
- `tileops/kernels/attention/__init__.py` (exports and `__all__`)

## Goal

No kernel class ships that no op can reach.

## Plan

1. Confirm each of the seven is unreachable at HEAD, including via `kernel_map` overrides in tests.
2. Delete the classes and their `__init__` exports.
3. Decide separately whether MHA should regain its own kernels or keep delegating to GQA — deleting the stranded set does not answer that.

## Constraints

- No behaviour change: every deleted class is currently unreachable.
- Do not delete a kernel that is merely unused on H200 but still the only implementation for its slot.

## Acceptance Criteria

- [ ] The seven unreachable classes and their exports are gone
- [ ] A verdict is recorded on the six non-Hopper fallbacks — removed, or kept with the reason stated
- [ ] A repo-wide scan finds no kernel class absent from every op's `default_kernel_map`

lcy-seso:
Closes #1830

Deletes 14 public `Kernel` classes no op could reach. −3567 / +211 lines; the additions are a PyTorch oracle replacing a kernel-vs-kernel test reference. No product logic rewritten.

## What was deleted

| Group | Classes | Why |
| --- | --- | --- |
| MHA-specific | `MHAFwdKernel`, `MHAFwdWgmmaPipelinedKernel`, `MHABwdKernel`, `MHABwdWgmmaPipelinedKernel` | `MultiHeadAttention*Op` delegates to GQA with `heads_kv == heads`; these are the pre-unification duplicates |
| Contiguous KV-cache prefill | `GQAPrefillWithKVCacheFwdKernel`, `GQAPrefillWithKVCacheRopeFwdKernel`, `GQAPrefillWithKVCacheRopeAppendKernel` | Superseded by the paged path; restore path recorded in #1839 |
| Non-Hopper fallbacks | `GQAFwdKernel`, `GQABwdKernel`, `GQASlidingWindowFwdKernel`, `GQASlidingWindowVarlenFwdKernel`, `MLADecodeKernel` | Hopper counterparts are the dispatched path — see the audit below |
| Structurally unusable | `FlashAttnBwdPostprocessKernel`, `GroupedGemmPersistentKernel` | Cannot be wired into any existing slot — see below |

**`FlashAttnBwdPostprocessKernel`** annotates the *global* `dq` with `make_dq_layout()`, the Ampere 8×8 fragment reorder. Only the two deleted Ampere backward kernels ever wrote `dq` that way; `GQABwdWgmmaPipelinedKernel` `atomic_add`s into plain row-major `dq`. Wiring it into the Hopper path would read the wrong layout and produce wrong numbers.

**`GroupedGemmPersistentKernel`** takes tight `(A, B, sizes, offsets)` with no padded offsets and no transpose parameters; `GroupedGemmFwdOp`'s slot passes `batch_padded_offsets` and supports `nn`/`tn`/`nt`. Tracked in #1835.

## Capability audit

`supported_archs` is hard-enforced — `Op._install_kernel_map` raises for any resolved entry excluding the current device, defaults and overrides alike, for every key in the map. Measured by forcing `get_sm_version()` to 80:

| op | `main` default | `main` + Ampere override | this branch |
| --- | --- | --- | --- |
| `GroupedQueryAttentionBwdOp` | rejected | **constructs** | rejected |
| `MultiHeadAttentionBwdOp` | rejected | **constructs** | rejected |
| `MultiHeadLatentAttentionDecodeWithKVCacheFwdOp` | rejected | **constructs** | rejected |
| `GroupedQueryAttentionSlidingWindowFwdOp` | rejected | **constructs** | rejected |
| `GroupedQueryAttentionSlidingWindowVarlenFwdOp` | rejected | **constructs** | rejected |
| `GroupedQueryAttentionFwdOp` / `MultiHeadAttentionFwdOp` | rejected | rejected — nested sub-op rebuilds a Hopper-only map | rejected |

Default-path Ampere support was already gone at `main` — removing `is_hopper()` took it, not this PR. What this PR removes is the manual override escape hatch, which for five ops was their last sm80/sm89 path.

That is acceptable because the deleted kernels were **not an H200 fallback**: no `_select_*_kernel_cls` returns them, and their shape guards are identical to the surviving Hopper counterparts, so no shape regime needs them. They declare `[80, 89, 90]` and do run on H200 — as slower duplicates reachable only by hand-writing a `kernel_map`.

**Not lost:** MHA (fully covered by the GQA delegation, including `(output, lse)`, backward, and the same autotune surface) and K-unaligned tight-packed NT grouped GEMM (`MoeGroupedGemmNopadKernel` / `MoeGroupedGemmNopadFwdOp`, tested at `K=96`).

**Genuinely removed:** the contiguous KV-cache prefill ABI. The paged survivors express the same mathematics but require a page pool and block table. No manifest entry — implemented or `spec-only` — describes it, so nothing in the spec dangles. Recorded in #1839.

## Also fixed

Dead import in `bench_gqa.py`; the varlen Ampere factory / wrapper / custom-op registration left behind by the class deletion; two stale references to deleted classes (`grouped_gemm_persistent_3wg.py` error string and module docstring); the `docs/design/manifest.md` example, which named two deleted kernels and paired `gqa_bwd_*` keys with `op: mha.py`.

`tests/kernels/test_grouped_gemm_persistent_3wg.py` validated the 3WG kernel against another kernel, which the trust model does not allow. Replaced with a PyTorch oracle; node count unchanged (21), one test renamed.

## Verification

- [x] Modified files pass unit tests — attention 192, 3WG 21, grouped_gemm + MoE + manifest 169, sliding-window 35; `scripts/validate_manifest.py` clean; CI green
- [x] The seven unreachable classes and their exports are gone
- [x] Verdict recorded on the six non-Hopper fallbacks — all removed, `FlashAttnBwdPostprocessKernel` on layout-coupling grounds rather than disuse
- [x] Repo-wide scan finds no kernel class absent from every op's `default_kernel_map` — 213 public classes, 0 unreachable, excluding four elementwise intermediate base classes reachable through their subclasses

The scan was a one-off verification, not committed as a script or CI check — say the word if you want it as a standing structural test.


tileops/kernels/attention/gqa_sliding_window_varlen_fwd.py
```diff
@@ -331,41 +330,6 @@ def _call_wrapped(self, wrapped_fn, q, k, v, cu_seqlens_q, cu_seqlens_k,
...
-        cu_seqlens_q: torch.Tensor,
-        cu_seqlens_k: torch.Tensor,
-        max_seqlen_q: int,
-    ) -> Tuple[torch.Tensor, torch.Tensor]:
-        return self._call_wrapped(
-            _gqa_sw_fwd_varlen_wrapped_kernel,
-            q, k, v, cu_seqlens_q, cu_seqlens_k, max_seqlen_q)
-
-
 # ── WGMMA Pipelined (SM90) ────────────────────────────────────────────────────
```
Ibuki-wind, line 154: Deleting `GQASlidingWindowVarlenFwdKernel` leaves its only implementation path — `_gqa_sw_fwd_varlen_kernel`, `_gqa_sw_fwd_varlen_wrapped_kernel`, and the fake registration above — with no caller, while import still registers the dead `top::gqa_sw_fwd_varlen_wrapped_kernel` custom op. → Delete that non-WGMMA factory/wrapper/fake-registration block as part of removing the class.

tileops/kernels/grouped_gemm/grouped_gemm_persistent_3wg.py
```diff
@@ -1029,8 +1029,7 @@ def _persistent_grouped_gemm_v2_kernel(numel, num_experts, N, K, dtype,
     if K % block_k != 0:
         raise ValueError(
             f"GroupedGemmPersistent3WGKernel requires K-alignment "
-            f"(K={K} must be divisible by block_k={block_k}). "
-            f"Use GroupedGemmPersistentKernel for K-unaligned shapes."
+            f"(K={K} must be divisible by block_k={block_k})."
```
Ibuki-wind, line 1032: The runtime error no longer recommends the deleted fallback, but the module docstring at line 13 still says K-unaligned callers should use `GroupedGemmPersistentKernel`; following that guidance now fails to import. → Remove or replace the stale fallback guidance in the module docstring too.

Ibuki-wind (changes requested):
### Overall
Two collateral cleanup gaps remain in the deletion and should be resolved before merge.

Ibuki-wind (changes requested):
### Overall
The code cleanup and prior inline gaps are resolved, but the Capability audit must correct its sm80 construction matrix before merge.

### Cross-cutting concerns
- **Medium — the MHA row and resulting path count are factually wrong.** At `main`, with `get_sm_version()` forced to 80, `MultiHeadAttentionBwdOp(batch=1, heads=8, seq_len=128, dim=64, kernel_map={"gqa_bwd_preprocess_kernel": FlashAttnBwdPreprocessKernel, "gqa_bwd_kernel": GQABwdKernel})` constructs successfully. The op explicitly accepts `gqa_bwd_*` keys and passes that map to `GroupedQueryAttentionBwdOp`; unlike MHA forward, there is no nested larger default map that defeats the override. Therefore the combined `MultiHeadAttention{Fwd,Bwd}Op` row cannot say both are rejected, and this PR removes the override-only Ampere path from five ops, not four. → Split the MHA row (forward rejected, backward constructs) and update the paragraph/count accordingly.

Ibuki-wind (approved):
Clean — no issues.

PR #1838, opened 2026-08-05
[Bench][Timing] Align benchmarks with native CUPTI

superAngGao:
Replace Kineto projection timing with native CUPTI activity timing, through NVIDIA's `cupti-python` rather than a hand-written extension.

A discovery pass records the per-call activity sequence (kernel, MEMCPY, MEMSET); timed repeats measure each call from its earliest to its latest activity, so a single-activity call gives pure kernel duration and a multi-activity call keeps its inter-kernel gaps. Prepare activities are validated, then excluded. Persisting L2 is reset and a 2x-L2 buffer cleared before every iteration, with inputs rotated through a shifting pool.

Iteration counts come off a wall-time budget rather than a constant: a calibration pass measures one iteration, then warmup and measurement each run for their budget, clamped to `[10, 200]`. A 5 us op is sampled 200 times instead of 50; a millisecond op stops paying for 50. `bench_kernel` returns every sample, so the caller owns the statistic.

Attribution fails closed: a missing, extra, or reordered activity raises and names the observed sequence. `TILEOPS_ALLOW_CUDA_EVENTS_FALLBACK=1` opts into diagnostic CUDA-event timing.

Each result carries `timing`, `n_samples` and `latency_p10_ms` / `latency_p90_ms` around its median, all of which reach the JUnit XML: a median whose spread is unknown, or one that came from the fallback, cannot be read as a latency.

`BenchmarkBase.compare()` times several implementations forward then reversed, so drift across a case lands on all of them instead of on whichever ran last.

## Validation

- Six full-suite runs on the fixed-count version: 1,153 cases each, zero fallbacks. Moving from `3 x 50` trial-means to a single trial's median shifted nothing — geomean `-0.53%`, median absolute difference `0.54%`.
- GQA decode `llama-3.1-70b-32k-float16`: `0.2172` / `0.2168` / `0.2220 ms`.
- After rebasing onto main (H200): `benchmarks/tests` 22 passed; `bench_norm.py` 30 passed / 3 xpassed, all 66 rows `timing: cupti`, no fallback.

Follow-up: baseline correctness and competitive-baseline coverage in #1880.


superAngGao:
Implementation notes

This PR changes the benchmark timing path in a few places:

1. Native CUPTI attribution replaces Kineto projection for the default timing path.

   - Discovery records the expected prepare and operator CUDA activity sequences separately.
   - One CUPTI session captures the complete ordered activity trace for each pytest benchmark case.
   - Timed trials use activity-index checkpoints plus a deterministic prepare/operator sequence ledger; prepare activities are validated but excluded from operator latency.
   - Latency is measured as the selected GPU activity envelope: earliest selected operator-kernel start to latest selected operator-kernel end.
   - For a single-kernel op, this is equivalent to pure kernel duration.
   - For a multi-kernel op, this intentionally includes inter-kernel gaps, matching the SOL-style operator latency semantics.

2. CUDA-event fallback is no longer silent by default.

   - Default behavior is fail-closed if native CUPTI attribution cannot produce a valid result or drops activity records.
   - CUDA-event fallback is still available for local diagnosis with `TILEOPS_ALLOW_CUDA_EVENTS_FALLBACK=1`.
   - JUnit/nightly artifacts include backend, benchmark protocol, sampled calls, expected operator-kernel count, case margin, input policy, and fallback reason.

3. Input policy is changed from 3-clone reuse to a SOL-style shifting tensor pool.

   - The goal is to avoid repeatedly benchmarking exactly the same data pointers while still keeping the benchmark deterministic.
   - Input-pool copies and L2 reset/flush kernels are captured as the expected prepare sequence and validated, but they are excluded from the selected operator envelope.
   - Conv1d/Conv1dBias keep weight/bias static while shifting only the dynamic input, because shifting weights interacted with the op's internal prepack/cache behavior and changed the observed activity sequence.

4. Case-scoped attribution details.

   - The default case margin is 1 ms at both ends (`TILEOPS_CUPTI_CASE_MARGIN_US=1000`).
   - Session start is `synchronize -> start/reset -> timestamp -> leading margin`; session stop is `synchronize -> trailing margin -> timestamp -> flush/disable`.
   - Per-repeat begin/end tolerances and the 96 us repeat guard are removed. They are not needed because logical calls are attributed by ordered activity-index ranges rather than overlapping CPU timestamp windows.
   - No extra pytest inter-case guard is required: the previous session is synchronized, flushed, and disabled before the next session resets collector state and starts its own leading margin.

   The 1 ms margin is supported by an empirical interval estimate of the CPU/GPU synchronization-ruler error. For one short logical call with ruler error `epsilon`:

   `GPU_end - CPU_end <= epsilon <= GPU_start - CPU_begin`

   On H200/CUDA 12.9, 14,550 clean repeats were grouped into 291 trials of 50 repeats. The known externally disturbed MLA process was excluded. Intersecting the 50 repeat intervals in every trial produced 291/291 valid ruler-error intervals:

   - Typical trial (median feasible interval): `[-4.10 us, +23.25 us]`
   - Approximate 98% empirical envelope: `[-51.28 us, +133.06 us]`
   - Full observed conservative envelope: `[-61.89 us, +161.00 us]`

   The 1 ms margin therefore retains at least `1000 / 161 ~= 6.2x` empirical headroom over the widest observed positive bound. This is an empirical bound for the tested software/hardware stack, not an NVIDIA-guaranteed maximum. GPU clock locking is not required for this timestamp-containment estimate; latency values from that diagnostic run were not used for performance comparison.

5. Runner/reporting changes.

   - `run_benchmarks.py` passes `--assert=plain` to avoid assert-rewrite/import behavior interfering with benchmark collection in this environment.
   - The runner assigns each benchmark file its own trace directory; each pytest case emits one compressed `.json.gz` CUPTI trace.
   - `nightly_report.py` parses and summarizes the benchmark provenance fields.

About the multi-kernel outliers in the validation table:

- The attribution/count/backend path was stable: all three full rounds had 3472/3472 native-CUPTI entries and 50/50 sampled calls.
- Single-kernel results were stable enough for the small-kernel timing use case this PR primarily protects.
- Some multi-kernel span entries had large relative range. That is not an attribution failure by itself: under the SOL-style metric, multi-kernel latency includes the gaps between kernels.
- We did not continuously sample GPU processes during the full run, so the PR does not claim whether those outliers came from external GPU contention or from op/runtime scheduling. A focused rerun of the outlier files kept native CUPTI 50/50 and did not reproduce the largest GQA sliding-window-varlen spike.

benchmarks/benchmark_base.py
```diff
@@ -50,66 +54,502 @@ def _workload_contract(op_name: str) -> tuple[str, frozenset[str]]:
...
+            f"CUPTI dropped {trace['dropped']} records during timing"
+        )
+
+    samples_us: list[float] = []
+    grouped = _kernels_by_cpu_window(trace)
+    for kernels in grouped:
+        selected = _select_expected_sequence(kernels, expected_sequence)
+        if selected is None:
+            continue
+        first = selected[0]
```
stelladuyx: The last kernel to start is not necessarily the last one to finish. Maybe we can use min(start) and max(end) across all selected kernels to compute the complete activity envelope, especially for kernels running concurrently on multiple streams.

benchmarks/benchmark_base.py
```diff
@@ -50,66 +54,502 @@ def _workload_contract(op_name: str) -> tuple[str, frozenset[str]]:
...
-    for evt in kineto_results.events():
-        if evt.device_type() != DeviceType.CUDA:
+        block_offset = self._offset_blocks[self._call_idx]
+        result: list[Any] = []
+        for entry in self._entries:
+            if "scalar" in entry:
+                result.append(entry["scalar"])
+                continue
+            start = block_offset * entry["block_numel"]
+            entry["pool"].narrow(0, start, entry["storage_span"]).copy_(entry["source"])
```
stelladuyx, line 148: Even when this is emitted as a MEMCPY and ignored by the kernel-only CUPTI
collector, it can repopulate or otherwise perturb L2, so the benchmarked
operator no longer starts from the intended cold-L2 state. If copy_() is
lowered to a CUDA kernel for a particular layout or backend, it may also become
part of the discovered kernel sequence and measured span.

benchmarks/benchmark_base.py
superAngGao: Good point. Updated the timing sample to use the selected activity envelope, `min(start_ns)` to `max(end_ns)`, instead of relying on selected-list order. This covers the concurrent-stream case where the last kernel to start is not the last one to finish.

benchmarks/benchmark_base.py
superAngGao: Agreed. Updated the shifting-pool path so `copy_()` happens in `prepare_one` before persisting-L2 reset, L2 flush, and the CUPTI begin timestamp. That keeps the input-copy work outside attribution and prevents it from warming the post-flush cache state used by the measured operator.

bigSheep123:
Hi @superAngGao — great to see the move off Kineto projection onto native CUPTI. I've been chasing the nightly partial-trace problem on our H200 and this PR removes the whole projection-window failure class. While validating it I hit one portability issue in the attribution primitive, reworked it on top of your branch, and ran a fairly heavy validation campaign. Sharing the branch and the data here.

**Branch:** `bigSheep123/TileOPs@pr-1838-extcorr` (3 commits on top of `65a3faf`) — happy to open a PR against your branch, or feel free to cherry-pick.

## Issue: CPU-window attribution trips over clock-domain skew

The PR attributes kernel activity records to repeats by testing GPU activity timestamps against a CPU timestamp window `[begin − 2µs, end + 8µs]` (plus a 16µs inter-repeat spin guard). That compares timestamps across the CPU and GPU clock domains, and the tolerances are empirical.

On our box the skew lands right on the boundary: measured CPU↔GPU offset is **8.1–8.9µs against the 8.0µs end tolerance**. The 0.1–0.9µs that falls on the wrong side pushes the last kernel out of the window, discovery sees an empty/incomplete sequence, and the test fails. In a 59-file A/B run, **~13 tests (rsqrt / rope / log_softmax / conv2d / dsa_decode) fail only with window attribution** — reproducibly. Any driver/machine with different skew will move this boundary, so the knobs can't be tuned once and trusted.

## Rework: attribute by launch-callback correlation instead of clocks

- Subscribe a CUPTI callback to just the **six kernel-launch cbids** (runtime: `cudaLaunchKernel` / `ExC` / `Cooperative`; driver: `cuLaunchKernel` / `Ex` / `Cooperative`). On API enter, record the launch's `correlationId` against the currently active scope — `(repeat, i)` or `prepare`.
- The scope is **process-global**, so backward kernels launched from the autograd engine worker thread attribute correctly.
- Timing values still come from `CONCURRENT_KERNEL` activity records exactly as in your PR — only the *grouping* changed: a pure `correlationId` equality join, no clock ever compared across domains. The three tolerance env knobs and the spin guard go away; provenance gains `cupti_attribution=launch-correlation`.
- **No RUNTIME/DRIVER activity tracing.** An earlier iteration of mine used `EXTERNAL_CORRELATION` push/pop instead; it had two problems worth recording: (a) the marker is thread-local, so autograd-thread backward launches missed it (~24 bwd/fwdbwd tests failed closed), and (b) it requires API activity tracing, which added 2–6µs per inter-kernel gap on driver-API multi-kernel ops (gqa_decode fp16 +9%, fused_topk +30%). The launch-callback design fixes both; the callback body is two relaxed atomic loads plus one vector append and showed no measurable perturbation (ratio 1.000 on the affected ops).
- Fail-closed is preserved and strengthened: kernels matching no scope raise an explicit attribution error with a trace dump — nothing is silently dropped or mis-binned.
- Known boundary: launches through unhooked APIs (e.g. `cudaGraphLaunch`) would land in `unattributed` and fail loudly rather than mis-time. No op in the current suite launches that way (zero fallbacks across all runs below).

## Validation data (H200, driver 595.71.05, torch 2.10.0+cu129, CUPTI 12.9.79)

**A/B, both versions** — 59 bench files × 1161 tests, warm TileLang cache, run concurrently on two GPUs:

- Outcomes: the rework has **zero unique failures**. The only outcome difference is the window-skew failure class above (fails on the original, passes on the rework). The remaining 33 failures are identical on both sides and pre-date this PR (list below).
- Latency: per-test ratio (rework/original), n=1096: **median 1.0000, p10/p90 = 0.989/1.008**. The handful of >20% outliers are bidirectional and sit on ops we separately confirmed to be clock/algorithm-bistable on this unlocked-clock box.
- Provenance: 100% `native-cupti` on both sides, zero CUDA-events fallbacks, full 50/50 sampling everywhere.

**Reliability, rework only** — 3 further full runs on three different GPUs, same day:

- Outcomes identical across all three runs and identical to the A/B run (1074 pass / 54 skip / 33 pre-existing fail — same test list each time).
- 3 × 3281 timing provenance entries: all `native-cupti` + `launch-correlation`, zero fallbacks, fail-closed never triggered.
- Latency reproducibility: same-clock-state GPU pair median ratio **1.0000** (p10/p90 0.957/1.021); cross-session (different day, different GPU, same clock state) median **1.0000** (p10/p90 0.978/1.013).
- Caveat: our machine has no clock locking (no sudo). We observed two GPU clock bins ~7–9% apart; the offset hits tileops and baseline entries symmetrically, so it's environmental. I'd recommend one locked-clock pass before merging — absolute numbers, not the mechanism, are what it affects.

## Pre-existing issues found during validation (all reproduce identically on both versions — unrelated to the timing rework)

1. **`bench_gqa_fp8.py` 70B configs deadlock on GPU.** `test_gqa_prefill_fp8_tensor_core_bench[llama-3.1-70b-prefill-fp8tc-bn224-s3584-float16]` (and the bf16 variant) hangs forever: GPU pegs 100% utilization at idle-level power (~120W), host spins in `synchronize`. It reproduces deterministically within 60s, and also hangs in a **plain `op(*inputs); torch.cuda.synchronize()` call with no profiler and no bench harness** — so it's a kernel-side deadlock (WS producer/consumer barrier is my guess), not a timing/CUPTI issue. The 8B configs are fine (4/4 pass on both versions). Suggest a per-test timeout on this file so CI can't hang indefinitely.
2. `bench_gqa_fp8.py` records via `BenchmarkReport.record` directly, bypassing the provenance injection in `BenchmarkBase._build_result` — its junit entries carry no `timing`/`cupti_attribution` keys, so the nightly report can't verify how it was measured.
3. Shifting-pool `as_strided` views drop `requires_grad` → the 12 `vs_fla_bwd` tests fail on both versions.
4. Three sigmoid/silu manifest configs fail TileLang compilation (consistent across five sweeps on this environment).
5. Cold-compile budgets: conv / grouped_gemm / gqa_fp8 exceed 1h per file without a warm cache; CI per-file budget or cache strategy needs adjusting.
6. From earlier review rounds, still open: `cuda-python` is an undeclared dependency; the CUDA-events fallback is detected by a `"CUPTI"` substring match on the exception; the input pool has no memory cap; nightly history contains false regressions from the pre-rework mixed-timing era.

## Commits

- `d3f234f` — external-correlation attribution (first pass, superseded; kept for history)
- `322afd0` — launch-callback attribution (the fix described above)
- `22baa75` — comment accuracy fix

Happy to walk through any of the data or rerun specific configs.

superAngGao:
### Phase-isolated benchmark measurement timeline

For one parametrized pytest benchmark case, the current measurement lifecycle is:

```text
pytest starts one benchmark case
│
├─ Initialize the case attribution ledger
│   ├─ Build/load the native CUPTI extension outside activity capture
│   └─ CUDA synchronize; no CUPTI collector is active yet
│
├─ Benchmark TileOps (`bench_kernel`)
│   ├─ Open one op-ledger entry
│   ├─ First call                         ← uncollected; absorbs lazy init/JIT
│   ├─ Warmup × 10                       ← uncollected
│   │
│   ├─ Discovery collector session
│   │   ├─ Enable CUPTI and clear collected activity
│   │   ├─ Repeat × 3
│   │   │   ├─ checkpoint → prepare iteration → checkpoint
│   │   │   └─ run operator → checkpoint
│   │   └─ Synchronize, force-flush, disable CUPTI, and retain the capture
│   │
│   ├─ Timing collector session
│   │   ├─ Enable CUPTI and clear collected activity
│   │   ├─ Begin activity-index checkpoint
│   │   ├─ Repeat × 50
│   │   │   ├─ Rotate to the next shifting-pool input
│   │   │   ├─ Reset persisting L2 and clear the 2×-L2 buffer
│   │   │   ├─ Run the operator
│   │   │   └─ CUDA synchronize          ← separates logical repeats
│   │   ├─ End activity-index checkpoint
│   │   └─ Synchronize, force-flush, disable CUPTI, and retain the capture
│   │
│   ├─ Match all 50 repeats against the discovered prepare/operator ledger
│   ├─ Report `median(50 per-repeat activity envelopes)`
│   └─ Close the op-ledger entry
│
├─ Benchmark each baseline
│   └─ Run the same first-call, warmup, discovery, and timing lifecycle
│
├─ Run any remaining test logic and attach benchmark metadata
│
├─ CUDA synchronize
├─ Serialize the case ledger and its phase captures
├─ Save one compressed case trace (`.json.gz`) when trace output is enabled
└─ Release the per-case CUDA cache
```

There is no CPU timestamp window or case-boundary sleep in the current protocol.
The case scope is a ledger and persistence boundary, not a CUPTI timing window.
Discovery and timing use separate native-CUPTI collector sessions. Attribution is
based on activity-index checkpoints plus the discovered prepare/operator sequence;
each reported sample spans from the first selected operator activity start to the
last selected operator activity end, so multi-activity gaps are preserved.

superAngGao:
Hi @bigSheep123 — thank you for the detailed investigation and validation data. One important update: your branch and analysis are based on `65a3faf`, while the current PR head is `e7cf495`, and the attribution primitive has since been replaced.

The current implementation no longer attributes repeats with `[CPU begin - tolerance, CPU end + tolerance]` windows. In particular:

- one CUPTI activity session covers the complete pytest benchmark case, with a 1 ms leading and trailing margin;
- discovery records the expected prepare and operator activity sequences separately;
- timed trials are partitioned by flushed CUPTI activity-index checkpoints and validated against a deterministic `prepare + operator` sequence ledger;
- the old per-repeat `2/8 us` tolerances and `16 us` spin guard have been removed;
- attribution uses neither launch callbacks nor external-correlation IDs, so it also avoids callback/API-tracing perturbation and thread-local propagation concerns;
- multi-kernel latency remains the GPU activity envelope from `min(start)` to `max(end)`, including real inter-kernel gaps.

The CPU/CUPTI timestamps now describe only the outer case envelope; they are not used to decide which kernels belong to an individual repeat. We also measured the synchronization-ruler residual on 14,550 clean repeats: the full observed conservative interval was `[-61.89 us, +161.00 us]`, leaving at least about 6.2x empirical headroom inside the 1 ms case margin on that H200/CUDA 12.9 setup.

So the clock-domain portability failure you found in the old commit should no longer apply to the current head, and the launch-callback rework should not be necessary for attribution. If you have bandwidth, rerunning the approximately 13 skew-sensitive cases against `e7cf495` would be especially useful as an independent check. Thanks again for catching and characterizing the original failure mode.

superAngGao:
### SOL-aligned timing follow-up

Commit [`8c5a28d`](https://github.com/superAngGao/TileOPs/commit/8c5a28df6ad93d5d43bb460a427406b3f6c2a917) makes the validated **phase-isolated native-CUPTI protocol** the default benchmark path:

- one phase-scoped CUPTI capture for discovery and one fresh capture for timing;
- `CONCURRENT_KERNEL`, `MEMCPY`, and `MEMSET` activity collection;
- one set of 50 timed samples summarized with `median(samples)`, matching the current SOL-ExecBench aggregation;
- complete protocol provenance in JUnit and the nightly report.

The **legacy case-scoped capture protocol** remains selectable for diagnosis through `TILEOPS_BENCHMARK_PROTOCOL=native-cupti-case-sequence-v1`.

Validation of the production diff:

- 37 attribution/provenance unit tests passed;
- the GQA decode-paged smoke file passed with the phase-isolated protocol selected implicitly (no protocol environment variable): 8 passed, 3 skipped;
- the resulting trace contained the expected four captures (`TileOps discovery`, `TileOps timing`, `baseline discovery`, `baseline timing`), with 50/50 timing samples and zero dropped records;
- six completed phase-isolated full-suite runs produced 1,153 case traces per run with zero CUPTI fallback, dropped records, malformed traces, or invalid phase traces.

#### Conclusions from the full-suite comparison

1. **Changing from three groups of 50 repeats summarized as the median of trial means to one group of 50 repeats summarized as `median(samples)` does not introduce a systematic latency shift.** In a same-isolation comparison of the legacy case-scoped protocol and the phase-isolated protocol over 1,107 TileOps latency entries, the new/legacy geometric-mean shift was about `-0.53%`, the signed median shift was `-0.22%`, and the median absolute per-case difference was `0.54%`. The new aggregation also matches the current SOL-ExecBench timing summary.

2. **Stronger pytest/import/process isolation does not systematically change GPU kernel latency, but it substantially increases benchmark wall time.** In the second-round comparison, benchmark-file process -> pytest case changed the geometric mean by `+0.34%`, and pytest case -> fresh process per case changed it by `+0.10%`. The main effect is harness cost, not measured GPU latency.

3. **Adding MEMCPY/MEMSET is a measurement-correctness change.** It does not add new GPU work. Pure-kernel operators retain the same kernel envelope; for an operator that already contains copies or sets, the phase-isolated protocol measures the complete selected GPU activity envelope instead of under-reporting a kernel-only interval. Real traces exercised both paths: a GQA operator discovery sequence contained `8 kernels + 1 memcpy`, and an MLA baseline discovery sequence contained `6 kernels + 1 memset`.

#### Observed full-benchmark wall time

These are runner wall-clock times for the same 59-file suite on the H200 host. They include collection/import/harness work and should be read as operational benchmark cost, not as an isolated kernel-performance metric. Compilation and shared cache state were not reset between every run.

| Timing protocol / process isolation | Round 1 | Round 2 | Average | Relative cost |
| Legacy case-scoped capture, benchmark-file subprocess | 55m10s | 52m57s | 54m04s | 1.00x |
| Phase-isolated capture, benchmark-file subprocess | 46m07s | 33m51s | 39m59s | 0.74x vs legacy case-scoped capture |
| Phase-isolated capture, per-case pytest session in one file process | 47m01s | 46m42s | 46m52s | 1.17x vs phase-isolated file process |
| Phase-isolated capture, fresh subprocess per case | 145m41s | 145m32s | 145m37s | 3.64x vs phase-isolated file process; 3.11x vs per-case pytest session |

The practical default therefore remains one subprocess per benchmark file. Fresh-process-per-case isolation is useful as a diagnostic mode, but its roughly 3-4x suite-time cost is not justified for nightly timing when it does not materially change the GPU latency distribution.

michaelwithu (dismissed):
Reviewed the current native-CUPTI/SOL timing implementation and found no correctness blockers. The phase boundaries, activity-envelope latency, and fail-closed attribution behavior are consistent with the intended design.

Non-blocking cleanup suggestions:

1. benchmarks/benchmark_base.py: the public tuple validation duplicates the identical check retained in _bench_kernel_impl. Since the wrapper is the only production caller, keeping validation in one layer (preferably the public wrapper) would avoid redundant code.
2. benchmarks/native_cupti.py: the Python torch.cuda.synchronize() immediately before NativeCUPTIExtension.stop() is redundant because native stop() already calls cudaDeviceSynchronize() before flushing CUPTI activity records.
3. benchmarks/native_cupti.py: the discovery-side torch.cuda.synchronize() immediately before checkpoint() is likewise redundant because the native checkpoint() already synchronizes before its CUPTI flush.

Local focused validation: 37 tests passed. One additional test could not exercise the change because this environment does not have ninja, so the native CUPTI extension build stopped early.

superAngGao:
Thanks for the careful review! I applied all three cleanup suggestions in bca51f5: tuple validation now lives only in the public wrapper, the redundant Python synchronizations before native stop/checkpoint are removed, and the C++ checkpoint comment is updated. Focused tests passed (38/38); H200 GQA sliding-window and BatchNorm smoke A/B both retained 50/50 samples with zero dropped records and no fallback.

PR #1850, opened 2026-08-07
[Bench][CUPTI] Add dynamic dummy warmup

michaelwithu:
## Summary

This PR makes the existing Kineto-based benchmark path robust enough for
repeated benchmark sessions and exposes two explicit kernel metrics from the
same trace:

- `kernel_sum_ms`: sum of attributed kernel durations;
- `kernel_span_ms`: last kernel end minus first kernel start.

`latency_ms` remains an alias of `kernel_sum_ms` for backward compatibility,
and `latency_metric=kernel_sum` makes that choice explicit.

The implementation also retains the dynamic dummy-prefix mitigation for the
Kineto/CUPTI timestamp issue tracked by
[pytorch/pytorch#192021](https://github.com/pytorch/pytorch/issues/192021),
bounds large traces by kernel fan-out, excludes memcpy/memset activities, and
accepts calls whose annotation is projected onto several CUDA streams.

This PR changes only `benchmarks/benchmark_base.py`.

## Dynamic Kineto collection

Repeated `torch.profiler.profile()` sessions can lose a growing prefix of CUDA
activities after Kineto re-registers its CUPTI timestamp callback. A dynamic,
indexed dummy prefix absorbs that invalid activity prefix. Dummy activities are
never included in the reported latency, and an incomplete set of measured
repeat IDs is never accepted.

Before collecting the full trace, a short discovery trace determines how many
kernels one call launches. Normal one/few-kernel calls retain the requested 50
repeats per trial. High-fan-out calls are bounded by a 10,000-activity trace
budget, with a minimum of 5 repeats per trial. Discovery results are cached by
call signature and reused when possible.

One Kineto trace produces both sum and span. Every kernel is attributed once to
an indexed repeat; memcpy/memset records are excluded. If one repeat is
projected once per CUDA stream, duplicate annotation IDs are accepted while
kernel activities are still counted only once.

## Comparison with SOL-style native CUPTI

The control implementation is the SOL-ExecBench-style native CUPTI collector:
SOL describes the benchmark protocol, while native CUPTI is the collection
backend. It does not use Kineto or `torch.profiler`.

Two full H200 runs were performed and then repeated after swapping GPU6/GPU7.
The comparison below uses only four-way parameter matches whose Dynamic and
native-SOL kernel counts agree: 772 records, consisting of 417 single-kernel
and 355 multi-kernel records.

### Single-kernel calls

For one kernel:

```text
kernel sum = kernel span = kernel duration
```

Dynamic/Kineto and native SOL agree closely:

| GPU | N | Median signed difference | Median absolute difference | P90 absolute difference | Within 3% | Within 5% |
| 6 | 417 | 0.000% | 0.298% | 1.923% | 93.8% | 96.9% |
| 7 | 417 | 0.000% | 0.289% | 1.639% | 95.4% | 98.8% |

This validates Kineto's device timestamp/duration extraction for a single
kernel. There is no aggregate directional bias between the collectors.

### Multi-kernel calls: kernel sum

For multiple kernels, duration sum remains consistent with native SOL:

| GPU | N | Median signed difference | Median absolute difference | P90 absolute difference | Within 3% | Within 5% |
| 6 | 355 | 0.000% | 0.161% | 1.020% | 98.6% | 100.0% |
| 7 | 355 | 0.000% | 0.178% | 1.724% | 92.7% | 94.6% |

Matching kernel count together with matching duration sum strongly supports
that both collectors measure the same kernel work. Kernel sum is therefore the
recommended metric for cross-collector kernel execution comparisons.

### Multi-kernel calls: span

Span behaves differently:

| GPU | N | Median signed Dynamic difference | Median absolute difference | P90 absolute difference |
| 6 | 355 | +8.718% | 8.718% | 56.404% |
| 7 | 355 | +8.018% | 12.454% | 55.672% |

Dynamic and native SOL use the same mathematical definition:

```text
span = max(kernel_end) - min(kernel_start)
```

The difference is not a different span formula. Kineto enables CPU operator
tracing, launch correlation, annotations, and profiler-buffer recording. That
runtime instrumentation delays submission of subsequent kernels. The GPU then
observes longer real idle gaps between short kernels, and span includes those
gaps. Native CUPTI collection is lighter and perturbs the launch path less.

Kineto does not add CPU time to the result during post-processing; it changes
the observed execution timeline while profiling is active. This observer
effect is small for one/few-kernel TileOps calls and can be large for fragmented
Torch references that launch thousands of kernels.

Example: `GatedDeltaNetFwdOp`, B=2, S=32768, H=4, FP16:

| Collector/GPU | Kernels | Kernel sum | Span | Span - sum |
| Dynamic GPU6 | 13,392 | 28.7776 ms | 277.9658 ms | 249.1882 ms |
| Native SOL GPU6 | 13,392 | 28.8060 ms | 178.4147 ms | 149.6087 ms |
| Dynamic GPU7 | 13,392 | 28.8157 ms | 265.0319 ms | 236.2162 ms |
| Native SOL GPU7 | 13,392 | 28.8340 ms | 171.0576 ms | 142.2236 ms |

Kernel sum differs by less than 0.1%, while Dynamic span is about 55% larger on
both GPUs. Nearly all of the difference is inter-kernel gap, not kernel
execution time.

The interpretation is therefore:

- single kernel: Dynamic and native SOL measure the same duration;
- multiple kernels, sum: Dynamic and native SOL remain closely aligned;
- multiple kernels, span: Kineto reports the span of the instrumented
  execution and can overestimate a lightweight native-CUPTI span.

`kernel_span_ms` is still useful for inspecting the Kineto timeline, but it is
not an unbiased replacement for native-SOL span on highly fragmented calls.

## Multi-stream projection

A grouped Torch Conv3D call exposed a separate validity issue: one repeat was
projected into eight overlapping annotation windows, one per CUDA stream, and
launched 32 stable kernels. The previous uniqueness check rejected this valid
trace regardless of dummy count.

The collector now requires every repeat ID to be present but permits several
annotation windows with the same ID. Each kernel is still accumulated once.
Targeted validation after the change:

```text
test_conv3d_bench[3d-resnext-grouped-k3-float16]: passed
TileOps: 2 kernels, sum 0.0346 ms, span 0.0817 ms
Torch:  32 kernels, sum 0.3011 ms, span 0.1380 ms
projection: 150/150 repeats, 0 retries
```

The Torch sum can exceed span because kernels overlap across streams.

## Full-run wall time

| Round | Collector | GPU | Wall time |
| --- | --- | ---: | ---: |
| 1 | Dynamic/Kineto | 6 | 1560.73 s |
| 1 | Native SOL | 7 | 842.43 s |
| 2 | Dynamic/Kineto | 7 | 1462.56 s |
| 2 | Native SOL | 6 | 815.41 s |

The GPU swap preserves the difference. Kineto's CPU/GPU trace instrumentation
and high-fan-out Torch references dominate the additional wall time; computing
sum and span from the same trace does not launch additional kernels or create
an additional profiler session.

## Validation

```text
pytest -q \
  benchmarks/tests/test_benchmark_base.py \
  benchmarks/tests/test_benchmark_boundaries.py

17 passed
```

Ruff and `git diff --check` pass for the single-file change.

## Limitations

- The dummy prefix masks an invalid activity prefix; it does not fix Kineto's
  timestamp callback behavior.
- The dummy limit of 100 and event budget of 10,000 are engineering
  guardrails, not CUPTI guarantees.
- Kernel-count equality and sum agreement strongly support matching activity
  sets, but the reports do not preserve every kernel name/correlation ID.
- Kineto span includes profiling-induced launch-gap perturbation for
  multi-kernel calls.
- The learned dummy count and fan-out cache are thread-local and assume the
  current single-threaded benchmark runner.

### DIFF
--- a/benchmarks/benchmark_base.py
+++ b/benchmarks/benchmark_base.py
@@ -4,6 +4,7 @@ import os
 import subprocess
 import sys
 import threading
+import time
 from abc import ABC, abstractmethod
 from datetime import datetime
 from typing import (
@@ -62,10 +63,60 @@ class _CuptiProjectionError(Exception):
 # so its device event cannot fall inside a window regardless of how the
 # projection behaves; kernels the timed call launches do.
 _KERNEL_REGION = "tileops_bench_kernel"
+_CUPTI_DUMMY_REGION = "tileops_bench_dummy"
+
+# CUPTI 28 may return bad timestamps for a growing prefix of kernel activities
+# after Kineto re-registers its timestamp callback for each profiler session.
+# Sacrifice a dynamically sized prefix of unmeasured calls so the annotated
+# measurement region remains complete. The learned count is retained on the
+# benchmark thread across bench_kernel() calls because the faulty CUPTI state
+# survives across profiler sessions in the same process.
+_CUPTI_DUMMY_INITIAL = 1
+_CUPTI_DUMMY_SAFETY_MARGIN = 2
+_CUPTI_DUMMY_MAX = 100
+_CUPTI_PROJECTION_RETRIES = 3
+_CUPTI_EVENT_BUDGET = 10_000
+_CUPTI_MIN_REPEATS = 5
 
 
-def _sum_kernel_time_us(kineto_results):
-    """Sum device time of the kernels the timed call launched.
+def _effective_repeats(requested: int, kernel_count: int, n_trials: int) -> int:
+    """Bound trace size without reducing ordinary one/few-kernel sampling."""
+    if requested <= 0 or kernel_count <= 0 or n_trials <= 0:
+        return requested
+    budgeted = _CUPTI_EVENT_BUDGET // (kernel_count * n_trials)
+    return min(requested, max(_CUPTI_MIN_REPEATS, budgeted))
+
+
+def _call_signature(fn, args: tuple[Any, ...]) -> tuple:
+    """Stable-enough in-process key for reusing activity fan-out counts."""
+    owner = fn.__self__ if hasattr(fn, "__self__") else fn
+    callable_id = (
+        type(owner).__module__,
+        type(owner).__qualname__,
+        getattr(fn, "__qualname__", type(fn).__qualname__),
+    )
+    arg_signature = tuple(
+        (tuple(arg.shape), str(arg.dtype), tuple(arg.stride()))
+        if isinstance(arg, torch.Tensor) else (type(arg).__qualname__, repr(arg))
+        for arg in args
+    )
+    return callable_id + (arg_signature,)
+
+
+def _trial_means_ms(
+    by_repeat_us: dict[int, float], repeats: int, trials: int
+) -> list[float]:
+    """Return one mean latency in milliseconds for each trial."""
+    return [
+        sum(by_repeat_us[i] for i in range(t * repeats, (t + 1) * repeats))
+        / repeats
+        * 1e-3
+        for t in range(trials)
+    ]
+
+
+def _kernel_time_us_by_repeat(kineto_results):
+    """Collect summed and elapsed device time for each timed call.
 
     Sums only kernels inside a :data:`_KERNEL_REGION` annotation window, so the
     L2-flush fill is excluded and the kernel under test is counted regardless of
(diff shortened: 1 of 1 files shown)


lcy-seso:
Closing as superseded by #1838.

This PR hardens the Kineto-based projection path and adds a dynamic dummy prefix to work around the Kineto/CUPTI timestamp issue. #1838 removed that path: timing now comes from CUPTI activity records directly, with no projection windows and no dummy prefix, so there is no longer a Kineto trace for this change to apply to.

The two metrics it exposes are covered: the timed span is the earliest-to-latest activity of the discovered sequence, and `latency_p10_ms` / `latency_p90_ms` / `n_samples` now travel with every result. Reopen if something here is not covered.

PR #1865, opened 2026-08-09
[Feat][Manifest] declare slot ABIs for the reduction family

lcy-seso:
## Summary

Starts B3 — the manifest field an outside author implements a slot from, without reading TileOPs
source. Reduction first, because it is the family that has been regularised.

- **Spec** (`docs/design/manifest.md`): a slot carries `build` — what the factory is specialized on
  — and `call` — what the returned callable receives and returns. Nothing else.
- **Names are semantics.** `SumFwdOp` declares `sum`, not the `reduce` dispatch key it shares with
  seven others behind an `op_kind` selector. `VarMeanFwdOp` is why: same class as `var`, two
  returned tensors against one. A parameter that changes return arity multiplexes operations.
- **`from` references, never restates.** `M` and `N` are the expressions `roofline.vars` already
  carries; a copy is a second thing to keep in step.
- **Validator**: a `from`, a call tensor, a shape term, a dtype or a layout that resolves to
  nothing is an error. 10 tests, one per failure.
- **14 of 19 reduction entries** declared, each checked against runtime rather than the schema.

## Deliberately absent

- `argreduce` — its factory takes `inner_stride`, a memory-layout parameter. Declaring it would
  put one backend's strided indexing into the contract.
- `all` / `any` / `count_nonzero` — the factory is built on the semantic dtype but the callable
  receives float32 whenever the input is bool, int32, int64 or complex. The storage dtype varies
  with the input, so no fixed value states it.

Both are the same shape of problem: a build fact that cannot be stated from what the entry
declares. Recorded here rather than papered over with an expression.

## Not declared yet

`revision`, `mutates`, `aliases`, `accepts_noncontiguous`, `provider_options`. Nothing consumes
them and the validator cannot fail on them, so they would be promises no one reads.

## Test plan

- [x] `pytest tests/test_validate_manifest.py` — 132 passed
- [x] `pytest tests/ops/test_{reduce,logical_reduce,vector_norm}.py` — 533 passed
- [x] `validate_manifest.py --strict` clean
- [x] Every declared slot built through its op on GPU; arity, shape and dtype compared against
      what the kernel actually returns

### DIFF
--- a/docs/design/manifest.md
+++ b/docs/design/manifest.md
@@ -247,6 +247,7 @@ Validator enforces `cls.__name__ == manifest_key` exactly — no heuristic resol
 | `workloads`               | yes      | Benchmark shapes/dtypes.                                      |
 | `roofline`                | yes      | Performance model.                                            |
 | `source`                  | yes      | Implementation paths.                                         |
+| `slots`                   | no       | Backend-neutral factory contract. See [Slots](#slots).        |
 
 ### `family`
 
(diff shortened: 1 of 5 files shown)


lcy-seso:
Closing: the field this adds is a mirror of the code, not a spec.

Attempting it produced the evidence. Roughly a fifth of the entries could not be declared at all —
argreduce's `inner_stride`, the logical ops' storage-dtype duality, GroupNorm and InstanceNorm
applying affine in the op layer — and every entry that could be declared was transcribed by running
the op and reading what the kernel received, which inverts the trust model the manifest rests on.

The cause is ordering. A slot describes the seam between op and implementation, and that seam is
still whatever each kernel constructor happens to take. Making it uniform is B1's job, and B1 has
not started. Writing the ABI first encodes the current irregularity into the manifest — the same
argument the RFC already uses for keeping B4 behind regularisation.

The three undeclarable cases are real layering findings and move to the RFC. Fixes for what is
fixable get their own PR.

PR #1879, opened 2026-08-11
[Feat][Attention] Restore contiguous GQA prefill with KV cache

### LINKED ISSUE #1839: [FEATURE][ATTENTION] restore GQA prefill with a contiguous KV cache if a caller needs it
### Symptom / Motivation

GQA prefill against a **contiguous** (non-paged) KV cache is no longer expressible. The three kernels that served it were deleted in #1836 as unreachable — correctly, since no op dispatched them — but the capability itself was real and has no surviving equivalent.

Recover the code from `82535c3d:tileops/kernels/attention/gqa_fwd.py`, lines 1036-1725:

| Class | Line at `82535c3d` |
| --- | --- |
| `GQAPrefillWithKVCacheFwdKernel` | 1291 |
| `GQAPrefillWithKVCacheRopeAppendKernel` | 1410 |
| `GQAPrefillWithKVCacheRopeFwdKernel` | 1658 |

### Root Cause Analysis

The three form a complete `kernel_map` for one op that was never written. They map 1:1 onto the paged op's kernel set, minus FP8:

| Contiguous (deleted) | Paged (live) |
| --- | --- |
| `GQAPrefillWithKVCacheFwdKernel` | `GQAPrefillPagedWithKVCacheFwdKernel` |
| `GQAPrefillWithKVCacheRopeAppendKernel` | `GQAPrefillPagedWithKVCacheRopeAppendKernel` |
| `GQAPrefillWithKVCacheRopeFwdKernel` | `GQAPrefillPagedWithKVCacheRopeFwdKernel` |
| — | `GQAPrefillPagedWithFP8KVCacheFwdKernel` |

The deleted ABI took `k_cache`/`v_cache` shaped `[batch, seqlen_kv, heads_kv, dim]` plus per-request `cache_seqlens`. The paged survivors express the same mathematics but not the same ABI: a contiguous-cache caller must build a page pool, choose a page size, and construct a block table — storage reorganisation, not a kernel rename.

Investment went to the paged path — manifest entry, tests, benchmarks and the later FP8 variant are all paged-only. The contiguous kernels never got any of it, which is why they ended up stranded.

## Goal

A caller holding a contiguous KV cache can run chunked prefill without repacking into a page pool.

## Plan

1. Add a `GroupedQueryAttentionPrefillWithKVCacheFwdOp` manifest entry — signature, shape_rules, dtype_combos, workloads, roofline, and a `source.kernel_map` naming the three classes. Model it on the paged entry, dropping the FP8 combos.
2. Restore the three kernels from the ref above.
3. Implement the op. Its `default_kernel_map` branches on `fuse_rope` exactly as the paged op does, minus the FP8 branch.

## Constraints

- Restoring kernels without the op re-creates the unreachable state #1836 removed. Steps 1-3 land together.
- A manifest entry alone does not make a kernel reachable; `status: implemented` without the op class fails the validator, and `status: spec-only` leaves the kernels stranded.

## Acceptance Criteria

- [ ] Manifest entry exists and `scripts/validate_manifest.py` passes
- [ ] The op dispatches all three kernels across the `fuse_rope` on/off paths
- [ ] Correctness tests against a PyTorch oracle, covering the RoPE-fused and unfused paths
- [ ] No kernel class ships absent from every op's `default_kernel_map`

## Trigger

Open work only when a caller's serving cache is deliberately contiguous and repacking into pages is unacceptable. Absent that, the paged path already covers prefill-with-cache and this stays closed.

superAngGao:
## Summary

- restore contiguous BSHD GQA prefill-with-KV-cache kernels, including fused Neox RoPE
- add `GroupedQueryAttentionPrefillWithKVCacheFwdOp` with current capability-based dispatch; the RoPE forward kernel owns and launches the append kernel, matching the live paged design
- add the manifest contract, representative workloads, roofline accounting, manifest-driven benchmark, and PyTorch-oracle tests

The op appends `k_new` / `v_new` in place at each request's `cache_seqlens` and returns the current chunk output without requiring page-pool repacking.

Closes #1839

## Validation

- `ruff check` on all changed Python files
- `pytest tests/ops/attention/test_gqa_prefill_contiguous.py -q` — 4 passed on H200 (FP16 causal + softcap, BF16 non-causal, partial-RoPE, capacity rejection)
- `pytest tests/ops/test_kernel_selection.py tests/perf/test_gqa_prefill_contiguous_roofline.py -q` — 38 passed
- manifest benchmark smoke on H200 — 0.1211 ms for B=4, S_new=512, prefix=4096, H/H_kv=8/2, D=64, FP16, softcap=50


superAngGao:
Closed because issue #1916 initiates a round of refactoring and feature enhancement in a broader range. Future PRs will address this issue more systematically.

PR #1887, opened 2026-08-11
[Perf][foundry][GEMM] Accelerate W4A16 decode on H200

zhen8838:
## Summary

- Add a fused M=1 W4A16 decode specialization that unpacks little-endian nibbles, applies group-128 affine dequantization with the contract's FP16 rounding point, and reduces directly to the FP16 output.
- Dispatch M=1 calls to the specialization while preserving the existing tensor-core implementation for M>1.
- Compare the candidate, exact TileOPs incumbent, and both runnable same-contract vLLM Marlin reduction modes on every primary manifest workload.

## TileFoundry Description

```python
@module(
    entry="gemm_w4a16",
    topologies=(Topology("cta", None),),
)
class GemmW4A16:
    @func
    def gemm_w4a16(
        activation: Tensor[(M, K), "f16"],
        packed_weight: Tensor[(N, K // 2), "u8"],
        weight_scale: Tensor[(N, K // 128), "f32"],
        weight_zero: Tensor[(N, K // 128), "u8"],
    ) -> Tensor[(M, N), "f16"]:
        packed_i32 = tf.cast(packed_weight, dtype="i32")
        low = tf.mod(packed_i32, 16)
        high = tf.floor_div(packed_i32, 16)
        q_i32 = tf.reshape(tf.stack(low, high, axis=-1), new_shape=(N, K))

        scale_expanded = tf.repeat_interleave(weight_scale, repeats=128, axis=1)
        zero_expanded = tf.repeat_interleave(
            tf.cast(weight_zero, dtype="f32"), repeats=128, axis=1
        )
        dequant_f32 = (tf.cast(q_i32, dtype="f32") - zero_expanded) * scale_expanded
        dequant_f16 = tf.cast(dequant_f32, dtype="f16")
        return tf.matmul(activation, tf.transpose(dequant_f16, perm=(1, 0)))
```

## Performance

Operator: `GemmW4A16Op`

| Environment | Value |
| --- | --- |
| image | ghcr.io/tile-ai/tileops-runner:cu132-torch2.13-tl-afcebed1-dev |
| digest | sha256:2590888968a870216e1ea076829b909e62e9053608f6a65d5ab5ecd7eb5561f7 |
| gpu | NVIDIA H200 |
| driver | 595.71.05 |
| cuda | 13.2 |
| torch | 2.13.0+cu132 |
| vllm | 0.27.1 |
| timer | native CUPTI |

Method: Candidate, incumbent, and both Marlin modes ran in one process on common logical inputs
with identical W4A16 work and FP16 output. Each was checked against the reference before timing,
then timed by `bench_kernel` through CUPTI — 25 ms warmup, 100 ms repeat policy, L2 reset before
every iteration. Layout conversion, compilation, tuning, and workspace setup were excluded; CUPTI
failure was fail-closed.

Ratio in comparator columns: implementation / candidate. &#x1F7E2; > 1 means the candidate is faster;
&#x1F534; <= 1 means it is not.

| Workload | M | N | K | Dtype | TileFoundry candidate (ms) | TileOPs incumbent (ms)<br>/ candidate | vLLM Marlin FP16 reduce (ms)<br>/ candidate | vLLM Marlin FP32 reduce (ms)<br>/ candidate |
| --- | ---: | ---: | ---: | --- | ---: | ---: | ---: | ---: |
| decode-l2-resident-ish | 1 | 8192 | 8192 | float16 | 0.025696 | **0.084832<br>&#x1F7E2;&nbsp;3.3014x** | 0.020192<br>&#x1F534;&nbsp;0.7858x | 0.020256<br>&#x1F534;&nbsp;0.7883x |
| decode-hbm-streaming-threshold | 1 | 8192 | 16384 | float16 | 0.046944 | **0.169536<br>&#x1F7E2;&nbsp;3.6115x** | 0.035360<br>&#x1F534;&nbsp;0.7532x | 0.0354245<br>&#x1F534;&nbsp;0.7546x |
| decode-non-power2-low-cta | 1 | 7168 | 20480 | float16 | 0.057344 | **0.205344<br>&#x1F7E2;&nbsp;3.5809x** | 0.038176<br>&#x1F534;&nbsp;0.6657x | 0.038176<br>&#x1F534;&nbsp;0.6657x |
| decode-long-k-pressure | 1 | 8192 | 81920 | float16 | 0.219678 | **0.815550<br>&#x1F7E2;&nbsp;3.7125x** | 0.125472<br>&#x1F534;&nbsp;0.5712x | 0.125471<br>&#x1F534;&nbsp;0.5712x |
| geometric mean |  |  |  |  | 0.0624353 | **0.221533<br>&#x1F7E2;&nbsp;3.5482x** | 0.0430038<br>&#x1F534;&nbsp;0.6888x | 0.0430574<br>&#x1F534;&nbsp;0.6896x |

## Result And Limitations

**improvement without SOTA**

- Classification is improvement without SOTA: the candidate is 3.3014x to 3.7125x faster than the TileOPs incumbent on every primary row, with a 3.5482x geometric-mean speedup, but is 1.2726x to 1.7508x slower than the fastest passing Marlin mode and 1.4519x slower by geometric-mean latency.
- Candidate p10-to-p90 widths are 1.747%, 1.431%, 1.004%, and 0.495% of the median in table order; every incumbent win and external-baseline loss is larger than measured noise.
- The specialization applies only to M=1; M>1 continues to use the existing TileOPs tensor-core path. Selection runs through `applies()`/`general`/`kernel_map`, so a device the decode kernel is not built for falls back to the tensor-core path instead of failing.
- A load-only twin of the kernel — same tiling and staging, no dequantization — runs the smallest workload in 0.0199 ms (1.69 TB/s), matching Marlin. The remaining gap is dequantization instruction count, not the weight stream.
- The exact TileFoundry description cannot currently be scheduled directly because the admitted TileFoundry checkout lacks unsigned-8 tensor support and dynamic RepeatInterleave shape inference, so the delivered TileLang kernel is a semantic runtime twin of that graph.


src/tileops/kernels/gemm_w4a16.py
```diff
@@ -186,6 +187,20 @@ def __init__(
...
+                n,
+                k,
+                dtype,
+                config=config,
+                tune=tune,
+                group_size=group_size,
+            )
+            self.kernel = self.decode_kernel.kernel
+            self.config = self.decode_kernel.config
+            return
```
lcy-seso: Please dispatch through `applies()`/`general`/`kernel_map` (+ `gemm.yaml`) instead of nesting: only the decode kernel constrains `supported_archs`, so `_check_arch` now makes M=1 fail on an unlisted arch instead of falling back to this kernel.

src/tileops/kernels/gemm_w4a16_decode.py
```diff
@@ -0,0 +1,193 @@
...
+                            dtype,
+                        )
+                        products_high[row, packed_k] = T.cast(
+                            activation_shared[packed_k * 2 + 1], "float32"
+                        ) * T.cast(dequant_high, "float32")
+
+                    T.reduce_sum(products_low, partial_low, dim=1)
+                    T.reduce_sum(products_high, partial_high, dim=1)
+                    for row in T.Parallel(block_n):
+                        accum[row] = accum[row] + partial_low[row] + partial_high[row]
```
lcy-seso: Main bottleneck: the cross-thread reduction is inside the K loop (640 iterations x2 at K=81920) — accumulate per-thread and reduce once after the loop.

src/tileops/kernels/gemm_w4a16_decode.py
```diff
@@ -0,0 +1,193 @@
...
+                zeros = T.alloc_fragment((block_n,), "float32")
+                products_low = T.alloc_fragment((block_n, block_k // 2), "float32")
+                products_high = T.alloc_fragment((block_n, block_k // 2), "float32")
+                partial_low = T.alloc_fragment((block_n,), "float32")
+                partial_high = T.alloc_fragment((block_n,), "float32")
+                accum = T.alloc_fragment((block_n,), "float32")
+
+                T.fill(accum, 0.0)
+                for group in T.Serial(k // block_k):
+                    T.copy(activation[pid_m, group * block_k], activation_shared)
```
lcy-seso: `T.Serial` with predicated global->fragment loads pays full global latency every iteration — stage the weights through shared and pipeline it. (Line 44: every other `T.copy` in the repo writes an explicit slice.)

src/tileops/kernels/gemm_w4a16_decode.py
```diff
@@ -0,0 +1,193 @@
...
+                            pid_n * block_n + row < n,
+                            T.cast(
+                                packed_weight[
+                                    pid_n * block_n + row,
+                                    group * (block_k // 2) + packed_k,
+                                ],
+                                "int32",
+                            ),
+                            0,
+                        )
```
lcy-seso: Per-element predicated loads likely degrade to 1-byte scalar loads; since every manifest N divides all candidate `block_n`, you could drop the predicate and `T.copy` into shared for vectorized loads.

src/tileops/kernels/gemm_w4a16_decode.py
```diff
@@ -0,0 +1,193 @@
...
+    def default_config(self) -> dict:
+        return {"block_n": 16, "threads": 128}
+
+    @property
+    def autotune_configs(self) -> list[dict]:
+        return [
+            {"block_n": block_n, "threads": threads}
+            for block_n in (4, 8, 16)
+            for threads in (128, 256)
+        ]
```
lcy-seso, line 189: No `num_stages` or split-K in the space, and `tune` defaults to `False` in the bench — so the reported numbers are untuned `default_config`.

src/tileops/kernels/gemm_w4a16_decode.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def forward(
+        self,
+        activation: torch.Tensor,
+        packed_weight: torch.Tensor,
+        weight_scale: torch.Tensor,
+        weight_zero: torch.Tensor,
+    ) -> torch.Tensor:
+        return _gemm_w4a16_wrapped(
```
lcy-seso: `GemmKernel`/`GemvKernel` keep the `custom_op` wrapper for torch.compile only and call the compiled JIT directly in eager; at ~50us/call the extra dispatch matters and CUPTI hides it.

tests/ops/test_gemm.py
```diff
@@ -239,6 +239,21 @@ def test_quantize_weight_int4_keeps_one_sided_groups_in_range() -> None:
...
+        [[0.0, 0.2, 0.4, 0.6, 0.0, -0.2, -0.4, -0.6]],
+        dtype=torch.float32,
+    )
+
+    packed, scale, zero, dequantized = quantize_weight_int4(weight, group_size=4)
+
+    assert torch.equal(packed, torch.tensor([[0x50, 0xFA, 0xAF, 0x05]], dtype=torch.uint8))
+    assert torch.equal(zero, torch.tensor([[0, 15]], dtype=torch.uint8))
+    torch.testing.assert_close(scale, torch.full((1, 2), 0.04))
+    torch.testing.assert_close(dequantized, weight)
```
lcy-seso: This tests `quantize_weight_int4`, which the PR doesn't touch, while the new kernel has no smoke coverage at all (`full-w4a16-m1` is marked `full`) — worth trading for a smoke M=1 case plus a "M=1 selects the decode kernel" assertion.

benchmarks/tests/test_gemm_w4a16_marlin.py
```diff
@@ -0,0 +1,61 @@
...
+
+import pytest
+import torch
+
+from benchmarks.ops.bench_gemm import _prepare_marlin_w4a16_baseline
+
+
+@pytest.mark.smoke
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA required")
+@pytest.mark.parametrize("use_fp32_reduce", [False, True], ids=["fp16-reduce", "fp32-reduce"])
```
lcy-seso: Suggest deleting: the inline assertion this PR adds to `bench_gemm.py` already checks Marlin against the ref on every workload with real data, and `use_fp32_reduce` doesn't affect the layout conversion under test here.

lcy-seso (commented):
Numerics check out. Three things, inline: it bypasses the repo's `applies()`/`general`/`kernel_map` selection (and thereby breaks the M=1 arch fallback), it's ~6.6x off roofline on untuned defaults, and none of the 76 new test lines cover the new kernel.

PR #1897, opened 2026-08-12
[Perf][foundry][MoE] Accelerate MoE experts decode on H200, and move its kernel selection to the dispatch site

zhen8838:
## Problems

- Decode ran the gate/up GEMM at BM128 with a separate activation kernel, leaving most of every tile idle when few rows route to each expert.
- The op chose kernels in its constructor, in vocabulary written for the occasion, and got the choice wrong: a caller's `kernel_map` entry replaced the class the shape check had just approved.
- The fused kernel read its shared-memory staging back with TMA at four sites with no proxy fence between them.
- Three copies of one CTA-tile floor decided schedules from a multiprocessor count, and measured, the floor refuses the faster schedule exactly where it still wins.
- `use_fused_activation` and `gemm_kernel=` asked the caller to pick what the shape already determines, and neither ever won a measurement.
- The fused gate/up op had no manifest entry, so it got no synthesized validation, no roofline and no manifest-driven coverage.

## Changes

- Kernel selection moves to `_get_kernel`, where `Op.select_kernel_key` picks over a `GroupedGemmCall` with `inputs=` passed, and each implementation states its own region through `applies` and `general`.
- The gate/up stage becomes one role with two candidates: `MoeGroupedGemmSeparateActKernel` wraps the wide GEMM and its activation into one kernel, as `SharedExpertMLPKernel` already wraps several launches, so the fused and unfused pipelines are ordinary siblings behind the thin `MoeGateUpFwdOp`.
- The parent op's constructor builds four stages and chooses nothing; `takes_shape`, `wants_fused_epilogue`, the override bookkeeping, the warn-and-substitute fallback, both caller-facing switches and the retired `MoeGroupedGemmNopad3WGFusedActFwdOp` are all gone, covered by the selector's own rules.
- The fused epilogue fences before its TMA store at all four sites, and all three CTA-tile floors are deleted.
- Both gate/up ops validate device and normalise to contiguous before the get-kernel boundary, and `MoeGateUpFwdOp` gains a manifest entry, input shapes, `_infer_output_shapes`, a test and a manifest-driven benchmark.
- The shared MoE protocol moves to `moe/abc.py` and the grouped-GEMM call and regimes move into the `grouped_gemm` package, so no half of the protocol imports its base class out of the other and nothing reaches into a private module across packages.

## TileFoundry Description

The HIR below preserves the routed-expert math and public output contract.
Production lowering fuses aligned sparse decode (`T * top_k / E_local <= 32`) and retains the existing unfused BM128 pipeline for denser prefill.

```python
@module(
    entry="kernel",
    topologies=(Topology("cta", 128),),
)
class OperatorModule:
    @func
    def kernel(
        hidden: Tensor[(T, H), "bf16"],
        w_gate_up: Tensor[(E, 2 * F, H), "bf16"],
        w_down: Tensor[(E, H, F), "bf16"],
        topk_weights: Tensor[(T, K), "f32"],
        topk_ids: Tensor[(T, K), "i32"],
    ) -> Tensor[(T, H), "bf16"]:
        flat_ids = tf.reshape(topk_ids, new_shape=(T * K,))
        gate_up_w = tf.reshape(
            tf.index_select(w_gate_up, flat_ids, dim=0),
            new_shape=(T, K, 2 * F, H),
        )
        down_w = tf.reshape(
            tf.index_select(w_down, flat_ids, dim=0),
            new_shape=(T, K, H, F),
        )
        token_col = tf.reshape(hidden, new_shape=(T, 1, H, 1))
        gate_up = tf.reshape(
            tf.matmul(gate_up_w, token_col),
            new_shape=(T, K, 2 * F),
        )
        gate = gate_up[:, :, :F]
        up = gate_up[:, :, F : 2 * F]
        activated = tf.silu(gate) * up
        expert_out = tf.reshape(
            tf.matmul(
                down_w,
                tf.reshape(activated, new_shape=(T, K, F, 1)),
            ),
            new_shape=(T, K, H),
        )
        weighted = tf.cast(expert_out, dtype="f32") * tf.reshape(
            topk_weights,
            new_shape=(T, K, 1),
        )
        output = tf.cast(
            tf.reduce(weighted, axes=(1,), keepdim=False, kind="sum"),
            dtype="bf16",
        )
        output_local = tf.reshard(
            output,
            layout=ShardLayout(
                Layout((T, H)),
                (S(0),),
                Mesh(
                    (Topology("cta", 128),),
                    Layout((128,)),
                    names=("block",),
                ),
            ),
            storage=rmem,
        )
        return tf.reshard(
            output_local,
            layout=ShardLayout(
                Layout((T, H)),
                (B(),),
                Mesh(
                    (Topology("cta", 128),),
                    Layout((128,)),
                    names=("block",),
                ),
            ),
            storage=gmem,
        )
```

## Performance

One H200, bf16, `benchmarks/ops/bench_fused_moe_experts.py` unmodified against
`upstream/main` (`7213c05`). Three alternating runs per side, each in a fresh
process — a process that has already run several multi-gigabyte shapes measures
the same shape about 2x slower. Medians and ranges of `device_busy_ms` under
native CUPTI.

| Workload | `main` (ms) | range | This PR (ms) | range | |
| --- | ---: | :--- | ---: | :--- | ---: |
| qwen3-235b-decode | 3.0925 | 3.092-3.095 | 2.7753 | 2.772-2.776 | **1.114x** |
| qwen3-235b-prefill | 5.9767 | 5.934-5.998 | 5.9755 | 5.959-5.996 | 1.000x |
| deepseek-v3-decode | 6.0966 | 6.077-6.145 | 5.4147 | 5.413-5.419 | **1.126x** |
| deepseek-v3-prefill | 8.4585 | 8.435-8.524 | 8.3886 | 8.357-8.441 | 1.008x |

Decode gains ~12% on both models, from 7 kernel launches down to 6: the
activation moves into the gate/up epilogue and the down GEMM takes a BM64
schedule. Prefill keeps the unfused BM128 pipeline and is unchanged — the same
kernel classes with the same configs run there, verified by dumping both sides
at that shape.

Image `ghcr.io/tile-ai/tileops-runner:cu132-torch2.13-tl-afcebed1-dev`, driver
595.71.05, CUDA 13.2, torch 2.13.0+cu132, tilelang 0.1.11+cu132.gitafcebed1.


lcy-seso:
Pushed two commits moving schedule selection back to the kernel layer.

**Problem.** The op held the decode policy: a threshold imported from a kernel module, two kernels' private `_DEFAULT_CONFIG`, a hardcoded `2 * 132` CTA floor, and a schedule dict injected into the down GEMM through a new `config=` parameter.

**Change.**
- `kernels/grouped_gemm/_config.py` holds the regime and the CTA check, shared by both 3WG kernels; both `default_config`s delegate to it.
- Each kernel answers what the op used to: `unsupported_reason()` for shape alignment, `uses_decode_schedule()` for whether the fused pipeline is faster here.
- The op keeps only the pipeline choice. `config=` and both cross-layer private imports are gone.

**The selector takes no role**, because both GEMM roles move together (H200, bf16, ms):

| rows/expert | gate_up default | gate_up decode | down default | down decode |
| --- | --- | --- | --- | --- |
| 32 | 2.0466 | 1.9032 | 1.0224 | 0.9817 |
| 16 | 4.0370 | 3.7484 | 1.9877 | 1.9150 |
| 256 | 3.0204 | 4.9874 | 1.6418 | 2.6102 |

The threshold is the largest row count measured positive, with no margin below it: misapplying the decode schedule costs 34-39%, staying in regime gains 4-7%.

**Two holes fixed.** A shape with `K % 128 != 0` could auto-fuse believing it had a decode schedule while the kernel fell back to the default one. And reading `sm_count` from the device would have made op construction probe a device, which this repo does not do — it now answers from the architecture the kernels declare, which is what the `132` was standing in for.

Timings unchanged: qwen3 decode 2.774 / prefill 5.945, deepseek decode 5.411 / prefill 8.277. MoE and grouped_gemm suites 137 passed, 11 skipped; ruff clean.

PR #1898, opened 2026-08-12
[Feat][GEMM] fp16/bf16 dense GEMM: small-batch kernel-mode and GemmOp/GemvKernel tuning

bigSheep123:
Part of #1902. Scope is `GemmOp` / `GemvKernel` / `SmallBatchGemmKernel` and their tests; `GemmFp8Op`, `GemmW4A16Op` and `GroupedGemmOp` are untouched.

## Problems

- `GemmOp`'s fp16/bf16 path had one kernel structure and a modal 128x128x64 config, so small-`m` and awkward-`n` shapes ran below `torch.matmul`.
- The two consumer epilogues used named barriers 4/5, but TileLang allocates from the bottom and took 3/4: one warpgroup's arrivals could release the other's barrier, and a store could then read a half-written `c_smem`. Five of six shipped coop2 configs aliased.
- `top::gemv_wrapped_kernel` advertised `a[k] -> c[n]` but forwarded to a body expecting `(1, k)`.
- A structure-flagged explicit `config=` was merged away, producing one structure at another's tile width — a combination `_enumerate` excludes.
- Rank was unchecked and the forward fast-path signature omitted `b`'s dtype, so a rank-3 operand or a changed `b` reached a kernel and died on a TileLang argument error instead of a named `ValueError`.
- The benchmark compared against cuBLAS's default heuristic alone, which credits wins that a cuBLASLt user with algorithm selection would not concede.

## Changes

- `SmallBatchGemmKernel`: CUDA-core weight-streaming body, `m == 2` NT.
- `GemvKernel`: cross-warp SMEM reduction with `reduce_threads > 32`, banded by `k`.
- `GemmKernel`: five structures (coop2, coop2s, coop2_splitk, simple, swap_ab), one config schema each, chosen by `gemm_heuristics.best_config` — an analytic score over L2 / L1 / tensor-core terms and wave efficiency.
- Kernel-mode selection goes through the candidate registry: each kernel states its region positively and `Op.select_kernel_key` picks. No kernel names a sibling; order decides nothing.
- `bench_gemm.py` adds a `cublaslt-best` column — cuBLASLt's heuristic picks ranked against `torch.matmul` through `timing.median_busy_ms`, the statistic that reports the row. `nvmath-python` comes from the runner image (#2003).
- Dispatch rules live in each kernel's `applies`, and `CallSpec` carries `sm_count`, so a predicate reads the call rather than probing the current device.
- The four defects above are fixed; `compute-sanitizer --tool synccheck` flags the old barrier ids and is clean on the new.
- `_validate_dtypes` is synthesized from the manifest and runs on every call; `_infer_output_shapes` is implemented, so the L2 shape-parity check runs instead of being skipped. The only manifest change is the third dispatch key.
- No fallback, by design: every `GemmKernel` structure loads through TMA, so the innermost extent must be a multiple of 8 elements for fp16/bf16. `_tma_misalignment` states it once. All 13 manifest workloads clear it; a sweep of 17 shapes x 4 layouts refused 28 of 68, all predicted by that one rule. The bandwidth kernels use `cp.async` and serve `m <= 2` at any extent.
- Test nodes 38 -> 56 (+18): first coverage for the five structures, the two `top::` wrapped ops, `trans_a=True`, and the four refusal paths.

## Performance

H200, torch 2.10.0+cu129, TileLang 0.1.13. Native-CUPTI kernel timing, span-inclusive, two interleaved rounds. `x-best` = ours vs `min(torch, cuBLASLt-best)` per shape; parity band 0.98-1.02. **14 ahead / 7 parity / 6 behind** over 27 anchored rows.

| group | rows | x-best |
| --- | --- | --- |
| gemv (m=1) | 4 | **1.124 - 1.390** |
| small-batch (m=2,4,8) | 9 | **1.052 - 1.182** |
| mid-m (m=16,32,64) | 3 | 1.001 - **1.029** |
| prefill-attn-proj fp16/bf16 | 2 | 0.993 / 0.987 |
| decode-gate-up / mid-m96-gate-up / k-dominant | 3 | 0.982 / 0.979 / 0.979 ⚠ |
| square-1k fp16/bf16 | 2 | 0.964 / 0.967 |
| prefill-down / decode-down / wide-n | 3 | 0.937 / 0.932 / 0.932 |

⚠ Undecided, not results. These three are launch-bound split-K or long-K rows whose own spread is wider than their distance to the decision line — one fixed binary read 0.979 to 1.030 across runs. Settling them needs a protocol that controls for run-to-run variation.

`decode-down` and `prefill-gate-up` want a fine `block_n` (112 / 176) that tilelang decomposes into narrow WGMMA instructions today — tile-ai/tilelang#2928, fix in #2931. `square-1k` needs async multicast for the last ~4%; the form available today measures a net loss — tile-ai/tilelang#2929.

## Follow-ups

1. A protocol that can decide the three undecided rows.
2. Pin the fine-`block_n` configs once tilelang#2931 releases.
3. `forward()`'s structure chain -> a table, which unblocks a per-structure `autotune()`.
4. `workloads` has no `m <= 8` row, so `bench_gemm.py` never exercises the gemv or small-batch kernel-modes.


benchmarks/cublaslt_baseline.py
```diff
@@ -0,0 +1,437 @@
...
+
+_lib: Optional[ctypes.CDLL] = None
+
+
+def _load() -> Optional[ctypes.CDLL]:
+    """Load libcublasLt once; return None if unavailable."""
+    global _lib
+    if _lib is not None:
+        return _lib
+    for name in ("libcublasLt.so.12", "libcublasLt.so"):
```
zhen8838: why so hardcode？

src/tileops/kernels/gemm_heuristics.py
```diff
@@ -0,0 +1,381 @@
...
+
+import functools
+import math
+from dataclasses import dataclass
+from typing import Optional
+
+__all__ = ["SWAP_AB_BLOCK_NN", "SWAP_AB_MPAD", "best_config", "gemv_config",
+           "small_batch_config", "swap_ab_stages"]
+
+_SMEM_BUDGET = 227 * 1024  # SM90 per-CTA opt-in SMEM ceiling
```
zhen8838, line 64: should respect the current hardware, do not hardcode here

zhen8838 (changes requested):

src/tileops/kernels/gemm.py
```diff
@@ -549,114 +1716,333 @@ def __init__(self,
...
-                    # a is tiny (fits in L1), load directly to registers
-                    for _k in T.vectorized(tile_k):
-                        a_local[_k] = a[bk * block_k + tk * tile_k + _k]
-                    # FMA
-                    for _k in T.serial(tile_k):
-                        c_accum[0] += a_local[_k].astype(accum_dtype) * b_shared[
-                            tn, tk * tile_k + _k].astype(accum_dtype)
+                    for mi in T.serial(m):
+                        for _k in T.vectorized(tile_k):
+                            a_local[mi, _k] = a[mi, bk * block_k + tk * tile_k + _k]
```
superAngGao, line 2288: The final K tile can make this global A load out of bounds. Zero-filling `b_shared` does not make the A access valid, and `NaN/Inf * 0` can still poison the accumulation. Please predicate the load with `k_idx < k` (or use a zero-filling copy), and exercise a non-aligned K under `compute-sanitizer --tool memcheck`.

benchmarks/ops/bench_gemm.py
```diff
@@ -267,6 +268,18 @@ def test_gemm_bench(
...
+    # Stronger baseline: cuBLASLt's fastest algorithm found by heuristic search
+    # (falling back to the default torch.matmul path when that is faster), timed
+    # by the same CUPTI harness. torch.matmul alone uses only cuBLAS's default
+    # top-1 heuristic, which is measurably slower than searchable algorithms on
+    # some shapes (small-M, tall-skinny, awkward-N) — comparing against it can
+    # credit a kernel with a win a cuBLASLt user would not concede. Conditional:
+    # skipped (leaving the torch-cublas baseline) when cuBLASLt is unavailable.
+    best_fn = make_cublaslt_best(m, n, k, dtype, trans_a, trans_b)
+    if best_fn is not None:
+        result_lt = bm.profile(best_fn, a, b)
```
superAngGao: This baseline is profiled separately, after algorithm search, while TileOps and torch are measured through the interleaved `compare()` pass. We do not yet know whether the resulting thermal/clock drift is material, so I would treat this as a benchmark-protocol follow-up rather than a correctness blocker. Could we either put `cublaslt-best` into the same compare plan, or add a small A/B showing that separate measurement does not move the result?

src/tileops/manifest/gemm.yaml
```diff
@@ -57,6 +57,7 @@ GemmOp:
     kernel_map:
       gemm_kernel: GemmKernel
       gemv_kernel: GemvKernel
+      small_batch_kernel: SmallBatchGemmKernel
```
superAngGao, line 67: Could we also add at least one `m=1` and one dispatched `m=2` workload? Registering the new key without those workloads means the manifest-driven nightly benchmark never exercises `GemvKernel` or `SmallBatchGemmKernel`, so the headline small-batch improvements have no ongoing regression coverage.

superAngGao (commented):
A few focused follow-ups from reviewing the newly added paths. I have not repeated the hardware/library hard-coding points already raised. **TileOps now has a documented dispatch contract, so please also re-check the implementation against the current public dispatch conventions.**

benchmarks/cublaslt_baseline.py
bigSheep123: _sonames() now derives the major version from torch.version.cuda — the torch that runs the comparison — and falls back to the unversioned name. A build on a different CUDA no longer silently drops this baseline.

src/tileops/kernels/gemm_heuristics.py
bigSheep123: 👌

src/tileops/manifest/gemm.yaml
bigSheep123: added in 8c8c5aa: ds-v3-decode-gate-up-b1 (m=1) and -b2 (m=2), on the same projection as the existing decode rows so the three token counts are comparable. Dispatch verified: m=1 → gemv_kernel, m=2 → small_batch_kernel. n=2112 is kept narrow deliberately — at wider n the operand-swapped generic config wins m=2 and the bandwidth kernel would never be exercised.

benchmarks/ops/bench_gemm.py
bigSheep123: Took the first option, in 279a595: cublaslt-best now goes into the same compare() plan, so all three entries are timed in one forward-then-reversed interleaved pass under the same CUPTI protocol. The algorithm search still runs once at construction, before any timed rep; the call site comments the reasoning.

benchmarks/ops/bench_gemm.py
```diff
@@ -320,15 +321,30 @@ def test_gemm_bench(
     # The benchmark framework warms up internally; eval_roofline() is read
     # lazily after profiling, by which point forward() has bound the dims.
 
-    # flag_gems' mm takes row-major operands; a transposed row is its own layout,
-    # which its kernel does not express, so those rows carry cuBLAS alone.
+    # Stronger baseline: cuBLASLt's fastest algorithm found by heuristic search
```
zhen8838: please cleanup the comments

benchmarks/cublaslt_baseline.py
```diff
@@ -0,0 +1,588 @@
...
+        _ck(
+            self._lib.cublasLtMatrixLayoutCreate(ctypes.byref(lay), dt, rows, cols, ld),
+            "MatrixLayoutCreate",
+        )
+        return lay
+
+    def _matmul(self, a: torch.Tensor, b: torch.Tensor, algo: _MatmulAlgo) -> int:
+        # A_lt := B, B_lt := A (see the __init__ layout note).
+        stream = ctypes.c_void_p(torch.cuda.current_stream().cuda_stream)
+        return self._lib.cublasLtMatmul(
```
zhen8838: we really need call ctypes here？currently, have no cublas python lib？

src/tileops/ops/gemm/gemm.py
```diff
@@ -84,28 +103,25 @@ def _infer_mnk(self, a: torch.Tensor, b: torch.Tensor) -> Tuple[int, int, int]:
...
+        accordingly), ``"small_batch"`` for the low-``m`` NT bandwidth kernel,
+        else ``"gemm"`` — ``GemmKernel`` (SM90), covering all four
+        ``(trans_a, trans_b)`` layouts.
+
+        Which one serves the call is stated by the candidates themselves
+        (``call_spec.gemv_region`` / ``small_batch_region``, read through
+        ``Kernel.applies``); this method owns only mechanism: mapping the
+        selected key to a kernel instance and caching it.
         """
         call = GemmCall(m=m, n=n, k=k, dtype=dtype, trans_a=self.trans_a, trans_b=self.trans_b)
```
zhen8838, line 118: The dispatch logic here is a little messy. You can receive it in GemmCall.

zhen8838 (changes requested):

benchmarks/ops/bench_gemm.py
bigSheep123: 👌

src/tileops/ops/gemm/gemm.py
bigSheep123: Done in e92899e9: GemmCall.gemv_mode states which operand is the vector — "lhs_row", "rhs_col", or None. gemv_region asks it, _get_kernel reads it instead of re-deriving the same flags two files away, and gemv_n gives the kernel its output length.

benchmarks/cublaslt_baseline.py
bigSheep123: No Python binding for it: torch exposes the backend choice, not algorithm selection, and nothing installed wraps cuBLASLt's algorithm API. The library itself is there — it comes with torch and torch-cublas runs on it; this module opens the same one. nvmath-python could replace the ctypes, but it is a new runner-image dependency, so that seemed a maintainer call rather than something to slip into this PR.

 I measured before defending it. The search beats torch's default on 5 of the 16 GemmFwdOp rows — wide-n-24576 1.10x, ds-v3-prefill-down 1.07x, 1.03–1.05x on three low-m rows — and those are the rows this PR's kernel is furthest behind on, so dropping the baseline would only flatter our own numbers.

Most of the ctypes did not earn its place, though. The pool was cuBLASLt's 8 heuristic picks plus 256 enumerated from every algorithm id × its capability grid; the 8 carry every win, so the enumeration is gone in f9aa52a5 — 588 lines down to 441. Measuring also surfaced three defects, fixed in d1bcd663 / 65609038 / 993942e0: the module's own L2 flush was smaller than the device's L2 (selection ranked warm, the report times cold), a searched algorithm could displace torch and then read slower than it, and the handles were never freed. The entry is asserted against the reference now, like the other baselines.

benchmarks/cublaslt_baseline.py
```diff
@@ -0,0 +1,441 @@
...
+        a = torch.randn(self.m, self.k, dtype=self.dtype, device="cuda")
+        b = torch.randn(
+            (self.n, self.k) if self.trans_b else (self.k, self.n), dtype=self.dtype, device="cuda"
+        )
+        # The harness's own flush buffer, sized from the device's L2. A private
+        # 64 MB one left H200's 60 MB L2 partly warm, so selection ranked
+        # candidates under a cache state the reported timing never sees.
+        flush = _get_l2_flush_cache()
+
+        def min_ms(run, iters):
```
zhen8838: Why do you need to construct benchmark manually? You should reuse the benchmark tool in tileops.

bigSheep123: ok

benchmarks/baselines.py
```diff
@@ -181,6 +184,26 @@ def baseline_fn(x, weight=None, bias=None):
     return baseline_fn
 
 
+def cublaslt_best(
```
zhen8838: Why do you have to pack it here again?

zhen8838 (dismissed):

benchmarks/baselines.py
bigSheep123: get

lcy-seso:
`nvmath-python` is in the runner image now, via #2003. Both tags are rebuilt at the same tilelang commit:

```
docker pull ghcr.io/tile-ai/tileops-runner:cu132-torch2.13-tl-afcebed1-dev
```

The dependency-file changes here (`requirements.in`, `constraints.txt`, the lock) are in that PR, so they can come out of this one.

zhen8838 (dismissed):

lcy-seso (changes requested):
Do not merge. I would like to clean this pr a bit more.

src/tileops/kernels/gemm/call_spec.py
```diff
@@ -21,14 +24,61 @@ class GemmCall(CallSpec):
...
+        return self.n if self.gemv_mode == "lhs_row" else self.m
+
 
 def gemv_region(call: GemmCall) -> bool:
-    """Whether the call is a matrix-vector product the GEMV kernel is written for.
+    """Whether the call is a matrix-vector product the GEMV kernel is written for."""
+    return call.gemv_mode is not None
+
+
+def small_batch_region(call: GemmCall) -> bool:
```
lcy-seso: Both region predicates have exactly one caller: the matching `Kernel.applies` in `dense.py`. `kernel_base.applies` is where the architecture says a kernel states the region it serves, so I moved both bodies there and dropped the functions.

src/tileops/kernels/gemm/call_spec.py
```diff
@@ -1,13 +1,16 @@
...
 
 import dataclasses
 from typing import Optional
 
 import torch
 
+from tileops.utils import get_sm_count
+
 from ..call_spec import CallSpec
+from .heuristics import swap_ab_grid_underfills
```
lcy-seso: This import inverts the dependency: a module describing a call now depends on kernel tuning heuristics. It goes away with `small_batch_region`; `dense.py` already imports `heuristics`, so inlining costs no cycle.

src/tileops/kernels/gemm/call_spec.py
```diff
@@ -21,14 +24,61 @@ class GemmCall(CallSpec):
...
+        if self.trans_a:
+            return None
+        if self.m == 1 and self.trans_b:
+            return "lhs_row"
+        if self.n == 1 and not self.trans_b:
+            return "rhs_col"
+        return None
+
+    @property
+    def gemv_n(self) -> int:
```
lcy-seso, line 42: Total function, meaningless value: a call with no GEMV form gets `m` back. Callers happen to guard by dispatch key, but the property is public. Now raises.

src/tileops/kernels/gemm/call_spec.py
```diff
@@ -21,14 +24,61 @@ class GemmCall(CallSpec):
     trans_a: bool = False
     trans_b: bool = False
 
+    @property
+    def gemv_mode(self) -> Optional[str]:
```
lcy-seso: Typed `Optional[Literal["lhs_row", "rhs_col"]]` — the two values are the contract, and callers switch on them.

src/tileops/ops/gemm/gemm.py
```diff
@@ -16,7 +17,9 @@
 
 from ..op_base import Op
 
-__all__ = ["GemmFp8FwdOp", "GemmFwdOp", "GemmW4A16FwdOp"]
+__all__ = ["GEMM_KEYS", "GemmFp8FwdOp", "GemmFwdOp", "GemmW4A16FwdOp"]
+
+GEMM_KEYS = ("gemv_kernel", "small_batch_kernel", "gemm_kernel")
```
lcy-seso: The MoE siblings keep the equivalent private (`_GEMM_KEYS` in `routed_expert/`). Made private here too; the dispatch test imports the private name.

src/tileops/kernels/gemm/heuristics.py
```diff
@@ -0,0 +1,403 @@
...
+from typing import Optional
+
+__all__ = [
+    "SWAP_AB_BLOCK_NN",
+    "SWAP_AB_MPAD",
+    "best_config",
+    "gemv_config",
+    "small_batch_config",
+    "swap_ab_grid_underfills",
+    "swap_ab_stages",
```
lcy-seso: `swap_ab_stages` is exported but called only inside this module, and `SWAP_AB_BLOCK_NN` below is imported by nobody. `__all__` overstates the surface; both are private now.

benchmarks/cublaslt_baseline.py
```diff
@@ -0,0 +1,163 @@
+"""cuBLASLt *searched-best* GEMM baseline for benchmarks.
```
lcy-seso: `bench_gemm.py` is the only consumer, and `baselines.py` holds the generic baselines several bench files share. Moved the module into `bench_gemm.py`; `baselines.py` goes back to identical with main.

benchmarks/cublaslt_baseline.py
```diff
@@ -0,0 +1,163 @@
...
+__all__ = ["CublasLtBestGemm", "make_cublaslt_best"]
+
+_SUPPORTED_DTYPES = (torch.float16, torch.bfloat16)
+
+_WORKSPACE_BYTES = 256 * 1024 * 1024
+
+_N_CANDIDATES = 8
+
+
+def _busy_ms(run: Callable[[], object]) -> float:
```
lcy-seso: This recomputes `median(device_busy_ms)`, which `benchmark_base._build_result` already computes. Two definitions of the number a row is selected on and reported with can drift apart, which is exactly what the docstring above promises cannot happen. Now one definition, `timing.median_busy_ms`, called by both.

lcy-seso (commented):
Review of the GEMM work, with the changes already applied on this branch — the comments below say why each one.

Two more, without a line to hang them on: `_FP8_BLOCK` names the block-scale `128` that was a repeated literal in `bench_gemm.py`, and the dependency-file changes are out, since `nvmath-python` reached the runner image through #2003.

Left alone on purpose: the numeric literals inside the TileLang kernel bodies, the candidate tile sets and wave-fill thresholds in `heuristics.py`, `GemmKernel`'s structure flags, and `_splitk_pair`'s parameter list. Those are implementation decisions, not cleanup.

src/tileops/kernels/gemm/dense.py
```diff
@@ -593,6 +630,1007 @@ def _gemm_main(
...
+    edge on compute-bound prefill shapes (large M, GPU-filling grid).
+
+    A static-wave persistent loop (grid = ``sm_count``; each CTA sweeps tile ids
+    ``flat_id = sm_count * w + pid``) overlaps a tile's TMA-store epilogue with
+    the next tile's mainloop prologue: the ring counters carry across waves so
+    the producer keeps prefetching. A Triton-style grouped tile order
+    (``group_size_m``) keeps concurrently-resident CTAs on a shared ``B`` column
+    stripe for L2 reuse.
+
+    Wave quantization is the dominant residual against cuBLAS here. Sweeping
```
lcy-seso: Measurement narrative and rejected alternatives in a docstring: the wave-quantization sweep, the two removal attempts and their numbers. That belongs in the PR body, not in code a caller reads. Trimmed here and in the other kernel builders.

src/tileops/kernels/gemm/dense.py
```diff
@@ -817,46 +2148,82 @@ def __init__(
...
+        dtype: Input/output torch dtype (fp16 / bf16); fp32 accumulation.
+        config: Optional explicit config; defaults to :attr:`default_config`.
+        tune: Whether to autotune over :attr:`autotune_configs`.
+    """
+
+    supported_archs: list[int] = [90]
+
+    @classmethod
+    def applies(cls, call) -> bool:
+        return small_batch_region(call)
```
lcy-seso: `applies` probed the current device through `get_sm_count()`, so a dispatch predicate read global CUDA state rather than the call. `CallSpec` now carries `sm_count`, resolved when unstated like `arch` and `h200`; this reads `call.sm_count`.

src/tileops/kernels/gemm/heuristics.py
```diff
@@ -0,0 +1,381 @@
...
+
+import functools
+import math
+from dataclasses import dataclass
+from typing import Optional
+
+__all__ = ["SWAP_AB_BLOCK_NN", "SWAP_AB_MPAD", "best_config", "gemv_config",
+           "small_batch_config", "swap_ab_stages"]
+
+_SMEM_BUDGET = 227 * 1024  # SM90 per-CTA opt-in SMEM ceiling
```
lcy-seso: Still open, despite the thread being resolved. The six ranking constants — `_L1_TBPS`, `_L2_TBPS`, `_TC_TFLOPS`, `_RED_TBPS`, `_LAUNCH_US`, `_ISSUE_NS` — are H200 measurements, and nothing here reads a device property. On an SM90 part with a different compute/bandwidth balance the scorer ranks against the wrong numbers and `supported_archs` admits it anyway. The branch's answer so far is a docstring stating the constants describe one device class, which records the limitation without removing it.

Fixing it in this PR. These are measured per-device values, and this repo already has a home for those: `src/tileops/perf/profiles/<gpu>.yaml`, loaded by `find_profile(torch.cuda.get_device_name())`, whose contract is to return `None` rather than guess a ceiling for a device it does not claim. The constants move into the profile, the selector reads the current device's, and a device with no calibration gets `GemmKernel`'s modal config instead of a ranking calibrated for another board.

lcy-seso (approved):
LGTM

PR #1908, opened 2026-08-13
[Maintain][Bench] Trim timing.py comments to what still holds

lcy-seso:
Comments in `benchmarks/timing.py` that no longer match the code.

`_l2_flush_cache` was declared twice, and a divider plus one comment named a benchmark
protocol and an input pool that are both gone.

The budget comment claimed a short kernel gets many samples and a long one few. On the CI
card the L2 flush costs 43.3 us alone, so a 6.5 us and a 28.2 us kernel both calibrate near
it and both clamp to `_MAX_ITERS` — 200 warmup, 200 repeat, identical. The comment now says
that, and that the flush is in the calibrated cost but not in the reading.

Two small cleanups alongside: `bench_kernel` warmed one iteration immediately before the
warmup loop, which opens with the same pair (one fewer warmup iteration, the only behaviour
change), and `_kernel_busy_us` now guards empty input like `_kernel_span_us` does.

`_MAX_ITERS` and the budget semantics are unchanged, so sample values are unchanged.

### DIFF
--- a/benchmarks/timing.py
+++ b/benchmarks/timing.py
@@ -24,8 +24,9 @@ import torch
 
 _logger = logging.getLogger("tileops.bench")
 
-# Per-phase wall-time budgets in ms; iteration counts derive from them, so a
-# short kernel gets many samples and a long one few.
+# Per-phase wall-time budgets in ms, divided by the cost of one cold-isolated iteration.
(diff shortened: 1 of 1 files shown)


lcy-seso:
Folded into #1909 (same commit, authorship preserved) to keep the two benchmark fixes and this trim in one review.

PR #1911, opened 2026-08-14
[Bench] Require a benchmark case to be named

lcy-seso:
109 benchmark cases collect as `sub-shape0-dtype0-output_dtype0` or `1-2-64-4-False-False-dtype0-False`. The id is the workload's name wherever it is read later — nightly report, published page, perf history key — so those name nothing.

Naming is the author's job, so this is a rule and a check, not a generator:

- `.claude/domain-rules/benchmark.md` — every case is named; take it from a manifest `label`, or pass `id=` to `pytest.param`. The name says which scenario the case stands for, not which parameters it holds.
- `benchmarks/tests/test_workload_names.py` — parses the bench files (no GPU, no baseline imports), counts cases with no id, holds each file to today's count. Never raises a number.

Two mechanical parts are fixed here:

- `pytest_make_parametrize_id` renders the values with no readable repr — dtype, shape tuple, boolean — so `shape0-dtype0` reads `1024x4096-float16`. Ids stay the same length as before (median 25, max 85, over 1087 cases, none lost or merged); a case still has no *name* until an author gives it one.
- Seven attention workloads ended their label with the dtype the id already appends, reading `…-fp16-float16`.

The work of naming the 109 is #1912.


lcy-seso:
Folded into #1910 — same commit, one PR instead of two.

PR #1930, opened 2026-08-17
[Perf][foundry][Attention] Accelerate paged MHA decode on H200 with a warp-specialized kernel, and select it at the dispatch site

zhen8838:
## Summary

- Paged MHA decode spent most of its time outside its kernels: one call issued 10 Torch operations -- an int reduction, a device-to-host memcpy that synchronises the host, two integer fills, two copies, a CUB scan pair and two half fills -- in front of the two TileLang launches that do the work. The new kernel needs none of them: the per-sequence length and the block table go to the device as they arrive, and the physical page base is computed inside the producer.
- The query is one token, so the score computation is a matrix-vector product and an MMA's M axis would be 94-98% padding. This kernel keeps both contractions on the CUDA cores and is written on TileLang's lower level: one producer and one consumer warp group split by warp specialization, TMA into a multi-stage shared-memory ring, mbarrier full/empty pairs with the phase parity carried by hand, an explicit register budget per group, a declared swizzle on the TMA destinations, and every reduction a warp shuffle chain over registers, so a consumer warp owns its rows outright and needs no block-wide reduction.
- Measured against the tensor-core alternative it is 8-16% faster on all four workloads, and against a single-launch variant that fuses the cross-split combine behind a device-scope ticket it is a wash on the largest workload and faster on the rest -- so the simpler two-launch form is what ships.
- Selection moves to the dispatch site: the warp-specialized kernel states the region it serves -- a one-token query, non-causal, no softcap, 16-bit, head dim a multiple of 32 up to 256, and a page size some tile height divides -- and the existing kernel is marked general, so a longer query, an odd head dim, an odd page size, a causal request or a pre-Hopper device keeps exactly the path it has today.

## TileFoundry Description

```python
@module(
    entry="mha_decode_paged",
    topologies=(Topology("cta", CTAS),),
)
class MhaDecodePagedWs:
    @func
    def mha_decode_paged(
        q: Tensor[(B, 1, H, D), DT],
        k: Tensor[(N_KV, H, D), DT],
        v: Tensor[(N_KV, H, D), DT],
        real_seqlen_kv: Tensor[(B,), "i32"],
        block_table: Tensor[(B, NUM_PAGES), "i32"],
    ) -> Tensor[(B, 1, H, D), DT]:
        with Mesh(
            ("cta",),
            layout=(B, H, SPLIT),
            names=("seq", "head", "split"),
        ) as cta:
            # Everything is authored inside the mesh, because a timeline has
            # nothing to say about an op whose result carries no position on the
            # level being analysed -- and the paging arithmetic is part of the
            # program, not a prologue to it. These first reshards place the
            # arguments without moving them: no mesh axis is named, so every
            # position sees the whole tensor.
            k_cache = tf.reshard(k, (N_KV, H, D), "gmem")
            v_cache = tf.reshard(v, (N_KV, H, D), "gmem")
            table = tf.reshard(block_table, (B, NUM_PAGES), "gmem")
            lengths = tf.reshard(real_seqlen_kv, (B,), "gmem")
            q_placed = tf.reshard(q, (B, 1, H, D), "gmem")

            # The cache is physical: logical page p of sequence b lives at
            # block_table[b, p], so the pages are gathered into each sequence's
            # logical order before anything attends over them.
            pages = tf.reshape(table, new_shape=(B * NUM_PAGES,))
            k_logical = tf.reshape(
                tf.index_select(
                    tf.reshape(k_cache, new_shape=(NUM_PAGES, PAGE, H, D)), pages, dim=0
                ),
                new_shape=(B, 1, N_KV, H, D),
            )
            v_logical = tf.reshape(
                tf.index_select(
                    tf.reshape(v_cache, new_shape=(NUM_PAGES, PAGE, H, D)), pages, dim=0
                ),
                new_shape=(B, 1, N_KV, H, D),
            )
            q_rows = tf.reshape(q_placed, new_shape=(B, 1, 1, H, D))

            # Staged where TMA delivers them: one CTA holds one head's worth of
            # one KV split of one sequence.
            k_tile = tf.reshard(
                k_logical, (B @ cta.seq, 1, N_KV @ cta.split, H @ cta.head, D), "smem"
            )
            v_tile = tf.reshard(
                v_logical, (B @ cta.seq, 1, N_KV @ cta.split, H @ cta.head, D), "smem"
            )

            # The arithmetic is in registers, so the query row and the staged
            # tiles are read out of their residency into rmem first.
            q_reg = tf.reshard(q_rows, (B @ cta.seq, 1, 1, H @ cta.head, D), "rmem")
            k_reg = tf.reshard(
                k_tile, (B @ cta.seq, 1, N_KV @ cta.split, H @ cta.head, D), "rmem"
            )
            v_reg = tf.reshard(
                v_tile, (B @ cta.seq, 1, N_KV @ cta.split, H @ cta.head, D), "rmem"
            )

            # f32 throughout: the inputs are 16-bit, but every accumulation this
            # operator performs is f32, and the tolerance is stated against that.
            q32 = tf.cast(q_reg, dtype="f32")
            k32 = tf.cast(k_reg, dtype="f32")
            v32 = tf.cast(v_reg, dtype="f32")
            raw = tf.reduce(q32 * k32, axes=(-1,), keepdim=True, kind="sum")
            score = raw * tf.full_like(raw, value=SCALE)

            # Cache rows past a sequence's real length hold nothing yet. The
            # coordinates and the lengths are placed too: a timeline has nothing
            # to say about an operand that names no position on its level.
            position = tf.reshard(
                tf.reshape(
                    tf.arange(type=Tensor[(N_KV,), "i32"]), new_shape=(1, 1, N_KV, 1, 1)
                ),
                (1, 1, N_KV @ cta.split, 1, 1),
                "rmem",
            )
            length = tf.reshard(
                tf.reshape(lengths, new_shape=(B, 1, 1, 1, 1)),
                (B @ cta.seq, 1, 1, 1, 1),
                "rmem",
            )
            live = tf.where(
                tf.cmp_lt(position, length), score, tf.full_like(score, value=MASKED)
            )

            # One softmax over the whole gathered cache. Every reduction here
            # runs over the axis the mesh split, so each leaves a pending
            # cross-CTA combine rather than a finished value.
            peak = tf.reduce(live, axes=(-3,), keepdim=True, kind="max")
            weight = tf.exp(live - peak)
            total = tf.reduce(weight, axes=(-3,), keepdim=False, kind="sum")
            blended = tf.reduce(weight * v32, axes=(-3,), keepdim=False, kind="sum")
            out = tf.cast(blended / total, dtype=DT)

            # Dropping the KV axis already left (B, 1, H, D); settling what the
            # split positions left pending is what this return to gmem is.
            return tf.reshard(out, (B @ cta.seq, 1, H @ cta.head, D), "gmem")
```

## Performance

Operator: `MultiHeadAttentionDecodePagedWithKVCacheFwdOp`

| Environment | Value |
| --- | --- |
| image | ghcr.io/tile-ai/tileops-runner:cu132-torch2.13-tl-afcebed1-dev |
| digest | sha256:2590888968a870216e1ea076829b909e62e9053608f6a65d5ab5ecd7eb5561f7 |
| gpu | NVIDIA H200 |
| driver | 595.71.05 |
| cuda | 13.2 |
| torch | 2.13.0+cu132 |
| flashinfer | 0.6.16.post3 |

Method: One process, one GPU, one set of inputs, the manifest benchmark's own comparison harness: same warmup and repeat budget, same L2 flush before every iteration, same forward-then-reversed implementation order, same CUPTI attribution, and both TileLang implementations autotuned. Device-busy time is the union of a call's kernel execution intervals, so it excludes the gaps between them. Reported latency is the median of three whole-comparison repeats; the percentage beside it is the min-to-max spread across those repeats. The candidate is reached by constructing the public Op with manifest arguments and letting dispatch choose, with no candidate-only switch.

Ratio in comparator columns: implementation / candidate. &#x1F7E2; > 1 means the candidate is faster;
&#x1F534; <= 1 means it is not.

| Workload | B | H | S_kv | D | page | Dtype | This PR (ms) | Base (ms)<br>/ candidate | FlashInfer (ms)<br>/ candidate |
| --- | ---: | ---: | ---: | ---: | ---: | --- | ---: | ---: | ---: |
| single-token-page128-float16 | 1 | 16 | 512 | 128 | 128 | float16 | 0.0060955 (+/-1.85%) | **0.025952 (+/-11.5%)<br>&#x1F7E2;&nbsp;4.2576x** | 0.009312 (+/-0%)<br>&#x1F7E2;&nbsp;1.5277x |
| batch2-page256-float16 | 2 | 8 | 1024 | 64 | 256 | float16 | 0.0057445 (+/-3.34%) | **0.022625 (+/-9.4%)<br>&#x1F7E2;&nbsp;3.9385x** | 0.009792 (+/-0%)<br>&#x1F7E2;&nbsp;1.7046x |
| longer-cache-float16 | 1 | 8 | 1024 | 64 | 256 | float16 | 0.0053115 (+/-0.56%) | **0.023647 (+/-10.2%)<br>&#x1F7E2;&nbsp;4.4520x** | 0.009568 (+/-0.33%)<br>&#x1F7E2;&nbsp;1.8014x |
| shorter-cache-float16 | 1 | 8 | 512 | 64 | 256 | float16 | 0.004672 (+/-3.07%) | **0.021152 (+/-3.18%)<br>&#x1F7E2;&nbsp;4.5274x** | 0.0091995 (+/-0.34%)<br>&#x1F7E2;&nbsp;1.9691x |
| geometric mean |  |  |  |  |  |  | 0.00542932 | **0.0232794<br>&#x1F7E2;&nbsp;4.2877x** | 0.00946509<br>&#x1F7E2;&nbsp;1.7433x |

## Result And Limitations

**measured SOTA**

- Every-row SOTA: all four manifest workloads are faster than the fastest runnable external baseline, by 53% to 97% of the candidate's own time, against a run-to-run spread of at most 3%. No regression remains on this operator's manifest surface.
- FA3 is the other external that can serve this contract, but only where the page size is a multiple of 256, so it has no result on the first row and is not a column above. Where it runs it is slower than FlashInfer: batch2-page256 0.01837 ms, longer-cache 0.01808 ms, shorter-cache 0.01763 ms.
- Kernels per call go from 11 to 2, and the remaining second launch is a cross-split combine whose device time equals this device's empty-kernel floor -- its body is not what it costs. Fusing it away was implemented and measured, and did not win.
- The candidate is also closer to the reference than the base path on the same inputs, because it carries the log-sum-exp and the partial outputs in float32 where the base path stores them in float16.
- Correctness is checked beyond the manifest shapes: a partial last page, a cache far shorter than its static bound, a ragged batch, a shuffled non-identity block table, one token of cache, head dim 256, bfloat16, and every configuration the autotuner may select on every manifest shape.
- Hopper only: the kernel needs TMA and warpgroup register control, so it declares SM90 and selection falls back to the existing kernel elsewhere.
- The operator's test file grows from 4 nodes to 9: one bfloat16 dtype-dispatch case, three parametrisations of a NaN regression on a cache shorter than its bound, and one assertion that a multi-token query is handed to the general kernel.

### DIFF
--- a/src/tileops/kernels/attention/__init__.py
+++ b/src/tileops/kernels/attention/__init__.py
@@ -31,6 +31,7 @@ from .gqa_sliding_window_varlen_fwd import (
 )
 from .mha_decode import MHADecodeKernel
 from .mha_decode_paged import MHADecodePagedKernel
+from .mha_decode_paged_ws import MHADecodePagedWsKernel
 
 __all__ = [
     "FlashAttnBwdPreprocessKernel",
@@ -54,6 +55,7 @@ __all__ = [
     "GQASlidingWindowVarlenFwdWgmmaPipelinedKernel",
     "MHADecodeKernel",
     "MHADecodePagedKernel",
+    "MHADecodePagedWsKernel",
     "MLADecodeWsKernel",
     "NSACmpFwdVarlenKernel",
     "NSAFwdVarlenKernel",
--- a/src/tileops/kernels/attention/call_spec.py
+++ b/src/tileops/kernels/attention/call_spec.py
@@ -21,6 +21,7 @@ __all__ = [
     "decode_bs1_region",
     "dense_prefill_region",
     "fp8_dtype",
+    "paged_decode_ws_region",
     "square_ws_prefill_region",
     "uses_sliding_window",
 ]
@@ -135,6 +136,40 @@ def causal_ws_prefill_region(call: AttentionCall) -> bool:
         and call.dim == 128
         and call.dtype in ATTENTION_DTYPES
     )
+
+
+#: Tile heights the warp-specialized paged decode kernel can pick from. A tile
+#: divides the page size, so one tile never straddles two pages, and it splits
+#: evenly across the four consumer warps.
+_WS_DECODE_TILES = (16, 32, 64, 128)
+#: Head dims that map onto one warp: the score reduction is a shuffle chain over
+#: 32 lanes, so a lane owns ``dim / 32`` elements of the head vector.
+_WS_DECODE_LANES = 32
+
+
+def paged_decode_ws_region(call: AttentionCall) -> bool:
+    """The paged-decode region the warp-specialized MHA kernel serves.
+
+    Stated positively, and only in terms the call already carries. What is left
+    to the general kernel: a query longer than one token (this kernel's whole
+    reason for skipping the tensor cores is that ``seqlen_q`` is 1), a head dim
+    that does not divide across a warp, a page size no tile height divides, a
+    softcap, and a causal request -- which for a one-token query against a
+    finished cache is not the same computation.
+    """
+    if call.max_seqlen_q != 1 or call.is_causal or call.softcap != 0.0:
+        return False
+    if call.dtype not in ATTENTION_DTYPES or call.is_fp8:
+        return False
+    if uses_sliding_window(call):
+        return False
+    if call.dim % _WS_DECODE_LANES != 0 or not 0 < call.dim <= 256:
+        return False
+    if call.page_size <= 0 or call.seqlen_kv <= 0:
+        return False
+    return any(
+        tile <= call.page_size and call.page_size % tile == 0 and tile <= call.seqlen_kv
+        for tile in _WS_DECODE_TILES)
 
 
 def decode_bs1_region(call: AttentionCall) -> bool:
--- a/src/tileops/kernels/attention/mha_decode_paged.py
+++ b/src/tileops/kernels/attention/mha_decode_paged.py
@@ -433,6 +433,10 @@ def _(batch: int, heads: int, seqlen_q: int, seqlen_kv: int, dim: int, page_size
 
 class MHADecodePagedKernel(Kernel):
     supported_archs: list[int] = [80, 89, 90]
+    #: The implementation behind the specialised one for this key: it serves any
+    #: paged decode call, including the query lengths, head dims and page sizes
+    #: the warp-specialized Hopper kernel does not claim.
+    general: bool = True
 
     def __init__(self,
                  batch,
--- a/src/tileops/kernels/attention/mha_decode_paged_ws.py
+++ b/src/tileops/kernels/attention/mha_decode_paged_ws.py
@@ -1,0 +1,467 @@
+"""Warp-specialized paged MHA decode for Hopper.
+
+``seqlen_q`` is 1, so the score computation is a matrix-vector product: the M
+axis of an MMA would be 94-98% padding, and the tensor cores buy nothing on a
+kernel whose whole KV cache is a few megabytes. This kernel therefore keeps the
+contractions on the CUDA cores and spends its budget on the two things that do
+decide the time -- how few launches reach the device, and how early the KV tiles
+are in flight.
+
+The pipeline is written out rather than scheduled:
+
+* one producer warp group and one consumer warp group, split by ``T.ws`` and
+  placed in the two arms of a single ``If`` (two sequential ``T.ws`` regions let
+  the thread-sync pass emit a block-wide barrier the producer never reaches);
+* ``T.tma_copy`` moves each K and V tile into a per-stage ring, announced on
+  ``T.alloc_barrier`` mbarriers with the phase parity carried by hand;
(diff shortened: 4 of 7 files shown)


lcy-seso (approved):
LGTM, cool

PR #1938, opened 2026-08-18
[Perf][Convolution] Split Conv2d kernels on has_bias at compile time

### LINKED ISSUE #1937: [Perf][Ops] Conv2dFwd: no-bias calls pay a zero-bias alloc and fill kernel on every invocation
## Problem

Every Conv2d kernel (`Conv2d1x1Kernel`, `Conv2dKernel`, `Conv2dSymmetricKernel`, `GroupConv2dKernel`) declares `bias` as a mandatory prim_func parameter. When the caller has no bias — the common case for bottleneck convolutions followed by normalization layers — `forward()` materializes `torch.zeros(c_out)` on **every call** just to satisfy the kernel ABI. That is one CUDA allocation plus one fill kernel launch of pure overhead: the kernel then adds zeros.

## Evidence

Measured on H200 (nightly-aligned environment: tilelang @ `afcebed1`, torch 2.13.0+cu132), no-bias conv2d workloads before/after removing the dead fill:

| workload | before (ms) | after (ms) | delta |
|---|---|---|---|
| resnext-grouped-3x3-float16 | 0.00509 | 0.00400 | -21.4% |
| bottleneck-expand-1x1-fp16-float16 | 0.00486 | 0.00390 | -19.8% |
| stem-3x3-s2-fp16-float16 | 0.00445 | 0.00360 | -19.1% |
| resnet-1x1-fp16-float16 | 0.00534 | 0.00440 | -17.6% |
| mobilenetv2-depthwise-float16 | 0.00765 | 0.00670 | -12.4% |
| midres-5x5-s1-fp16-float16 | 0.01700 | 0.01610 | -5.3% |

The gain scales inversely with kernel duration (fixed ~1-2 µs fill launch), and the bias variants are bit-identical before/after (control group within ±1.6%).

## Prior art

- PyTorch `conv2d(bias=None)` never materializes a bias tensor — the ATen kernels branch on `bias.defined()` at zero cost.
- This repo's GEMM FP8 kernels already split prim_func variants on `has_bias` at trace time.

## Proposal

Split each Conv2d kernel's prim_func into two variants — `_conv2d_*_main(x, weight, out)` and `_conv2d_*_bias_main(x, weight, out, bias)` — selected by `has_bias` at trace time (compile-time dispatch, zero runtime cost). The wrapped custom op takes `Optional[bias]` and `forward()` passes `None` through instead of allocating zeros.

## Scope

- `Conv2d1x1Kernel`, `Conv2dKernel`, `Conv2dSymmetricKernel`, `GroupConv2dKernel` (this PR)
- Follow-up: the same `bias is None → torch.zeros` pattern exists in 8 more conv kernels (`Conv1dPointwiseKernel`, `Conv1dKernel`, `GroupConv1dKernel`, `Conv3dKernel`, `GroupConv3dKernel`, and the Conv1d direct path) and can receive the identical treatment.

RMLYC:
## Motivation

Closes #1937

Every Conv2d kernel declared `bias` as a mandatory prim_func parameter, so no-bias calls materialized `torch.zeros(c_out)` on every invocation — one allocation plus one fill kernel launch of dead work per call. This PR gives each Conv2d kernel two prim_func variants (with/without the bias parameter) selected by `has_bias` at trace time, aligning with PyTorch's zero-overhead `bias=None` handling and the existing GEMM FP8 pattern in this repo.

## Changes

- `src/tileops/kernels/convolution.py` only; no interface changes.
- `Conv2d1x1Kernel`, `Conv2dKernel`, `GroupConv2dKernel`: prim_func split into `_main` / `_bias_main`; wrapped custom op's `bias` parameter becomes `Optional[torch.Tensor]`; `forward()` passes `None` through instead of allocating a zero bias.
- `Conv2dSymmetricKernel`: same split, with `bias` moved to the last prim_func parameter so both variants share `out_idx=[5]`; the implicit-GEMM macro is split into bias / no-bias variants accordingly.

## Testing

- Unit tests (isolated processes): all conv2d dispatch/numerics tests pass, including `test_conv2d_no_bias_matches_torch`, `test_conv2d_no_bias_grouped_matches_torch`, and the 1x1/3x3/5x5 dispatch tests; verified numerics against `F.conv2d` for every kernel × {bias, no-bias}.
- Verified via profiler and kernel counts that the no-bias path launches exactly one kernel (no fill) and never calls `torch.zeros`.
- Full conv benchmark suite (`benchmarks/ops/bench_convolution.py`, 57 cases, single-process nightly-style run on H200): **57/57 PASS**.

## Performance (H200, tilelang @ afcebed1, torch 2.13.0+cu132)

No-bias workloads vs the published numbers (bias variants unchanged, within ±1.6%):

| workload | before (ms) | after (ms) | delta | ratio before→after |
|---|---|---|---|---|
| resnext-grouped-3x3-float16 | 0.00509 | 0.00400 | -21.4% | 3.74 → 4.71 |
| bottleneck-expand-1x1-fp16-float16 | 0.00486 | 0.00390 | -19.8% | 0.82 → 1.04 |
| stem-3x3-s2-fp16-float16 | 0.00445 | 0.00360 | -19.1% | 1.45 → 1.83 |
| resnet-1x1-fp16-float16 | 0.00534 | 0.00440 | -17.6% | 0.80 → 0.97 |
| bottleneck-reduce-1x1-fp16-float16 | 0.00551 | 0.00470 | -14.7% | 0.79 → 0.94 |
| late-stage-1x1-fp16-float16 | 0.00589 | 0.00510 | -13.4% | 1.08 → 1.25 |
| mobilenetv2-depthwise-float16 | 0.00765 | 0.00670 | -12.4% | 0.40 → 0.46 |
| classifier-1x1-fp16-float16 | 0.01020 | 0.00910 | -10.8% | 0.90 → 0.99 |
| midres-5x5-s1-fp16-float16 | 0.01700 | 0.01610 | -5.3% | 1.12 → 1.19 |

No workload regresses beyond noise (±3%).

## Checklist

- [x] PR title and commits follow the `[Perf][Scope]` convention.
- [x] Implementation follows Google Python Style; `ruff check` passes.
- [x] No manifest changes (no interface or signature change).
- [x] No new public tests; existing conv2d tests cover both bias modes.
- [x] Benchmarks record the `torch` baseline for every workload.

---

## Update: extended to Conv1d/Conv3d per review feedback

@superAngGao pointed out the Conv2d-only coverage left the optimization inconsistent — the Op layer already accepts an optional bias for all three dimensions, but the Conv1d/Conv3d kernel layer still allocated a zero bias on every no-bias call. Commit `7526f71` extends the identical treatment to `Conv1dPointwiseKernel`, `Conv1dKernel`, `GroupConv1dKernel` (group + direct paths), `Conv3dKernel`, and `GroupConv3dKernel`, making the optimization consistent across all convolution dimensions.

### Conv1d/Conv3d no-bias workloads vs the published numbers

| workload | before (ms) | after (ms) | delta | ratio before→after |
|---|---|---|---|---|
| encodec-init-bfloat16 | 0.00445 | 0.00355 | -20.2% | 1.51 → 1.88 |
| encodec-init-float16 | 0.00445 | 0.00355 | -20.2% | 1.50 → 1.86 |
| wav2vec2-layer1-bfloat16 | 0.00765 | 0.00662 | -13.4% | 2.46 → 2.83 |
| wav2vec2-layer1-float16 | 0.00765 | 0.00666 | -13.0% | 2.43 → 2.79 |
| encodec-deep-bfloat16 | 0.01290 | 0.01194 | -7.5% | 1.32 → 1.43 |
| encodec-deep-float16 | 0.01290 | 0.01200 | -7.0% | 1.32 → 1.42 |
| 3d-resnext-grouped-k3-float16 | 0.01600 | 0.01511 | -5.6% | 16.22 → 16.76 |
| r3d-stem-k3-s1-fp16-float16 | 0.02380 | 0.02262 | -4.9% | 4.81 → 5.05 |
| whisper-large-conv1-float16 | 0.04980 | 0.04787 | -3.9% | 1.14 → 1.18 |
| video-stage-downsample-k3-s2-fp16-float16 | 0.03580 | 0.03462 | -3.3% | 0.74 → 0.76 |

Within ±3% (unchanged): whisper-large-conv1-bfloat16 (-2.3%), 3d-unet-aspp-3x3x3-rate6-float16 (-2.8%), unet-encoder-k3-s1-bf16-bfloat16 (-0.3%).

### Conv1d/Conv3d bias variants (control group, generated code unchanged)

| workload | before (ms) | after (ms) | delta |
|---|---|---|---|
| encodec-deep-bfloat16 | 0.0124 | 0.01232 | -0.6% |
| encodec-deep-float16 | 0.0124 | 0.01232 | -0.6% |
| encodec-init-bfloat16 | 0.00358 | 0.00358 | 0.0% |
| encodec-init-float16 | 0.00362 | 0.00358 | -1.1% |
| wav2vec2-layer1-bfloat16 | 0.00691 | 0.00691 | 0.0% |
| wav2vec2-layer1-float16 | 0.00688 | 0.00691 | +0.4% |
| whisper-large-conv1-bfloat16 | 0.0482 | 0.04762 | -1.2% |
| whisper-large-conv1-float16 | 0.0482 | 0.04758 | -1.3% |
| r3d-stem-k3-s1-fp16-float16 | 0.0231 | 0.02291 | -0.8% |
| unet-encoder-k3-s1-bf16-bfloat16 | 0.3537 | 0.35314 | -0.2% |
| video-stage-downsample-k3-s2-fp16-float16 | 0.0354 | 0.03507 | -0.9% |

### Verification for the extension

- Numerics vs `F.conv1d`/`F.conv3d` for every Conv1d/Conv3d kernel path (pointwise, general, group, depthwise-direct, 3d general, 3d group) × {bias, no-bias}: all match.
- No-bias path verified to launch exactly one kernel and never call `torch.zeros`.
- Existing conv1d/conv3d unit tests: 33/33 pass.
- Full conv benchmark suite re-run (57 cases, single-process nightly-style, H200): **57/57 PASS**; no workload regresses beyond noise.

### DIFF
--- a/src/tileops/kernels/convolution.py
+++ b/src/tileops/kernels/convolution.py
@@ -118,6 +118,70 @@ def _conv1d_kernel(
             x: T.Tensor((n, c_in, l_in), dtype),  # type: ignore
             weight_flat: T.Tensor((c_out, k_total), dtype),  # type: ignore
             out: T.Tensor((n, c_out, out_l), dtype),  # type: ignore
+        ):
+            with T.Kernel(
+                T.ceildiv(out_l, block_n),
+                T.ceildiv(c_out, block_m),
+                n,
+                threads=threads,
+            ) as (bx, by, bz):
+                weight_shared = T.alloc_shared((block_m, block_k), dtype)
+                data_shared = T.alloc_shared((block_k, block_n), dtype)
+                out_local = T.alloc_fragment((block_m, block_n), accum_dtype)
+                out_shared = T.alloc_shared((block_m, block_n), dtype)
+
+                T.use_swizzle(10, enable=enable_rasterization)
+                T.clear(out_local)
+
+                tile_ol_start = bx * block_n
+                tile_ol_end = tile_ol_start + block_n - 1
+                tile_input_start = tile_ol_start * stride_l - pad_left
+                tile_input_end = tile_ol_end * stride_l + (kernel_l - 1) * dilation_l - pad_left
+                tile_spatial_full = (
+                    (tile_ol_end < out_l) & (tile_input_start >= 0) & (tile_input_end < l_in)
+                )
+
+                for k_iter in T.Pipelined(T.ceildiv(k_total, block_k), num_stages=num_stages):
+                    T.copy(weight_flat[by * block_m, k_iter * block_k], weight_shared)
+
+                    for i, j in T.Parallel(block_k, block_n):
+                        k_idx = k_iter * block_k + i
+                        ol = bx * block_n + j
+                        kw = k_idx // c_in
+                        ci = k_idx % c_in
+                        il = ol * stride_l + kw * dilation_l - pad_left
+                        if tile_spatial_full & ((k_iter + 1) * block_k <= k_total):
+                            data_shared[i, j] = x[bz, ci, il]
+                        else:
+                            in_bound = (k_idx < k_total) & (ol < out_l) & (il >= 0) & (il < l_in)
+                            data_shared[i, j] = T.if_then_else(
+                                in_bound,
+                                x[bz, ci, il],
+                                T.cast(0.0, dtype),
+                            )
+
+                    T.gemm(weight_shared, data_shared, out_local)
+
+                for i, j in T.Parallel(block_m, block_n):
+                    oc = by * block_m + i
+                    ol = bx * block_n + j
+                    out_shared[i, j] = T.if_then_else(
+                        (oc < c_out) & (ol < out_l),
+                        T.cast(out_local[i, j], dtype),
+                        T.cast(0.0, dtype),
+                    )
+
+                for i, j in T.Parallel(block_m, block_n):
+                    oc = by * block_m + i
+                    ol = bx * block_n + j
+                    if oc < c_out and ol < out_l:
+                        out[bz, oc, ol] = out_shared[i, j]
+
+        @T.prim_func
+        def _conv1d_bias_main(
+            x: T.Tensor((n, c_in, l_in), dtype),  # type: ignore
+            weight_flat: T.Tensor((c_out, k_total), dtype),  # type: ignore
+            out: T.Tensor((n, c_out, out_l), dtype),  # type: ignore
             bias: T.Tensor((c_out,), dtype),  # type: ignore
         ):
             with T.Kernel(
@@ -166,18 +230,11 @@ def _conv1d_kernel(
                 for i, j in T.Parallel(block_m, block_n):
                     oc = by * block_m + i
                     ol = bx * block_n + j
-                    if has_bias:
-                        out_shared[i, j] = T.if_then_else(
-                            (oc < c_out) & (ol < out_l),
-                            T.cast(out_local[i, j] + T.cast(bias[oc], accum_dtype), dtype),
-                            T.cast(0.0, dtype),
-                        )
-                    else:
-                        out_shared[i, j] = T.if_then_else(
-                            (oc < c_out) & (ol < out_l),
-                            T.cast(out_local[i, j], dtype),
-                            T.cast(0.0, dtype),
-                        )
+                    out_shared[i, j] = T.if_then_else(
+                        (oc < c_out) & (ol < out_l),
+                        T.cast(out_local[i, j] + T.cast(bias[oc], accum_dtype), dtype),
+                        T.cast(0.0, dtype),
+                    )
 
                 for i, j in T.Parallel(block_m, block_n):
                     oc = by * block_m + i
@@ -185,7 +242,7 @@ def _conv1d_kernel(
                     if oc < c_out and ol < out_l:
                         out[bz, oc, ol] = out_shared[i, j]
 
-        return _conv1d_main
+        return _conv1d_bias_main if has_bias else _conv1d_main
 
     return _conv1d_func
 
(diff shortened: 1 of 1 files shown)


superAngGao:
# Review: Incomplete Optimization Coverage Across Convolution Dimensions

## Summary
This PR optimizes Conv2d kernels by splitting them into separate variants with/without bias at compile time, eliminating unnecessary zero-bias tensor allocations. However, the optimization is **incomplete** — Conv1d and Conv3d kernels were left behind.

## Issue: Inconsistent Coverage Across Convolution Dimensions

### Current State After This PR

| Kernel | Op Layer Signature | Kernel Layer Signature | Zero-Bias Allocation |
|--------|-------------------|------------------------|---------------------|
| Conv1d | `bias: Optional[Tensor] = None` ✓ | `bias: Tensor` (required) | **Yes** ❌ |
| Conv2d | `bias: Optional[Tensor] = None` ✓ | `bias: Optional[Tensor] = None` ✓ | **No** ✓ |
| Conv3d | `bias: Optional[Tensor] = None` ✓ | `bias: Tensor` (required) | **Yes** ❌ |

### Problem

The Op layer (`src/tileops/ops/convolution.py`) **already supports** optional bias for all three dimensions (Conv1d, Conv2d, Conv3d), but this PR only updates the Conv2d **kernel layer**.

This creates an inconsistency where:
- **Conv2d**: When called with `bias=None`, no zero tensor is allocated ✓
- **Conv1d/Conv3d**: When called with `bias=None`, the kernel layer still creates `torch.zeros(c_out)` ❌

### Affected Kernels

The following kernels still materialize zero-bias tensors in their `forward()` methods:

**src/tileops/kernels/convolution.py:**

```python
class Conv1dPointwiseKernel(Kernel):
    def forward(self, x, weight, bias: Optional[Tensor] = None):
        if bias is None:
            bias = torch.zeros(self.c_out, device=x.device, dtype=x.dtype)  # ❌
        ...

class Conv1dKernel(Kernel):
    def forward(self, x, weight, bias: Optional[Tensor] = None):
        if bias is None:
            bias = torch.zeros(self.c_out, device=x.device, dtype=x.dtype)  # ❌
        ...

class Conv3dKernel(Kernel):
    def forward(self, x, weight, bias: Optional[Tensor] = None):
        if bias is None:
            bias = torch.zeros(self.c_out, device=x.device, dtype=x.dtype)  # ❌
        ...
```

### Impact

For Conv1d and Conv3d operations without bias:
- One unnecessary tensor allocation per call
- One unnecessary fill kernel launch per call
- Missed optimization opportunity already supported by the Op layer

## Recommendation

To complete this optimization, the same treatment should be applied to:

1. **Conv1d kernels**: `Conv1dPointwiseKernel`, `Conv1dKernel`, `GroupConv1dKernel`
2. **Conv3d kernels**: `Conv3dKernel`, `GroupConv3dKernel`

Each should:
- Split the prim_func into two variants (with/without bias)
- Update the wrapped kernel to accept optional bias and dispatch to the correct variant
- Remove the `torch.zeros()` fallback in `forward()`

This would make the optimization **consistent across all convolution dimensions** and fully leverage the optional-bias support already present in the Op layer.

---

**Suggested Action**: Either extend this PR to cover Conv1d/Conv3d, or file a follow-up issue to track the remaining work.

RMLYC:
Thanks for the thorough review — agreed the Conv2d-only coverage was inconsistent. Commit 7526f71 extends the same compile-time `has_bias` split to all Conv1d/Conv3d kernels (`Conv1dPointwiseKernel`, `Conv1dKernel`, `GroupConv1dKernel` including its direct path, `Conv3dKernel`, `GroupConv3dKernel`). Numerics verified against `F.conv1d`/`F.conv3d` for every path × {bias, no-bias}, existing conv1d/conv3d tests 33/33 pass, and the full conv benchmark suite (57 cases) passes with no regressions; per-workload numbers are in the updated PR description.

lcy-seso:
Closing in favour of #1942, which carries this optimization. The problem this PR reported is real and the approach it chose — splitting the prim_func on `has_bias` at trace time rather than materializing a zero bias — is the one that shipped. Three reasons this PR is not the vehicle:

1. **Half the diff edits a layer that is being deleted.** Adding `Optional[torch.Tensor]` to the ten kernel-level `custom_op` schemas and branching the call arity at each call site is work on `src/tileops/kernels/convolution.py`'s compile boundary. Threshold 4 of the multi-backend migration cookbook (RFC D6) moves that boundary to the op layer, and the RMSNorm migration in #1914 deleted the kernel-level `custom_op` outright. Conv1d/2d/3d now do the same, so those edits have nothing to attach to.

2. **545 lines of duplicated prim_func bodies.** Each of the ten pairs differs in exactly one expression — the bias term in the epilogue. A single `@T.macro` body feeding two prim_func signatures gives the same trace-time split with no duplication; the symmetric kernel in this PR already shows the macro form, the other nine are copies.

3. **`has_bias` ends up with two sources.** Which variant is compiled comes from the constructor flag while the argument count comes from `bias is not None` at the call. They agree today only because the op layer happens to build with `bias is not None`; a kernel constructed directly with the other value is an arity error. `gemm.py` already derives it from the call, and #1939 removed exactly this kind of second source.

The measured gain is unchanged by any of this: the first commit here (`81c55993`, caching the zero bias, ~10 lines) reported bottleneck-expand at 4.7 → 3.8 us, and the final table reports 4.86 → 3.90 us for the same workload. The 590 lines bought interface cleanliness, not speed — worth having, but it is one `@T.macro` away.

#1937 stays open until the replacement lands. Thanks for finding and measuring this; the numbers in the issue are what made the fill launch visible.

PR #1942, opened 2026-08-20
[Refactor][Ops] Move conv and pool onto the op/backend boundary, and make conv bias optional

### LINKED ISSUE #1937: [Perf][Ops] Conv2dFwd: no-bias calls pay a zero-bias alloc and fill kernel on every invocation

lcy-seso:
Supersedes #1938. Closes #1937.

## Problems

- The conv and pool ops handed over no tensors and took no `target=`, so no target could serve them, and the node a traced graph carried belonged to a kernel (conv) or to a whole family (pool).
- `bias` was a mandatory prim_func parameter: every no-bias call allocated `torch.zeros(C_out)` and paid a launch to add zeros.
- Depthwise Conv2d ran the grouped GEMM with `M = 1` against a 64-row tile, and the grouped kernels staged their weight tile element by element.

## Changes

- Conv1d/2d/3d and the twelve pool ops each register one operator with a manifest-driven fake, hand over `signature.inputs` in order, normalize contiguity, and take a keyword-only `target=`. The kernel-level `custom_op`s — ten conv, fifteen pool — and the two shared pool operators are deleted: one compile boundary per family, and it belongs to the op.
- Bias is optional end to end: two prim_func signatures over one `@T.macro` body selected by `has_bias` at trace time, presence in the kernel cache key, and a kernel that refuses a call built for the other side. An absent `bias` crosses to a builder as `None` in its own argument, not as a shorter argument list — with two optional inputs the count cannot say which one is missing.
- Depthwise Conv2d gets a direct multiply-accumulate kernel, and the grouped kernels lay `k` out in the weight's own order, which makes the weight staging a tile copy.
- Inputs on different devices are rejected before kernel selection, the CUDA requirement moves into the pool kernels, and call state is recorded after the launch returns rather than before.
- `torch_compile_fullgraph: true` on the conv entries, backed by a cold fullgraph test per dimension; those and the three pool fullgraph tests assert the graph holds nothing but the op's own node.
- `MeanPoolingForwardOp` stays in-tree: it has no manifest entry, so a target's builder could not be told what to build.
- Test nodes: `tests/ops/test_convolution.py` +9, `tests/test_op_backend_seam.py` +5.

## Performance

H200, dev runner image, autotune on, all 57 conv workloads, `device_busy_ms`, baseline `main` at 50e3fe58. Every no-bias row drops one launch — the zero-bias fill, present in 31 of the 57 rows. No row regresses. Pool is measured too, since it loses its kernel-level operators: 96 rows against `main`, none slower.

| workload | dtype | bias | before (ms) | after (ms) | delta |
|---|---|---|---|---|---|
| Conv2d (1,32,56,56) depthwise | float16 | no | 0.00760 | 0.00230 | -69.7% |
| Conv2d (1,128,28,28) grouped | float16 | no | 0.00500 | 0.00340 | -32.0% |
| Conv1d (1,1,24000) | float16 | no | 0.00440 | 0.00300 | -31.8% |
| Conv3d (1,3,16,112,112) | float16 | no | 0.02380 | 0.01890 | -20.6% |
| Conv3d (1,32,32,64,64) | bfloat16 | yes | 0.35500 | 0.28920 | -18.5% |

Conv1d keeps its original k order: laying it over `(c_in, kernel_l)` to make the weight staging a tile copy cost 49% on `(1,128,600)`.

## Not in this PR

- Route the general Conv2d/Conv3d staging through `T.im2col` the way the symmetric path does — needs the NHWC transform on those paths.
- Decide what float32 conv should do about TF32; it changes numerics, so it needs a tolerance decision first.
- Kernel constructors still take launch-time sizes, so a decode-shaped sweep rebuilds a kernel per size.

### DIFF
--- a/src/tileops/kernels/convolution.py
+++ b/src/tileops/kernels/convolution.py
@@ -113,13 +113,8 @@ def _conv1d_kernel(
         threads: int,
         enable_rasterization: bool,
     ):
-        @T.prim_func
-        def _conv1d_main(
-            x: T.Tensor((n, c_in, l_in), dtype),  # type: ignore
-            weight_flat: T.Tensor((c_out, k_total), dtype),  # type: ignore
-            out: T.Tensor((n, c_out, out_l), dtype),  # type: ignore
-            bias: T.Tensor((c_out,), dtype),  # type: ignore
-        ):
+        @T.macro
+        def _conv1d_body(x, weight_flat, out, bias):
             with T.Kernel(
                 T.ceildiv(out_l, block_n),
                 T.ceildiv(c_out, block_m),
@@ -185,6 +180,27 @@ def _conv1d_kernel(
                     if oc < c_out and ol < out_l:
                         out[bz, oc, ol] = out_shared[i, j]
 
+        if has_bias:
+
+            @T.prim_func
+            def _conv1d_bias_main(
+                x: T.Tensor((n, c_in, l_in), dtype),  # type: ignore
+                weight_flat: T.Tensor((c_out, k_total), dtype),  # type: ignore
+                out: T.Tensor((n, c_out, out_l), dtype),  # type: ignore
+                bias: T.Tensor((c_out,), dtype),  # type: ignore
+            ):
+                _conv1d_body(x, weight_flat, out, bias)
+
+            return _conv1d_bias_main
+
+        @T.prim_func
+        def _conv1d_main(
+            x: T.Tensor((n, c_in, l_in), dtype),  # type: ignore
+            weight_flat: T.Tensor((c_out, k_total), dtype),  # type: ignore
+            out: T.Tensor((n, c_out, out_l), dtype),  # type: ignore
+        ):
+            _conv1d_body(x, weight_flat, out, None)
+
         return _conv1d_main
 
     return _conv1d_func
@@ -216,13 +232,8 @@ def _conv1d_direct_kernel(
         threads: int,
         enable_rasterization: bool,
     ):
-        @T.prim_func
-        def _conv1d_direct_main(
-            x: T.Tensor((n, c_in, l_in), dtype),  # type: ignore
-            weight: T.Tensor((c_out, 1, kernel_l), dtype),  # type: ignore
-            out: T.Tensor((n, c_out, out_l), dtype),  # type: ignore
-            bias: T.Tensor((c_out,), dtype),  # type: ignore
-        ):
+        @T.macro
+        def _conv1d_direct_body(x, weight, out, bias):
             with T.Kernel(
                 T.ceildiv(out_l, block_n),
                 T.ceildiv(c_out, block_m),
@@ -257,6 +268,27 @@ def _conv1d_direct_kernel(
                             )
                         else:
                             out[bz, oc, ol] = T.cast(out_local[i, j], dtype)
+
+        if has_bias:
+
+            @T.prim_func
+            def _conv1d_direct_bias_main(
+                x: T.Tensor((n, c_in, l_in), dtype),  # type: ignore
+                weight: T.Tensor((c_out, 1, kernel_l), dtype),  # type: ignore
+                out: T.Tensor((n, c_out, out_l), dtype),  # type: ignore
+                bias: T.Tensor((c_out,), dtype),  # type: ignore
+            ):
+                _conv1d_direct_body(x, weight, out, bias)
+
+            return _conv1d_direct_bias_main
+
+        @T.prim_func
+        def _conv1d_direct_main(
+            x: T.Tensor((n, c_in, l_in), dtype),  # type: ignore
+            weight: T.Tensor((c_out, 1, kernel_l), dtype),  # type: ignore
+            out: T.Tensor((n, c_out, out_l), dtype),  # type: ignore
+        ):
+            _conv1d_direct_body(x, weight, out, None)
 
         return _conv1d_direct_main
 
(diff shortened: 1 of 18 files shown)


lcy-seso:
Two places read bias presence differently. Pushed two commits that make them agree.

`_infer_output_shapes` reads it off a `None` argument — the argument stays, its value says
whether the call passed a bias:

```python
shapes = op._infer_output_shapes(
    tuple(input.shape),
    tuple(weight.shape),
    None if bias is None else tuple(bias.shape),
)
```

The get-or-build reads it off the argument count instead:

```python
inputs = (input, weight) if bias is None else (input, weight, bias)
```

Same op, same fact, a dozen lines apart. Now both do the first thing:
`inputs = (input, weight, bias)`, three sites, and
`test_a_missing_optional_input_is_absent_from_the_hand_over` is renamed and asserts the
third argument is `None`.

**Why the argument stays rather than the other way round.** Conv has one optional input and
it is last, so the count does recover presence and this PR is self-consistent either way.
It does not generalize:

| op | optional inputs | ways to call it | distinct argument counts |
| --- | --- | --- | --- |
| `Conv2dFwdOp` | `bias` | 2 | 2 |
| `ClampFwdOp` | `min`, `max` | 4 | 3 |
| `GroupNormFwdOp` | `weight`, `bias` | 4 | 3 |
| `Mamba2FwdOp` | `dt_bias`, `initial_states` | 4 | 3 |
| `InstanceNormFwdOp` | `running_mean`, `running_var`, `weight`, `bias` | 16 | 5 |

A builder handed `(x_spec, some_spec)` cannot tell whether the second one is `min` or
`max`, and neither can the memory key: a clamp with only a lower bound and one with only
an upper bound hand over the same dtypes and shapes, so the second call is served the
first one's kernel while `has_min` / `has_max` decide what the PrimFunc looks like. Wrong
result, no error. The information is lost at the boundary, so no key can recover it.

**Why two commits.** The `op_base.py` one — letting a get-or-build take `None` for an
absent optional input — is the same change #1941 makes. Keeping it separate means whichever
of the two lands first, the other's rebase drops it. Without it this PR would fail its own
seam tests, since `TensorSpec.of(None)` raises.

Verified locally: `tests/test_op_backend_seam.py` 24 passed; `tests/ops/test_convolution.py
-m smoke` 36 passed; `scripts/validate_manifest.py --strict` clean; `tests/test_validate_manifest.py`
148 passed with no GPU visible.

lcy-seso:
Pushed `e032737d`, which applies this PR's pattern to the pool family. Three findings came out of reviewing it against the conv changes, and all three were shared-code problems rather than conv ones, so fixing them here keeps the pattern in one place:

- Nine pool ops shared two operators. A traced graph then named the family, not the op, and no pool op declared a name a test could check. Each of the twelve concrete classes registers its own operator now, with a manifest-driven fake; the three existing parametric fullgraph tests gained a graph-ownership assertion and cover all twelve.
- The four get-or-build sites handed over no tensors, so no target could serve a pool op.
- The op layer rejected a non-CUDA input. That is the in-tree kernel's requirement; `kernels/pool/common.require_cuda` states it in the kernel layer, and the three error-message cases read the kernel's message.
- `_eager_forward` recorded the kernel, the extents and the dtype before the launch, so a call that raised left state describing a call that never ran.

Pool ops still take no `target=`; an external pool kernel is reachable by device detection only. That is the one gate left for the family.

`tests/ops/test_pool.py` 261 passed in the dev runner image; mean-pooling, seam and compile suites 33 passed; the CPU manifest gates are clean.

PR #1948, opened 2026-08-21
[Bench][Pool] Time pooling against the strongest baseline, not just torch eager

### LINKED ISSUE #1947: [Bench] Pool benchmarks compare against suboptimal PyTorch baselines
Related to #1806 — this issue is the Pooling-family concrete case of that umbrella request for external baselines.

## Problem

All pooling benchmarks in `benchmarks/ops/bench_pool.py` currently use PyTorch's native pooling (`test.ref_program`) as the baseline. PyTorch's pooling kernels are not the fastest available implementations for several of these operators, so the TileOps-vs-baseline comparison understates (or for some shapes misrepresents) TileOps' actual standing.

Measured on H200 (fp16/bf16/fp32, the full pool workload set in `workloads/pool.py`):

- **2D pooling**: FlagGems' triton kernels beat `torch.nn.functional` by ~1.4-1.7x; direct cuDNN v9 (Resample node) beats torch by ~1.1-1.5x.
- **3D pooling**: direct cuDNN beats torch by ~1.1-1.8x (FlagGems has no 3D pooling).
- **1D pooling**: torch's native 1D kernels are actually the fastest — both cuDNN and FlagGems can only emulate 1D as 2D and lose by ~2-3x.

## Proposal

Introduce cuDNN (v9 graph API Resample node, called directly through the backend C API — not via PyTorch) and FlagGems as candidate baselines, measure both across the existing workloads, and select per operator:

- 1D → keep torch (fastest)
- 2D → FlagGems
- 3D → cuDNN
- unsupported shapes (indices, adaptive pooling, dilation≠1, …) → torch fallback

No new test/benchmark cases; only the baseline callable changes.

RMLYC:
Closes #1947. Part of #1806.

## Problems

- torch eager is not the strongest pooling implementation available, so timing TileOPs against it alone overstates where TileOPs stands. Measured on an H200: torch through inductor beats eager by 2–6x on 1D, FlagGems takes 0.63–0.77 of eager for 2D, cuDNN 0.58–0.70 for 3D.
- Nothing checked that a baseline computes what the reference computes.

## Changes

- Every row times four implementations: `tileops`, the library where one applies (FlagGems 2D, cuDNN v9 Resample 3D), `torch-ref`, and `torch-compile`. The strongest is per row, not per op — inductor is 2.7x worse than eager on one 3D max-pool shape and 6x worse on another, so both torch readings stay.
- The compiled baseline resets dynamo per case: eight graphs per code object, 48 cases, one `ref_program`. Without the reset the later cases would time eager under a tag that says compiled.
- The library baseline is asserted against the reference before the case is timed, and a selected library that is missing raises rather than reporting torch under a tag that claims the library.
- cuDNN goes through the v9 backend in ctypes: cudnn-frontend 1.27 binds no resample node, and the legacy `cudnnPoolingForward` rejects bfloat16 while reaching the same kernel (0.0867 vs 0.0865 ms).
- `trust-model.md` §Benchmark and the benchmark domain rule now cover a baseline that is a different implementation, which the `ref_program` override could not name in the report.
- `flag_gems` is not in the `bench` extra: its metadata pins `numpy==1.26.4` against this stack's 2.2.6, so pip cannot resolve it there. #1951 installs it in the runner image with `--no-deps`.
- Test nodes: `benchmarks/tests/test_benchmark_boundaries.py` +1, a CPU-only check that torch is the baseline only where it is meant to be.

## What the numbers say now

`device_busy_ms` summed per op, H200. "strongest" is the per-row minimum over the three baselines.

| op | torch/tileops | compile/tileops | strongest/tileops | strongest source |
|---|---|---|---|---|
| AvgPool1d | 2.24 | 0.84 | **0.84** | compile 3/3 |
| MaxPool1d | 2.96 | 0.73 | **0.73** | compile 3/3 |
| MaxPool1dIndices | 1.73 | 0.62 | **0.62** | compile 3/3 |
| AvgPool2d | 2.11 | 2.39 | 1.27 | flaggems 1, compile 2 |
| MaxPool2d | 2.71 | 0.94 | **0.94** | compile 9/9 |
| MaxPool2dIndices | 1.30 | 0.68 | **0.68** | compile 9/9 |
| AvgPool3d | 2.30 | 0.88 | **0.88** | compile 3/3 |
| MaxPool3d | 4.05 | 4.38 | 1.74 | cudnn 1, compile 2 |
| MaxPool3dIndices | 1.38 | 3.16 | **0.76** | compile 2, torch 1 |
| AdaptiveAvgPool2d | 1.70 | 1.64 | 1.64 | torch 2, compile 1 |
| AdaptiveMaxPool2d | 4.33 | 2.09 | 2.09 | torch 2, compile 1 |
| AdaptiveMaxPool2dIndices | 1.46 | 0.74 | **0.74** | compile 1, torch 2 |

Against the strongest baseline TileOPs trails on seven of the twelve ops. The 1D ops are the clearest: all three lose to inductor, `MaxPool1dIndices (64,256,128)` by 3.2x. The with-indices variants are the pattern — TileOPs pays 2.07x for writing indices where FlagGems pays 1.00x.

Not addressed here: the nightly's ratio alert reads the first baseline in the row, which is the library or eager torch, not the strongest. Making it read the strongest is a change to `scripts/nightly_report.py`.

`bench_pool.py`: 48 passed. Numerics: FlagGems and cuDNN match the reference on 2D avg/max, `ceil_mode`, `count_include_pad=False` and 3D avg/max across float16, bfloat16 and float32 — exact in half precision, 2e-7 in float32.


lcy-seso:
Please hold the merge — the FlagGems baselines cannot run in CI until the runner image changes.

- No job installs the `bench` extra. The nightly and gpu-smoke run `install_tileops.sh` (`pip install -e . --no-deps`); preflight installs `.[dev]`. The runner image is what supplies bench dependencies.
- The image has no `flag_gems`, and `_pool_baselines.py` falls back to `("torch-ref", test.ref_program)` on `ImportError`. Nothing fails, so the nightly would keep reporting torch-ref for every 2D row.
- cuDNN needs nothing: `nvidia-cudnn-cu13 9.20.0.48` is already in the image, so that line in the extra is redundant.
- `flag_gems==5.0.2` cannot be resolved into the image: it pins `numpy==1.26.4` against the image's 2.2.6. The same pin breaks `pip install -e ".[bench]" -c constraints.txt`.
- What works, measured in the image on an H200: `--no-deps flag_gems==5.0.2` plus `sqlalchemy`, after which `ops.avg_pool2d` imports and matches `F.avg_pool2d` under numpy 2.2.6 / torch 2.13. So the Dockerfile takes the `--no-deps` install, `requirements.in` takes `sqlalchemy`, and the lock is regenerated.

The image change has to land and the runners have to be on the new tag before the 2D rows here take effect. Happy to open that PR.

lcy-seso:
Reviewed against the repo's own rules and measured on an H200 before touching anything, then pushed `13184351` onto this branch. The direction here is right — I verified it — and the fixes are about what the code says when something is missing.

**What I verified first**

- Both baselines compute what torch computes: 27 probes over 2D avg/max, `ceil_mode`, `count_include_pad=False`, 3D avg/max, fp16/bf16/fp32. fp16 and bf16 agree bit for bit, fp32 within 2e-7. Nothing in `compare()` checks this, so it was worth checking.
- The baselines really are stronger than torch, per-row against `main` on the same machine: flaggems takes 0.63–0.77 of torch's device time for 2D, cudnn 0.58–0.70 for 3D. The reported TileOps advantage drops accordingly — AvgPool2d 2.13 -> 1.58, MaxPool2d 2.70 -> 1.70, MaxPool3d 4.05 -> 2.83, and MaxPool2dIndices 1.30 -> 0.82. That last number reproduces your geomean exactly.
- The raw v9 backend calls are the only route to a bf16 cuDNN pooling baseline. cudnn-frontend 1.27 exposes 141 graph ops and no resample/pooling, and the legacy `cudnnPoolingForward` rejects `CUDNN_DATA_BFLOAT16`. It also reaches the same kernel as the graph API (0.0867 vs 0.0865 ms), so there is no cheaper path being left on the table. Recorded in the file so the next reader does not re-litigate it.
- `_PROPAGATE_NAN` is the right choice and it is not free: the non-propagating mode is 19% faster and returns a number where torch returns NaN. Also recorded.

**What changed**

- A selected library that is missing now raises, naming the library. Before, `except Exception` turned it into a silent torch baseline — the runner image ran these benchmarks for a day without flag_gems and the report just said `torch-ref`. Falling back to torch stays for the two cases where it is the answer: an op the table does not name, and a workload the library cannot express.
- torch is timed alongside the library baseline instead of replaced. `scripts/nightly_report.py` alerts when `baseline_ratio < 0.80` against the *first* baseline, and its history was measured against torch; with torch gone that alert silently changes meaning. Worth knowing either way: 3 of 47 rows are already below 0.80, two of them MaxPool2dIndices.
- The cuDNN descriptor calls are methods on the handle that owns them rather than seven free functions taking `lib` first, which also gives the cached descriptors somewhere to be destroyed.
- One table drives the per-op choice. `_SELECTED_BASELINE` named twelve ops but `_OP_KIND_NDIM` gated six of them out first, and `_EXCLUDED_CASES` was an empty set that `pool_baseline`'s docstring cited as a fallback reason.
- `flag_gems` is out of the `bench` extra: its metadata pins `numpy==1.26.4` against this stack's 2.2.6, so `pip install -e ".[bench]" -c constraints.txt` cannot resolve it at all. #1951 installs it in the runner image with `--no-deps` and holds the version in `constraints.txt`. `nvidia-cudnn-cu13` was redundant — torch already brings it.
- One CPU-only routing test: an unlisted op and an inexpressible workload take torch, and the five listed ops still name a library. It caught a real ordering bug while being written — parameter support has to be decided before the library is required.

`benchmarks/ops/bench_pool.py`: 48 passed, and the report now carries `tileops` + the library baseline + `torch-ref` per row.

Issue #1916, opened 2026-08-15
[Refactor][Attention] Consolidate dense GQA prefill topology


superAngGao:
Parent: #1915
Integrated reference: #1913
Dispatch-contract reference: #1896

## Objective

Consolidate the complete Dense BSHD GQA Prefill topology behind one manifest-backed public Op and one external target role. This is one topology sub-issue: it includes the Dense semantic variants and implementation cleanup needed to replace the old fixed-length surfaces.

## Public contract

- Public Op: `GroupedQueryAttentionPrefillDenseFwdOp`.
- External role: `gqa_prefill_dense`, following #1896.
- Fixed tensor order:
  `q, k, v, q_scale, k_scale, v_scale, rope_cos, rope_sin`.
- Dense BSHD Q/K/V only; no KV-cache mutation.
- Full/causal/window, softcap, custom scale, RoPE layout, precision, and NVIDIA schedule remain internal dispatch dimensions.
- Disabled FP8 and RoPE features use normalized identity scales and dummy tables so the callable signature stays fixed.

### Op and manifest

- Introduce `GroupedQueryAttentionPrefillDenseFwdOp`.
- Replace `GroupedQueryAttentionFwdOp`.
- Absorb `GroupedQueryAttentionSlidingWindowFwdOp` instead of keeping sliding window as a public Op.
- Support square and rectangular Dense Q/KV lengths.
- Support full, causal, left/right window, softcap, and custom softmax scale.
- Support fused RoPE with explicit `neox` and `interleaved` layouts.
- Validate target-neutral shape, dtype, device, contiguity, scale, and RoPE-table semantics.
- Migrate `MultiHeadAttentionFwdOp` to the Dense delegate.
- Implement the #1896 target seam without exposing NVIDIA kernel names to external backends.

### FP8

- Incorporate the applicable optimized Dense FP8 work from #1873.
- Retain the SM90 BN224 native-FP8 fast path.
- Extend that kernel family to the declared causal/non-causal, square/rectangular, window, softcap, scale, and RoPE contract rather than adding a dequantize-to-16-bit correctness fallback.
- Keep native FP8 Tensor Core execution as an internal Dense specialization.
- Ensure roofline accounting uses actual input dtype rather than output dtype.

### Sliding and kernel ownership

- Retain `GQASlidingWindowFwdWgmmaPipelinedKernel` as an internal Dense specialization.
- Normalize it to the Dense callable semantics, including custom scale and softcap.
[...]

PR #1975, opened 2026-08-24
[Refactor][Attention] Establish dense GQA Op boundary

superAngGao:
﻿## Summary

- Define the spec-only public boundary for `GroupedQueryAttentionDenseFwdOp` over dense BSHD tensors.
- Standardize rectangular causal/window/RoPE semantics on bottom-right alignment.
- Keep validation and optional-input normalization in the Op, then resolve implementations through the existing `Op.get_or_build_kernel` boundary. A future BUILTIN PR will provide the shape/dtype key and concrete kernel construction.

The ABI covers rectangular prefill and contiguous decode, FP16/BF16, scaled FP8, causal/window/softcap, and NeoX/interleaved RoPE. This PR adds no BUILTIN kernel, benchmark, compatibility adapter, design-document change, `Op` base change, runtime test, or `torch.compile` promise; kernel families and their public execution tests will migrate separately under #1916.

## Validation

- Manifest validator suite: 154 passed
- Ruff, Python compilation, and `git diff --check` passed
- Independent review found no remaining high- or medium-severity issue


src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+            _validate_attention_dtype(dtype)
+        self.dtype = dtype
+        self.target = target
+        self.dispatch_kernel()
+
+    @property
+    def default_kernel_map(self) -> Dict[str, Kernel]:
+        return {}
+
+    def _infer_output_shapes(
```
zhen8838, line 221: Why do these two functions use only one parameter but pass so many unused ones?

zhen8838 (dismissed):

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+        self,
+        q: torch.Tensor,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        q_scale: Optional[torch.Tensor],
+        k_scale: Optional[torch.Tensor],
+        v_scale: Optional[torch.Tensor],
+        rope_cos: Optional[torch.Tensor],
+        rope_sin: Optional[torch.Tensor],
+    ) -> torch.Tensor: ...
```
lcy-seso: A protocol feels too heavy for a single-use interface — worth it only if the interface is meant to be shared across many APIs.

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+        k: torch.Tensor,
+        v: torch.Tensor,
+        q_scale: Optional[torch.Tensor] = None,
+        k_scale: Optional[torch.Tensor] = None,
+        v_scale: Optional[torch.Tensor] = None,
+        rope_cos: Optional[torch.Tensor] = None,
+        rope_sin: Optional[torch.Tensor] = None,
+    ) -> torch.Tensor:
+        """Run Dense BSHD GQA through the selected callable."""
+        batch, seq_len_q, heads, dim = q.shape
```
lcy-seso, line 286: `forward` runs 26 branches; 21 of them are `raise ValueError` on the arguments.

- Two sibling ops in this file already extract that: `GroupedQueryAttentionPrefillVarlenFwdOp._validate_forward_inputs` (L891) and `GroupedQueryAttentionPrefillPagedWithKVCacheFwdOp._validate_forward_inputs` (L1178).
- Here only dtype is extracted (`_validate_dtypes`); shape, device, and combination checks stay inline. Half-extracted reads worse than not extracting.
- Suggested shape: `_validate_forward_inputs` + `_resolve_optional_inputs` returning the 8-tuple, leaving `forward` as unpack shapes -> validate -> resolve -> `self._get_callable(inputs)(*inputs)`.

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+        batch_kv, seq_len_kv, heads_kv, dim_kv = k.shape
```
lcy-seso, line 287: No ndim check on `q` / `k` / `v`.

- A 3-D `q` fails here as `ValueError: not enough values to unpack`, which does not name the argument.
- Both sibling `_validate_forward_inputs` methods check ndim explicitly.

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+        if k.shape != v.shape:
+            raise ValueError("k and v must have the same shape")
+        if batch_kv != batch or dim_kv != dim:
+            raise ValueError("q and k/v must have matching batch and head dimension")
+
+        _validate_positive(batch=batch, seq_len_q=seq_len_q, seq_len_kv=seq_len_kv)
+        _validate_gqa_dims(heads, heads_kv, dim)
+
+        if self.is_causal and seq_len_q > seq_len_kv:
+            raise ValueError("causal dense attention requires seq_len_q <= seq_len_kv")
```
lcy-seso, line 296: Blocking: `seq_len_q < seq_len_kv` is admitted, but nothing defines where query `i` sits on the key axis.

- The existing square kernel masks with `q_idx >= k_idx`, i.e. top-left (`kernels/attention/gqa_fwd.py:102`).
- Decode needs bottom-right (`i + S_kv - S_q >= j`), otherwise it attends to key 0 only.
- The same offset decides which RoPE position `q_i` gets, and there is no position/offset input.
- Neither the docstring nor `shape_rules` states it, so every target will pick its own. Please fix the alignment here.

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+        if self.pos_encoding_mode == "rope" and seq_len_q > seq_len_kv:
+            raise ValueError("fused RoPE requires seq_len_q <= seq_len_kv")
+
+        resolved_sm_scale = _attention_scale(dim, self.sm_scale)
```
lcy-seso: `resolved_sm_scale` is computed and discarded.

- Its only use is the `isfinite` check on the next line, which `__init__:183` already performs.
- `_manifest_params()` hands the builder `self.sm_scale`, possibly `None`, so every target re-derives `1/sqrt(D)`.
- `softcap` is handled the other way: normalized once in `__init__` via `_score_softcap`. Pick one convention for both.

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+        if not math.isfinite(resolved_sm_scale):
+            raise ValueError("sm_scale must be finite")
+        if self.pos_encoding_mode not in ("none", "rope"):
+            raise ValueError("pos_encoding_mode must be 'none' or 'rope'")
```
lcy-seso: Duplicate of `__init__:179`. `pos_encoding_mode` is not mutated between construction and call.

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+        if self.rope_layout not in ("neox", "interleaved"):
+            raise ValueError("rope_layout must be 'neox' or 'interleaved'")
```
lcy-seso: `rope_layout` is validated here, `pos_encoding_mode` in `__init__`. Both are constructor parameters with a closed value set; validate both in `__init__`.

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+        if self.pos_encoding_mode == "rope":
+            resolved_rotary_dim = _rope_rotary_dim(dim, self.rotary_dim)
+        else:
+            if self.rotary_dim is not None:
+                raise ValueError("rotary_dim requires pos_encoding_mode='rope'")
```
lcy-seso: Duplicate of `__init__:181`.

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+        normalized_scales = []
+        for name, scale in zip(("q_scale", "k_scale", "v_scale"), scales, strict=True):
+            if scale is None:
+                continue
+            if scale.device != q.device:
+                raise ValueError(f"{name} must be on the same device as q")
+            if tuple(scale.shape) != (batch, heads_kv):
+                raise ValueError(f"{name} must have shape {(batch, heads_kv)}")
+            normalized_scales.append(scale.contiguous())
+        resolved_scales = tuple(normalized_scales) if normalized_scales else (None, None, None)
```
lcy-seso: `normalized_scales` is built by append, so its length depends on the all-or-none check at L314 holding.

- If that check moves, this splats a 1- or 2-tuple into `inputs` at L359 and the failure surfaces at the callable, not at validation.
- `tuple(s.contiguous() if s is not None else None for s in scales)` removes the coupling.

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+            strict=True,
+        ):
+            if scale is not None and scale.dtype != torch.float32:
+                raise ValueError(f"{name} must have float32 dtype")
+        for name, table in (("rope_cos", rope_cos), ("rope_sin", rope_sin)):
+            if table is not None and table.dtype != output_dtype:
+                raise ValueError(f"{name} must have dtype {output_dtype}")
+
+    def eval_roofline(self) -> tuple[int, int]:
+        raise NotImplementedError("Dense GQA has no in-tree implementation yet")
```
lcy-seso, line 269: This override can go: `Op.eval_roofline` (`op_base.py:191`) already raises `NotImplementedError`, with a message pointing at `roofline.md` 4.4.6.

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
 
 
+class _DenseFwdCallable(Protocol):
+    """Target-owned second dispatch layer for the Dense GQA boundary.
+
+    A target builder returns this callable for one input signature. Any finer
+    selection and caching of concrete kernels belongs here, not in the Op.
+    """
+
+    def __call__(
```
lcy-seso: `_DenseFwdCallable` has one use site, and it is the target of a `cast`, which does not check at runtime.

As written it buys what a `Callable[...]` alias would. Worth a Protocol only if a target-side implementation gets checked against it.

src/tileops/ops/attention/gqa.py
```diff
@@ -128,6 +131,236 @@ def _build_packed_prefill_kernel(
...
+
+
+class GroupedQueryAttentionDenseFwdOp(Op):
+    """Shape-agnostic dense BSHD GQA for prefill and contiguous decode.
+
+    The public ABI covers rectangular Q/KV lengths, causal/window/softcap modes,
+    FP16/BF16 and scaled FP8 inputs, and optional caller-owned RoPE tables.
+    This target-neutral layer caches a target-owned :class:`_DenseFwdCallable`
+    under the standard input-signature rules; concrete kernel dispatch remains
+    inside that callable. The in-tree callable is deferred to a follow-up.
```
lcy-seso: This class defines the public contract for dense GQA but never states what it computes. Please write the linear algebra into the class docstring.

The docs site renders LaTeX now; `ops/norm/layer_norm.py:22` is the in-repo precedent for the format (`$$ ... $$`, escaped backslashes). The equation should pin down what prose leaves open:

- which KV group head `h` reads
- where query `i` sits on the key axis, i.e. the offset the causal and window masks compare against (see the alignment comment on L291)
- the score scale and its default
- how `softcap` enters
- where the FP8 `q_scale` / `k_scale` / `v_scale` factors apply
- what accumulates in FP32, and where the cast to `dtype` happens

src/tileops/manifest/attention.yaml
```diff
@@ -98,6 +98,71 @@ MultiHeadAttentionBwdOp:
...
+      dtype: {type: "torch.dtype | None", default: null}
+      pos_encoding_mode: {type: str, default: none}
+      rotary_dim: {type: "int | None", default: null}
+      rope_layout: {type: str, default: neox}
+    dtype_combos:
+      - {q: float16, k: float16, v: float16, o: float16}
+      - {q: bfloat16, k: bfloat16, v: bfloat16, o: bfloat16}
+      - {q: float8_e4m3fn, k: float8_e4m3fn, v: float8_e4m3fn, o: float16}
+      - {q: float8_e4m3fn, k: float8_e4m3fn, v: float8_e4m3fn, o: bfloat16}
+    shape_rules:
```
lcy-seso, line 133: `forward` enforces two constraints missing from `shape_rules`:

- `is_causal` requires `S_q <= S_kv` (gqa.py:291)
- `pos_encoding_mode == 'rope'` requires `S_q <= S_kv` (gqa.py:293)

The manifest is the spec, so both belong here, plus the rule stating query-to-key alignment once it is decided.

lcy-seso (commented):
Reviewed the Dense GQA boundary. One blocking item, the rest are placement and duplication.

**Blocking**

- Query-to-key alignment is undefined for `S_q < S_kv` (inline on `gqa.py:291`). It decides both the mask and the RoPE positions, and every target will guess differently until it is written down.

**Also**

- `forward` carries 21 argument checks inline while two sibling ops in the same file use `_validate_forward_inputs`.
- Three checks duplicate `__init__`; `resolved_sm_scale` is computed and never used.
- No test. `grep` finds the class only in `ops/__init__.py`, `attention/__init__.py`, and the manifest; `source.test` names `tests/ops/attention/test_gqa.py`, which has no case for it. `status: spec-only` covers the missing kernel, but the 95 lines of validation ship constructible and callable, unexercised.
- The class docstring never states what the op computes.

superAngGao:
Updated at `192ec155`:

- defined bottom-right alignment (`p_i = i + S_kv - S_q`) in the Op contract and manifest;
- moved runtime checks into `_validate_forward_inputs` and constructor-only invariants into `__init__`;
- added explicit BSHD rank errors;
- replaced append-based optional handling with `_canonicalize_inputs`, which returns contiguous tensors in all eight manifest slots while preserving `None`;
- removed the single-use `Protocol`, cast, and second-cache framing;
- documented the Dense GQA linear algebra, FP8 scales, softcap, accumulation, and output-cast semantics;
- added the missing causal/RoPE manifest constraints.

I retained the minimal `eval_roofline` override because the base method is abstract and codegen skips `status: spec-only`; without it the public Op cannot be instantiated.

The PR has also been narrowed: it no longer changes `Op` base, design docs, backend-seam tests, or existing GQA tests, and it adds no BUILTIN kernel. Runtime correctness tests will accompany the first concrete kernel migration. Manifest validation, Ruff, Python compilation, and `git diff --check` pass.

src/tileops/ops/attention/gqa.py
superAngGao: The full signatures are intentional because they mirror the manifest input slots. `_infer_output_shapes` is required by the `Op` abstract contract and manifest parity checks; Dense GQA output shape depends only on `q_shape`. `_validate_dtypes` uses the tensor inputs to enforce Q/K/V, scale, and RoPE dtype relations.

src/tileops/ops/attention/gqa.py
superAngGao: Agreed. I removed `_DenseFwdCallable`, the `Protocol`, and the cast. The Op now uses the existing `get_or_build_kernel` boundary directly.

src/tileops/ops/attention/gqa.py
superAngGao: Done. Runtime checks now live in `_validate_forward_inputs`; contiguous conversion and fixed manifest-slot construction live in `_canonicalize_inputs`. `forward` is now validate -> canonicalize -> get kernel -> invoke.

src/tileops/ops/attention/gqa.py
superAngGao: Done. `q`, `k`, and `v` now receive explicit rank-4 BSHD checks before shape unpacking, with the argument name in each error.

src/tileops/ops/attention/gqa.py
superAngGao: Fixed. The class contract now defines bottom-right alignment as `p_i = i + S_kv - S_q`; causal/window masking and fused-RoPE query positions are defined from `p_i`. The manifest records the same constraints.

src/tileops/ops/attention/gqa.py
superAngGao: The discarded local has been removed. `sm_scale=None` remains constructor state because its default depends on the current head dimension; `softcap` is shape-independent and is normalized at construction. The future BUILTIN build closure (or a target builder using `TensorSpec`) resolves the dimension-dependent scale.

src/tileops/ops/attention/gqa.py
superAngGao: Done. `pos_encoding_mode` is validated only at construction now.

src/tileops/ops/attention/gqa.py
superAngGao: Done. Both `pos_encoding_mode` and `rope_layout` closed sets are validated in `__init__`.

src/tileops/ops/attention/gqa.py
superAngGao: Done. The duplicate runtime check was removed; the constructor owns this invariant.

src/tileops/ops/attention/gqa.py
superAngGao: Done. `_canonicalize_inputs` now constructs all eight manifest slots with one tuple comprehension, preserving `None` independently of the all-or-none validation.

src/tileops/ops/attention/gqa.py
superAngGao: I retained this minimal override after checking the class contract: `Op.eval_roofline` is abstract, and roofline codegen intentionally skips `status: spec-only`. Removing the override makes the public class non-instantiable. The override can disappear once an implemented entry supplies generated roofline code.

src/tileops/ops/attention/gqa.py
superAngGao: Agreed and removed. There is no dedicated callable type or second callable-owned cache now.

src/tileops/ops/attention/gqa.py
superAngGao: Done. The class docstring now specifies KV-group head mapping, bottom-right coordinates, mask/window semantics, scale default, softcap placement, FP8 scale placement, FP32 accumulation, and output casting.

src/tileops/manifest/attention.yaml
superAngGao: Done. The manifest now includes the causal and fused-RoPE `S_q <= S_kv` rules and documents bottom-right query-to-key alignment.

lcy-seso (approved):

PR #1976, opened 2026-08-24
[Refactor][Backend] Allow resolved external builder params

superAngGao:
## Summary

- Let an Op provide input-signature-resolved manifest parameters when constructing an external target callable.
- Preserve the existing `_manifest_params()` path when no override is supplied.
- Leave BUILTIN keying and construction unchanged.

The override is intentionally excluded from the external cache key: resolved values must be deterministic for an Op instance and tensor device/dtype/shape signature. This supports defaults such as `D**-0.5`, output dtype derived from an input spec, and shape-resolved dimensions without requiring every backend to repeat public Op semantics.

## Validation

- Added one focused backend-seam test in the existing test file.
- Ruff, Python syntax compilation, and `git diff --check` passed.
- Independent review found no correctness issue; pytest is unavailable in the local Windows environment, so CI remains the executable test gate.


superAngGao:
Closed because according to our design, external backends should take charge of dispatching kernels on the specific devices, while tileops must not interfering their behaviours through passing "params".

PR #1982, opened 2026-08-25
[Perf][FFT] Reuse twiddle LUT in complex128 SMEM stages

Ruihan11:
## Summary

This is a follow-up to [#1884](https://github.com/tile-ai/TileOPs/pull/1884), which fused complex output packing into the final FFT butterfly. This PR adds the remaining shared-memory twiddle optimization that was not part of the merged implementation.

- `complex128` shared-memory butterfly stages now read twiddles from the pre-computed GPU LUT instead of evaluating FP64 `sin`/`cos` at runtime.
- `complex64` intentionally retains inline FP32 `sin`/`cos`: measurements show that the extra LUT loads do not improve FP32 latency.
- The public `FFTC2CFwdOp` API, split-input kernel contract, output layout, and existing scratch buffers are unchanged.

## Performance

Measured on an NVIDIA H200 with the official CUDA 13.2 runner:

| Item | Value |
| --- | --- |
| PyTorch | 2.13.0+cu132 |
| CUDA | 13.2 |
| TileLang | 0.1.11+cu132.gitafcebed1 |
| Timer | Native CUPTI |
| CUPTI fallback | Disabled |
| Comparison | Four alternating candidate/baseline rounds |
| Metric | `device_busy_ms` |

The baseline is current `main` (which already contains the output-packing fusion from #1884). Speedup is `baseline / candidate`.

| Workload | Baseline mean (ms) | Candidate mean (ms) | Speedup |
| --- | ---: | ---: | ---: |
| c64, unbatched | 0.0074 | 0.0074 | **1.003x** |
| c64, batch 64 | 0.0149 | 0.0144 | **1.029x** |
| c128, batch 64 | 0.0267 | 0.0186 | **1.438x** |

The autotuner selects different configurations between some very short runs. A fixed `{"block_size": 512, "threads": 512}` native-CUPTI comparison isolates the kernel change:

| Workload | Baseline (ms) | Candidate (ms) | Speedup |
| --- | ---: | ---: | ---: |
| c64, unbatched | 0.007904 | 0.007841 | **1.008x** |
| c64, batch 64 | 0.012672 | 0.012640 | **1.002x** |
| c128, batch 64 | 0.026816 | 0.017984 | **1.491x** |

These measurements show that the optimization is neutral for FP32 and provides a substantial FP64 gain. The dtype-specific dispatch avoids turning the LUT trade-off into a c64 regression.

## Correctness and checks

- 14 FFT correctness tests pass, including a new `N=4096` Op-level case that reaches the final LUT stage.
- The kernel-level layout test continues to cover the radix-specific final-stage paths for both dtypes.
- `ruff check` passes.
- `ruff format --check` passes.
- `compileall` passes.
- `git diff --check` passes.
- Test-node delta:

  ```text
  tests/kernels/test_fft_output_layout.py   4 -> 4   +0
  tests/ops/test_fft.py                     9 -> 10  +1
  TOTAL                                    13 -> 14  +1
  ```

## Scope and limitations

- No public API or benchmark infrastructure changes are included.
- Input split-copy removal and scratch-buffer elimination are intentionally left for separate follow-up work.
- PyTorch cuFFT remains faster; this PR optimizes the TileOPs implementation relative to the current TileOPs baseline.

### DIFF
--- a/src/tileops/kernels/fft.py
+++ b/src/tileops/kernels/fft.py
@@ -29,8 +29,10 @@ def _fft_c2c_kernel(n: int, batch_size: int = 1, dtype: str = "complex64") -> Ca
       Each block reads input elements at their bit-reversed positions directly
       into shared memory, eliminating the separate bit-reversal kernel.  It
       then performs all log2(min(n, 2*threads)) butterfly stages entirely in
-      SMEM using on-the-fly trig.  Result: one global-memory read + one write
-      for the entire SMEM phase, regardless of how many stages it covers.
+      SMEM.  Complex128 uses the pre-computed twiddle LUT; complex64 retains
+      inline trig because its FP32 trig is cheaper than the extra LUT loads.
+      Result: one global-memory read + one write for the entire SMEM phase,
+      regardless of how many stages it covers.
 
     Phase 2 — LUT stages (remaining stages, one kernel each):
       For stages where butterfly strides exceed the SMEM chunk size, each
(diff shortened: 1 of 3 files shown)


lcy-seso (approved):
LGTM, Thank you 😊

PR #2000, opened 2026-08-28
[Perf][Elementwise] Reuse the sinusoidal divisor down the rows, and clamp Mish

lcy-seso:
Stacked on #1997, whose commits this PR's diff also contains. The part to review here is the last two commits:

```
git diff perf/elementwise-transcendental...perf/elementwise-roofline-2
```

which touches `kernels/elementwise/sinusoidal.py`, `MishFwdKernel` in `kernels/elementwise/activations.py`, `manifest/elementwise_generative.yaml` and `tests/ops/test_special_elementwise.py`. Merge #1997 first and this rebases to just those.

## problems

The five ops #1997 fixed were the ones losing to their PyTorch baseline. Ranking the same nightly (run 33134732113, H200) by achieved bandwidth against the copy roof instead turns up ops that beat torch comfortably and are still nowhere near the hardware:

| op | achieved / roof | vs torch |
| --- | --- | --- |
| SinusoidalFwdOp | 9.6% | 2.43x |
| LogSumExpFwdOp | 58.1% | 6.36x |
| PowFwdOp | 61.0% (worst row 20.5%) | 1.20x |
| MishFwdOp | 61.1% | 1.57x |
| LogSoftmaxFwdOp | 68.4% | 8.17x |

Sinusoidal writes 33.6 MB in 81.8us. Stripping it one operation at a time against a write-only kernel at the same launch config put 8.2us on the store, 20.4us on `pow` plus the division, and 53.2us on trigonometry — the last because `T.if_then_else(is_even, sin, cos)` evaluates both and discards one.

The `pow` is worse than it looks: the divisor depends on the dimension pair, not the position, so 2048 distinct values were being recomputed 8.4M times.

Mish spent 9.5% of its kernel on a `T.if_then_else` guarding the saturation point. `T.if_then_else` in an element body makes TileLang emit a scalar loop instead of float4 lanes.

## changes

- `sinusoidal.py`: one block now covers 64 positions by 32 dimension pairs. `pow` runs once per pair into shared memory and is reused down the rows; the angle is staged in a fragment so each pair computes it once; the tile leaves through a single `T.copy`. The arithmetic per element is unchanged, so the output is bit-identical — including on a ragged shape, where it reproduces the current kernel's own 1-ulp disagreement with `torch.pow` exactly (52378 differing elements at 500x250, before and after).
- `sinusoidal.py`: an odd `d_model` now raises. The manifest already declared `d_model % 2 == 0`; the kernel silently dropped the unpaired dimension.
- `activations.py`: Mish clamps the exponential's argument at the saturation point instead of selecting on it. Past that point the rational is 1 to every bit fp32 carries, so `x * 1.0` is what the discarded branch returned anyway. Bit-identical, and NaN still propagates through the outer multiply.
- Two tests, node delta +2 (73 -> 75 on `test_special_elementwise.py`): a sinusoidal shape where neither extent divides the block tile — the aligned cases never reach the partial tile — and the odd-`d_model` rejection.
- The manifest's `flops` comment read as an account of what the kernel issues per element, which stopped being true. `6 * S * D` is unchanged: roofline.md §1.3 fixes it as a per-element convention count and §1.2 says the implementation's instruction mix does not feed back into it. Only the wording moved.

## measured

H200, CI image, CUPTI device-busy median, `benchmarks/` timing. "roof" is a TileLang kernel at the same shape and launch config doing only the store.

| op / workload | before | after | roof | baseline |
| --- | --- | --- | --- | --- |
| Sinusoidal 4096x4096 float16 | 81.8us | 25.2us | 8.2us | 111.6us |
| Sinusoidal 2048x4096 float16 | 42.4us | 13.6us | 4.1us | 56.7us |
| Mish 16x256x80x80 float16 | 40.2us | 36.2us | 26.3us | 63.6us |
| Mish 16x512x40x40 float16 | 21.5us | 19.2us | 14.3us | 33.2us |

Sinusoidal goes from 9.6% of the write roof to 31%, ratio 2.43 -> 4.43. Mish from 61% to 68%, ratio 1.58 -> 1.76.

## the other three, and why they are not here

- **PowFwdOp** is `powf`-bound and cannot be fixed without changing what the op computes. It costs 4.4 ns/element on both the broadcast and the same-shape rows, so the two rows differ only in byte accounting, not in per-element cost — there is no broadcast-path defect behind the 20.5% row. `exp2(b * log2(a))` would be an order cheaper and would return NaN for `pow(-2.0, 3.0)`, which `torch.pow` answers -8.0; the pow tests draw positive bases only, so nothing would have caught it. This is the case roofline.md §1.2 already carves out: the FLOP convention counts a transcendental as 1, and the metric does not certify an SFU-bound kernel as at its limit.
- **LogSumExpFwdOp** and **LogSoftmaxFwdOp** are reductions, and their worst rows are parallelism-bound rather than instruction-bound: `lm-head-logits` is 4 rows of 102400, so the whole GPU gets 4 units of work and reads 1.6 MB in 10.5us. That is a split-reduction question, in a family that has had six perf PRs land in the last week. Left alone deliberately rather than half-done.

### DIFF
--- a/src/tileops/kernels/elementwise/_erf.py
+++ b/src/tileops/kernels/elementwise/_erf.py
@@ -4,31 +4,13 @@ import tilelang.language as T
 
 __all__ = ["erf"]
 
-#: |x| at which erf is taken as saturated. erf(3.6) = 1 - 7.4e-7, three hundred
-#: times finer than half a float16 ulp at 1.0, so the clamp costs neither dtype
-#: anything it could store.
+#: |x| taken as saturated. erf(3.6) = 1 - 7.4e-7, far below half a float16 ulp.
 _CLAMP = 3.6
 
-#: erf(x) = clip(t * P(w), -1, 1) for t = clamp(x, -_CLAMP, _CLAMP) and
-#: w = 1 - (t / _CLAMP)**2, coefficients highest degree first. Three properties of
-#: this form, in this order:
-#:
-#: - erf saturates to exactly +-1 at the clamp, which GELU depends on: it scales
-#:   the 1 - erf(x) residual by x, so a tail off by eps carries an error of
-#:   |x| * eps / 2, unbounded in |x|. Two things make it exact. _CLAMP is scaled
-#:   out of x before squaring rather than after, so w is exactly zero at the clamp
-#:   and P is exactly its constant term float32(1 / _CLAMP), whose product with
-#:   _CLAMP is 1.0; and the clip stops the backend contracting that product into
-#:   the caller's add, which would evaluate it to full width and land 7e-9 short.
-#: - w rather than x**2 as the polynomial variable keeps the Horner chain
-#:   conditioned in float32. The same fit in unnormalised x**2 loses four digits
-#:   to cancellation at the clamp.
-#: - t * P(w) rather than a polynomial in x alone keeps the *relative* error small
-#:   near zero, where erf(x) -> x.
-#:
-#: Worst case over the real line is 1.7e-5, an order below half a float16 ulp at
-#: 1.0 (2.4e-4). tests/ops/test_unary_math.py measures the resulting ulp distance
-#: on device, over every float16 and every bfloat16 value.
+#: erf(x) = clip(t * P(w), -1, 1), t = clamp(x, -_CLAMP, _CLAMP), w = 1 - (t / _CLAMP)**2,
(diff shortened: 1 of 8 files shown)


lcy-seso:
Folding into #1997 — same work, one PR.

PR #2012, opened 2026-08-30
[Test][GQA] Rotate the paged-prefill reference in torch, not through the RoPE op

lcy-seso:
## Problems

- `_apply_neox_rope_position_ids` in `tests/ops/attention/test_gqa_prefill_paged.py` built a `RopeNeoxPositionIdsFwdOp` and called it. The fused-rope test therefore checked one implementation against another, not against a reference. Any change to that op's arithmetic moves the reference and fails this test with the code under test unchanged — on a branch that makes the op accumulate in f32 and round once, the four `test_gqa_prefill_paged_with_kv_cache_fused_rope` cases fail with a max absolute difference of 4.88e-04.
- The appended-k assertion carried no tolerance, so it compared two float16 rotations at `assert_close`'s default `atol=1e-5` — below one float16 ulp at these magnitudes.
- Five module-level helpers in the test built the paged inputs and addressed the page pool. trust-model.md §Test puts input construction in `workloads/`, and `tests/test_workload_placement.py` gates it by the method name `gen_inputs`, which a free function never carries, so these were outside that rule's view.
- The duplicate that rule exists to prevent was already there: `GQAPrefillPagedWithKVCacheFwdWorkload.gen_inputs` built its own `cu_seqlens` inline, a second construction of the same vector for the same op.

## Changes

- `_apply_neox_rope_position_ids` is plain torch: the frequency table is built in the test, read in `x`'s dtype as the kernel reads it, and the rotation runs in f32 with a single rounding at the boundary.
- The varlen offset vector, the interleaved block table, the logical-to-physical row map, the unit cache scales and the page-pool scatter move to `workloads/attention/gqa.py`. The payload tensors stay in the test, which is what decides whether a page holds noise or a known logical row. `paged_cache_row` replaces `_physical_pos`: it returns a row of the page pool, not a general position.
- `GQAPrefillPagedWithKVCacheFwdWorkload.gen_inputs` reads `make_cu_seqlens` instead of its inline copy. Its block table stays an identity mapping and says why: a timed run walks a cache filled in order, and whether the walk honours a permuted table is the test's question.
- The fused-rope assertions take `_PREFILL_PAGED_TOLERANCE[dtype]` rather than repeating its float16 pair as literals, and the appended-k assertion states why it needs a tolerance: it compares two rotations of the same input, not a copy.

### DIFF
--- a/tests/ops/attention/test_gqa_prefill_paged.py
+++ b/tests/ops/attention/test_gqa_prefill_paged.py
@@ -1,60 +1,23 @@
 """Tests for packed GQA prefill with paged KV cache append."""
 
-from itertools import accumulate
-
 import pytest
 import torch
 
 from tileops.manifest import load_workloads
 from tileops.ops import GroupedQueryAttentionPrefillPagedWithKVCacheFwdOp
 from tileops.perf.formulas import gqa_prefill_paged_with_kv_cache_fwd_roofline
+from workloads.attention.gqa import (
+    fill_paged_cache_from_logical,
+    make_cu_seqlens,
+    make_interleaved_block_table,
+    make_unit_cache_scales,
+    paged_cache_row,
+)
 
 _PREFILL_PAGED_TOLERANCE = {
     torch.float16: (5e-3, 1e-5),
     torch.bfloat16: (8e-2, 1e-2),
 }
-
-
-def _make_cu_seqlens(lengths: list[int]) -> torch.Tensor:
-    return torch.tensor([0, *accumulate(lengths)], device="cuda", dtype=torch.int32)
-
-
-def _physical_pos(
-    block_table: torch.Tensor, batch_idx: int, logical_pos: int, page_size: int
-) -> int:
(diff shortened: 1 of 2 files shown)


lcy-seso:
This is the unblocker for the precision finding in #2010's Result And Limitations. Verified the combination on one idle H200 in `tileops-runner:cu132-torch2.13-tl-afcebed1-dev`:

| Tree | Selection | Result |
| --- | --- | --- |
| #2010 at bc29db95 + this patch | `tests/ops/attention/test_gqa_prefill_paged.py` | 30 passed |
| the same, with #2010's kernel accumulating in f32 again | that file + `tests/ops/test_rope.py` | 85 passed |

The two PRs touch no file in common and merge in either order — #2010 currently keeps the storage-dtype arithmetic, so it passes with or without this one. Once this merges, restoring the f32 accumulation in the RoPE kernel is a standalone change, and by the second row above it is already known to pass.

PR #2040, opened 2026-09-02
[Feat][GEMM] Add the SM90 FP8 1D2D block-scaled GEMM kernel

michaelwithu:
# [Feat][GEMM] Add the SM90 FP8 1D2D block-scaled GEMM kernel

## How this started: two findings

**1. The FlashInfer baseline in `bench_gemm.py` computes something else.** One run, one reference:

```
reference (workload.ref_program) == tileops (GemmFp8BlockScaledKernel)   PASS
reference (workload.ref_program) != flashinfer-fp8-blockscale-sm90       FAIL
                                    97.8% of elements differ, max abs 15-19 (0.016 allowed)
```

The kernel is correct. The defect is in the call site, `_flashinfer_fp8_blockscale_ref`, which
gets both scales wrong:

- `scale_a` goes in K-contiguous, but that entry point assumes M-contiguous — it never reads
  the tensor's strides, the layout is assumed.
- `scale_b` goes in as `(N, K/128)` per row, but on SM90 that entry point only has 128x128
  granularity. Probing row by row confirms what it does instead: it reads the leading
  `(N/128, K/128)` region of that buffer, each value governing 128 output columns, and ignores
  the rest. (The docstring advertises a per-token `(N, K//128)` form; the SM90 path does not
  honour it.)

FlashInfer's Python layer does not validate `weight_scale`'s shape at all, `input_scale`'s
checks happen to pass, and the row carries no `assert_matches_reference` — so nothing warns.
Timing is unaffected (1.00-1.02x), which is why the recorded milliseconds still measure
FlashInfer's speed correctly; what is invalid is treating them as the same computation.

**2. `GemmFp8BlockScaledKernel` is slow** (4096x4096x7168, one instrument): 310 TF/s as it
stands; 322 with the scales made M/N-contiguous, so the strided gather its K-contiguous layout
implies costs only 8% (20% on K-dominant shapes); 585 is the ceiling of that structure with
the scale work removed entirely, putting the per-K-step promotion pass at roughly 40%; a
Hopper-pipelined kernel reaches 987 (TMA / warp specialization / persistent, 1.7x). Separately,
`default_config` leaves `block_n=128` for every shape it does not special-case, so
`(128, 2112, 7168)` launches 17 CTAs on 132 SMs.

## Why a new kernel rather than a change to an existing one

FP8 block scaling comes in three granularities; TileOPs covers per-tensor
(`GemmFp8EpilogueKernel`) and 1D1D (`GemmFp8BlockScaledKernel`). The missing one — 1D2D, A per
1x128 and B per 128x128 — is what a served model actually produces: DeepSeek-V3 quantizes
weights per 128x128 block; FlashInfer implements only this granularity on SM90 (which is where
the `scale_b` half of finding 1 comes from); and DeepGEMM, an independent implementation,
states the same requirement outright in `csrc/apis/layout.hpp:41-46`. Once the granularity is
fixed, `scale_a` being M-contiguous follows — it is the assumption hard-coded in
`sm90_fp8_gemm_1d2d.hpp:123`, not a choice.

The three granularities nest, so the existing 1D1D kernel **can** compute a 1D2D call:
expanding `scale_b` 128x along N is enough, and we verified the results agree. The reverse does
not hold — what 1D1D needs is not present in 1D2D scales. But the price is transposing
`scale_a` on every call, which is activation-side work (+41-51% on decode shapes), and 1.28-4.10x
end to end. How much of that gap is inherent to 1D1D being the more general form and how much
is the old kernel's implementation debt (finding 2) is not separated here. The 1D1D kernel
stays: DeepGEMM ships a 1d1d impl too, so that contract has its own callers. It does, however,
have no upstream implementation to compare against on SM90, so it cannot be validated against
the ecosystem the way 1D2D can.

## Contract

| tensor | logical shape | dtype | contiguous dim |
| --- | --- | --- | --- |
| `a` / `b` | `[M, K]` / `[N, K]` | `float8_e4m3fn` | K |
| `scale_a` | `[ceil(K/128), M]` | `float32` | **M** |
| `scale_b` | `[ceil(N/128), ceil(K/128)]` | `float32` | K |
| out | `[M, N]` | `bfloat16` | N |

Refused with a stated reason: `M >= 128`; `M % 4 == 0` (`scale_a`'s 16-byte TMA row — DeepGEMM
pads internally, this kernel asks the caller for a multiple of 4); `K % 16 == 0` (the operands'
16-byte TMA row); `N % 2 == 0`, or `N % 8 == 0` under the shared-memory epilogue, which also
needs `block_n % 16 == 0`. TileLang's packed ABI validates both shape and strides, so a layout
mistake of the kind behind finding 1 is a hard error here, never a silent wrong result.

## Performance

Seven of the eight shapes the manifest already declares as `block128` workloads. The `m=1` one is
out of this PR's scope: below 128 rows the 128-row tile pads M, and the padded kernel measured a flat
0.0102 ms for every M from 4 to 64 (1.38x FlashInfer). A swapped-operand sibling was prototyped and
reaches 1.03x at M <= 32 on 7168x2048 but 1.4-2.3x on long K, so it is held for its own PR; this
kernel states `M >= 128` in its refusal. Same granularity, same layouts, no adapter on either
side, and both are asserted against a torch reference before being timed. Repo instrument
(CUPTI kernel time, L2 flushed per iteration), idle H200, best of a forward and a reverse pass.

| m | n | k | tileops | flashinfer | ratio |
| --- | --- | --- | --- | --- | --- |
| 128 | 2112 | 7168 | 0.01168 ms | 0.01158 ms | 1.009x |
| 128 | 7168 | 2048 | 0.00864 ms | 0.00835 ms | 1.034x |
| 4096 | 2112 | 7168 | 0.12792 ms | 0.12123 ms | 1.055x |
| 4096 | 7168 | 2048 | 0.11904 ms | 0.11885 ms | 1.002x |
| 4096 | 4096 | 7168 | 0.20757 ms | 0.19262 ms | 1.078x |
| 4096 | 7168 | 16384 | 0.82115 ms | 0.82813 ms | 0.992x |
| 4096 | 24576 | 1536 | 0.29541 ms | 0.33090 ms | 0.893x |

Between 11% ahead and 8% behind FlashInfer. Ahead on the many-wave shapes, at parity on
128x2112x7168 and 4096x7168x2048, behind on 128x7168x2048, 4096x4096x7168 and 4096x2112x7168. No ratio
against `GemmFp8BlockScaledKernel` is reported: the two take different granularities in different
layouts, so such a number would be the same mistake as finding 1.

Four changes got the kernel here, the first three taken from the upstream `sm90_fp8_gemm_1d2d`:
scales read into registers before the WGMMA and the stage released before the promotion; one
`empty` arrival per consumer warp instead of per thread; the TMA store committed without waiting,
the wait deferred to the next tile's epilogue so the store overlaps its mainloop; and, on
single-wave tiles, the producer's K loop run by one warp instead of the whole warp-group. Measured
against the commit before them (`a17fce0b`), interleaved before/after/after/before on an idle H200
with the repo instrument -- the two "before" and the two "after" readings agree to within 0.5% on
every row:

| m | n | k | before | after | after/before |
| --- | --- | --- | --- | --- | --- |
| 128 | 2112 | 7168 | 0.01258 ms | 0.01171 ms | 0.931x |
| 128 | 7168 | 2048 | 0.00880 ms | 0.00870 ms | 0.989x |
| 4096 | 2112 | 7168 | 0.13210 ms | 0.12810 ms | 0.970x |
| 4096 | 7168 | 2048 | 0.12282 ms | 0.11896 ms | 0.969x |
| 4096 | 4096 | 7168 | 0.20552 ms | 0.20677 ms | 1.006x |
| 4096 | 7168 | 16384 | 0.87008 ms | 0.82842 ms | 0.952x |
| 4096 | 24576 | 1536 | 0.31578 ms | 0.29414 ms | 0.931x |

Six of seven shapes 1-7% faster, the seventh flat within noise. The store overlap pays on the
many-wave shapes; the single-warp producer pays on the decode shape.

Tried and rejected, with numbers: `num_stages` 5 or 6 (slower on every shape); FlashInfer's tile
picks `block_n=56` for 128x7168x2048 and `block_n=112` for 4096x2112x7168 (1.13x and 1.12x against
our 1.04x and 1.08x); and 2-CTA TMA multicast of the B tile, which FlashInfer uses for m >= 1024
(5-9% slower on all three prefill shapes tested -- the traffic it saves is not what bounds these
shapes, and the cluster lockstep costs more than it returns). Nsight Compute on 128x2112x7168, where both kernels run the identical schedule, showed no register
spills (SASS carries no LDL/STL) and put the gap in the producer warp-group: 0.6 more active warps
per cycle, `branch_resolving` stalls 9x and `long_scoreboard` stalls 3x FlashInfer's. TileLang
elects one lane per TMA, so the other 127 producer threads only polled `empty`, 56 times per tile.
Running that loop on one warp is the fourth change above; it closed the decode gap from 1.067x to
1.009x and, gated on `max_waves == 1`, leaves the multi-wave shapes' code byte-identical (narrowing
them too measured 2.3% slower). A lone-thread producer deadlocks -- TileLang's election is a warp
collective -- which is why it is a warp. Profiled again after the change, the three metrics that named the
cause sit on FlashInfer's values: active warps per cycle 2.89 -> 2.28 (theirs 2.27), `branch_resolving`
0.283 -> 0.034 (0.032), `long_scoreboard` 2.99 -> 1.96 (1.09). What remains of the gap is the last
warp's 32 lanes polling `empty` where FlashInfer polls from one, and four `expect_tx` + TMA pairs per
stage where it issues one `arrive_and_expect_tx`. The decode shapes sit at roughly 9.7 TB/s of L2-to-SMEM
traffic broadcasting the single A tile to all 132 CTAs, the regime both kernels share.

## Correctness

The reference is independent of the implementation: each 128-wide K block scaled by definition
in fp32, taking neither this kernel nor any library as ground truth. Seven smoke nodes cover
the single-wave global-store epilogue; the per-K B-scale ring under `max_waves > 1`; the
STSM + TMA-store epilogue, which had no coverage at all; `block_n=56` being refused under that
epilogue; both branches of `default_config`, measured-table hit and heuristic band; and the
four unaddressable dimensions.

`test_gemm_fp8_1d2d_bench` asserts both tileops and FlashInfer against the reference before
timing — the step finding 1 is missing. Beyond the tests, the seven manifest `block128` shapes
and the edges around them (`m`/`n` not multiples of 128, a partial trailing K block,
`scale_k=1`, several waves, `block_n` at 16/32/56/64/128) were each checked against that
reference.

One defect was found and fixed while measuring: `block_n=56` under the shared-memory epilogue
produced all-NaN output. `fp8_gemm_raw_acc_stsm_bf16` covers 16 columns per STSM atom, so 56
leaves columns 48-55 unwritten, and the combination was reachable through `config=` on any shape
in the measured table.

`block_n=56` is now gone from the allowlist entirely, because it turned out to be the only value
that does not divide 128 and so the sole reason a second code path existed. Poisoning the second
B scale and re-running with caching disabled confirmed it: every `block_n` that divides 128
passes with that scale set to NaN, single-wave and staged alike, and only 56 fails. Removing it
drops `uniform_scale_b` and its false branches from all three warp-group bodies, the second
staged scale row, and `fp8_gemm_1d2d_promote_shared_ab` in full -- the staged call site passed
what the `_uniform` variant already computes. A test pins the refusal so re-adding 56 fails
loudly rather than publishing an all-NaN tile.

`tests/ops/test_gemm.py` node count 60 -> 67 (+7, +11.7%).

## Filed separately, not fixed here

1. Drop the `flashinfer-fp8-blockscale-sm90` tag from the `block128` rows with a stated reason
   (`.claude/domain-rules/benchmark.md`: "where it cannot express the case, drop the tag and say
   why"), move it onto the 1D2D workloads this PR introduces, and give the baselines that stay
   an `assert_matches_reference`.
2. `GemmFp8BlockScaledKernel.default_config`'s `block_n` choice: 27 -> 47 TF/s on decode.
3. Its per-K-step promotion pass: roughly 40% of the kernel.
4. `GemmFp8EpilogueKernel` is slower than the real `torch._scaled_mm` on 4 of its 6 declared
   workloads (worst 3.75x, peaking at 23% of H200 FP8 dense). This is invisible today because
   the only per-tensor baseline that runs on SM90 is an fp32 dequantize-and-matmul carrying the
   `torch-scaled-mm` tag; `torch._scaled_mm` itself is never called anywhere in the repo, though
   the manifest declares it as `ref_api`.
5. Nothing in the repo produces 1x128 activation scales in either layout — `FP8QuantFwdOp` is
   the per-row `[B, S, G, D]` quantizer for the lightning indexer.

## Still needed before merge

A manifest 1D2D granularity, a `source.kernel_map` entry, and `applies()` / `refusal()` stating
the region this kernel serves. The module now lives at `src/tileops/kernels/gemm/fp8_1d2d.py`
and is re-exported from `tileops.kernels.gemm` and `tileops.kernels`, but no op reaches it yet:
today only the tests and the benchmark import it.


src/tileops/kernels/__init__.py
```diff
@@ -42,6 +42,7 @@
     BmmKernel,
     GemmFp8BlockScaledKernel,
     GemmFp8EpilogueKernel,
+    GemmFp81D2DKernel,
```
lcy-seso: **[Blocker] Wire this into an Op/manifest before exporting it.** The new kernel is currently reachable only by direct imports from tests/benchmarks, so no public TileOPs call can select the advertised 1D2D path. If dispatch is intentionally deferred, keep the kernel unexported until that follow-up lands.

PR #2052, opened 2026-09-03
[Feat][Attention] Support RoPE in dense GQA

superAngGao:
## Summary

- add NeoX and interleaved RoPE to `GroupedQueryAttentionDenseFwdOp`
- rotate Q and K once in a preprocessing kernel, then run the selected SM90 Dense attention kernel
- support partial/full rotary dimensions, rectangular causal attention, and sliding-window attention
- keep the no-RoPE execution paths unchanged
- place the Dense causal, sliding-window, and RoPE implementations together in `kernels/attention/gqa_dense.py`; the former `gqa_prefill_fwd_ws.py` and `gqa_sliding_window_fwd.py` file paths are removed, but their Dense kernels remain

This PR is rebased on #2058 and does not restore the removed packed-THD Op or adapter.

The public operation is fused from the caller's perspective. The current GPU implementation launches a Q/K RoPE preprocessing kernel followed by the selected Dense attention kernel; native-CUPTI timing covers the complete two-kernel envelope, including the gap.

## Validation

H200:

- Dense RoPE/sliding-window cases: **6 passed**
- full GQA smoke file: **10 passed, 2 deselected**
- strict manifest validation passed
- ruff and diff checks passed
- covered FP16/BF16, NeoX/interleaved, partial/full rotary dimensions, rectangular `Sq < Skv`, and sliding window

Native-CUPTI, cold L2, 50 timed repeats (median), H200 SM clock 1500 MHz:

| Workload | TileOps | FlashInfer `ROPE_LLAMA` | Relative |
|---|---:|---:|---:|
| B4, Sq/Skv 512, H32/Hkv8 | 65.5 us | 76.4 us | 116.6% |
| B1, Sq512, Skv8K, H64/Hkv8 | 262.6 us | 572.4 us | 218.0% |
| B1, Sq512, Skv16K, H64/Hkv8 | 469.4 us | 1133.4 us | 241.4% |
| B1, Sq1024, Skv32K, H64/Hkv8 | 1739.3 us | 4491.3 us | 258.2% |
| B1, Sq/Skv4K, H32/Hkv8 | 323.6 us | 726.9 us | 224.6% |
| B1, Sq/Skv8K, H32/Hkv8 | 1028.5 us | 2546.9 us | 247.6% |

The separate rotation pass adds 2.5–13.8% over TileOps no-RoPE latency across the long-context cases.

Sliding-window + RoPE (`window_left=1024`):

| Sequence length | TileOps | FlashInfer `ROPE_LLAMA` | Relative |
|---|---:|---:|---:|
| 4K | 245.8 us | 325.7 us | 132.5% |
| 8K | 440.9 us | 629.9 us | 142.9% |


lcy-seso (commented):
A few suggestions on the plumbing, plus one question about `gqa_legacy_packed_dense.py`. Left inline.

src/tileops/ops/attention/gqa.py
```diff
@@ -494,6 +492,7 @@ def _get_kernel(
     ) -> Callable[..., torch.Tensor]:
         """Resolve the implementation stored in the Op's single cache layer."""
         q, k = inputs[:2]
+        rope_cos = inputs[6]
```
lcy-seso: `inputs[6]` ties this to the position of `rope_cos` in `_canonicalize_inputs`. Adding or reordering an input silently reads the wrong tensor. Unpack by name instead:

```suggestion
        q, k, _v, _q_scale, _k_scale, _v_scale, rope_cos, _rope_sin = inputs
```

(then drop the `q, k = inputs[:2]` line above)

src/tileops/ops/attention/gqa.py
```diff
@@ -527,10 +534,23 @@ def build() -> Kernel:
...
                 sm_scale=self.sm_scale,
                 softcap=self.softcap,
+                fuse_rope=self.pos_encoding_mode == "rope",
+                max_position=rope_cos.shape[0] if rope_cos is not None else 1,
+                rotary_dim=(
+                    _rope_rotary_dim(dim, self.rotary_dim)
+                    if self.pos_encoding_mode == "rope"
+                    else 0
+                ),
+                rope_layout=self.rope_layout,
```
lcy-seso: These four arguments are byte-identical to lines 517-524 in the windowed branch. Hoist them once before `build()` and splat them into both calls:

```python
rope_on = self.pos_encoding_mode == "rope"
rope_kwargs = {
    "fuse_rope": rope_on,
    "max_position": rope_cos.shape[0] if rope_cos is not None else 1,
    "rotary_dim": _rope_rotary_dim(dim, self.rotary_dim) if rope_on else 0,
    "rope_layout": self.rope_layout,
}
```

That keeps the two branches from drifting when a fifth RoPE parameter shows up.

src/tileops/kernels/attention/gqa_dense/attention.py
```diff
@@ -439,69 +433,7 @@ def main(
...
-        k_scale: Optional[torch.Tensor] = None,
-        v_scale: Optional[torch.Tensor] = None,
-    ) -> torch.Tensor:
-        q_bshd, k_bshd, v_bshd = self._bshd(q, k, v)
-        return self.kernel(q_bshd, k_bshd, v_bshd).reshape(q.shape)
-
-
-class GQADensePrefillCausalWsKernel(Kernel):
+class GQADenseCausalWsKernel(Kernel):
     """H200 causal Dense prefill using the FA3 two-consumer pipeline."""
```
lcy-seso: Drop `H200` from the docstring. `supported_archs = [WS_ARCH]` already declares what this kernel runs on, and it is sm90 as a whole rather than one board -- naming a specific card here is both a duplicate and narrower than the truth. `"""Causal dense prefill using the FA3 two-consumer pipeline."""` says everything the reader needs.

Same wording in `gqa_legacy_packed_dense.py:20`, and the skip messages in `tests/ops/attention/test_gqa.py` gate on `is_h200()` while the kernel itself only asks for sm90.

src/tileops/kernels/attention/gqa_dense.py
```diff
@@ -517,14 +449,18 @@ def __init__(
         dtype: torch.dtype,
         sm_scale: Optional[float] = None,
         softcap: float = 0.0,
+        fuse_rope: bool = False,
```
lcy-seso, line 612: The four new parameters land ahead of `config` and `tune`, so any caller that passed `config` positionally now passes it as `fuse_rope`. The class already has a `*` before `device_index` -- put these after it and the change cannot be misread:

```python
        config: Optional[dict] = None,
        tune: bool = False,
        *,
        fuse_rope: bool = False,
        max_position: int = 1,
        rotary_dim: int = 0,
        rope_layout: str = "neox",
        device_index: Optional[int] = None,
```

Same in `gqa_dense/sliding_window.py:281`.

src/tileops/kernels/attention/gqa_dense/attention.py
```diff
@@ -535,6 +471,22 @@ def __init__(
...
+                heads_kv,
+                seq_len_q,
+                seq_len_kv,
+                dim,
+                max_position,
+                rotary_dim,
+                rope_layout,
+                self.dtype_str,
+            )
+            if fuse_rope
```
lcy-seso: This 14-line ternary is copied verbatim into `gqa_dense/sliding_window.py:302`. A factory in `rope.py` removes both copies and gives the `fuse_rope=True` / `rope_cos=None` combination one place to be rejected:

```python
def make_qk_rope_preprocessor(
    *, fuse_rope: bool, batch, heads, heads_kv, seq_len_q, seq_len_kv,
    dim, max_position, rotary_dim, rope_layout, dtype,
) -> Optional["DenseQKRoPEPreprocessor"]:
    ...
```

Call site becomes `self.rope = make_qk_rope_preprocessor(fuse_rope=fuse_rope, ...)`.

src/tileops/kernels/attention/gqa_dense/sliding_window.py
```diff
@@ -294,6 +299,22 @@ def __init__(
         self.dtype = dtype
         self.sm_scale = dim**-0.5 if sm_scale is None else sm_scale
         self.softcap = softcap
+        self.rope = (
```
lcy-seso: Second copy of the block in `gqa_dense/attention.py:474`; see the factory suggestion there. The only difference is that this one passes `seq_len` for both sequence lengths.

src/tileops/kernels/attention/gqa_legacy_packed_dense.py
```diff
@@ -0,0 +1,67 @@
...
+
+import torch
+
+from .call_spec import (
+    ATTENTION_DTYPES,
+    WS_ARCH,
+    causal_ws_prefill_region,
+    square_ws_prefill_region,
+)
+from .gqa_dense.attention import _gqa_dense_causal_ws_kernel
```
lcy-seso: This reaches into `gqa_dense/` for a private symbol. `gqa_dense/__init__.py` re-exports the two kernel classes and nothing else, so the jit builder is now shared across a subpackage boundary through a name that is marked private on both sides.

Two ways out, either fine: export the builder from `gqa_dense/__init__.py` deliberately, or move this module into `gqa_dense/` (see the comment on the class below) so the import stays inside the package.

src/tileops/kernels/attention/gqa_legacy_packed_dense.py
```diff
@@ -0,0 +1,67 @@
...
+from .packed_prefill import PackedPrefillKernel
+
+__all__ = ["GQALegacyPackedDenseAdapter"]
+
+
+class GQALegacyPackedDenseAdapter(PackedPrefillKernel):
+    """Legacy packed-THD adapter for the Dense H200 causal kernel."""
```
lcy-seso: What makes this `Legacy`? The class is what `GroupedQueryAttentionPrefillFwdOp` selects for `gqa_prefill_causal_fwd_kernel`, `applies()` still owns a live region of `causal_ws_prefill_region`, and two tests assert on it. Nothing in this PR retires it or points at a replacement, and the name reaches users through the manifest `kernel_map`. If it is a supported path, name it for the ABI it adapts -- `GQAPackedCausalWsKernel` or similar -- and if it is on its way out, the PR should say what replaces it.

The module split has the same question attached: it holds one class whose only reason to sit outside `gqa_dense/` is the name. `gqa_dense/packed_adapter.py` would keep the packed-THD adapter next to the kernel it wraps and make the import above a package-internal one.

Two lines from the old docstring should come back either way -- they are caller-facing, not implementation notes:

- the FlashInfer `single_prefill_sm90` provenance;
- "emits output only; a backward pass that needs log-sum-exp must compute it separately".

tests/ops/attention/test_gqa.py
```diff
@@ -143,29 +143,95 @@ def _ones_prefill_scales(
...
+    )
+    return torch.cat((rotated.to(x.dtype), x[..., rotary_dim:]), dim=-1).contiguous()
+
+
+@pytest.mark.parametrize(
+    "rope_layout, rotary_dim, dtype",
+    [
+        (None, None, torch.float16),
+        ("neox", 64, torch.float16),
+        ("interleaved", None, torch.float16),
```
lcy-seso: `interleaved` is only exercised at full rotary width, so the combination that actually stresses the index math -- `d0 = 2f`, `d1 = 2f + 1` bounded by `rotary_dim`, plus the `[rotary_dim:]` tail copy -- is never run. Changing this case to `("interleaved", 64, torch.float16)` covers it at no extra node; `neox` keeps its partial case in the row above.

tests/ops/attention/test_gqa.py
```diff
@@ -143,29 +143,95 @@ def _ones_prefill_scales(
...
-def test_gqa_dense_h200_main_kernel_matches_reference() -> None:
+def test_gqa_dense_h200_main_kernel_matches_reference(
+    rope_layout: Optional[str], rotary_dim: Optional[int], dtype: torch.dtype
+) -> None:
     if not is_h200():
         pytest.skip("Dense warp-specialized prefill is currently supported on H200")
-    batch, seq_len_q, seq_len_kv, heads, heads_kv, dim = 1, 96, 150, 8, 2, 128
-    q = torch.randn(batch, seq_len_q, heads, dim, device="cuda", dtype=torch.float16)
-    k = torch.randn(batch, seq_len_kv, heads_kv, dim, device="cuda", dtype=torch.float16)
+    batch, seq_len_q, seq_len_kv, heads, heads_kv, dim = 1, 160, 270, 8, 2, 128
```
lcy-seso: The no-RoPE case moved from `96, 150` to `160, 270`. Both are partial-tile against `BLOCK_M = 128`, but 96 was the only shape covering `S_q < BLOCK_M`, and a feature PR silently changing an existing regression shape makes it hard to tell whether the old one still passes. Either keep `96, 150` as a parameter row or say in the PR body why it had to move.

src/tileops/kernels/attention/gqa_legacy_packed_dense.py
lcy-seso: Filling in what I checked, since the answer decides whether the file stays.

Nothing here is new code. On `main`, `gqa_prefill_fwd_ws.py` holds two classes over one shared `_gqa_prefill_fwd_fa3_kernel` builder:

| class on `main` | ABI | consumer |
| --- | --- | --- |
| `GQADensePrefillCausalWsKernel` | BSHD dense | `GroupedQueryAttentionDenseFwdOp` |
| `GQAPrefillFwdWsPersistentCausalKernel` | packed THD | `GroupedQueryAttentionPrefillFwdOp` |

This PR moves the first into `gqa_dense/attention.py`, moves the second into this new file under a new name, and then imports the private builder back across the subpackage boundary.

So, three separate answers:

- **The class has to stay.** It is what `gqa_prefill_causal_fwd_kernel` resolves to, `applies()` still owns a live slice of `causal_ws_prefill_region`, and `test_gqa_prefill_fwd_q_lt_kv_uses_prefill_ws_kernel` plus `test_gqa_legacy_packed_dense_path_matches_reference` both assert on it. Dropping it regresses the prefill op on sm90 causal dim-128.
- **`Legacy` does not describe it.** No replacement is named in this PR and nothing is scheduled for removal, while the name reaches users through the manifest `kernel_map`. A live path should be named for the ABI it adapts.
- **The module does not need to exist.** It holds exactly one class, and the only thing keeping that class outside `gqa_dense/` is the new name. `gqa_dense/packed_adapter.py` puts the packed-THD adapter beside the kernel it wraps and turns the import above into a package-internal one.

If the intent is that the prefill op migrates off this path later, say so in the PR body and open the follow-up -- that is a fine reason to keep the class, but not a reason to name it `Legacy` today.

src/tileops/kernels/attention/gqa_dense.py
lcy-seso, line 1: These predate the PR, but the rename is the moment to drop them -- they are the ones that only restate the line they sit on.

Delete outright, four lines, both consumer epilogues:

```
287:  T.copy(acc_o, Os[0, :, :])  # registers -> smem
290:  )  # smem -> global (coalesced)
428:  T.copy(acc_o, Os[1, :, :])  # registers -> smem
431:  )  # smem -> global (coalesced)
```

The source and destination buffers are on the line; naming the direction adds nothing, and `(coalesced)` is an unverified claim about the store rather than a fact about the code.

Collapse to one mention, six near-identical labels:

```
215, 249, 277  # pipelined rescale (prev alpha) covers QK latency
354, 388, 418  # pipelined rescale: final alpha before last PV
```

That the epilogue carries `alpha` from the previous tile is one fact about the pipeline, worth stating once in the kernel docstring rather than at each of its six mentions.

Worth keeping as they are: the FlashInfer provenance on line 118, `# register-P (rs-wgmma)` on 159/298, and the `set_max_nreg` pair on 130/154, which explain the 24/240 split rather than restate it.

src/tileops/ops/attention/gqa.py
superAngGao: Done — `_get_kernel()` now unpacks the complete tuple by name, including `rope_cos`.

src/tileops/ops/attention/gqa.py
superAngGao: Done — `rope_on` and the shared `rope_kwargs` dictionary are computed once before `build()` and used by both branches.

src/tileops/kernels/attention/gqa_dense/attention.py
superAngGao: Done — the kernel wording and test gate now describe/check SM90 rather than H200. The packed adapter mentioned here was subsequently removed by #2058.

src/tileops/kernels/attention/gqa_dense.py
superAngGao: Done — the RoPE parameters are keyword-only after `config` and `tune` in both Dense kernel constructors.

src/tileops/kernels/attention/gqa_dense/attention.py
superAngGao: Done — both Dense paths now use the shared `make_dense_qk_rope_preprocessor()` factory.

src/tileops/kernels/attention/gqa_dense/sliding_window.py
superAngGao: Done — the duplicate construction block was replaced by the same shared RoPE preprocessor factory.

src/tileops/kernels/attention/gqa_legacy_packed_dense.py
superAngGao: Resolved by rebasing onto #2058: the packed adapter and its cross-package private import no longer exist in this PR.

src/tileops/kernels/attention/gqa_legacy_packed_dense.py
superAngGao: Resolved by #2058: the old packed-prefill Op and this adapter were removed. This PR now contains only the Dense implementations.

tests/ops/attention/test_gqa.py
superAngGao: Done — the interleaved case now uses partial `rotary_dim=64`, covering both pair indexing and the unrotated tail.

src/tileops/kernels/attention/gqa_dense.py
superAngGao: Done in ea49b05a — the redundant copy/rescale comments are gone, and the carried-alpha behavior is stated once in the WS builder docstring.

lcy-seso (approved):

PR #2066, opened 2026-09-06
[Perf][foundry][Pool] Fold the 1d average-pool window into one output thread

lcy-seso:
## Summary

- Rewrite the `AvgPool1dFwdOp` forward kernel as one flat pass over output positions: a thread owns one output, accumulates its window in a register, and reads its taps under a compile-time bounds claim rather than a per-element branch.
- Decide the divisor when the kernel is built. Without ceil mode a floor-mode window's padded extent is the whole kernel for every output, so counting the padding makes the divisor a constant.
- Let `AvgPool1dSpatialKernel` delegate to the general kernel rather than carry a second copy of the scan: it serves exactly the region the general kernel emits for `ceil_mode=False, count_include_pad=True`.

## TileFoundry Description

```python
@module(
    entry="avg_pool1d",
    topologies=(Topology("cta", 132),),
)
class AvgPool1d:
    @func
    def avg_pool1d(x: Tensor[(N, C, L_IN), "f16"]) -> Tensor[(N, C, L_OUT), "f16"]:
        edge = tf.full_like(x[:, :, 0:PAD], value=0.0)
        padded = tf.concat([edge, x, edge], axis=2)
        taps = tf.stack(
            [padded[:, :, k : k + (L_OUT - 1) * STRIDE + 1 : STRIDE] for k in range(KERNEL)],
            axis=-1,
        )
        return tf.reduce(taps, axes=(-1,), keepdim=False, kind="mean")
```

`count_include_pad=False` makes the divisor the number of real input positions the window
covers, so the same window runs over a mask of ones and the two sums are divided:

```python
        total = tf.reduce(tf.cast(taps, dtype="f32"), axes=(-1,), keepdim=False, kind="sum")
        live_taps = tf.stack([live_padded[:, :, k : k + 2047 : 2] for k in range(4)], axis=-1)
        count = tf.reduce(tf.cast(live_taps, dtype="f32"), axes=(-1,), keepdim=False, kind="sum")
        return tf.cast(total / count, dtype="bf16")
```

`tilefoundry check` holds the delivered kernel to those programs, run by the evaluator, on
every primary manifest workload: `--fn allclose --atol 1e-3 --rtol 1e-3` reports
`max_violation 0` for `f16[4,128,2048]`, `f16[2,256,8000]` and `bf16[2,128,1024]`.

## Performance

Operator: `AvgPool1dFwdOp`

| Environment | Value |
| --- | --- |
| gpu | NVIDIA H200 |
| driver | 595.71.05 |
| cuda | 13.2 |
| torch | 2.13.0+cu132 |
| tilefoundry | 0.1.dev131+g2306bf050 |
| timer | native CUPTI |

Method: candidate, incumbent and both torch baselines ran in one process on common inputs.
Each TileLang implementation was searched over its own config space and is reported at its
best; each was checked against `torch.nn.functional.avg_pool1d` before timing, then timed by
`bench_kernel` through CUPTI, which clears L2 before every iteration. Compilation and tuning
were excluded; CUPTI failure was fail-closed.

| Workload | N | C | L | Params | Dtype | Candidate (ms) | TileOPs incumbent (ms)<br>/ candidate | torch-compile (ms)<br>/ candidate | torch-ref (ms)<br>/ candidate |
| --- | ---: | ---: | ---: | --- | --- | ---: | ---: | ---: | ---: |
| audio-downsample | 4 | 128 | 4096 | k3 s2 p1 | float16 | 0.004576 | **0.005504<br>&#x1F7E2;&nbsp;1.2028x** | **0.005248<br>&#x1F7E2;&nbsp;1.1469x** | **0.012000<br>&#x1F7E2;&nbsp;2.6224x** |
| long-temporal | 2 | 256 | 32000 | k5 s4 p2 | float16 | 0.017281 | **0.019072<br>&#x1F7E2;&nbsp;1.1036x** | **0.020097<br>&#x1F7E2;&nbsp;1.1629x** | **0.045792<br>&#x1F7E2;&nbsp;2.6499x** |
| ceil | 2 | 128 | 2048 | k4 s2 p1 ceil cip=False | bfloat16 | 0.002176 | **0.003008<br>&#x1F7E2;&nbsp;1.3824x** | **0.002784<br>&#x1F7E2;&nbsp;1.2794x** | **0.004288<br>&#x1F7E2;&nbsp;1.9706x** |
| geometric mean |  |  |  |  |  | 0.005788 | **0.006999<br>&#x1F7E2;&nbsp;1.2092x** | **0.007058<br>&#x1F7E2;&nbsp;1.2194x** | **0.014487<br>&#x1F7E2;&nbsp;2.5029x** |

## Result And Limitations

**SOTA**

- Classification is SOTA on this device: the candidate is ahead of every comparator on every
  primary workload, by 1.10x to 1.38x against the incumbent and 1.15x to 1.28x against
  torch-compile.
- What is left is one division per output, and that is measured rather than inferred. The
  same kernel with the division replaced by an addition of the same divisor — everything
  else held fixed — runs 3.841us / 16.480us / 1.953us against 4.481us / 17.408us / 2.208us
  in table order.
- A load-only twin — same grid, same input stream, the same tap count, no divisor — runs
  3.840us / 15.745us / 1.952us. Two of the three land on the no-division kernel, so on
  `audio-downsample` and `ceil` the division is the whole residue over the memory stream.
  `long-temporal` keeps 0.6us beyond it, in the per-tap select.
- The division is kept. `total / k` and `total * (1 / k)` differ by up to one fp32 ulp for a
  divisor that is not a power of two, and PyTorch divides; buying 5% to 14% with a deviation
  from the reference is not a trade this op should make silently.
- Three guard strategies were measured. A per-tap branch: 4.640us / 17.728us / 2.240us. The
  per-tap select this PR ships: 4.512us / 17.280us / 2.176us. A per-CTA predicate over an
  unguarded body: 4.640us / 17.313us / 2.272us.
- 261 `tests/ops/test_pool.py` cases pass, unchanged in count.

### DIFF
--- a/src/tileops/kernels/pool/avg_pool1d.py
+++ b/src/tileops/kernels/pool/avg_pool1d.py
@@ -26,75 +26,75 @@ def _avg_pool1d_kernel(
 ):
     accum_dtype = "float"
     out_l = pool_output_dim(l_in, kernel_l, stride_l, pad_l, ceil_mode)
-    total_output = n * c_in * out_l
+    rows = n * c_in
+    total = rows * out_l
+    # With no padding and no ceil overshoot every window lies inside its row, so
+    # the per-tap bounds test is a compile-time truth and drops out.
+    window_inside = pad_l == 0 and (out_l - 1) * stride_l + kernel_l <= l_in
+    # Without ceil mode the last window ends at l_in + pad_l at the furthest, so
+    # counting the padding gives the whole window for every output. Ceil mode can
+    # push a window past that, and not counting the padding shortens the windows
+    # that overhang, so both take the divisor from the window's own extent.
+    whole_window_divides = window_inside or (count_include_pad and not ceil_mode)
 
     @tilelang.jit(out_idx=[1], compile_flags=["-O3", "-DENABLE_BF16"])
     def _avg_pool1d_func(block_m: int, threads: int):
+        tile_full = total % block_m == 0
+
+        def safe(il):
+            """*il* as an index that is always in range, clamping only when it can
+            leave the row. The clamped value is read under a predicate that already
+            discarded it, so which position it names does not matter."""
+            return il if window_inside else T.max(0, T.min(il, l_in - 1))
+
         @T.prim_func
         def _avg_pool1d_main(
-            x: T.Tensor((n, c_in, l_in), dtype),  # type: ignore
-            out: T.Tensor((n, c_in, out_l), dtype),  # type: ignore
+            x: T.Tensor((rows, l_in), dtype),  # type: ignore
+            out: T.Tensor((rows, out_l), dtype),  # type: ignore
         ):
-            with T.Kernel(T.ceildiv(total_output, block_m), threads=threads) as bx:
-                T.use_swizzle(10)
-                tile_out_start = bx * block_m
-                tile_out_end = tile_out_start + block_m - 1
-                tile_ol_start = tile_out_start % out_l
-                tile_ol_end = tile_out_end % out_l
-                tile_same_row = tile_out_start // out_l == tile_out_end // out_l
-                tile_input_start = tile_ol_start * stride_l - pad_l
-                tile_input_end = tile_ol_end * stride_l + kernel_l - 1 - pad_l
-                tile_spatial_full = (
-                    tile_same_row & (tile_input_start >= 0) & (tile_input_end < l_in)
-                )
+            with T.Kernel(T.ceildiv(total, block_m), threads=threads) as tile:
                 for i in T.Parallel(block_m):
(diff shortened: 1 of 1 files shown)


lcy-seso:
Superseded by #2073, which carries this change unchanged plus whole-row shared staging: audio-downsample 5.5 -> 3.8 us (0.67x -> 1.37x of torch.compile) and ceil 3.0 -> 2.1 us (0.67x -> 1.33x), against 4.6 us and 2.2 us here. Merged with #2067 because both applied the same window_inside test and the same safe() helper to sibling kernels; together the shared part is written once.

PR #2067, opened 2026-09-06
[Perf][foundry][Pool] Read a 1d max-pool window once per output thread

lcy-seso:
## Summary

- Rewrite the `MaxPool1dFwdOp` forward kernel as one flat pass over output positions: a thread owns one output, folds its window into a register, and reads its taps under a compile-time bounds claim rather than a per-element branch.
- Add `block_k`, which spreads window positions across lanes and finishes with a cross-lane max, for global pooling — a window as long as the row leaves no output positions to spread over threads.
- Carry the window's NaN beside its value in the lane path, because a cross-lane max drops it.

## TileFoundry Description

```python
@module(
    entry="max_pool1d",
    topologies=(Topology("cta", 132),),
)
class MaxPool1d:
    @func
    def max_pool1d(x: Tensor[(N, C, L_IN), "f16"]) -> Tensor[(N, C, L_OUT), "f16"]:
        left = tf.full_like(x[:, :, 0:PAD], value=NEG_INF)
        right = tf.full_like(x[:, :, 0:OVERHANG], value=NEG_INF)
        padded = tf.concat([left, x, right], axis=2)
        taps = tf.stack(
            [
                padded[:, :, k * DILATION : k * DILATION + (L_OUT - 1) * STRIDE + 1 : STRIDE]
                for k in range(KERNEL)
            ],
            axis=-1,
        )
        return tf.reduce(taps, axes=(-1,), keepdim=False, kind="max")
```

`tilefoundry check` holds the delivered kernel to that program, run by the evaluator, on
every primary manifest workload: `--fn equal` reports 0 mismatched elements for
`f16[32,80,1365]`, `f16[64,256,1]` and `bf16[16,128,1023]`.

`tilefoundry analyze` prices the description at `bound-by=memory` on all three, which is
what the schedule below is written against. The description states the semantics, not the
placement: it materializes the stacked window, so its `ideal-ns` is the cost of an unfused
program and not the floor the kernel is read against. That floor is measured below.

## Performance

Operator: `MaxPool1dFwdOp`

| Environment | Value |
| --- | --- |
| gpu | NVIDIA H200 |
| driver | 595.71.05 |
| cuda | 13.2 |
| torch | 2.13.0+cu132 |
| tilefoundry | 0.1.dev131+g2306bf050 |
| timer | native CUPTI |

Method: candidate, incumbent and both torch baselines ran in one process on common inputs.
Each TileLang implementation was searched over its own config space and is reported at its
best; each was checked against `torch.nn.functional.max_pool1d` before timing, then timed by
`bench_kernel` through CUPTI, which clears L2 before every iteration. Compilation and tuning
were excluded; CUPTI failure was fail-closed.

| Workload | N | C | L | Params | Dtype | Candidate (ms) | TileOPs incumbent (ms)<br>/ candidate | #2064 staged (ms)<br>/ candidate | torch-compile (ms)<br>/ candidate | torch-ref (ms)<br>/ candidate |
| --- | ---: | ---: | ---: | --- | --- | ---: | ---: | ---: | ---: | ---: |
| sincnet-speaker-local | 32 | 80 | 4096 | k3 s3 | float16 | 0.010784 | 0.010720<br>&#x1F534;&nbsp;0.9941x | 0.010721<br>&#x1F534;&nbsp;0.9942x | 0.010656<br>&#x1F534;&nbsp;0.9881x | **0.038752<br>&#x1F7E2;&nbsp;3.5935x** |
| textcnn-global | 64 | 256 | 128 | k128 s1 | float16 | 0.003392 | **0.010624<br>&#x1F7E2;&nbsp;3.1321x** | **0.003584<br>&#x1F7E2;&nbsp;1.0566x** | 0.003360<br>&#x1F534;&nbsp;0.9906x | **0.020672<br>&#x1F7E2;&nbsp;6.0943x** |
| ecg-cnn-dilated | 16 | 128 | 2048 | k5 s2 p2 d2 ceil | bfloat16 | 0.006752 | **0.007872<br>&#x1F7E2;&nbsp;1.1659x** | **0.007872<br>&#x1F7E2;&nbsp;1.1659x** | **0.006912<br>&#x1F7E2;&nbsp;1.0237x** | **0.027200<br>&#x1F7E2;&nbsp;4.0284x** |
| geometric mean |  |  |  |  |  | 0.006392 | **0.009588<br>&#x1F7E2;&nbsp;1.5000x** | **0.007074<br>&#x1F7E2;&nbsp;1.1067x** | 0.006389<br>&#x1F534;&nbsp;0.9995x | **0.027554<br>&#x1F7E2;&nbsp;4.3106x** |

## Result And Limitations

**improvement without SOTA**

- Classification is improvement without SOTA. The candidate is 3.1321x the incumbent on the
  global-pooling workload and 1.1659x on the dilated one, and is ahead of torch-compile on
  the dilated one. On the other two it neither gains nor loses: 0.9906x and 0.9881x
  torch-compile, inside the run-to-run spread.

### Overlap with #2064

#2064 rewrites the same file. Its answer to `textcnn-global` is to stage whole rows of one
image in shared memory; this PR's is to spread the window across lanes and reduce, with no
shared memory and no staging preconditions. The two cannot both land as written, and the
table above times both against this branch in one process, each at its own best config.

| | this PR | #2064 | serves |
| --- | ---: | ---: | --- |
| sincnet-speaker-local | 0.010784 | 0.010721 | both, unchanged from main |
| textcnn-global | 0.003392 | 0.003584 | both |
| ecg-cnn-dilated | 0.006752 | 0.007872 | this PR only; #2064 leaves it at main |

#2064's `max_pool1d` path needs `out_l <= 32`, `block_m % out_l == 0` and `c_in % rows == 0`
to engage, plus a shared row pad tuned to a bank count. This PR's window split needs a window
of 8 positions or more and is otherwise unconditional. The recommendation is that #2064 keeps
its `adaptive_max_pool2d` half, which this PR does not touch, and drops `max_pool1d.py`.

### The rest of the gap

- A load-only twin — same grid, same input stream, one tap instead of the window — runs
  `sincnet-speaker-local` in 9.54us against the candidate's 10.78us, and `ecg-cnn-dilated` in
  5.44us against 6.75us. The residue is the per-tap load, not the access pattern: the twin
  already moves every byte the full kernel does.
- Five other schedules were measured and rejected on that basis. Shared-memory staging of the
  input span: 10.66us / 8.16us (sincnet / ecg). The same with 16-byte-aligned spans: 10.90us
  on ecg. A three-dimensional grid without staging: 11.42us / 7.58us. One thread owning V
  consecutive outputs with `T.vectorized` loads: 11.71us at V=4 and 16.06us at V=8 on sincnet.
- `block_k` splits the window only where the tuner is offered the choice — windows of 8
  positions or more. Below that a cross-lane reduction costs more than the serial run it
  replaces. Both families stay in the tuning space whenever the window can divide, so the
  choice is measured rather than assumed.
- NaN propagation cannot use plain `fmax`: `fmax` returns the non-NaN operand. Folding the
  test into the update costs about 5% against a NaN-unaware `fmax` on `ecg-cnn-dilated`
  (6.66us against 6.30us), measured. The arithmetic alternative — accumulating `v - v` beside
  the maximum — is 5% faster and wrong: the compiler folds `v - v` to zero.
- The 4.8 TB/s HBM figure the target publishes is not reachable under this measurement. A
  same-shape elementwise copy, timed the same way, reaches 3.49 TB/s at 41.9 MB and 2.22 TB/s
  at 8.4 MB; the L2 clear before every iteration is part of what the kernel is timed against.
- 261 `tests/ops/test_pool.py` cases pass, unchanged in count.

### DIFF
--- a/src/tileops/kernels/pool/max_pool1d.py
+++ b/src/tileops/kernels/pool/max_pool1d.py
@@ -11,6 +11,15 @@ from tileops.kernels.pool.common import pool_output_dim
 
 __all__ = ["MaxPool1dKernel", "MaxPool1dWithIndicesKernel"]
 
+# A window shorter than this has too few positions to fill even one lane group,
+# and the cross-lane reduction costs more than the serial run it replaces.
+_MIN_SPLIT_WINDOW = 8
+
+# Output positions below which a launch cannot fill the device from the output
+# axis alone: 132 SMs at 2048 resident threads is the widest device this kernel
+# is built for.
+_SMALL_LAUNCH = 132 * 2048
+
 
 @functools.lru_cache(maxsize=32)
 def _max_pool1d_kernel(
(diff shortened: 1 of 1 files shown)


lcy-seso:
Superseded by #2073. Two of the three changes carry over: the (rows, l_in) addressing and the NaN form, which is where this PR's whole gain on ecg-cnn-dilated comes from -- window_inside alone measures 7.8 -> 7.8, 10.8 -> 10.7 and 3.6 -> 3.6 us, since ecg has pad_w = 2 and the other two rows already had it.

The block_k lane split is dropped. On the manifest it fires only on textcnn-global, and there it ties the shared staging #2064 already merged: 3.50 us against 3.60 us. Keeping the merged path costs one fewer config dimension and drops the hard-coded 132-SM count the review here flagged.

The NaN form also turned out not to be free everywhere: applied to both launches it took ecg-cnn-dilated to 1.03x but textcnn-global to 0.85x, so #2073 gives the two launches different forms depending on whether their taps come from global or from shared.

PR #2085, opened 2026-09-06
[Refactor][Norm] Share the row reduction and the partial predicate

lcy-seso:
## Problems

- `layer_norm` reduces a loaded fp32 row block to mean and rstd inline, term for term the same as `_config.make_row_reduce`, which `group_norm` and `ada_layer_norm` already call.
- The predicate choosing the per-thread-partial path is five lines, comment included, byte for byte in `layer_norm` and `rms_norm`. It differs only in a threshold both already pass as a parameter.

## Changes

- `layer_norm`'s non-partial branch calls `make_row_reduce`.
- The predicate moves to `_config` as `use_per_thread_partial`; both files call it.

## Not shared

`layer_norm`'s per-thread-partial branch stays written out. It accumulates across a fragment `threads` wide, walking shared memory with a serial stride, where the shared macro rewrites the full row in place — two different reductions that happen to share a tail. `rms_norm`'s partial branch reduces a sum of squares, so it has nothing to share either.

## No behaviour change

Config and generated CUDA dumped for 102 specializations — seven kernel classes, three dtypes, eight shapes, the row counts chosen to build both sides of the predicate and the widths to cover an aligned and a padded row. **Zero differ.** 249 norm tests pass, 5 skipped.

### DIFF
--- a/src/tileops/kernels/norm/_config.py
+++ b/src/tileops/kernels/norm/_config.py
@@ -27,6 +27,7 @@ __all__ = [
     "select_row_config",
     "select_row_config_by_width",
     "select_row_configs",
+    "use_per_thread_partial",
 ]
 
 # Powers of two only (tl::AllReduce is an XOR butterfly) that also divide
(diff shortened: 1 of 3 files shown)


lcy-seso:
Closing: net zero lines. The layer_norm half removes a third copy of the mean/rstd reduction (-13); the predicate half adds more than it removes (+14) to deduplicate a five-line boolean, and costs a jump to read. Neither is worth a review on its own. The duplication is recorded here for whoever next opens layer_norm for a real reason.

PR #2088, opened 2026-09-07
[Perf][foundry][Pool] Drop two redundant terms from the max-pool index scan

lcy-seso:
## Problems

The with-indices window scan carried two terms it does not need:

- `not is_nan` guarded every compare, and
- a `has_nan` accumulator rode alongside `nan_idx` for the whole scan.

## Changes

`max_val` starts at `-inf` and only a passing compare replaces it, so it is never
NaN, and NaN fails `>`. The compare alone therefore rejects NaN, and `not is_nan`
comes out of the update where the window is known to lie inside the row. Where it is
not, `first_valid` admits the first element whatever it is, so that arm keeps the
test.

`nan_idx` starts at `-1`, which already says no NaN has been seen, so `has_nan` goes
with it: one fewer register live across the scan, one fewer operation per tap.

`eff_w` names the window's span once instead of spelling it inline.

## Result And Limitations

23 lines in, 29 out. Values and indices are bit-identical to PyTorch on `ecg-cnn-dilated`,
`sincnet-speaker-local` and `textcnn-global`, for plain input, random NaN, and an
all-`-inf` input.

`benchmarks/ops/bench_pool.py` on an idle H200, this branch against `main`, each side
from its own empty `TILELANG_CACHE_DIR`:

| workload | before | after | vs torch-compile |
| --- | ---: | ---: | --- |
| textcnn-global MaxPool1dIndices f16 | 4.40 us | 3.90 us | 1.09x -> **1.23x** |

No other row moves beyond the report's 0.1 us resolution.

The gain lands on `textcnn-global` because its padding is zero, so
`always_in_bounds` holds and it takes the arm both terms were removed from.
`ecg-cnn-dilated` has `pad_w = 2` and keeps the guarded arm, so it is unchanged at
0.86x. A per-output runtime version of that test was built and measured against it:
13.12 us to 11.20 us in a standalone harness, but 13.50 us to 13.63 us in the kernel
itself, so it was dropped. Decomposing that row shows why the two do not agree --
removing its 16.8 MB of int64 index writes saves 0.64 us, while removing the window
scan saves 4.19 us, and in the kernel the four integer divisions per output that the
harness does not have hide the scan's savings.

### DIFF
--- a/src/tileops/kernels/pool/max_pool1d.py
+++ b/src/tileops/kernels/pool/max_pool1d.py
@@ -53,7 +53,8 @@ def _max_pool1d_kernel(
     accum_dtype = "float"
     out_l = pool_output_dim(l_in, kernel_w, stride_w, pad_w, ceil_mode, dilation_w)
     total_output = n * c_in * out_l
-    always_in_bounds = pad_w == 0 and (out_l - 1) * stride_w + (kernel_w - 1) * dilation_w < l_in
+    eff_w = dilation_w * (kernel_w - 1) + 1
+    always_in_bounds = pad_w == 0 and (out_l - 1) * stride_w + eff_w - 1 < l_in
 
     @tilelang.jit(out_idx=[1], compile_flags=["-O3", "-DENABLE_BF16"])
     def _max_pool1d_func(block_m: int, threads: int):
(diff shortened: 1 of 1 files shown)


lcy-seso:
Folded into #2087 as `52dea843`, cherry-picked unchanged, and closing this one so the two do
not both edit the pool kernels.

Verified independently before folding it in, since #2087 now reports it:

- Values and int64 positions equal to `F.max_pool1d` over 7 shapes x 3 dtypes x 3 input kinds
  (plain, random NaN and -inf, all -inf) x 9 configs, 0 mismatches. The shapes split
  deliberately across both arms of the bounds specialization: unpadded and window-inside rows
  take the arm the two terms came out of, while padded, dilated, ceil-overshoot and all three
  combined keep the guarded arm.
- Re-measured on the manifest workloads for `MaxPool1dIndicesFwdOp`, main against this change,
  one process on one idle H200:

| workload | main | with this change | torch-compile |
| --- | --- | --- | --- |
| sincnet-speaker-local f16 | 16.160us | 16.096us | 15.424us |
| textcnn-global f16 | 4.448us | 3.872us | 4.864us |
| ecg-cnn-dilated bf16 | 12.608us | 10.944us | 10.816us |

`textcnn-global` reproduces what this PR claimed, 4.45 to 3.87us. One correction: this PR's
report says `ecg-cnn-dilated` is unchanged because it keeps the guarded arm, but it gains
1.15x, 12.608 to 10.944us. `has_nan` left both arms, not only the unguarded one, so the
guarded arm also loses one accumulation per tap. That takes the row from 0.86 to 0.99 against
torch-compile.

PR #2100, opened 2026-09-07
[Feat][MoE] Add an SM90 GEMM template over dense, batched and expert-grouped layouts with per-call tile selection

michaelwithu:
## Problem

TileOPs' Hopper GEMMs are one hand-written kernel per layout. The MoE grouped GEMMs (`GroupedGemmPersistent3WGKernel`) serve one tight layout, bind their shape at construction, and pick a schedule from a fixed regime table rather than from the shape at hand; there is no masked or batched variant, and the dense kernel shares nothing with them. Adding a layout meant another kernel body, and there was no model to say which tile a new shape should run.

This PR adds one SM90 GEMM template, `SM90GemmFwdKernel`, whose single kernel body is specialised by a spec and whose tile is chosen per call by a calibrated cost model. It covers dense, batched and four expert-grouped layouts with the same mainloop and epilogue, takes bf16 or fp16 operands with fp32 accumulation, and keeps `M` dynamic so one compiled kernel serves every row count of a shape.

No op calls it yet: the staged grouped-GEMM op interface it will sit behind is a separate PR, and the adapter, manifest `kernel_map` and 3WG replacement follow both.

## Changes

- `src/tileops/kernels/moe/sm90_gemm.py` (re-exported with `GemmType` from `kernels/moe/__init__.py`): one `@T.prim_func` body specialised by `SM90GemmSpec` — GEMM type, operand majorness (read off the strides, so a transposed view is a layout, not a copy), `ab_dtype` bf16/fp16, `cd_dtype` operand or fp32, static vs dynamic dims, tile, stages, one or two math warp-groups, 2-CTA TMA multicast. Persistent grid, L2-swizzled tile order, TMA/WGMMA pipeline, TMA-store epilogue.
- Six GEMM types over three schedulers: `NORMAL` and `M_GROUPED_ALIGNED_PER_ROW` (flat grid); `BATCHED` (grid per batch); `M_GROUPED_MASKED`, `M_GROUPED_ALIGNED_PSUM` and `M_GROUPED_TIGHT_PSUM` (per-group prefix sum in shared memory, binary search, in-group swizzle). The tight layout stores a group's ragged last tile under a row mask so it never overwrites the next group. K-grouped types, `with_accumulation` and small M (`block_m` 16/32) are out of scope by decision.
- The math warp-groups keep one WGMMA group in flight (`wgmma.wait_group 1`, release the previous stage) instead of draining every k-step. Draining cost Mixtral-shaped grouped GEMMs 4–7% against the 3WG kernel at the same tile (CUPTI); with one group in flight they are level.
- `sm90_gemm_heuristics.py`: the candidate enumeration and an L1/L2 cycle model, calibrated on H200 against a sweep of ~600 of its own candidates over 15 shapes. Findings that went into the rules: `block_n` in multiples of 64 (other widths are 1.46x+ slower through TileLang's epilogue swizzle); per-group tile round-up in the tile count, so 32-row experts pick `block_m` 64; a 2-CTA cluster only for dense GEMMs and only from 16 waves (1–6% slower at 8 waves and fewer, 0.6–5% faster from 16); shared-memory slack for TileLang's 1024-byte buffer alignment (a stress case asked for 231680 of 232448 bytes and failed to launch). Selector gap to the measured best: mean 0.7%, max 2.3%.
- `tests/kernels/test_sm90_gemm.py`: 29 cases, one per code path (six types, four operand layouts, fp16 operands, fp32 output, dynamic M, both multicast directions and odd tile groups, a cluster across a group boundary, tight ragged tiles, selector rules, spec validation). A 200-case randomised stress over types, layouts, dtypes, unaligned shapes and off-selector tiles found one bug — a shared-memory budget a few hundred bytes under the 227 KB cap failing at launch, fixed by the alignment slack above — and nothing else.

## Performance

Idle H200, one GPU, `do_bench` 100 iterations, outputs compared element-wise against torch fp32.

Dense bf16 NT, cuBLAS time / this kernel's time (>1 is faster):

| Weight (N x K)             | M=2048 | 4096 | 8192 | 16384 |
| -------------------------- | -----: | ---: | ---: | ----: |
| llama7b qkvo 4096x4096     |   0.97 | 1.01 | 0.97 |  1.03 |
| llama70b qkvo 8192x8192    |   1.02 | 1.03 | 1.09 |  0.98 |
| llama70b gate 28672x8192   |   1.00 | 1.02 | 1.19 |  1.19 |
| dsv3 dense gate 18432x7168 |   1.01 | 1.03 | 1.04 |  1.08 |
| dsv3 dense down 7168x18432 |   0.92 | 0.99 | 0.98 |  1.00 |

Grouped tight psum on gate_up shapes with skewed routing, vs `GroupedGemmPersistent3WGKernel` and `torch._grouped_mm` (>1 is faster):

| Model     | tokens | experts |     N |    K | vs 3WG | vs grouped_mm |
| --------- | -----: | ------: | ----: | ---: | -----: | ------------: |
| dsv3      |    256 |     256 |  4096 | 7168 |  1.05x |         1.29x |
| dsv3      |   1024 |     256 |  4096 | 7168 |  1.22x |         1.22x |
| dsv3      |   4096 |     256 |  4096 | 7168 |  1.00x |         1.39x |
| mixtral   |    256 |       8 | 28672 | 4096 |  0.97x |         1.36x |
| mixtral   |   4096 |       8 | 28672 | 4096 |  0.98x |         1.03x |
| qwen2-moe |   1024 |      64 |  2816 | 2048 |  1.00x |         1.33x |
| qwen2-moe |   4096 |      64 |  2816 | 2048 |  1.21x |         0.96x |

Masked and batched match `torch.bmm` within 1–2% on weight-bandwidth-bound DeepSeek-V3 shapes.

Where this kernel is not good enough: small M on a weight that does not fill the machine. Its smallest tile is `block_m` 64 and it has no split-K, so when `M × N` yields fewer tiles than SMs the GEMM runs one partial wave and nothing hides the K loop. Measured on the same H200 (cuBLAS time / this kernel's time, >1 is faster):

| Weight (N x K)             | M=16 | M=64 | M=128 | M=256 | M=512 |
| -------------------------- | ---: | ---: | ----: | ----: | ----: |
| llama7b qkvo 4096x4096     | 0.35 | 0.89 |  0.91 |  0.52 |  0.52 |
| llama70b gate 28672x8192   | 0.94 | 0.96 |  0.98 |  0.94 |  0.88 |
| dsv3 dense down 7168x18432 | 0.68 | 1.05 |  0.92 |  0.86 |  1.00 |

On the narrow 4096² weight the kernel is half of cuBLAS from M=256 up to M=512 (128 tiles on 132 SMs, one wave); on the wide 28672-column weight, where even M=16 fills the machine, it stays within 2–12%. Small N (N=2112, which only 64-wide tiles divide) is the same problem from the other side. Grouped small token counts are not affected the same way: tight psum at 8–128 tokens with skewed routing is 1.02–1.27x the 3WG kernel and 1.3–1.5x `torch._grouped_mm` on DeepSeek-V3 and Mixtral gate_up shapes, because the per-expert scheduler already runs the ragged tiles the shape has. Below the wave boundary this kernel should not be selected for dense shapes: a dedicated small-M kernel (split-K or a swapped-operand schedule with `block_m` 16/32) is a follow-up rather than a tuning of this one, and the short-group 3WG fused-activation kernel keeps the MoE decode regime.

Second commit, after the staged grouped-GEMM wiring measured the template inside the experts pipeline: a tight group's ragged last tile was written straight from the accumulator fragment under a row predicate (TMA cannot clip it against the group). On DeepSeek-V3's down projection (N 7168, K 2048) under random routing that cost 12–24% against `GroupedGemmPersistent3WGKernel`; the tile now stages through the warp-group's shared buffer like a full tile and writes back with row-predicated vector stores. Interleaved on H200, E=128, 32768 rows, random routing: down 2.19–2.28 → 1.89–1.91 ms, gate_up 3.78–3.83 → 3.67 ms; the template is now 1.02–1.05x the 3WG kernel on both projections. Only the tight type compiles this branch; the tables above were measured before it, so the tight rows there are conservative.

Third commit, from a boundary audit of the kernel (52 offline cases: ragged N and K, 0- and 1-row groups, 1000 groups, masked counts past `max_m`, clusters with odd tile counts, every host refusal): a zero-row call divided by zero in the selector and now returns the empty output; three inputs that computed a wrong answer or failed at TMA descriptor encode are refused with a message — a pinned `block_m` that differs from an aligned layout's `m_alignment` (the kernel rounds group starts to `block_m`), a cluster pinned on a per-group type (a CTA pair straddled two groups), and `N` not a multiple of 8 or `K = 0`. Also refused: `num_groups` or `m_alignment` below 1, operands under two dims, `out` aliasing an operand, operands on different devices, operands or `grouped_layout` not contiguous in their physical layout. The aligned per-row peer-row read is clamped inside the metadata, and the tight ragged store carries an explicit column guard. No selector-chosen spec was affected; every refusal names its reason.

## Follow-ups

- Third PR, after this and the interface PR: an adapter class over the staged op's `MGroupedGemmCall` (its `applies` claims the grouped layouts this kernel has a type for — tight psum, aligned psum, aligned per-row, masked — its constructor looks the layout up into a `gemm_type` and builds this kernel, its `forward` passes the metadata as `grouped_layout`), the manifest `kernel_map` and workloads, a benchmark, and the replacement of `GroupedGemmPersistent3WGKernel` where this kernel covers it (the fused-activation variant stays: it needs a gate/up epilogue this kernel does not have, and small M is out of scope).
- A small-M kernel for the decode regime (M ≤ 128 dense, short groups in MoE): split-K or a swapped-operand schedule with `block_m` 16/32. Out of this template's scope by decision; until it exists the 3WG fused-activation kernel and `GemmKernel`'s small-batch paths keep those shapes.
- `GemmFwdOp` dense adapter, after a measured comparison against `GemmKernel` on the dense workloads.
- Mid-batch `BATCHED` einsum layout; re-run the calibration sweep on other devices.

### DIFF
--- a/src/tileops/kernels/moe/__init__.py
+++ b/src/tileops/kernels/moe/__init__.py
@@ -8,6 +8,7 @@ from .moe_grouped_gemm_separate_act import MoeGroupedGemmSeparateActKernel
 from .permute_align import MoePermuteAlignKernel
 from .permute_contiguous import MoePrePermuteContiguousKernel
 from .shared_expert_mlp import SharedExpertMLPKernel
+from .sm90_gemm import SM90GemmFwdKernel
 from .unpermute import MoeUnpermuteKernel
 
 __all__ = [
@@ -21,5 +22,6 @@ __all__ = [
     "MoeUnpermuteKernel",
     "PostPermuteCall",
     "PrePermuteCall",
+    "SM90GemmFwdKernel",
     "SharedExpertMLPKernel",
 ]
--- a/src/tileops/kernels/moe/__init__.py
+++ b/src/tileops/kernels/moe/__init__.py
@@ -8,11 +8,12 @@ from .moe_grouped_gemm_separate_act import MoeGroupedGemmSeparateActKernel
 from .permute_align import MoePermuteAlignKernel
 from .permute_contiguous import MoePrePermuteContiguousKernel
 from .shared_expert_mlp import SharedExpertMLPKernel
-from .sm90_gemm import SM90GemmFwdKernel
+from .sm90_gemm import GemmType, SM90GemmFwdKernel
 from .unpermute import MoeUnpermuteKernel
 
 __all__ = [
     "FusedTopKKernel",
+    "GemmType",
     "MGroupedGemmCall",
     "MoeGroupedGemmNopadKernel",
     "MoeGroupedGemmPersistent3WGFusedActKernel",
--- a/src/tileops/kernels/moe/sm90_gemm.py
+++ b/src/tileops/kernels/moe/sm90_gemm.py
@@ -325,23 +325,30 @@ def _make_prim_func(
 
     @T.macro
     def store_tile(C, C_src, C_s, group, row0, col0, rows, wg):
-        """TMA-store one warp-group's rows; a tight group's ragged tail is masked."""
+        """Store one warp-group's rows of a tile from its shared staging buffer.
+
+        Every tile is staged fragment -> shared first; the previous tile's TMA store
+        may still be reading ``C_s``, hence the barrier before. A full tile then
+        goes out through TMA. A tight group's ragged last tile cannot: TMA clips
+        against the tensor, not the group, and would overwrite the next group's
+        rows, so its valid rows are written back with row-predicated vector stores.
+        (Writing the accumulator fragment straight from registers scattered 4-byte
+        stores across the tile and cost 12% to 24% on short-K grouped GEMMs.)
+        """
+        T.sync_threads(barrier_id=_EPILOGUE_BARRIER_BASE + wg, arrive_count=128)
+        T.copy(C_src, C_s)
         if tight_psum:
             if rows < T.int32(wg_rows):
+                T.sync_threads(barrier_id=_EPILOGUE_BARRIER_BASE + wg, arrive_count=128)
                 if rows > 0:
                     for i, j in T.Parallel(wg_rows, block_n):
                         if i < rows:
-                            C[row0 + i, col0 + j] = C_src[i, j]
+                            C[row0 + i, col0 + j] = C_s[i, j]
             else:
-                T.sync_threads(barrier_id=_EPILOGUE_BARRIER_BASE + wg, arrive_count=128)
-                T.copy(C_src, C_s)
                 T.fence_proxy_async()
                 T.sync_threads(barrier_id=_EPILOGUE_BARRIER_BASE + wg, arrive_count=128)
                 T.copy(C_s, C[row0, col0])
         else:
-            # The previous tile's TMA store may still be reading C_s.
-            T.sync_threads(barrier_id=_EPILOGUE_BARRIER_BASE + wg, arrive_count=128)
-            T.copy(C_src, C_s)
             T.fence_proxy_async()
             T.sync_threads(barrier_id=_EPILOGUE_BARRIER_BASE + wg, arrive_count=128)
             if a_has_group:
--- a/src/tileops/kernels/moe/sm90_gemm.py
+++ b/src/tileops/kernels/moe/sm90_gemm.py
@@ -1,0 +1,905 @@
+"""SM90 GEMM template, a TileLang port of DeepGEMM's ``sm90_bf16_gemm_impl``.
+
+One kernel body, specialised by ``SM90GemmSpec`` the way the CUDA template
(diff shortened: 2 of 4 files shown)


michaelwithu:
Withdrawing for now to rethink what this kernel should be solving before it lands; the branch stays for reference.

PR #1873, opened 2026-08-10
[Perf][Attention] Optimize FP8 GQA prefill pipeline

superAngGao:
## Summary

- consume FA3-compatible `[batch, heads_kv]` descales directly instead of materializing expanded per-head/per-tile scale tensors
- keep PV accumulation in the WGMMA-native register layout, rescale it in place, and defer the row-sum quad reduction for schedules shorter than 32 BN224 tiles
- retain the proven per-tile reduction path at and beyond the 32-tile liveness boundary
- remove the superseded PV accumulate helper and add repeated-launch coverage around the 28/32/36-tile boundary

## Validation

On an H200 locked to a 1500 MHz SM clock:

- `ruff check`: passed
- `ruff format --check`: passed
- `pytest -q tests/ops/attention/test_gqa_fp8.py`: 11 passed

The long-sequence tests use non-zero values and launch each boundary shape twice.

## Benchmark

Measured with the native-CUPTI V3 protocol: case-scoped trace with 1 ms margins, isolated discovery/timing captures, 10 warmups, 3 discovery iterations, 50 timed repeats summarized by their median, and kernel/MEMCPY/MEMSET activity enabled. All eight cases passed with zero dropped records or attribution errors.

| Shape | Output | TileOps | FA3 | TileOps / FA3 throughput |
|---|---:|---:|---:|---:|
| S=896, H=32 | FP16 | 0.0318 ms | 0.0285 ms | 89.50% |
| S=896, H=32 | BF16 | 0.0312 ms | 0.0285 ms | 91.33% |
| S=1792, H=32 | FP16 | 0.0959 ms | 0.0856 ms | 89.27% |
| S=1792, H=32 | BF16 | 0.0956 ms | 0.0855 ms | 89.47% |
| S=3584, H=64 | FP16 | 0.5975 ms | 0.5300 ms | 88.70% |
| S=3584, H=64 | BF16 | 0.5973 ms | 0.5300 ms | 88.74% |
| S=7168, H=64 | FP16 | 2.3601 ms | 2.0324 ms | 86.11% |
| S=7168, H=64 | BF16 | 2.3572 ms | 2.0348 ms | 86.32% |

PR #2104, opened 2026-09-08
[Feat][Attention] Add native FP8 Dense GQA

superAngGao:
## Summary

- register one native-FP8 implementation under `GroupedQueryAttentionDenseFwdOp`
- port the direct-descale BN224 fast schedule from #1873
- retain a native FP8 Tensor Core path for causal, rectangular/tail, window, softcap, and fused-RoPE calls
- keep FP8 inputs in FP8; there is no dequantize-to-FP16 attention fallback

This is stacked on #2095 and should be rebased after that PR merges. #1926 remains the broad design reference; this PR extracts only the Dense FP8 migration.

## Validation

Clock-locked H200 (SM 1500 MHz):

- `28 passed, 2 deselected` across the public Dense GQA and FP8 smoke suites
- FP8 correctness covers FP16/BF16 outputs, the BN224 fast path, causal/tail, rectangular window + softcap + custom scale, and fused interleaved RoPE
- existing FP16/BF16 prefill, decode, sliding-window, and RoPE cases remain green
- `ruff check` and `git diff --check` pass

## Performance

Native-CUPTI, cold L2 per repeat. The regression A/B interleaves the original #1873 wrapper and the new Dense Op for 3 rounds x 50 repeats.

| Output | S=896, H=32 | S=1792, H=32 | S=3584, H=64 | S=7168, H=64 |
|---|---:|---:|---:|---:|
| FP16 Dense Op vs #1873 | -0.095% | +0.061% | +0.015% | -0.047% |
| BF16 Dense Op vs #1873 | +0.154% | +0.154% | +0.066% | -0.042% |

The original BN224 region therefore shows no measurable regression after adding the wider feature family.

The general feature schedule was also measured on larger workloads against FA3:

| Workload | TileOps | FA3 | TileOps / FA3 throughput |
|---|---:|---:|---:|
| causal, Sq=1024, Skv=4096, H=32 | 0.4730 ms | 0.0901 ms | 19.1% |
| non-causal window (1024, 64) + softcap, Sq=2048, Skv=4096, H=32 | 0.3171 ms | 0.1106 ms | 34.9% |
| causal, Sq=Skv=4096, H=64 | 2.0398 ms | 0.3886 ms | 19.1% |

The general kernel is functionally broad but not yet performance-ready. This PR remains draft while that schedule is optimized.


superAngGao:
Closed as PR#2107 provides a better solution.

PR #2114, opened 2026-09-09
[BugFix][Attention] Align dense GQA decode split dispatch

Guan-jeans:
## Summary

- Fix Dense GQA decode so the tuned split ceiling is reduced to a runtime-valid, finite split tier: `{1, 2, 4, 8, 16, 32}`.
- Partition whole KV tiles inside the Dense split kernel and remove its `split_length` tensor input, so autotuning and runtime use the same partitioning rule.
- Route the measured plain-FP16 Qwen3 region through the BS1 kernel below `Skv=1024` and through the updated generic split kernel from `Skv=1024` onward.
- Add the finite split-capacity tier—not exact `Skv`—to the Dense Op cache key.
- Paged decode is intentionally unchanged; it has a different layout and dispatch policy and should be evaluated separately.

## Validation

- `pytest tests/ops/attention/test_gqa.py tests/ops/attention/test_gqa_decode_paged.py -m smoke`: **49 passed, 8 deselected** on H200.
- Ruff lint/format, manifest validation, and the repository-specific Python linters pass.
- `tests/ops/attention/test_gqa.py`: **23 -> 35** collected tests (**+12**). Existing parametrized coverage was adjusted so the dispatch/cache boundary is covered without adding a separate test function.

## Native-CUPTI benchmark

Clock-locked H200, public `GroupedQueryAttentionDenseFwdOp`, `B=1, H=32, Hkv=4, D=128, fp16`:

| `Skv` | Selected path | GPU busy | Full Op envelope |
| ---: | --- | ---: | ---: |
| 256 | BS1 | 6.19 us | 6.19 us |
| 512 | BS1 | 9.28 us | 9.28 us |
| 1024 | generic BN64/split16 | 5.73 us | 8.77 us |
| 4096 | generic BN64/split32 | 8.37 us | 10.11 us |
| 16384 | generic BN64/split32 | 16.32 us | 17.79 us |
| 32768 | generic BN64/split32 | 26.64 us | 27.95 us |

The previously reported `19.6 us -> 6.6 us (2.96x)` result is a forced generic-kernel busy-time comparison, not the latency change of the B=1 public Op, which previously selected the BS1 kernel. The full CUPTI envelope above is the relevant Op latency for the two-kernel split implementation.

RoPE and other head configurations retain their previous routing until they are measured independently.

### DIFF
--- a/src/tileops/kernels/attention/gqa_decode.py
+++ b/src/tileops/kernels/attention/gqa_decode.py
@@ -17,6 +17,20 @@ from .online_softmax import (
 
 __all__ = ["GQADecodeKernel"]
 
+
+def _effective_num_split(num_split: int, block_N: int, real_seqlen_kv: int) -> int:
+    """Split count the runtime can use for a KV extent of *real_seqlen_kv*.
+
+    The tuned or default ``num_split`` is only a ceiling: shrink it until every
+    split keeps at least one full KV tile. ``1`` means the sequence is too
+    short to split and the no-split kernel should run. Gating dispatch on the
+    tuned value itself (the old ``real_seqlen_kv < num_split * block_N`` test)
+    let a large tuned ``num_split`` push execution into the never-tuned
+    no-split kernel.
+    """
+    return max(1, min(num_split, real_seqlen_kv // block_N))
+
+
 # JIT kernel: no-split variant
 
 
@@ -262,7 +276,6 @@ def _gqa_decode_split_kernel(batch, heads, groups, dim, sm_scale, softcap, dtype
             V: T.Tensor(shape_v, dtype),
             glse: T.Tensor([batch, heads, num_split], dtype),
             Output_partial: T.Tensor(part_shape, dtype),
-            split_length: T.Tensor(num_split, "int32"),
         ):
             with T.Kernel(batch, heads // valid_block_H, num_split, threads=threads) as (
                 bx,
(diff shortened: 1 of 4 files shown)


superAngGao (commented):
Thanks for tracking down the tuned-split/runtime-dispatch mismatch and for moving the whole-tile partition into the kernel. We reproduced the intended generic-kernel improvement on a clock-locked H200, so the underlying issue and the performance opportunity are both real.

There is one integration issue with the current Dense GQA dispatch/cache design that I think we should address before merging.

The reported `B=1, H=32, Hkv=4, Skv=1024, D=128, fp16` workload does not normally reach `GQADecodeKernel`. `GroupedQueryAttentionDenseFwdOp` selects `gqa_dense_decode_bs1 -> GQADecodeBs1Kernel` for that region, and this PR does not change that kernel. Our native-CUPTI Op-level A/B measured approximately the same GPU busy time on base and HEAD (`6.22us` vs `6.21us`). Forcing the public Op to use the generic kernel reproduced the kernel-level improvement (`18.06us -> 5.57us` GPU busy), but the full two-kernel envelope was about `12us`, rather than a 3x Op-latency improvement.

On a shape that really uses the generic path (`B=2`), the improvement is real: GPU busy changed from `15.62us` to `6.91us`, while the complete CUPTI envelope changed from `15.62us` to `12.19us` because the split/combine gap remains part of Op latency.

The current implementation also makes `GQADecodeKernel.forward()` choose between two different GPU programs. This leaves the Op cache unaware of the actual strategy and creates two related risks:

1. `_effective_num_split()` can produce values such as 3, 7, or 15, although autotuning only measures `{1, 2, 4, 8, 16, 32}`. These become additional JIT specializations without a corresponding Op cache entry or measurement.
2. `num_split=1` is measured through `split_jit` (split + combine), but `forward()` executes the separate no-split kernel, so the tuner still measures one program and runtime executes another.

We suggest keeping the useful whole-tile partition and `split_length` removal, while moving the strategy choice to `GroupedQueryAttentionDenseFwdOp`:

```text
BS1-eligible and Skv < 1024 -> gqa_dense_decode_bs1
otherwise                   -> gqa_dense_decode
```

For the generic path, select only finite, measured split tiers (for example `split16` / `split32`) and include that strategy tier—not the exact `Skv`—in the Op cache key. This keeps dynamic sequence-length reuse, prevents first-call/cache-order dependence, and guarantees that the measured kernel is the one `forward()` executes.

Our H200 measurements support 1024 as an initial boundary for the tested plain-fp16 `H=32/Hkv=4/D=128` region: BS1 clearly wins at 256/512, while the updated generic path is at least as fast from 1024 through 32768. Other head ratios and RoPE can retain their current routing until measured separately.

Thank you again for identifying this issue. We are happy to help prepare the Op-level dispatch/cache patch on top of this PR.

superAngGao:
Benchmark correction for the table above:

The three main rows in the PR body are behavior-neutral regression checks: they select the same `split x16` path before and after this change. The reported `2.96x` row is a forced generic-kernel experiment and does not follow the public Op's normal B=1 dispatch.

| Public Op workload | Actual path | Base busy / envelope | HEAD busy / envelope | Op-envelope speedup |
| --- | --- | ---: | ---: | ---: |
| `B=1, Skv=1024, H=32, Hkv=4, D=128, fp16` | BS1 on both revisions | `6.22us / ~11-12us` | `6.21us / ~11-12us` | ~`1.00x` |
| `B=2, Skv=1024, H=32, Hkv=4, D=128, fp16` | generic no-split -> generic split | `15.62us / 15.62us` | `6.91us / 12.19us` | `1.28x` |

For diagnosis only:

| Forced path | Base busy / envelope | HEAD busy / envelope | Busy speedup | Envelope speedup |
| B=1 forced through generic | `18.06us / 18.06us` | `5.57us / ~12us` | `3.24x` | ~`1.5x` |

Thus the original `19.6us -> 6.6us (2.96x)` result demonstrates a real generic-kernel busy-time improvement, but it is not the latency improvement of the stated B=1 public-Op workload. For a multi-kernel implementation, the complete CUPTI envelope—including the split/combine gap—is the relevant Op latency.

superAngGao:
I pushed follow-up commit `3d5e7a07` to implement the Op-level integration discussed above:

- the measured plain-FP16 `B=1, H=32, Hkv=4, D=128` region uses BS1 below `Skv=1024` and the updated generic kernel from `Skv=1024` onward;
- RoPE and other head shapes retain their existing dispatch until they are measured independently;
- Dense runtime split counts are quantized to the finite `{1, 2, 4, 8, 16, 32}` candidates;
- the Dense Op cache keys on that finite split-capacity tier, not exact `Skv`;
- Paged decode keeps its own existing split policy and is not coupled to the new Dense policy.

Native-CUPTI H200 recheck of the public Op:

| `Skv` | Selected path | GPU busy | Full envelope |
| ---: | --- | ---: | ---: |
| 256 | BS1 | 6.19 us | 6.19 us |
| 512 | BS1 | 9.28 us | 9.28 us |
| 1024 | generic BN64/split16 | 5.73 us | 8.77 us |
| 4096 | generic BN64/split32 | 8.37 us | 10.11 us |
| 16384 | generic BN64/split32 | 16.32 us | 17.79 us |
| 32768 | generic BN64/split32 | 26.64 us | 27.95 us |

Validation: `49 passed, 8 deselected` for the Dense/Paged GQA smoke suites; Ruff lint/format and the repository-specific Python linters pass. The adjusted parametrization does not increase the test-node count relative to the previous PR head.

lcy-seso (approved):

PR #2123, opened 2026-09-11
[Feat][Kernels] Support pre-Hopper archs in GEMM, sparse MLA and FP8 BMM

Guan-jeans:
## Summary

- Add `GemmBasicKernel` (pipelined plain-T.gemm, sm80+, all four `(trans_a, trans_b)` layouts); `GemmFwdOp` routes to it wherever `get_sm_version() != 90`. `k` only needs `k * itemsize >= 4` — the backend zero-pads K tails — and M/N tails are guarded in the epilogue.
- Add `SparseMlaBasicKernel`: the WGMMA sparse-MLA decode math re-implemented on one homogeneous thread group with `T.Pipelined` + per-row `T.copy` KV gather; the DSA decode op dispatches to it off Hopper. Its default `block_i` is 32 below sm90 — the Hopper-sized 64 stages 230KB of SMEM for `kv_group=1` shapes and fails to launch on sm80's 163KB cap (verified on real hardware), while 32 halves the pipelined KV tiles to 148KB.
- `SharedExpertMLPKernel` mirrors `GemmFwdOp`'s dispatch when it builds its gate/up and down GEMMs (its ctor bypasses the op-level arch check). The default config stays Hopper-sized; its 3-stage 208KB footprint fails to launch on every pre-Hopper card (verified on real sm80, 163KB / 99KB caps), so pre-Hopper callers pass a smaller explicit config (e.g. a 2-stage 64x128 tile budgeted to 64KB) — documented at the config site.
- Open `BmmFp8Kernel` to sm89: the WS and persistent variants stay Hopper-only, the classic 3D-grid variant serves sm89 with shared-memory-budgeted configs (100KB cap).
- Correctness of the K-tail zero-padding and the `block_i` SMEM budget was verified on real sm80 (A800) and sm90 (H200) hardware.

## Test plan

- [x] `pytest` on the affected files, on real GPUs:
  - **H200 (sm90, CUDA_VISIBLE_DEVICES=4)**: `test_gemm.py` 29 passed / 2 failed — the same 2 (`test_explicit_structure_config_is_taken_verbatim`, `test_structure_routing_matches_test_ids`) fail identically on the untouched `main` baseline in the same container, so they are environment-specific (heuristic drift with the container's tilelang build), not regressions; the 2 new GemmBasicKernel cases pass. `test_deepseek_dsa_decode.py`, `test_bmm.py` smoke all pass. `test_moe_shared_fused_moe.py` 4 failed — the identical 4 fail on the `main` baseline too (a kernel-cache aliasing bug in the container's tilelang 0.1.11 build: `gating_output ... expected: 32, got: 16`), pre-existing and unrelated.
  - **A800 (sm80, CUDA_VISIBLE_DEVICES=0)**: `test_gemm.py` 19 passed / 7 failed — all 7 fail identically on the `main` baseline, and the change additionally fixes 7 cases that failed on the baseline (`smoke-fp16/bf16-square` and the four GEMV-boundary cases), a net +7 with zero new failures; the GEMV-boundary cases run on the fallback as-is (their k values all clear the `k * itemsize >= 4` vectorized-load floor). `test_deepseek_dsa_decode.py` smoke passes (after the `block_i=32` shared-memory fix). `test_bmm.py` 10 failed / 11 passed — the identical set fails on the baseline (sm80 has no FP8 tensor cores; `BmmKernel` small-batch cases fail on the baseline too). `test_moe_shared_fused_moe.py` fails with the same environment-specific cache bug as on H200; the shared-expert `GemmBasicKernel` GEMMs were exercised on sm80 with a 64x128 2-stage tile (the config pre-Hopper callers pass explicitly) and matched the reference.
- [x] `pre-commit run --all-files` — locally ran ruff 0.14.13 check + format (matching the pinned rev), `tilelang_idioms_lint`, `shipped_refs_lint`, `op_docstrings_lint`, and a trailing-whitespace scan on the touched files; the remaining hooks (gitleaks, mdformat, codespell) need the CI run.

## Test node delta

`python scripts/test_node_delta.py --base upstream/main` cannot run on the authoring machine (no torch). Manual count over the diff:

| File | Delta | New cases |
| ---- | ----- | --------- |
| tests/ops/test_gemm.py | +2 | `test_gemm_basic_kernel_k_tail_padding` (k=2/8/24 bf16 on the zero-padded K tail, smoke), `test_gemm_basic_kernel_rejects_narrow_k` (k=1 bf16 refused, smoke) |

`tests/ops/test_moe_shared_fused_moe.py` is unchanged from `main`.

## Regression

A pre-Hopper device previously failed kernel selection (`GemmKernel` / `SparseMlaKernel` are `[90]`) or hit a raw nvcc/PTX error at JIT time with no fallback — on A800 the `main` baseline fails 14 of 26 collected smoke cases in `test_gemm.py`; this change turns 7 of those green (including the four GEMV-boundary cases, which now run through the fallback unconditionally) and adds the two K-tail guard cases.

### DIFF
--- a/src/tileops/kernels/__init__.py
+++ b/src/tileops/kernels/__init__.py
@@ -19,6 +19,7 @@ from .attention import (
     NSACmpFwdVarlenKernel,
     NSAFwdVarlenKernel,
     NSATopkVarlenKernel,
+    SparseMlaBasicKernel,
     SparseMlaKernel,
 )
 from .convolution import (
@@ -40,6 +41,7 @@ from .fp8_quant import FP8QuantKernel
 from .gemm import (
     BmmFp8Kernel,
     BmmKernel,
+    GemmBasicKernel,
     GemmFp8BlockScaledKernel,
     GemmFp8EpilogueKernel,
     GemmKernel,
@@ -163,6 +165,7 @@ __all__ = [
     "GatedDeltaNetFwdKernel",
     "GatedDeltaNetFwdProductionKernel",
     "GatedDeltaNetPrefillFwdKernel",
+    "GemmBasicKernel",
     "GemmFp8BlockScaledKernel",
     "GemmFp8EpilogueKernel",
     "GemmKernel",
@@ -199,6 +202,7 @@ __all__ = [
     "RopeNonNeoxKernel",
     "RopeYarnKernel",
     "SmallBatchGemmKernel",
+    "SparseMlaBasicKernel",
     "SparseMlaKernel",
     "TopkSelectorKernel",
     "UnaryKernel",
--- a/src/tileops/kernels/attention/__init__.py
+++ b/src/tileops/kernels/attention/__init__.py
@@ -1,4 +1,4 @@
-from .deepseek_dsa_decode import SparseMlaKernel
+from .deepseek_dsa_decode import SparseMlaBasicKernel, SparseMlaKernel
 from .deepseek_mla_decode import MLADecodeWsKernel
 from .deepseek_nsa_cmp_fwd import NSACmpFwdVarlenKernel
 from .deepseek_nsa_fwd import NSAFwdVarlenKernel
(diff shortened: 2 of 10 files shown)


lcy-seso (approved):
Thank you for this contribution.

PR #2135, opened 2026-09-15
[Refactor][LinearAttention] Define unified Gated DeltaNet Op

superAngGao:
## Summary

- replace seven legacy Gated DeltaNet public Ops with the spec-only `GatedDeltaNetFwdOp`
- define one FLA-aligned BTHD contract for equal-length prefill, packed-varlen prefill, and decode
- support optional initial state, both FLA state layouts, and fused gate/beta/QK-normalization inputs
- always return the FP32 recurrent state for inference continuation
- remove the superseded tests, benchmarks, workloads, and roofline entries while retaining kernels for staged migration in #2143

This PR establishes the public Op boundary only. It intentionally adds no BUILTIN kernel or benchmark. The retained kernels are marked as staged rollout until they are migrated or deleted.

## Validation

- strict targeted manifest validation passed
- Gated DeltaNet Op, Op base, backend seam, and manifest suites: **263 passed**
- Ruff and `git diff --check` passed


src/tileops/manifest/linear_attention.yaml
```diff
@@ -3,167 +3,58 @@
...
-  # Inference prefill for Gated DeltaNet. This is intentionally separate from
-  # GatedDeltaNetBHTDFwdOp, which exposes training-forward artifacts needed by
-  # backward. Prefill returns only the prompt output and the recurrent state
-  # consumed by decode.
-  ref_api: "none"
+GatedDeltaNetFwdOp:
+  # Dense inference contract shared by prefill and single-token decode. The
+  # implementation dispatches on T; the public interface does not expose a
+  # separate decode Op.
+  ref_api: "fla.ops.gated_delta_rule.chunk_gated_delta_rule"
```
lcy-seso, line 10: `inspect.signature` on the pinned fla 0.5.2, in the CI runner image, gives `chunk_gated_delta_rule(q, k, v, g, beta, scale, initial_state, output_final_state, use_qk_l2norm_in_kernel, use_beta_sigmoid_in_kernel, allow_neg_eigval, state_v_first, cu_seqlens, cu_seqlens_cpu, cp_context, **kwargs)`, with `use_gate_in_kernel`, `A_log` and `dt_bias` arriving through `**kwargs`.

manifest-spec.md: names match the reference verbatim, and every supported parameter is declared.

| this entry | reference |
| --- | --- |
| `use_qk_l2norm` | `use_qk_l2norm_in_kernel` |
| `use_gate_activation` | `use_gate_in_kernel` |
| `use_beta_sigmoid` | `use_beta_sigmoid_in_kernel` |
| absent | `output_final_state`, `state_v_first`, `cu_seqlens`, `cu_seqlens_cpu` |

Without `output_final_state` the fp32 state is a mandatory write on every prefill call; without `cu_seqlens` varlen is out of the contract. `spec-only` skips the L1 parity check, so nothing catches this at the status flip either.

src/tileops/manifest/linear_attention.yaml
```diff
@@ -3,167 +3,58 @@
...
     inputs:
-      q: {dtype: "float16 | bfloat16 | float32", shape: "[B, S, H, DK]"}
-      k: {dtype: "same_as(q)", shape: "[B, S, H, DK]"}
-      v: {dtype: "same_as(q)", shape: "[B, S, H, DV]"}
-      g: {dtype: "same_as(q)", shape: "[B, S, H]"}
-      beta: {dtype: "same_as(q)", shape: "[B, S, H]"}
+      q: {dtype: "float16 | bfloat16", shape: "[B, T, H, K]"}
+      k: {dtype: "same_as(q)", shape: "[B, T, H, K]"}
+      v: {dtype: "same_as(q)", shape: "[B, T, HV, V]"}
+      g: {dtype: "same_as(q)", shape: "[B, T, HV]"}
```
lcy-seso, line 19: `g` is log-space decay when `use_gate_activation` is off and the raw pre-activation gate when it is on. Neither this entry nor the class docstring says so, and the transforms `use_beta_sigmoid` and `allow_neg_eigval` name are undefined too. The contract is this entry's only deliverable.

src/tileops/manifest/linear_attention.yaml
```diff
@@ -3,167 +3,58 @@
...
+      beta: {dtype: "same_as(q)", shape: "[B, T, HV]"}
+      initial_state: {dtype: "float32", shape: "[B, HV, V, K]", optional: true}
```
lcy-seso: Value-major is the reference's `state_v_first`, which defaults to `False` and which fla's own `GatedDeltaNet` layer passes as `True`. The entry names neither the option nor the fact that it is pinned, so this reads as a TileOPs invention. Every retained GDN kernel produces key-major `[B, H, DK, DV]`.

src/tileops/manifest/linear_attention.yaml
```diff
@@ -3,167 +3,58 @@
...
 
   roofline:
-    func: "tileops.perf.formulas.gated_deltanet_prefill_fwd_roofline"
+    func: "tileops.perf.formulas.gated_deltanet_fwd_roofline"
 
   source:
-    kernel: tileops/kernels/linear_attention/gated_deltanet/gated_deltanet_prefill.py
-    kernel_map:
-      GatedDeltaNetPrefillFwdKernel: GatedDeltaNetPrefillFwdKernel
+    kernel: []
```
lcy-seso: manifest-spec.md: a kernel-less op sets `source.kernel` to `source.op`. `[]` passes the schema but reads as "no kernel file exists".

It matters because the `kernel_map` went with it: seven kernel classes are now reachable only through the `tileops.kernels` re-exports, and nothing else references them — the prefill, backward and three decode kernels have no test left at all. code-style.md defines `FIXME(staged-rollout)` for exactly this; one per orphaned file makes them greppable when the migration PRs start.

src/tileops/perf/formulas.py
```diff
@@ -308,15 +216,25 @@ def deltanet_decode_roofline(op: Any | None = None, **kwargs: Any) -> tuple[int,
...
+    batch, seq_len, heads, dim_k = data["q_shape"]
+    _batch, _seq_len, value_heads, dim_v = data["v_shape"]
     elem_bytes = _dtype_itemsize(data.get("dtype", data.get("dtypes", "float16")))
 
-    flops = 2 * batch * heads * (3 * dim_k * dim_v + dim_k)
-    nbytes = batch * heads * (2 * dim_k + 2 * dim_v + 2 + 2 * dim_k * dim_v)
-    return int(flops), int(nbytes * elem_bytes)
+    # Per recurrent head and token: two state matvecs, one state outer-product
+    # update, and the q-k dot product.
+    flops = 2 * batch * seq_len * value_heads * (3 * dim_k * dim_v + dim_k)
```
lcy-seso: The `+ dim_k` is commented as "the q-k dot product", but the gated delta rule computes no q-k dot product — the per-token work is `k · S`, the rank-1 update and `q · S`, already counted as `3 * dim_k * dim_v`.

The per-token gate decay `S *= exp(g)` costs `B * T * HV * K * V` multiplies and is not counted. At V=128 that is two orders of magnitude above the `dim_k` term, so this is not the lower bound the docstring claims.

src/tileops/ops/linear_attention/gated_deltanet.py
```diff
@@ -1,230 +1,70 @@
...
-
-    Head-major (BHTD) inputs: ``q/k [B, H, S, DK]``, ``v [B, H, S, DV]``,
-    ``g/beta [B, H, S]``. ``GatedDeltaNetBwdOp`` consumes the ``S`` this returns,
-    in the same layout. Token-major callers want ``GatedDeltaNetBTHDFwdOp``.
+    ``q`` and ``k`` use ``[B, T, H, K]``. ``v``, ``g``, ``beta`` and the
+    output use ``HV`` recurrent heads, where ``HV`` is a multiple of ``H``.
+    The recurrent state is always FP32 and value-major: ``[B, HV, V, K]``.
 
+    One interface covers prefill and decode. The target callable selects the
+    execution path from the current sequence length; in particular, ``T == 1``
```
lcy-seso: `default_kernel_map` is `{}` and `source.kernel` is empty, so an in-tree call raises `OpNotAvailableError` — only an external target selects a path. Callers read this docstring; say that no in-tree implementation exists yet.

src/tileops/ops/linear_attention/gated_deltanet.py
```diff
@@ -233,953 +73,104 @@ def _infer_output_shapes(
...
-    ) -> Kernel:
-        key = (
+        if q.ndim != 4 or k.shape != q.shape:
+            raise ValueError("q and k must have the same [B, T, H, K] shape")
+        if v.ndim != 4 or v.shape[:2] != q.shape[:2]:
+            raise ValueError("v must have shape [B, T, HV, V]")
+
+        batch, seq_len, heads, dim_k = q.shape
+        value_heads, dim_v = v.shape[2:]
+        if heads == 0 or value_heads % heads != 0:
```
lcy-seso: `heads == 0` raises "HV must be divisible by H", which does not describe it.

tests/ops/test_family_dispatch.py
```diff
@@ -118,47 +114,6 @@ def test_deltanet_decode_dispatch(
     assert op.select_kernel_key(DELTANET_DECODE_KEYS, call) == expected
 
 
-# --- Gated DeltaNet decode: the same shape, and the raw-CUDA kernel additionally
-# declines when the caller asked to autotune, having no knobs to tune.
-
-_GATED_ROWS = [
```
lcy-seso, line 124: This was the only written record of the decode selection policy: fp32 → `GatedDeltaNetDecodeFP32Kernel`; bf16 with `dim_k == dim_v == 128` on sm90 → `GatedDeltaNetDecodeRawCudaFlaStyleKernel`; `tune=True` declines the raw kernel; otherwise `GatedDeltaNetDecodeKernel`. Please copy the four rules into the follow-up issue.

lcy-seso (commented):
The boundary is right: one BTHD op in place of seven. Findings are what the spec-only entry leaves unwritten, plus one wrong term in the roofline. See inline.

src/tileops/manifest/linear_attention.yaml
superAngGao: Aligned the supported names with FLA (use_qk_l2norm_in_kernel, use_gate_in_kernel, and use_beta_sigmoid_in_kernel) and added state_v_first, cu_seqlens, and cu_seqlens_cpu. This inference contract intentionally pins output_final_state=True because manifest outputs cannot vary with a switch and serving callers need the recurrent state; the deviation is now explicit in the manifest and class docstring. cp_context remains outside this non-distributed Op boundary.

src/tileops/manifest/linear_attention.yaml
superAngGao: Defined the input semantics in the public class docstring: precomputed log-space g by default, raw gate plus A_log/dt_bias under use_gate_in_kernel, post-sigmoid beta by default, and the exact sigmoid / 2 * sigmoid transforms.

src/tileops/manifest/linear_attention.yaml
superAngGao: Added the FLA state_v_first parameter with its reference default (False) and declared both key-major [N, HV, K, V] and value-major [N, HV, V, K] layouts. The migration tracker (#2143) records that retained kernels are key-major and that value-major must become a kernel specialization rather than a public-boundary transpose.

src/tileops/manifest/linear_attention.yaml
superAngGao: Changed the kernel-less entry to point source.kernel at source.op, added FIXME(staged-rollout) ownership markers to each orphaned GDN kernel file, and opened #2143 with the retained-kernel inventory and cleanup conditions.

src/tileops/perf/formulas.py
superAngGao: Fixed the formula: the three contraction/update terms remain six FLOPs per state element, and the missing state-decay multiply adds one, for 7 * K * V per recurrent head and token. Varlen state traffic now uses len(cu_seqlens) - 1 states.

src/tileops/ops/linear_attention/gated_deltanet.py
superAngGao: The class docstring now explicitly says that there is no in-tree implementation yet and calls require an external target until the retained kernels are migrated.

src/tileops/ops/linear_attention/gated_deltanet.py
superAngGao: Split the validation: H <= 0 now reports H must be positive; divisibility has its own HV must be divisible by H error.

tests/ops/test_family_dispatch.py
superAngGao: Copied all four decode selection rules into the kernel migration tracker #2143 and required restoring them as Op-level dispatch tests when decode is migrated.

PR #2140, opened 2026-09-15
[Perf][MoE] Add indexed small-route expert path

michaelwithu:
## Summary

- add an SM90 indexed expert-MLP path for up to eight routed tokens, reading expert weights directly from post-routing top-k ids
- compute active-expert reuse on device and dispatch at routes / active_experts >= 8 / 7 between direct M=1 tiles and same-expert M16 tiles
- fuse gate/up activation, keep route-major intermediates, and perform weighted top-k reduction without permute/unpermute
- preserve the public FusedMoEExpertsFwdOp interface and retain the tight grouped pipeline as the shape and non-SM90 fallback
- fix the SharedFusedMoE vLLM baseline to use correction-bias routing and validate outputs before timing

## Validation

- `pytest -q tests/ops/test_fused_moe_experts.py -m smoke`: 30 passed
- `pytest -q tests/ops/test_moe_compile.py -m smoke`: 12 passed
- CUDA Graph replay test changes top-k ids from dispersed to reused after capture
- BF16 and FP16 correctness cover dispersed and reused routes
- pre-commit passes on all changed files

## H200 results

Kimi K2 SharedFusedMoE `T=1, E=384, K=8, H=7168, F=2048, shared_F=18432, BF16`:

| implementation | device busy | GPU span | kernels |
|---|---:|---:|---:|
| TileOps | 0.4200 ms | 0.4432 ms | 7 |
| vLLM | 0.4246 ms | 0.8582 ms | 13 |

Reuse boundary for the routed expert pipeline:

| routes / active experts | T=2 vs vLLM | T=4 vs vLLM |
|---:|---:|---:|
| 1.143 | 0.340 / 0.353 ms | 0.660 / 0.655 ms |
| 1.333 | 0.327 / 0.352 ms | 0.567 / 0.653 ms |
| 2.000+ | 0.246 / 0.353 ms | 0.325 / 0.653 ms |

Measurements use CUPTI device-busy timing with L2 flushing. The installed vLLM reports that no H200-specific E=384/N=2048 config file is present.


michaelwithu:
Closing this early submission. The small-route dispatch still has unresolved performance regressions at the low-reuse boundary and needs more optimization and validation before review.

Issue #2143, opened 2026-09-16
[Refactor][LinearAttention] Migrate Gated DeltaNet kernels to the unified inference Op


superAngGao:
## Context

PR #2135 replaces the legacy Gated DeltaNet public Ops with one inference-oriented `GatedDeltaNetFwdOp`. The new contract follows FLA's BTHD interface across equal-length prefill, packed-varlen prefill, and single-token decode, while always returning the FP32 recurrent state.

The old kernels remain exported temporarily, but no in-tree Op, test, or benchmark owns them after #2135. This issue tracks migrating the useful implementations behind the new Op and deleting the remainder.

## Public contract to implement

- `q`, `k`: `[B, T, H, K]`
- `v`: `[B, T, HV, V]`, with `HV % H == 0`
- `g`, `beta`: `[B, T, HV]`
- optional FP32 `initial_state`
- optional `cu_seqlens` and `cu_seqlens_cpu` for packed varlen (`B == 1`)
- key-major state `[N, HV, K, V]` by default
- value-major state `[N, HV, V, K]` when `state_v_first=True`
- FLA-compatible `use_qk_l2norm_in_kernel`, `use_gate_in_kernel`, `use_beta_sigmoid_in_kernel`, and `allow_neg_eigval` semantics
- always return `(o, final_state)`; this pins FLA's `output_final_state=True` for the inference API

## Retained implementation inventory

- `GatedDeltaNetPrefillFwdKernel`
- `GatedDeltaNetFwdKernel`
- `GatedDeltaNetFwdProductionKernel`
- `GatedDeltaNetDecodeKernel`
- `GatedDeltaNetDecodeFP32Kernel`
- `GatedDeltaNetDecodeRawCudaFlaStyleKernel`
- `GatedDeltaNetBwdKernel`

The backward kernel is not part of the inference Op migration. Retain it only if a later training Op owns it; otherwise delete it during final cleanup.

## Decode dispatch policy that must be preserved

The removed family-dispatch test was the only executable record of the legacy decode policy:

1. FP32 selects `GatedDeltaNetDecodeFP32Kernel`.
2. BF16 with `K == V == 128` on SM90 selects `GatedDeltaNetDecodeRawCudaFlaStyleKernel`.
3. `tune=True` declines the raw-CUDA specialization because it has no tunable knobs.
4. Remaining supported decode calls select `GatedDeltaNetDecodeKernel`.

[...]

PR #2144, opened 2026-09-16
[Refactor][LinearAttention] Migrate dense Gated DeltaNet prefill

superAngGao:
Part of #2143.

## Summary

- wire the Hopper BTHD dense-prefill pipeline into `GatedDeltaNetFwdOp`
- place the prefill implementation under `gated_deltanet/prefill/`, split into dense entry, common lowering/cache helpers, preparation/state-correction, and fused-forward stages; a future decode migration will use a sibling `decode/` package
- preserve the unified Op ABI and return the native FP32 final state
- add public-API correctness coverage and manifest-driven FLA benchmarks

This stage intentionally supports equal-length, zero-state prefill with `H == HV`, `K == V == 128`, precomputed gate/beta values, and sequence lengths divisible by 64. Decode, initial state, GVA, fused input transforms, packed varlen, and `state_v_first=True` remain follow-ups in #2143. The manifest stays `spec-only` until the declared contract is fully covered.

This is also an explicit pre-release kernel cleanup: `GatedDeltaNetPrefillFwdKernel` is removed from `tileops.kernels` rather than retained behind an incompatible constructor alias. The replacement specialization is `GatedDeltaNetDensePrefillFwdKernel`.

## Validation

H200, locked SM clock at 1500 MHz:

- Gated DeltaNet tests: `5 passed`, covering FP16, BF16, and a forced partition/warmup/state-correction path
- all 6 manifest benchmark cases passed one-shot TileOps-vs-FLA output/final-state checks and timing
- `scripts/validate_manifest.py --check-op GatedDeltaNetFwdOp`: passed (advisory synthetic-input warnings only)
- `scripts/test_node_delta.py --base origin/main tests/ops/test_gated_deltanet.py`: `2 -> 5` nodes (`+3`, `+150.0%`)
- benchmark-contract tests: `36 passed, 3 skipped`
- ruff and diff checks passed
- BF16 numerical parity with FLA:
  - output max abs error: `1.37e-4`
  - FP32 final-state max abs error: `2.31e-3`

Native-CUPTI full-op envelope:

| Workload | TileOps | FLA | Speedup |
| --- | ---: | ---: | ---: |
| B1, S=4096, H=16, D=128, BF16 | 0.3855 ms | 0.5435 ms | 1.41x |
| B1, S=32768, H=16, D=128, BF16 | 0.5224 ms | 1.5384 ms | 2.94x |


src/tileops/kernels/linear_attention/gated_deltanet/_dense_prefill_kernels.py
```diff
@@ -0,0 +1,2205 @@
...
+from collections.abc import Callable
+from typing import Any
+
+import tilelang
+import tilelang.language as T
+import torch
+
+from tileops.kernels.constants import LOG2E
+
+MULTI_PROCESSOR_COUNT = torch.cuda.get_device_properties().multi_processor_count
```
lcy-seso: - `get_device_properties()` reads the current CUDA device, which may differ from the input tensor's device. In multi-GPU or heterogeneous-GPU processes, this can permanently derive `TARGET_NUM_CTAS` from the wrong device. Please derive it from `k.device.index` or the kernel's `device_index`.

src/tileops/kernels/linear_attention/gated_deltanet/_dense_prefill_kernels.py
```diff
@@ -0,0 +1,2205 @@
...
+            a,
+            b,
+            c,
+            transpose_A=transpose_A,
+            transpose_B=transpose_B,
+            clear_accum=clear_accum,
+        )
+
+    T.gemm_v1 = _gemm_v1  # type: ignore[attr-defined]
+
```
lcy-seso: - `install_gemm_v1()` mutates the global `tilelang.language` namespace at import time, so one GDN call can affect unrelated kernels in the same process. Please scope this lowering to this implementation, or at least avoid an import-time global monkey patch.

src/tileops/kernels/linear_attention/gated_deltanet/dense_prefill.py
```diff
@@ -0,0 +1,280 @@
...
+        cp_seq_map=cp_seq_map,
+        raw_cu_seqlens=raw_cu_seqlens,
+        chunk_size=chunk_size,
+        state_head_first=output_states,
+        chunks_per_sequence=seq_len // chunk_size if output_states else 0,
+    )
+    return o.reshape(batch, seq_len, head, v.shape[-1]), states, final_state, g_cum
+
+
+class GatedDeltaNetDensePrefillKernel(Kernel):
```
lcy-seso: - Repository naming rules require kernels to include a direction suffix. Please rename this to `GatedDeltaNetDensePrefillFwdKernel` and update the manifest and exports before this becomes another public name to migrate.

tests/ops/test_gated_deltanet.py
```diff
@@ -19,6 +25,22 @@ def isolated_registry():
     registry.restore(state)
 
 
+@pytest.mark.skipif(
+    not torch.cuda.is_available() or torch.cuda.get_device_capability()[0] != 9,
+    reason="the migrated dense-prefill specialization requires SM90",
+)
+def test_gated_deltanet_dense_prefill_matches_reference() -> None:
+    torch.manual_seed(42)
+    test = GatedDeltaNetFwdTest(1, 64, 2, 128, torch.bfloat16)
```
lcy-seso: - This case has only one chunk, so `has_partition` is always false and the migrated warmup/state-correction/partition path is not exercised. Please lower `MAX_LOCAL_CHUNKS` via the environment and add a correctness case that actually enters the partitioned path.

src/tileops/manifest/linear_attention.yaml
```diff
@@ -59,16 +59,21 @@ GatedDeltaNetFwdOp:
       - "state_v_first or final_state.shape[1:] == (HV, K, V)"
       - "not state_v_first or final_state.shape[1:] == (HV, V, K)"
 
-  workloads: []
+  workloads:
+  - {q_shape: [1, 4096, 16, 128], k_shape: [1, 4096, 16, 128], v_shape: [1, 4096, 16, 128], g_shape: [1, 4096, 16], beta_shape: [1, 4096, 16], dtypes: [float16, bfloat16], label: "dense-prefill-b1-s4k-h16-d128"}
+  - {q_shape: [1, 32768, 16, 128], k_shape: [1, 32768, 16, 128], v_shape: [1, 32768, 16, 128], g_shape: [1, 32768, 16], beta_shape: [1, 32768, 16], dtypes: [float16, bfloat16], label: "dense-prefill-b1-s32k-h16-d128"}
```
lcy-seso: - Benchmark conventions require at least three shapes and a non-power-of-two case when supported. The current 4096/32768 workloads are both powers of two. Please add a realistic non-power-of-two workload and use scenario-based labels rather than parameter strings.

src/tileops/kernels/linear_attention/gated_deltanet/_dense_prefill_kernels.py
```diff
@@ -0,0 +1,2205 @@
+# Copyright (c) 2026 The Qwen team, Alibaba Group.
+# Licensed under the MIT License.
+# Adapted and modified for TileOps GatedDeltaNet prefill integration.
+"""Private TileLang stages used by the dense Gated DeltaNet prefill kernel."""
```
lcy-seso: - This 2.2K-line module now combines GEMM lowering, caching, sequence metadata, warmup/correction, prepare-H, and fused forward stages; it is also consumed by the retained training-forward path despite being named for dense prefill. Please keep the consolidation smaller and split it by responsibility into a few private modules.

src/tileops/kernels/__init__.py
```diff
@@ -74,9 +74,9 @@
         GatedDeltaNetDecodeFP32Kernel,
         GatedDeltaNetDecodeKernel,
         GatedDeltaNetDecodeRawCudaFlaStyleKernel,
+        GatedDeltaNetDensePrefillKernel,
         GatedDeltaNetFwdKernel,
         GatedDeltaNetFwdProductionKernel,
```
lcy-seso, line 80: - This replaces an exported kernel symbol with a narrower class and a different constructor, so downstream imports of `GatedDeltaNetPrefillFwdKernel` will break. Please confirm that `tileops.kernels` is not a compatibility surface; otherwise add a deprecation/compatibility path or call out the breaking change explicitly.

lcy-seso:
- This branch is currently conflicting with `upstream/main`. Please rebase onto the latest `upstream/main` and resolve the conflicts before further review.
- The manifest entry remains `spec-only`, so the normal validator does not exercise the Op's L2-L4 parity checks. Please run and report `scripts/validate_manifest.py --check-op GatedDeltaNetFwdOp`.
- This PR changes tests, but the required `scripts/test_node_delta.py` report is missing from the PR description.
- The only currently visible GitHub check is `apply-label`; please ensure the actual test, lint, and manifest-validation jobs run before merge.

src/tileops/kernels/linear_attention/gated_deltanet/_dense_prefill_kernels.py
superAngGao: Fixed in 4b10ad90. The module-level current-device probe is gone; `fused_gdr_fwd()` now derives the CTA target from `k.device.index` at runtime.

src/tileops/kernels/linear_attention/gated_deltanet/_dense_prefill_kernels.py
superAngGao: Fixed in 4b10ad90. `_gemm_v1` is now a GDN-local helper called directly by these stages; importing GDN no longer mutates `tilelang.language`.

src/tileops/kernels/linear_attention/gated_deltanet/dense_prefill.py
superAngGao: Fixed in 4b10ad90. Renamed the class and all manifest/export/Op references to `GatedDeltaNetDensePrefillFwdKernel`.

tests/ops/test_gated_deltanet.py
superAngGao: Added a forced-partition correctness case (`MAX_LOCAL_CHUNKS=4`, S=512). It exercises warmup, state correction, prepare-H, and partitioned forward together; the H200 targeted suite is now 4/4 passing.

src/tileops/manifest/linear_attention.yaml
superAngGao: Fixed in 4b10ad90. The manifest now has three scenario-labeled workloads, including the non-power-of-two S=12288 case.

src/tileops/kernels/linear_attention/gated_deltanet/_dense_prefill_kernels.py
superAngGao: Fixed in 4b10ad90. The implementation is now flat under `gated_deltanet/` but split by responsibility into `_dense_prefill_common.py`, `_dense_prefill_prepare.py`, and `_dense_prefill_forward.py`; the small `_dense_prefill_kernels.py` file is only a compatibility facade for the retained training-forward consumer.

src/tileops/kernels/__init__.py
superAngGao: Agreed that an alias with an incompatible constructor would be misleading. I kept the intentional pre-release cleanup and explicitly called the breaking removal out in the PR body; the replacement exported symbol is `GatedDeltaNetDensePrefillFwdKernel`.

superAngGao:
Thanks for the review. Addressed on the rebased branch:

- rebased onto current `upstream/main`; GitHub now reports the PR mergeable
- ran `scripts/validate_manifest.py --check-op GatedDeltaNetFwdOp` successfully (only advisory synthetic-input precondition warnings)
- added the requested test-node report to the PR body: the new test file contributes 4 nodes (`0 -> 4`)
- added a forced partition-path correctness test; the H200 targeted suite is `4 passed`
- pushed the implementation changes discussed in the inline threads and resolved each thread

The push has triggered the full CI suite. Manifest, benchmark-contract, and compile-contract jobs have already passed; the initial pre-commit failure was only `ruff format` changing the newly split private modules, and that formatting fix is now pushed in `7479cd1d` for a clean rerun.

superAngGao:
Follow-up validation after rebasing onto `0bedc999`: the targeted GDN suite also passes on the restored 183 host (`4 passed` in 92.04s on a clock-locked H200, GPU 0 at 1500 MHz), including the forced partition/warmup/state-correction case. GitHub reports the branch mergeable; only the latest `gpu-smoke` rerun is still pending.

superAngGao:
Follow-up organization change: the implementation now lives under `gated_deltanet/prefill/` (`dense.py`, `common.py`, `prepare.py`, and `forward.py`). This keeps the responsibility split from the review while making the future decode migration a symmetric sibling `gated_deltanet/decode/` package. The temporary `_dense_prefill_kernels.py` facade has been removed. Ruff passes and the H200 targeted suite remains `4 passed`.

tests/ops/test_gated_deltanet.py
```diff
@@ -19,6 +25,36 @@ def isolated_registry():
...
+    inputs = test.gen_inputs()
+    op = GatedDeltaNetFwdOp()
+    test.check(op, *inputs, atol=5e-2, rtol=5e-2)
```
lcy-seso: The test contract requires coverage of both supported dtypes, but both real-kernel cases currently exercise BF16 only. Please add an FP16 correctness case and use the repository tolerances (FP16 `1e-3`, BF16 `1.6e-2`) instead of `5e-2`. I cold-ran this PR on the official `cu132-torch2.13-tl-afcebed1-dev` image and also probed FP16/BF16, B>1, H=16, custom scale, and partitioned execution; observed max errors fit the standard tolerances, so the wider threshold is not needed.

benchmarks/ops/bench_gated_deltanet.py
```diff
@@ -0,0 +1,42 @@
...
+    dim: int,
+    dtype: torch.dtype,
+) -> None:
+    workload = GatedDeltaNetFwdWorkload(batch, seq_len, heads, dim, dtype)
+    inputs = workload.gen_inputs()
+    op = GatedDeltaNetFwdOp()
+
+    def fla(*args):
+        return chunk_gated_delta_rule(*args, output_final_state=True)
+
```
lcy-seso, line 42: `ManifestBenchmark.compare()` only times these callables; it does not verify their outputs. Please run a one-shot TileOps-vs-FLA correctness assertion before publishing timings, with dtype-appropriate tolerances. These manifest rows (H=16 and long/partitioned sequences) cover materially different regions from the current H=2 unit cases, so a bad row could otherwise still produce a benchmark report.

src/tileops/kernels/linear_attention/gated_deltanet/prefill/prepare.py
```diff
@@ -0,0 +1,1272 @@
...
+                seq_end_idx = seq_map_r2c[bb + 1]
+                num_iters = seq_end_idx - seq_start_idx
+
+                h_fragment = T.alloc_fragment((DK, block_DV), dtype=accum_dtype)
+                T.copy(
+                    raw_h0[bb, bh, 0:DK, bv * block_DV : (bv + 1) * block_DV],
+                    h_fragment,
+                )
+                # The loop never writes the partition a sequence starts on, and a
+                # fragment-to-global T.copy before it is dropped 鈥?the fragment's
```
lcy-seso: This em dash was corrupted during the module consolidation (`dropped 鈥?the`). Please restore the original text (`dropped — the`).

lcy-seso (commented):
One PR-description correction remains: against the current `upstream/main`, `scripts/test_node_delta.py` reports `tests/ops/test_gated_deltanet.py: 2 -> 4 (+2, +100.0%)`, not `0 -> 4` / a new file. Please update the validation section accordingly.

tests/ops/test_gated_deltanet.py
superAngGao: Fixed in 454906a6. The real-kernel test now covers FP16 and BF16 with the repository-standard 1e-3 and 1.6e-2 tolerances; the forced partition case uses 1.6e-2. The updated suite passes 5/5 on H200.

benchmarks/ops/bench_gated_deltanet.py
superAngGao: Fixed in 454906a6. The benchmark now calls assert_matches_reference(op, fla, ...) with reference_tolerance(dtype) before timing, which checks both the output and FP32 final state. All six manifest cases passed on H200.

src/tileops/kernels/linear_attention/gated_deltanet/prefill/prepare.py
superAngGao: Fixed in 454906a6; the original em dash is restored.

superAngGao:
Thanks for the careful cold-run review. Commit 454906a6 addresses the three inline findings. I also corrected the PR validation section against current main: after adding FP16 coverage, the file is now 2 -> 5 nodes (+3, +150.0%), rather than being described as a new 0 -> 4 file.

lcy-seso (approved):

PR #2157, opened 2026-09-19
[Perf][GEMM] Chunk the grouped GEMM epilogue, and let a caller state a padded row layout

michaelwithu:
## Why

`nt bf16` is the only one of the four workloads with a real opponent — under fp16 `torch._grouped_mm` falls back to 16 separate `mm` launches, and only bf16 reaches its CUTLASS grouped kernel. We lost it by 6.4%.

Two things cost us. The 128x256 tile holds a whole output tile in shared memory, 64 KiB, which leaves room for three mainloop stages against CUTLASS's four. And `epilogue_stage_n`, the mechanism that already trades that buffer for a fourth stage, was offered to dense and batched GEMM only.

Opening it to every grouped type does not work: a tight group's last tile is ragged, those rows are stored under a row mask rather than in one wide store, so chunking makes them pay a staging round for a store that never widens. On MoE's real routing (`min=1 max=663` rows per group, 55% of tiles ragged) that costs 3–5.5%. How many tiles are ragged is a property of the routing, not of the shape, so the selector cannot tell the two cases apart — and `GroupedGemmFwdOp` had no way for a caller to say which one it is holding.

## What

Offer the staged epilogue by whether a type's output tiles are whole: aligned m-grouped, K-grouped, dense and batched take it; tight m-grouped does not.

Give the caller a way to state that its groups are already aligned. Where they are, tight and padded are the same bytes — the only difference is whether the kernel must keep a row mask for a ragged tail. `batch_padded_offsets` was already in the signature and read by nothing, so it becomes `optional`: passing it states the padded layout and runs `M_GROUPED_ALIGNED_PSUM`, omitting it keeps today's tight path byte for byte. `GroupedGemmFwdOp.row_block` exposes the block to pad against.

Check the statement, because it is a contract. `torch._grouped_mm` device-asserts when its offsets disagree with the rows; this op computed 30.8% of rows wrong in silence. A row count that cannot be the padded layout is now refused on the host, and `layout_guard()` answers the rest asynchronously, shaped like `MoeGroupedGemmFwdOp.layout_guard`, for tests and benchmarks to assert on.

## Benchmark

**Environment**: H200, CUDA 13.2, PyTorch 2.13.0+cu132, TileLang 0.1.12

Baseline is `torch._grouped_mm`; times are CUPTI device-busy medians, taken on an idle GPU. Both speedups use the same baseline measurement, so the before and after columns are comparable.

### GroupedGemmFwdOp `(batch_sum=4096, groups=16, N=4096, K=4096)`

| Workload | baseline (ms) | before (ms) | after (ms) | before speedup | after speedup |
| --- | --- | --- | --- | --- | --- |
| nt fp16 | 0.2451 | 0.2199 | 0.2202 | 1.115 | 1.113 |
| **nt bf16** (padded after) | 0.2064 | 0.2171 | **0.1995** | **0.951** | **1.035** |
| nn fp16 | 0.2476 | 0.2111 | 0.2110 | 1.173 | 1.173 |
| **tn fp16** | 0.3037 | 0.2895 | **0.2731** | 1.049 | **1.112** |

**Takeaways:**

- `nt bf16` goes from behind to ahead, 0.951 to 1.035. The row states the padded layout after this PR and carries not one byte more — 256 rows per group already sit on the row block — so the gain is the row mask the kernel no longer keeps plus the fourth mainloop stage.
- `tn fp16` gains 5.7% from the staged epilogue alone: a K-grouped output tile is whole, so nothing about it is ragged.
- `nt` / `nn` fp16 run the tight path, whose generated kernel is identical to main's; 1.115→1.113 and the unchanged 1.173 are noise, as the absolute times show.
- The two aligned layouts have no bench workload, measured at kernel level: `aligned_psum` 91.0 → 83.2 us (+9.4%), `aligned_per_row` 90.4 → 84.3 us (+7.3%).

### Consumers this does not touch

Of the 66 selector inputs the template's four consumers produce, three change spec and the other 63 generate identical kernel code. Measured rather than only argued:

| Op | Workloads | Worst ratio vs best baseline | vs main |
| --- | --- | --- | --- |
| `MoeGroupedGemmFwdOp` | 6 | 1.224 (torch-grouped-mm) | 0.999–1.022, noise |
| `BmmFwdOp` | 16 | 1.008 (torch-cublas) | unchanged |
| `FusedMoeSharedExpertFwdOp` | 21 | 0.803 (vllm) | matches main to the microsecond |

## Regression

A benchmark assertion would have caught the silent miscompute: `bench_grouped_gemm.py` now runs `torch._assert_async(op.layout_guard(...))` before timing, so a workload whose stated layout disagrees with its data fails the row instead of reporting a fast wrong number.

## Test plan

- [x] `pre-commit run --all-files`
- [x] `pytest` on the affected files, on a real GPU — 237 passed
- [x] `python scripts/validate_manifest.py --levels schema,signature,shape,dtype,bench --strict`

New cases: `chunked-epilogue` (staged epilogue over a tight layout with ragged groups, the interaction the two changes create); `test_padded_layout_runs_the_aligned_kernel[groups-already-on-a-block]` / `[groups-need-padding]` (a padded call's result and that it reaches `M_GROUPED_ALIGNED_PSUM`, the second padding for real); `test_padded_offsets_are_refused_when_the_groups_split_k`; `test_the_general_kernel_does_not_serve_a_padded_call`.


src/tileops/ops/gemm/grouped_gemm.py
```diff
@@ -233,6 +290,7 @@ def _eager_forward(
             k,
             dtype,
             device_index,
+            padded,
         )
         return self.kernel(a, b, batch_sizes, batch_offsets, batch_padded_offsets)
```
lcy-seso, line 237: Blocking: the tight-layout fallback now passes `None` into a TileLang parameter that is still a required tensor. When the persistent candidate refuses a valid tight call (for example, `K=4100`), dispatch selects the general kernel, `batch_padded_offsets` is `None`, and that kernel forwards this value directly to its JIT ABI. Please have the general adapter substitute a valid unused tensor (for example, `batch_offsets`) and add an execution test for this fallback rather than only testing kernel selection.

src/tileops/ops/gemm/grouped_gemm.py
```diff
@@ -141,18 +162,51 @@ def _get_kernel(
...
+            batch_offsets: Start row of each group under a tight layout.
+            batch_padded_offsets: Start row of each group under a padded layout,
+                or ``None`` for a tight one.
+
+        Returns:
+            A 0-d ``torch.bool`` tensor on ``a``'s device.
+        """
+        starts = batch_offsets if batch_padded_offsets is None else batch_padded_offsets
+        ends = starts + batch_sizes
+        ok = (starts[1:] >= ends[:-1]).all() & (ends[-1] <= a.shape[0]) & (starts[0] == 0)
```
lcy-seso: Blocking: this predicate is weaker than the layout that the template scheduler actually consumes. The tight scheduler starts group `g` exactly at the previous end, while the aligned scheduler starts it at `align_up(previous_end, row_block)`; neither scheduler reads an arbitrary later `starts[g]`. For example, `sizes=[1, 1]`, padded starts `[0, 256]`, and 384 rows passes this guard, but the kernel reconstructs the second start as 128 and computes the wrong rows. Tight gaps have the same problem. Please validate the exact recurrence, non-negative sizes, and exact final coverage (`last_end == rows` for tight and `align_up(last_end) == rows` for padded), rather than using `>=`/`<=`.

src/tileops/kernels/grouped_gemm/grouped_gemm.py
```diff
@@ -190,6 +190,11 @@ class GroupedGemmKernel(Kernel):
     supported_archs: list[int] = [80, 86, 89, 90]
     general: bool = True
 
+    @classmethod
+    def applies(cls, call) -> bool:
+        """Serves a tight row layout: every group starts where the last one ended."""
+        return not call.padded
```
lcy-seso: Blocking: rejecting every padded call here leaves part of the manifest-declared API with no implementation. A padded call with a valid but non-TMA-addressable extent such as `K=4100` is rejected by the persistent candidate and also by this general candidate, although the manifest declares no such restriction. Please provide a padded general fallback, or make the supported region an explicit user-visible contract and validate it before dispatch. The former would preserve the existing general-fallback design.

tests/ops/test_grouped_gemm.py
```diff
@@ -193,3 +194,54 @@ def test_selection_prefers_the_template_where_tma_can_address_the_operands(
...
+    call = GroupedGemmCall(
+        arch=get_sm_version(),
+        numel=4096,
+        num_experts=16,
+        n=4096,
+        k=4096,
+        dtype=torch.float16,
+        padded=True,
+    )
+    assert not GroupedGemmKernel.applies(call)
```
lcy-seso: Please add direct tests for `layout_guard()`. None of the new op tests call it, and the benchmark assertion only exercises metadata generated by the workload, so false positives in the guard remain invisible. At minimum, cover valid and gapped tight layouts, valid padded recurrence, an extra-block padded gap, negative sizes, incomplete final coverage, malformed metadata shape/dtype, and zero-group behavior.

workloads/grouped_gemm.py
```diff
@@ -1,7 +1,9 @@
 import math
+from typing import Optional
 
 import torch
 
+from tileops.ops.gemm.grouped_gemm import GroupedGemmFwdOp
```
lcy-seso: Please avoid making the shared workload/reference layer import the Op implementation to learn its contract. This reverses the repository's manifest/workload-to-implementation trust direction and can make both tests and benchmarks follow an implementation mistake. The fixed alignment should come from a spec-owned or otherwise neutral contract constant, not from `GroupedGemmFwdOp`.

src/tileops/ops/gemm/grouped_gemm.py
```diff
@@ -167,7 +221,7 @@ def forward(
         b: torch.Tensor,
         batch_sizes: torch.Tensor,
         batch_offsets: torch.Tensor,
-        batch_padded_offsets: torch.Tensor,
+        batch_padded_offsets: Optional[torch.Tensor] = None,
```
lcy-seso: This is a breaking semantic change, not only an optionalization: the argument was previously required and ignored, so existing NT/NN callers already pass it and will now silently select the padded interpretation, while existing TN/TT callers will raise. If this break is intentional before the first stable release, please state it explicitly and document the required caller migration; otherwise the layout choice needs a non-ambiguous API transition.

lcy-seso (changes requested):
The staged-epilogue expansion looks useful, but the new padded-layout contract has correctness and fallback gaps that need to be resolved before merge. I have left the actionable findings inline.

michaelwithu:
Thanks — you were right that the padded contract wasn't ready. I've withdrawn it rather than patched it, so ②③④⑤⑥ go with it: all five existed only because a caller could state a padded row layout, and both start tables turn out to be derivable from `batch_sizes`, which neither kernel reads anyway — they rebuild the starts.

Inferring the layout instead doesn't help: it selects which kernel is compiled, on the host, from device-resident sizes. I measured the speculate-and-assert alternative at +3.0% per call against a 5.4% gain, paid on every call and repaid only by block-aligned ones. The right design takes segment lengths with `m_alignment` as a construction-time guarantee, like `MoeGroupedGemmFwdOp`, and belongs in its own PR.

① is fixed on its own. Your premise didn't reproduce — with `K=4100` the general kernel runs and is numerically correct (max err 1.95e-3), because `batch_padded_offsets` is declared in both of its prim_funcs and dereferenced in neither. The parameter is gone.

What's left is the staged-epilogue and selector work: `ops/`, `manifest/`, `workloads/` and `benchmarks/` are byte-identical to main.

lcy-seso (approved):

Issue #2131, opened 2026-09-15
[META][PERF] Ops below 0.80x of their best external comparator


lcy-seso:
## Scope

Fourteen implemented ops lose to their best external comparator by more than 20% on at least one declared workload. This issue tracks closing those gaps. It creates no new operators.

## How the list was produced

Nightly run [34919253127](https://github.com/tile-ai/TileOPs/actions/runs/34919253127), 2026-09-15, commit 15e4feea, H200, image `ghcr.io/tile-ai/tileops-runner:cu132-torch2.13-tl-afcebed1-2`. Every declared workload of every implemented op: 1072 cases across 182 ops, of which 1066 record at least one non-`tileops` baseline.

Ratio is the best non-`tileops` baseline's `device_busy_ms` over ours. The cut is 0.80 on an op's worst case.

| Band | Ops |
|---|---|
| below 0.50 | 7 |
| 0.50 to 0.80 | 7 |
| 0.80 to 1.00 | 69 |
| at or above 1.00 on every case | 92 |

The 69 ops in the 0.80-1.00 band are almost entirely elementwise kernels sitting between 0.96 and 1.00 — inside measurement noise on a card that drifts several percent over hours. They are not in scope here. Chasing them would mostly measure the weather.

Seven training-path ops are excluded: `MultiHeadAttentionBwdOp`, `GroupedQueryAttentionBwdOp`, `GatedDeltaNetBwdOp`, `DeltaNetBwdOp`, `GLABwdOp`, `DeltaNetAutogradOp`, `GatedDeltaNetAutogradOp`. They are frozen — correctness is maintained, performance is not pursued.

## The fourteen

| Op | Family | Worst | Cases below 1.00 | Baseline |
|---|---|---|---|---|
| `EngramDecodeFwdOp` | sequence_modeling | 0.31 | 3 / 3 | torch-compile |
| `FFTC2CFwdOp` | spectral | 0.37 | 3 / 3 | torch-cufft |
| `BmmFp8KNFwdOp` | gemm | 0.39 | 4 / 5 | flashinfer-bmm-fp8 |
| `GroupedQueryAttentionPrefillVarlenFwdOp` | attention | 0.47 | 6 / 6 | fa3 |
| `GroupedQueryAttentionDecodePagedWithKVCacheFwdOp` | attention | 0.49 | 7 / 8 | fa3 |
| `GemmW4A16FwdOp` | gemm | 0.49 | 6 / 6 | marlin-fp16 |
[...]

Issue #2130, opened 2026-09-15
[META][GEMM] Scheduling paths and coverage inside the existing GEMM ops


lcy-seso:
## Scope

Everything here changes an existing op. No new manifest entry, no new class in `ops/gemm/__init__.py`, no movement in the operator count. The work is scheduling paths inside the ops that already exist, plus the dtype and workload coverage those ops declare.

`GemmFwdOp` already dispatches six schedules — tiled, persistent static wave (`_gemm_coop2_kernel`), split-K, swap-AB, `GemmvKernel`, `SmallBatchGemmKernel` — behind `heuristics.best_config`. Callers do not choose a schedule and must not start to. Everything below lands the same way.

## Measured state

Nightly run [34919253127](https://github.com/tile-ai/TileOPs/actions/runs/34919253127), 2026-09-15, commit 15e4feea, H200, image `ghcr.io/tile-ai/tileops-runner:cu132-torch2.13-tl-afcebed1-2`. Ratio is the best non-`tileops` baseline's `device_busy_ms` over ours; below 1.00 means we lose.

| Op | Cases below 1.00 | Worst | Worst case | Baseline |
|---|---|---|---|---|
| `BmmFp8KNFwdOp` | 4 / 5 | 0.39 | `mha-decode-b32-pv-per-tensor` | flashinfer-bmm-fp8 |
| `GemmW4A16FwdOp` | 6 / 6 | 0.49 | `decode-long-k-pressure` | marlin-fp16 |
| `BmmFwdOp` | 10 / 16 | 0.73 | `square-b8-2k-bfloat16` | torch-cublas |
| `GemmFwdOp` | 8 / 16 | 0.90 | `ds-v3-prefill-gate-up-bfloat16` | cublaslt-best |
| `GroupedGemmFwdOp` | 1 / 4 | 0.96 | `nt-batch16-m4096-n4096-k4096-bfloat16` | torch |
| `GemmFp8FwdOp` | 1 / 15 | 0.97 | `small-batch-down-m8-per-tensor` | deepgemm |

`GemmFp8FwdOp` is in good shape and is not a target here. The other five are.

### 1. Stream-K schedule for `GemmFwdOp`

[...]

PR #2159, opened 2026-09-20
[Perf][Gemm] Speed up the W4A16 W4 GEMV case

superAngGao:
## Summary

- Reduce complete `GemmW4A16FwdOp` latency for W4 `(M, N, K) = (1, 8192, 8192)`, FP16, from 28.83 to 23.84 us on H200 (1.21x versus main).
- Assign each output row to one warp and retain its FP32 sums across K, then finish with warp shuffles; use 16 output rows per CTA for this case. Preserve the existing packed input format, FP32 scale arithmetic, and FP16 weight rounding.
- Enable this mapping only for the measured W4 shape on SM90 and add one regression case covering it. The change is limited to the GEMV kernel and its test fixture.

Refs #2130, #2131. **W4 remains 9.4% slower than the fastest Marlin baseline; this PR stays in draft.**

## Test plan

- [x] `pytest tests/ops/test_gemm.py -k w4a16 -q`: 14 passed, 64 deselected on H200.
- [x] Exact FP32-affine-to-FP16 basis checks on W4 at K indices 0, 1, 127, 128, 255, 256, 1023, 1024, and 8191, with zero tolerance; three additional tuning configurations passed the same exact check.
- [x] `SKIP=gitleaks pre-commit run --all-files`: all other hooks passed. The Gitleaks hook's Go bootstrap was unavailable; the matching official Gitleaks 8.30.0 binary passed `git --pre-commit --staged --redact --no-banner`.
- [x] Three paired complete-Op benchmark runs, alternating measurement order, with correctness checked before timing.

## Test node delta

`python scripts/test_node_delta.py --base origin/main` (base `0bedc99937592915be4920e974a938a11c2690b7`):

```text
File                      Base    HEAD    Delta
-----------------------------------------------
tests/ops/test_gemm.py      77      78       +1
-----------------------------------------------
TOTAL                       77      78       +1

Growth: +1.3%
```

- `full-w4a16-warp-rows`: covers the new SM90 W4 mapping through the public Op and the shared workload reference.

## Benchmark

**Environment:** NVIDIA H200, SM 1500 MHz / memory 3201 MHz, driver 595.71.05, PyTorch 2.10.0+cu129 (CUDA runtime 12.9), NVCC 13.2.78, TileLang 0.1.11+cu129.git65dbc983, vLLM 0.19.1.

### GemmW4A16FwdOp

Existing W4 workload and `ManifestBenchmark` protocol: CUPTI timing, cold L2, 200 samples per measurement. Values below are the median of three paired measurements. Both main and this PR run the complete Op, each with one device kernel; main uses its original GEMV kernel through the existing kernel override API. Marlin uses the existing benchmark adapter.

| Shape (M, N, K) | dtype | Main (ms) | PR (ms) | Marlin FP32 (ms) | Marlin FP16 (ms) | vs main | vs fastest Marlin | BW (TB/s) |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| (1, 8192, 8192) | float16 | 0.028832 | 0.023840 | 0.021792 | 0.021920 | 1.21x | 0.914x | 1.519 |

PR medians range from 23.68 to 23.872 us; main from 28.80 to 28.9275 us; Marlin FP32 from 21.7915 to 21.824 us. The PR and main both have maximum absolute error 0.0307999 against the independent FP32 GEMM reference using FP16-rounded weights.

**Takeaway:** W4 improves by 17.3% in latency, with a remaining 2.05 us gap to Marlin. Performance coverage in this PR is restricted to W4.


superAngGao:
Closed because PR#2168 provides a better solution.

PR #2160, opened 2026-09-20
[Refactor][Attention] Migrate varlen GQA kernels to the unified op

superAngGao:
## Summary

- migrate the existing 16-bit regular and sliding-window kernels to `GroupedQueryAttentionVarlenFwdOp`
- remove the two superseded public Ops and fold their workloads, tests, benchmark coverage, and roofline paths into the unified Op
- derive request lengths and flat tile ownership from `cu_seqlens` on the GPU; `total_q` and `total_kv` remain TileLang dynamic extents
- align the public result with Dense GQA: return only `o`; online-softmax `logsum` remains kernel-internal
- make fully masked causal rows deterministic (`o = 0`) and cover `q_len > kv_len`

## Validation

- H200 GQA suites: **71 passed total** — Dense/Bwd **41**, Varlen **30**
- H200 smoke coverage, including fully masked rows and D=128 sliding: **19 passed**
- H200 compile-boundary Varlen coverage: **6 passed**
- H200 native-CUPTI D=128 sliding workload: **1 passed**, one kernel, `0.0699 ms`
- roofline-oracle, manifest API, and kernel-map suites: **54 passed**
- strict manifest validation and Ruff passed

No performance improvement is claimed; this PR migrates and consolidates the existing implementations.


src/tileops/manifest/attention.yaml
```diff
@@ -221,15 +221,27 @@ GroupedQueryAttentionVarlenFwdOp:
       - "o.shape == (T_q, H, D)"
       - "H % H_kv == 0"
 
-  workloads: []
+  workloads:
+    - {total_q: 2048, total_kv: 2048, batch: 4, heads: 32, heads_kv: 8, dim: 128, q_lens: [512, 512, 512, 512], kv_lens: [512, 512, 512, 512], is_causal: true, dtypes: [float16, bfloat16], label: "llama-8b-uniform"}
```
lcy-seso, line 228: Blocking: this PR adds workloads and in-tree kernels for this Op, but the manifest still marks it `spec-only` while deleting two public `implemented` Ops. That downgrades validation to L0 and drops dtype/shape/benchmark and `torch.compile(fullgraph=True)` coverage. Please either keep the old implemented Ops until the unified contract is fully migrated, or narrow this contract to the supported feature set and mark it implemented.

src/tileops/kernels/attention/varlen.py
```diff
@@ -0,0 +1,81 @@
...
+        self.heads_kv = heads_kv
+        self.dim = dim
+        self.is_causal = is_causal
+        self.dtype = dtype
+        self.sm_scale = dim**-0.5 if sm_scale is None else sm_scale
+        self.softcap = softcap
+        self.window_size_left = window_size_left
+        self.window_size_right = window_size_right
+        self.accum_dtype = accum_dtype
+        self.init_config(config, tune)
```
lcy-seso, line 79: Blocking: autotuning fails deterministically here because `init_config(..., tune=True)` expects `self.kernel`, but these dynamic kernels never set it. Both regular and sliding-window calls raise `AttributeError`. Please make the dynamic program factory compatible with the tuning path and provide a valid tuning input supplier for the integer cumulative-length tensors.

tests/ops/attention/test_gqa_varlen.py
```diff
@@ -1,21 +1,22 @@
...
+    gqa_varlen_fwd_roofline,
 )
 from workloads.attention.gqa import (
     GroupedQueryAttentionSlidingWindowVarlenFwdWorkload,
 )
 
 
-class GroupedQueryAttentionSlidingWindowVarlenFwdTest(
+class GroupedQueryAttentionVarlenFwdTest(
     GroupedQueryAttentionSlidingWindowVarlenFwdWorkload, TestBase
```
lcy-seso: Please move the canonical reference implementation into the workload class and reuse it from both tests and benchmarks. The current test-local reference plus a separate benchmark-local copy can drift, and the workload still carries the old sliding-window name instead of representing the unified Op contract.

tests/ops/attention/test_gqa_varlen.py
```diff
@@ -312,19 +341,13 @@ def test_gqa_sliding_window_varlen_fwd_op(
...
-        heads_kv=heads_kv,
-        dim=dim,
-        max_seqlen_q=test.max_seqlen_q,
+    op = GroupedQueryAttentionVarlenFwdOp(
         is_causal=is_causal,
         window_size_left=wl,
         window_size_right=wr,
-        tune=tune,
     )
     test.check(op, *test.gen_inputs(), atol=1e-2, rtol=1e-2)
```
lcy-seso: The FP16 cases should use the project-standard `1e-3` tolerance; the observed tail-case error is about `1.22e-4`, so `1e-2` is unnecessarily loose. Please also add partial-tile sliding-window cases, dynamic-total reuse, regular-path `sm_scale`/`softcap`, invalid-input, and empty-request/KV regressions, and record the required test-node delta in the PR description.

benchmarks/ops/attention/bench_gqa.py
```diff
@@ -387,43 +400,51 @@ def _run(q, k, v, cu_seqlens_q, cu_seqlens_kv):
...
+    op = GroupedQueryAttentionVarlenFwdOp(
+        is_causal=causal,
+        window_size_left=window_size_left,
+        window_size_right=window_size_right,
     )
     bm = ManifestBenchmark(op, test)
 
-    functors = {"tileops": op, "torch-ref": _torch_gqa_prefill_varlen_ref(test)}
-    fa3_fn = _fa3_gqa_prefill_varlen(test)
+    functors = {
```
lcy-seso, line 512: Blocking: only FA3 is checked against the reference. `ManifestBenchmark.compare` times the functors but does not validate their outputs, so an incorrect TileOps result can still be benchmarked and published. Please assert `tileops` against `torch-ref` before timing.

src/tileops/ops/attention/gqa.py
```diff
@@ -608,9 +606,13 @@ class GroupedQueryAttentionVarlenFwdOp(Op):
     ``cu_seqlens_q`` and ``cu_seqlens_kv`` delimit each request. The interface
     covers both prefill and decode; tensor geometry and sequence metadata come
     from each call, while mask, score, out_dtype, and RoPE semantics are fixed at
-    construction. This spec-only shell intentionally has no BUILTIN kernel.
+    construction. The current BUILTIN path implements 16-bit regular and
+    sliding-window attention; FP8 and fused RoPE remain part of the public
+    contract for later kernel migrations.
```
lcy-seso, line 604: Before activating this unified public contract, please align the constructor with the manifest positional ABI. The manifest declares `softcap, out_dtype, pos_encoding_mode, rotary_dim, rope_layout`, while the constructor places `out_dtype` last. Also, `kernel_map` should be keyword-only together with `target` and `tune`.

src/tileops/manifest/attention.yaml
superAngGao: Addressed in 98b437d0. The unified FP8/RoPE-capable Op remains `spec-only`, and both previous `implemented` public Ops, their manifest contracts, compile boundaries, workloads, tests, and benchmarks are retained until the remaining capabilities migrate.

src/tileops/kernels/attention/varlen.py
superAngGao: Addressed in 98b437d0. Both dynamic program factories are now stored as `self.kernel`, and Varlen supplies valid Q/K/V plus `int32` cumulative-length tensors during tuning. I also ran full H200 sweeps successfully: 54 regular candidates and 16 sliding candidates.

tests/ops/attention/test_gqa_varlen.py
superAngGao: Addressed in 98b437d0. `GroupedQueryAttentionVarlenFwdWorkload.ref_program` is now the canonical regular/window reference shared by tests and benchmarks. The old sliding workload name remains only as a compatibility subclass while its implemented Op is retained.

tests/ops/attention/test_gqa_varlen.py
superAngGao: Addressed in 98b437d0. FP16/BF16 checks now use `1e-3`; coverage adds partial tiles, one-Op dynamic-total reuse, regular scale/softcap, invalid cumulative-length inputs, and empty request/KV cases. The PR body records the node deltas: Varlen 30→36 and compile boundary 6→12.

benchmarks/ops/attention/bench_gqa.py
superAngGao: Addressed in 98b437d0. Every Varlen benchmark validates TileOps against the canonical reference before `bm.compare`; FA3 is validated against the same reference before being added. Representative regular and sliding benchmark nodes both pass on H200.

src/tileops/ops/attention/gqa.py
superAngGao: Addressed in 98b437d0. The unified constructor now follows the manifest positional ABI (`softcap`, `out_dtype`, `pos_encoding_mode`, `rotary_dim`, `rope_layout`), and `kernel_map` is keyword-only with `target`.

src/tileops/perf/formulas.py
```diff
@@ -653,6 +660,16 @@ def gqa_sliding_window_varlen_fwd_roofline(
     return int(flops), int(nbytes)
 
 
+def gqa_varlen_fwd_roofline(op: Any | None = None, **kwargs: Any) -> tuple[int, int]:
```
lcy-seso, line 663: Blocking: this routes manifest workload rows straight into `gqa_prefill_varlen_fwd_roofline`, which after #2167 reads `data["q_shape"]` unconditionally. Workload rows declare `total_q`/`heads`/`dim`, not `q_shape`, so `test_varlen_manifest_workloads_are_evaluable` fails with `KeyError: 'q_shape'` — that is the current gpu-smoke failure. 3 of the 7 unified workloads fail (the ones without a window); the 4 windowed rows pass because the sliding formula reads totals. Please normalise the row into `q_shape`/`k_shape` here before delegating.

Two follow-ups from the same place. Lines 666-667 are dead now: `_shape_or_attrs` already pops `_roofline_kwargs` and merges it. And `GroupedQueryAttentionPrefillVarlenFwdOp` has the same breakage on `main` — 0 of its 3 workloads evaluate at b07a259f — which is an upstream defect this PR only inherits and deserves its own issue.

src/tileops/ops/attention/gqa.py
```diff
@@ -599,22 +599,26 @@ class GroupedQueryAttentionVarlenFwdOp(Op):
...
```
lcy-seso, line 604: Blocking: the unified contract has no check that the offsets span the packed tensors, and no opt-in equivalent of the compatibility Op's `validate_inputs`. A call whose `cu_seqlens_q[-1]` disagrees with `q.shape[0]` is accepted silently and returns wrong numbers: with q of 128 rows and `cu_seqlens=[0, 64]`, `out[64:]` came back 99.2% NaN after the allocation was poisoned, because those rows are never written. The over-claiming direction stays in bounds — `compute-sanitizer --tool memcheck` reports 0 errors, also under `PYTORCH_NO_CUDA_MEMORY_CACHING=1` — so this is wrong results, not corruption.

Please carry the guard over, and tighten the manifest entry: its shape_rules state only `cu_seqlens_q.ndim == 1`, while both compatibility entries state `cu_seqlens_q.shape == (B + 1,)`.

tests/ops/attention/test_gqa_varlen.py
```diff
@@ -312,21 +281,100 @@ def test_gqa_sliding_window_varlen_fwd_op(
...
+) -> None:
+    test = GroupedQueryAttentionVarlenFwdTest(
+        2, q_lens, kv_lens, 8, 2, 64, True, -1, -1, torch.float16
+    )
+    op = GroupedQueryAttentionVarlenFwdOp(is_causal=True)
+    test.check(op, *test.gen_inputs(), atol=1e-3, rtol=1e-3)
+
+
+@pytest.mark.smoke
+def test_varlen_rejects_invalid_cumulative_lengths_contract() -> None:
```
lcy-seso, line 368: Blocking: this checks an int32 dtype error and a shape mismatch, both already enforced structurally. It stands in for `test_gqa_prefill_varlen_rejects_bad_contract_inputs` and `test_gqa_prefill_varlen_rejects_unsupported_dtype`, which this PR deletes from `tests/ops/attention/test_gqa.py` (43 -> 41 nodes) and which covered `cu_seqlens_q[-1] must equal`, non-decreasing offsets, `max_seqlen_q` too small, and float32 rejection.

After the PR `grep -rn validate_inputs tests/ benchmarks/` returns nothing, so the `validate_inputs=True` branch that still ships has no coverage. Please restore those cases against the new contract, and record the -2 in the body — `scripts/test_node_delta.py --base b07a259f` reports 197 -> 238.

src/tileops/kernels/attention/gqa_sliding_window_varlen_fwd.py
```diff
@@ -101,51 +102,35 @@ def apply_mask(acc_s, k_idx, bx, q_len, kv_len, offset):
...
+class _GQASlidingWindowVarlenFwdKernelBase(VarlenKernel):
     """Shared base for variable-length GQA sliding window forward kernels."""
 
     @classmethod
     def applies(cls, call) -> bool:
         return (
             uses_sliding_window(call)
             and not call.is_fp8
+            and not call.fuse_rope
+            and call.sm_scale is None
```
lcy-seso, line 114: With these two conditions a windowed call carrying `sm_scale` or `softcap` matches no kernel: this one refuses it, and `GQAPrefillVarlenFwdKernel` refuses any window. Measured: `GroupedQueryAttentionVarlenFwdOp(is_causal=True, window_size_left=32, softcap=5.0)` raises `no implementation serves this call`, and the same for `sm_scale`. The unified manifest entry declares `window_size_left/right`, `sm_scale` and `softcap` with no stated exclusion. Please either state the exclusion in the manifest or reject it in `__init__` naming the missing capability; this has to be settled before the entry flips to `implemented`.

src/tileops/kernels/attention/gqa_prefill_varlen_fwd.py
```diff
@@ -77,19 +80,13 @@ def _gqa_prefill_varlen_fwd_main(
...
-            max_seqlen_kv: T.int32,  # type: ignore
             output: T.Tensor(q_shape, dtype),  # type: ignore
-            lse: T.Tensor([heads, total_q], accum_dtype),  # type: ignore
         ) -> None:
-            with T.Kernel(T.ceildiv(max_seqlen_q, block_m), heads, batch, threads=threads) as (
-                bx,
-                by,
-                bz,
-            ):
+            with T.Kernel(num_q_tiles, heads, threads=threads) as (q_tile, by):
```
lcy-seso, line 85: The grid changed from `(ceildiv(max_seqlen_q, block_m), heads, batch)` to this packed 2-D form, and every block now runs a serial O(batch) prefix sum plus a binary search and holds `batch+1` int32 in shared memory. That removes idle blocks on ragged batches and adds a per-block scheduling cost, so it can move either way. "No performance improvement is claimed" is not the same as no regression — please post before/after on one card in one window, quoting `device_busy_ms`, for `llama-8b-uniform`, one ragged shape, and a batch >= 64 shape where the prefix sum is visible.

src/tileops/kernels/attention/varlen.py
```diff
@@ -0,0 +1,129 @@
...
+
+def varlen_entry(cls: type, call: AttentionCall) -> Entry:
+    """Build one reusable Varlen kernel object from call-independent facts.
+
+    Packed totals are deliberately absent. The object reads them from Q/K on
+    every forward call; the TileLang program factory then caches the compiled
+    specialization it needs.
+    """
+    args = dict(
+        batch=call.batch,
```
lcy-seso, line 25: `batch` stays in the cache key and in the compiled program (`GroupTiling(batch, block_m)`, `T.alloc_shared([batch + 1])`), so the PR removes recompilation per packed total but not per request count — the axis that changes every step under continuous batching. Worth a follow-up issue rather than this PR.

src/tileops/kernels/attention/varlen.py
```diff
@@ -0,0 +1,129 @@
...
+        from tilelang.utils.device import get_current_device
+
+        batch = self.batch
+        heads = self.heads
+        heads_kv = self.heads_kv
+        dim = self.dim
+        dtype = self.dtype
+        # Exercise more than one 128-token tile without making tuning inputs
+        # depend on any particular manifest workload.
+        tokens_per_request = 256
```
lcy-seso, line 97: The kernel is shape-dynamic now, so one config tuned at 256 tokens per request is reused at every total. Either derive the tuning lengths from the call this kernel was built for, or say in the docstring why 256 is representative.

benchmarks/ops/attention/bench_gqa.py
```diff
@@ -148,6 +122,99 @@ def test_gqa_bwd_bench(
...
+        tune=tune,
+    )
+    bm = ManifestBenchmark(op, test)
+    assert_matches_reference(op, test.ref_program, *inputs, **reference_tolerance(dtype))
+    functors = {"tileops": op, "torch-ref": test.ref_program}
+    fa3_fn = _fa3_gqa_varlen(test, window_size_left, window_size_right)
+    if fa3_fn is not None:
+        assert_matches_reference(fa3_fn, test.ref_program, *inputs, **reference_tolerance(dtype))
+        functors["fa3"] = fa3_fn
+    bm.compare(functors, *inputs)
```
lcy-seso, line 218: The two new varlen bench functions were inserted above this line, so the `# No FlashInfer baseline for bwd` comment that follows now trails the sliding-window test instead of `test_gqa_bwd_bench`.

workloads/attention/gqa.py
```diff
@@ -605,3 +613,47 @@ def gen_inputs(
...
+            kv_pos = torch.arange(kv_len, device=q.device)[None, :]
+            visible = torch.ones((q_len, kv_len), dtype=torch.bool, device=q.device)
+            if self.is_causal:
+                visible &= kv_pos <= q_pos
+            if self.wl >= 0:
+                visible &= kv_pos >= q_pos - self.wl
+            if self.wr >= 0:
+                visible &= kv_pos <= q_pos + self.wr
+            scores = scores.masked_fill(~visible.view(1, q_len, kv_len), float("-inf"))
+            probs = torch.softmax(scores, dim=-1).nan_to_num()
```
lcy-seso: `nan_to_num` also converts NaN arriving from the inputs, not only the fully-masked-row NaN this intends to handle. `torch.where` on the all-masked row states the intent.

lcy-seso (commented):
Reviewed at 5d63272c, rebased onto b07a259f. gpu-smoke is red on this head and the first comment is the cause.

Three smaller points not worth their own thread. The unified entry is benchmarked alongside both compatibility Ops, so nightly gets three rows measuring the same two kernels. The `T.sync_threads()` after `cumsum_offsets` is redundant, since every thread computes the whole prefix sum itself and `grouped_gemm.py` omits it. And the PR adds an intermediate Kernel base class and re-parents two public implemented Ops without touching `docs/design/` or recording the base-class decision in the body, which `.claude/domain-rules/ops-design.md` asks for.

What landed well: dropping the unused `lse` output, `T.dynamic` for the packed totals, the single shared `ref_program`, and `assert_matches_reference` before every `bm.compare`.

superAngGao:
Thanks for the detailed review. I have pushed fixes for the smoke/roofline failure, restored opt-in offset-span and monotonicity checks (including compatibility-path tests), explicitly rejected unsupported window + custom-scale/softcap calls, and addressed the reference-NaN, misplaced comment, tuning-sample explanation, and redundant barrier points. The Varlen test file passes on H200 (39 passed), and the latest CI GPU smoke and all other checks pass at `d85f1c59`.

I also ran the requested same-card, clock-locked H200 native-CUPTI A/B and put the exact numbers in the PR body. Relative to the baseline, the migration-only implementation is 26.3% lower in latency on the ragged case, 3.4% higher on manifest `llama-8b-uniform`, and 20.5% higher at B=64. This PR establishes the unified Op/kernel migration; kernel-performance work is prepared separately in the follow-up draft PR #2162. We will compare that work against these A/B cases rather than attribute an unmeasured speedup to it.

The PR body now also explains the intermediate `VarlenKernel`, compatibility inheritance, and temporary duplicate nightly rows. The batch-dependent compiled key remains an explicit follow-up rather than being silently described as fully shape-dynamic.

src/tileops/kernels/attention/varlen.py
superAngGao: Thanks. The supply-program docstring now explains that 256 tokens per request covers two 128-token tiles while keeping the candidate sweep bounded. It is a synthetic tuning point, not evidence that one config is optimal across all packed totals. Call-derived tuning is a separate improvement.

src/tileops/kernels/attention/gqa_sliding_window_varlen_fwd.py
superAngGao: Thanks. The constructor now rejects window plus custom sm_scale or nonzero softcap before dispatch, with an explicit error, and the manifest records the same exclusion. There is a regression test for both combinations. We can lift the restriction when a window kernel implements those modifiers.

tests/ops/attention/test_gqa_varlen.py
superAngGao: Agreed. The current test file covers truncated Q/K totals, decreasing offsets, the compatibility max_seqlen_q guard, and float32 rejection. The PR body now records that two old nodes were removed while the new Varlen file has 39 nodes.

src/tileops/kernels/attention/varlen.py
superAngGao: Agreed. The cache key and compiled program still depend on batch, so this PR removes total-length recompilation but not request-count recompilation. We are leaving that limitation explicit and out of this migration; #2162 does not by itself resolve it.

src/tileops/ops/attention/gqa.py
superAngGao: Thanks for the concrete reproducer. The current head adds opt-in validate_inputs checks for zero start, exact final totals, and non-decreasing offsets, plus the manifest B+1 shape rule. The default remains false, matching the compatibility Op, to avoid a CPU synchronization on every hot-path call. Please let us know if you consider that opt-in contract insufficient.

benchmarks/ops/attention/bench_gqa.py
superAngGao: Fixed in the current head: the no-FlashInfer-backward comment is back inside test_gqa_bwd_bench, directly after its compare call.

src/tileops/perf/formulas.py
superAngGao: Thanks for identifying the roofline failure. The current head normalizes total_q/total_kv into q_shape/k_shape before delegating, removes the redundant payload merge, and adds a manifest-workload evaluation test. GPU smoke is green. The legacy PrefillVarlen Op has the separate pre-existing issue you noted; this PR does not claim to fix it.

workloads/attention/gqa.py
superAngGao: Agreed. The reference now uses torch.where only for rows with no visible keys, preserving NaNs that originate in the inputs instead of applying nan_to_num to every result.

src/tileops/kernels/attention/gqa_prefill_varlen_fwd.py
superAngGao: Agreed that the packed grid needs a measured A/B. The PR body now reports same-card locked-H200 device busy: uniform 60.27 to 62.32 us (+3.4%), ragged 168.19 to 123.98 us (-26.3%), and B=64 107.52 to 129.58 us (+20.5%). We acknowledge both regressions rather than claiming no regression. The performance follow-up #2162 will need its own remeasurement before merge.

src/tileops/ops/attention/gqa.py
```diff
@@ -855,532 +962,590 @@ def forward(
...
         )
-        kernel = self.kernel_for("gqa_paged", inputs)
-        return kernel(*inputs)
+        if tuple(q.shape[1:]) != (self.heads, self.dim):
+            raise ValueError("q shape does not match the legacy Op constructor")
+        if tuple(k.shape[1:]) != (self.heads_kv, self.dim):
+            raise ValueError("k/v shape does not match the legacy Op constructor")
+        if cu_seqlens_q.shape[0] != self.batch + 1:
+            raise ValueError("cu_seqlens_q length does not match batch")
+        lengths = cu_seqlens_q[1:] - cu_seqlens_q[:-1]
```
lcy-seso: Blocking: this compatibility Op no longer preserves its previous cumulative-offset validation. Before this refactor it unconditionally rejected a nonzero start, decreasing offsets, and a final offset beyond the packed Q/K extent. The shared validator only performs those checks when `validate_inputs=True`, but this compatibility constructor neither exposes nor enables that option; the code here now checks only the maximum Q length. On this head I confirmed that `[1, 9, 17]`, `[0, 12, 8]`, and `[0, 8, 20]` are all accepted for 16-row tensors, whereas the base rejects each case. The over-claiming case can launch tiles whose reads/writes extend past the packed tensors. Please restore the legacy checks here (or enable an equivalent validation path by default) and add regressions for nonzero, decreasing, and over-claiming offsets.

lcy-seso (changes requested):
One blocking compatibility regression remains; details inline.

src/tileops/ops/attention/gqa.py
superAngGao: Thanks, this was a real compatibility regression. Restored the legacy sliding-window Op checks unconditionally at 22be932f: both offset arrays must start at zero, remain non-decreasing, and not over-claim their packed tensor. The unified Varlen Op, kernels, manifest, and benchmarks are unchanged. Added regressions for a nonzero Q start, decreasing KV offsets, and an over-claiming Q endpoint; the full Varlen test file passes 42/42 on H200.

lcy-seso (approved):
Re-reviewed the latest compatibility fix. The legacy sliding-window Varlen op again rejects nonzero starts, decreasing offsets, and packed-tensor overclaims unconditionally, with regressions for all three cases. The full Varlen test file passes locally (42 tests), and CI is green.

